PR #9, opened 2023-01-25
Proper modularization

Cyprex:
## About The Pull Request

So, the includes in the .dme are storted alphabetically. But we want our modular files to override Skyrat's ones, loading after them, as you'd expect. To that end I've renamed modular_bubbers to modular_zubbers.

This PR also removes the old modularization guides that were largely copied from Skyrat. We've already discussed we'll do otherwise.

In addition to these it cleans up the veteran status changes, by deleting the now unnecessary module for them and commenting out the Unreachable code in my override.

It also adds @miniusAreas and @KathrinBailey to Codeowners for maps, if you'd rather not please tell me or make a PR!


Cyprex:
Let's just merge this and see what happens, I guess.

PR #34, opened 2023-02-11
Max detective spot raised from 1 to 2

567Turtle:
## About The Pull Request

Makes it so there can be two detectives

## How This Contributes To The Skyrat Roleplay Experience

They give us the Harry and Kim outfit but only one detective, this is criminal, there can now be two, detective is kind of a "no matter how many you have one is just as good as 200" role so this won't really effect balance, theres still only one revolver.

## Proof of Testing

<details>
<summary>Open Enclosed
</summary>
  Lieutenant Double Yefreitor Barry Du Boys and Lieutenant Jim Catsuragi arriving at the scene of the lynching of the warden.

image

</details>

## Changelog

:cl:
add: There can now be two detectives
/:cl:

### DIFF
--- a/code/modules/jobs/job_types/detective.dm
+++ b/code/modules/jobs/job_types/detective.dm
@@ -5,8 +5,8 @@
 	auto_deadmin_role_flags = DEADMIN_POSITION_SECURITY
 	department_head = list(JOB_HEAD_OF_SECURITY)
 	faction = FACTION_STATION
-	total_positions = 1
-	spawn_positions = 1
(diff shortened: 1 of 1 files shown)


Cyprex:
This is better done in config via job config.toml

567Turtle:
didn't know that was a thing, I guess I'll ask bubbers if he can do it in a bit

Cyprex:
Ask me in 16h or so, I'll have a repo up for the live server configs.

PR #43, opened 2023-02-14
Adds a gamemode vote

567Turtle:
## About The Pull Request

Adds a gamemode vote which works (fully)

THINGS I NEED 2 DO
[x] Make it so you can start the vote
[x] Make it so if the vote passes for an extended round, make the current round extended
[x] Make it so the vote starts automatically at the start of the round

**DONE AND READY TO MERGE**

image
image
image
image
image

## Changelog

:cl:
add: You can now vote if the round will be extended or dynamic
/:cl:


mcmeiler:
The vote option for extended is misleading. "Low threats" implies threats still roll. This, however, sets `dynamc_forced_extended = TRUE`, which prevents ALL rulesets from drawing.

567Turtle:
> The vote option for extended is misleading. "Low threats" implies threats still roll. This, however, sets `dynamc_forced_extended = TRUE`, which prevents ALL rulesets from drawing.

opforces still exist, but I can change that to say minimal or no threats if you'd prefer

mcmeiler:
> opforces still exist

Please. When's the last time you saw an accepted OPFOR.

567Turtle:
> > opforces still exist
> 
> Please. When's the last time you saw an accepted OPFOR.

When I was a min I accepted opfors all the time but... I'll change it

modular_zubbers/code/datums/votes/gamemode_vote.dm
```diff
@@ -0,0 +1,30 @@
+// BUTTERSTATION ADDITION
+
+#define CHOICE_EXTENDED "Extended Ruleset (Minimal Threats)"
```
Cyprex, line 3:
```suggestion
#define CHOICE_EXTENDED "Extended, potentially with OPFORs"
```

567Turtle: My line of thought was there's always gonna be some amount of threat, whether that's admeme intervention, self antaggers or opfors, so I feel like minimal is the best way to put it.

Cyprex:
This doesn't even work right now, does it?

567Turtle:
> This doesn't even work right now, does it?

It worked fine for a day and then uhh... It just stopped working? It wasn't anything I committed because I went back through on a local host and tried running previous versions of this code and they also just didn't start? I have absolutely no clue why. It worked perfectly for a day, on both localhosts and in-game and I have absolutely no clue what just caused it to stop

PR #51, opened 2023-02-21
Bubber jobs branch

ReturnToZender:
## About The Pull Request
Makes it so that Warden and Blueshield require Security experience, rather than Command experience. The CE would probably not make a good Blueshield. Warden players should have been a sec-off at least once.

## Changelog

:cl:
add: Makes it so that Warden requires security experience
add: Makes it so that Blueshield requires security experience rather than Command experience.
/:cl:


ReturnToZender:
Can I just say "Thanks, I hate it" to Skyrat's modularity?

567Turtle:
I think this is good, if you tested it and it didn't implode then it wouldn't hurt that bad to TM

PR #55, opened 2023-02-26
Cucked all kill OBJs

Kilmented:
I haven't coded in a hot minute. Please give review. 
My solution to the "removal" of the code was to just comment it out. Truthfully, simply deleting the code is an option too.


Cyprex:
I agree that kill objectives should be removed, but this is not the way to do it. If anything I'm surprised this passes CI. Check out https://github.com/Bubberstation/Bubberstation/blob/master/code/modules/antagonists/traitor/datum_traitor.dm#L168 and others

PR #56, opened 2023-02-26
[MODULAR] Nerfs Lavaland

NullDagaf:
## About The Pull Request

Lowers Lavaland fauna HP and damage output of KAs/crushers by 1/3. When an exact 1/3 split wasn't possible, HP was adjusted so the amount of hits required stayed the same.

Also deletes a duplicate file and renames a different one so it follows conventions.

## Why It's Not Bad For The Game

It allows alternatives to mining weaponry for if, for example, a megafauna finds it's way onto the station.

## Changelog

:cl:
balance: reduced fauna health and mining gear damage by 1/3
/:cl:


mcmeiler:
This is very much player-facing.

NullDagaf:
> This is very much player-facing.

Eh. Miners will still be able to kill fauna in the same amount of hits.
But sure.

tgstation.dme
```diff
@@ -6537,7 +6537,32 @@
 #include "modular_skyrat\modules\xenos_skyrat_redo\code\xeno_types\sentinel.dm"
 #include "modular_skyrat\modules\xenos_skyrat_redo\code\xeno_types\spitter.dm"
 #include "modular_skyrat\modules\xenos_skyrat_redo\code\xeno_types\warrior.dm"
-#include "modular_zubbers\code\modules\Dynamic\MidroundRulesets.dm"
+#include "modular_zubbers\code\modules\dynamic\midround_rulesets.dm"
```
mcmeiler, line 6540:
```suggestion
#include "modular_zubbers\code\modules\Dynamic\midround_rulesets.dm"
```
Fairly sure this is case-sensitive.

mcmeiler (changes requested):

tgstation.dme
NullDagaf: Did the folder name change not go through..?

NullDagaf:
For the record, the checks fail due to unrelated issues.

PR #65, opened 2023-03-02
Repaths a thing

NullDagaf:
## About The Pull Request
Someone for some reason PascalCase'd both the file and it's dir when the standard for both is snake_case.

## Changelog
Not player facing.


Cyprex:
Looks good to me, can be merged in 13h

PR #69, opened 2023-03-03
Tesh QOL update

Gr33d-y:
This moves the teshari to a modular file, changes their diet to be better and make more sense, removes the extra damage they take for no reason (anyone can set their size to be smaller than teshari) and fixes a bug where their hats are one pixel to the right for some reason.

These are simply the QOL updates from the now canned rework. There is no increase in speed or increase in damage in this PR.


mcmeiler (changes requested):
PR template. At least the Changelog.

Serxule:
teshari are birds, birds have hollow bones, they take extra damage because they are literally weak (also they're smaller therefore harder to hit)

projectkepler-ru:
this PR has been inactive for a long time now with no sign of updates

PR #74, opened 2023-03-03
Cigar*ette*

Kilmented:
1. Makes the cigarettes not blow your lungs out after smoking six. How? Changed the default lung damage, plus the ones for the special cigarettes. Healing cigarettes now heal your lungs as flavor or whatever, I dunno I thought it would be a fun addition.


Cyprex:
Sure, whatever

BobDanger-Coding:
Wait, why are some of the values changed into negatives?

Cyprex:
To heal instead of harm, presumably, though this should probably be tested or the code looked at.

KathrinBailey (changes requested):
make modular pl0x

also, lung harm -1?
that should make smoking heal you?

Cyprex:
Please check that code actually handles negative values correctly and there aren't any bugs from lungs healing over their cap or things like that.

PR #75, opened 2023-03-03
energy revolver & laser changes

W0LF-101:
Energy Revolver Changes:
No longer selfcharges 
Cell charge reduced from 1500 to 1000
Takes more power to fire lethal and taser shots (10% & 20% vs 5% and 13%) 

SC-1 Laser Gun Changes:
Takes less power to fire lethal shots (from 6% to 5%)

### DIFF
--- a/code/modules/projectiles/ammunition/energy/laser.dm
+++ b/code/modules/projectiles/ammunition/energy/laser.dm
@@ -3,7 +3,6 @@
 	e_cost = 83
 	select_name = "kill"
 /obj/item/ammo_casing/energy/laser/revolver
-	projectile_type = /obj/projectile/beam/laser
 	e_cost = 100
 	select_name = "kill"
 /obj/item/ammo_casing/energy/laser/hellfire
--- a/code/modules/projectiles/ammunition/energy/laser.dm
+++ b/code/modules/projectiles/ammunition/energy/laser.dm
@@ -4,7 +4,6 @@
 	select_name = "kill"
 /obj/item/ammo_casing/energy/laser/revolver
 	e_cost = 100
-	select_name = "kill"
 /obj/item/ammo_casing/energy/laser/hellfire
 	projectile_type = /obj/projectile/beam/laser/hellfire
 	e_cost = 100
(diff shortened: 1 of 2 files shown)


W0LF-101:
Please testmerge this alongside PR #98 Energy Sec as this change buffs the SC-1s to encourage picking those when the nonlethal ability of the SC-2s is not desired.

code/modules/projectiles/ammunition/energy/laser.dm
```diff
@@ -2,7 +2,10 @@
 	projectile_type = /obj/projectile/beam/laser
 	e_cost = 83
 	select_name = "kill"
-
+/obj/item/ammo_casing/energy/laser/revolver
+	projectile_type = /obj/projectile/beam/laser
```
Cyprex, line 6:
```suggestion

```

code/modules/projectiles/ammunition/energy/laser.dm
```diff
@@ -2,7 +2,10 @@
...
+	e_cost = 100
+	select_name = "kill"
```
Cyprex, line 8:
```suggestion
```

Cyprex: Removing these takes the values from `/obj/item/ammo_casing/energy/laser` instead, which is what you want here.
Also this ought to be modularised, I can handle that if you wish.

Cyprex (changes requested):
Looks good otherwise

modular_skyrat/modules/blueshield/code/special.dm
```diff
@@ -14,11 +14,11 @@
 	obj_flags = UNIQUE_RENAME
 	cell_type = /obj/item/stock_parts/cell/blueshield
 	pin = /obj/item/firing_pin/implant/mindshield
-	selfcharge = TRUE
+	selfcharge = FALSE
 
 /obj/item/stock_parts/cell/blueshield
 	name = "internal revolver power cell"
-	maxcharge = 1500
+	maxcharge = 1000
```
projectkepler-ru, line 21: keep the max charge at 1500

projectkepler-ru (changes requested):

projectkepler-ru (changes requested):
Scratch that, change it to 800 and reenable self charging, I'm reintroducing it as a blueshield sidearm then this is good to merge

projectkepler-ru (changes requested):

PR #79, opened 2023-03-03
Let the Skrell DRINK

Gr33d-y:
All this PR does is change Skrell from taking 5,000 times more damage in their liver compared to others to a nice and simple 2 times more damage in their livers. As it stands, if a Skrell drinks .5 units of whisky, their liver shuts down instantly. This is a problem for them because 1. Their diet is strict and contains toxic food, doing damage to replacement livers, and 2. some of the cool toxic foods they could enjoy, are made with a little bit of vodka. All this would do is allow them to partake in moderate booze and give a use to the toxic food that is made with vodka.

Edit for part two of the update: 
These changes make more sense to me as lorehead for skrell and what I want skrell to feel like. I would like feedback if possible for these changes as a few of them might be a bit strong or might need some better numbers.

Lungs are better in both cold and heat as of this update, the damage was changed from being level 2 at level 1 to just being at level 1 and the temperatures have been lowered and raised, skrell are creatures that adapt to their environment so this makes sense to me. 
Changes the diet to include seafood and give them water breathing, they are amphibious after all and wonderful fishers.
Their diet can be a lot of toxic stuff, so their liver should be able to deal with toxic stuff easier. Their liver took 20% more toxin damage for some reason, this made little sense to me as they are able to eat toxic foods, so this was changed to -20%. Keep in mind this is not toxin resistance, this is simply changing how much damage the liver itself takes from toxins.

As for the most controversial trait here, Hardly wounded, I feel as though skrell having 70% more brute damage is a bit much, but fair! It makes sense for the flavor of the species that they would have skin that is easy to bruise. That being said, the skin is thick and the bones are softer with more of a rubber feel to them. Both less likely to break, but are weaker in return. A good tradeoff IMO.

Edit for part three of the update: 
This adds in game lore to the skrell and changes their cold resistance to be more inline with what the Skrell are.


Cyprex:
Don't copy paste stuff like this, that's not modularization. In this case just edit the base files directly

Cyprex:
Default value for this for tg is 0.005 for all species, Skyrat changed it to 0.001. Do with that knowledge what you will, I'm cool with your change!

Cyprex:
I suppose this could be merged in its current state

Gr33d-y:
why are people mentioning unrelated PR's here?

PR #92, opened 2023-03-07
Better health viewing

NullDagaf:
## About The Pull Request
dreamseeker_1vl2ak1eDT

## Changelog

:cl:
qol: you now know your exact health as a simple/basic mob
/:cl:


Cyprex:
I really like this but if you could PR it upstream(ideally to tg) that would be most appreciated. If they refuse we can still merge this.

NullDagaf:
> I really like this but if you could PR it upstream(ideally to tg) that would be most appreciated. If they refuse we can still merge this.

Not happening.

PR #100, opened 2023-03-11
working out gives you STRONG_GRABBER for 15 seconds

NullDagaf:
## About The Pull Request

After using a weightlifting machine, you for 15 seconds, get the STRONG_GRABBER trait - meaning, your grabs start at aggressive. For the funsies.

## Changelog

:cl:
add: added an easter egg to weight lifting machines
/:cl:

### DIFF
--- a/code/game/objects/structures/gym.dm
+++ b/code/game/objects/structures/gym.dm
@@ -130,6 +130,9 @@
 		user.add_mood_event("exercise", /datum/mood_event/exercise)
 		to_chat(user, finishmessage)
 		user.apply_status_effect(/datum/status_effect/exercised)
+		// BUBBERSTATION EDIT START
+		user.apply_status_effect(/datum/status_effect/workout_buff)
+		// BUBBERSTATION EDIT END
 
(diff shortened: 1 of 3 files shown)


Cyprex:
Sorry, this is fun and seems very hard to abuse but I also see no good reason to potentially have this.

NullDagaf:
> Yeah uh, they want to do this because they play perma prisoner, and there is workout machines right next to the entry doorway. This is literally a powergame PR from someone calling people the nword in OOC.
> 
> This has massive balance implications as I will now explain. Strong-grabber allows you to instantly disarm/stun whoever you grab as it is instantly a violent grab (like CQC).

For the record, I haven't played prisoner ONCE on this server.
Also, this is specifically for the machines that stun you for 5 seconds or so on use.

NullDagaf:
I'm fine with lowering it to 5 seconds or closing it altogether if you think it's really such an issue.

Cyprex:
Yeah I don't think this is something we need, independently of what Yawet said.

PR #110, opened 2023-03-17
Genesis 11:9 (Required language tweaks)

567Turtle:
Now the whole world had one language and a common form of speech. 

And as people journeyed eastward, they found a plain in the land of Shinar and settled there.

And they said to one another, “Come, let us make bricks and bake them thoroughly.” So they used brick instead of stone, and tar instead of mortar.

“Come,” they said, “let us build for ourselves a city with a tower that reaches to the heavens, that we may make a name for ourselves and not be scattered over the face of all the earth.”

Then the LORD came down to see the city and the tower that the sons of men were building.

And the LORD said, “If they have begun to do this as one people speaking the same language, then nothing they devise will be beyond them.

**"Come, let Us go down and confuse their language, so that they will not understand one another’s speech.”**

So the LORD scattered them from there over the face of all the earth, and they stopped building the city.

That is why it is called Babel, for there the LORD confused the language of the whole world, and from that place the LORD scattered them over the face of all the earth.

:cl:
add: You can join most jobs without knowing Sol
/:cl:


projectkepler-ru:
@Cyprex  one of these days I'll get perm to add label >:(

projectkepler-ru (changes requested):
Please fixes your description and title

567Turtle:
> Please fixes your description and title

What's wrong with them?

NullDagaf:
> What's wrong with them?

Doesn't follow the template, also it's basically completely unclear as to what this PR does without looking at the code changes.

567Turtle:
> > What's wrong with them?
> 
> Doesn't follow the template, also it's basically completely unclear as to what this PR does without looking at the code changes.

Doesn't follow the template meh meh meh meh meh

mcmeiler:
> Doesn't follow the template meh meh meh meh meh

Following the template also makes reviews easier.

Apart from that, the Changelog part of the template actually shows up as the in-game changelog. Not adding it means players do not know these changes are added once they are merged.

modular_skyrat/modules/customization/modules/jobs/_job.dm
```diff
@@ -117,25 +117,83 @@
...
+
+/datum/job/paramedic
+	required_languages = null
+
+/datum/job/geneticist
+	required_languages = null
+
+/datum/job/virologist
+	required_languages = null // damara dent-head
+
```
mcmeiler, line 184: This is a bad idea, in my opinion. A virologist cannot release diseases unapproved by the CMO - How are they going to get them approved if they can't talk to one another?

567Turtle: Dunno ask psychotic moth chick

Cyprex:
I don't really have an opinion on whether or not we should disable language requirements, but regardless these should be done in a new modular file for our codebase.
Note that you can also set it on `/datum/job` to do it for every job.
Aside from these code changes, I'll leave the decision on whether or not this is a good idea to the other maintainers

567Turtle:
I don't really think modularizing this further is necessary. That would just spread out the code way too much and make maintaining this already shit code even shittier, it's commented correctly.

mcmeiler:
> I don't really think modularizing this further is necessary.

It makes upstream merges easier, as their code won't cause merge conflicts on their own modular files.

> it's commented correctly.

If I were really pedantic, it's not even commented correctly as it lacks end comments.

All in all I still think this should be in a modular file, too, if at all.

PR #112, opened 2023-03-18
Eradicate Poly

Myphicbowser:
## About The Pull Request

Removes poly from every Map, because it looks like Poly and her stupid fucking headset are load bearing codes, for SOME reason

## Changelog

:cl:
del: Poly
/:cl:


mcmeiler:
...What? I don't think I follow correctly.

> because it looks like Poly and her stupid fucking headset are load bearing codes

Elaborate further on this, please.

mcmeiler:
If this was the huge performance issue you make it out to be, other servers would have long removed Poly.

Removing Poly from the maps doesn't make the game not load messages, either. This happens during round setup. Saving messages happens after the round ended. The performance impact does not affect the game in any significant way.

KathrinBailey:
Pretty sure this was done in response to an admin spawning a ton of parrots and putting headsets on them and nothing else.
Additionally, modularisation + L.

PR #117, opened 2023-03-20
That ruleset got hands

ReturnToZender:
## About The Pull Request
Raises the population per threat requirement, to allow for more players to be considered not active. 
TG has a higher requirement because all the players do is fight antags. Constantly.
I think the actual number should be at or between 8 and 10, I feel like over 10 would be over-kill.

## Changelog

:cl:
add: Pop_per_requirement equal to 10, instead of 6.
/:cl:


ReturnToZender:
Can edit this in config :)

PR #124, opened 2023-03-22
Changes Antimov to include crew members instead of humans

NullDagaf:
## About The Pull Request
`"You must injure all crew members and must not, through inaction, allow a crew member being to escape harm.",`
`"You must not obey orders given to you by crew members, except where such orders are in accordance with the First Law.",`
`"You must terminate your own existence as long as such does not conflict with the First or Second Law."`

## Changelog

:cl:
fix: antimov now counts all crew members, instead of only humans
/:cl:


projectkepler-ru (approved):
Seems fine

PR #128, opened 2023-03-23
Gives Ghost Cafe Spawners Uses

Icarus-The-Sun:
## Gives Ghost Cafe Spawners a ton of uses.
-1 uses appears to not work in game, so this gives an extreme numbers of charges to the cafe spawners. There's no feasible reason ghosts should go through 10000000000000 charges through the round, so it should do more than enough to act as infinite charges.

## Changelog

:cl:
add: Gives Cafe Spawners more uses
/:cl:


Cyprex:
I fixed this for upstream skyrat but the upstream merge caused issues this time, unfortunately

PR #131, opened 2023-03-24
QM gets Blacksmith access

SingingSpock:
## About The Pull Request

Gives the QM access to the blacksmithing area. This is part of cargo, they should have access. 

Tested by walking into the blacksmithy on ice box. Didn't test on other maps, but I see no reason it'd be different.

## Changelog

:cl: SingingSpock
tweak: The QM gets access to the blacksmith area
/:cl:


SingingSpock:
As far as I can tell the CI tripped for something happening with a monkey's left arm hitting a disposals unit and something in tramstation's maints (The second one also appears on other PRs). I don't think it has to do with my PR at all.

PR #139, opened 2023-03-28
blueshield buff

projectkepler-ru:
an unapologetic buff to the blueshield and returning it to it's former glory

## About The Pull Request
This PR add back stuff blueshield had to be on them at all time (excluding their hell laser rifle)

## Changelog

:cl:
balance: added back the energy revolver to the blueshield arsenal and restored it's intended functionality as the founding father intended
imageadd: a new blueshield revolver sprite ported from project kepler
/:cl:

image


ReturnToZender:
Blueshield does not need a CENTCOM ID, and has a silver ID for a reason >:(

But other than that, I'm a fan!

modular_skyrat/modules/blueshield/code/special.dm
```diff
@@ -1,24 +1,26 @@
 /obj/item/gun/energy/e_gun/revolver //The virgin gun.
-	name = "energy revolver"
-	desc = "An advanced energy revolver with the capacity to shoot both electrodes and lasers."
-	force = 7
-	ammo_type = list(/obj/item/ammo_casing/energy/electrode, /obj/item/ammo_casing/energy/laser/revolver)
+	name = "X-R12 energy revolver"
+	desc = "An experimental model advanced energy weapon with the capacity to shoot both electrodes and lasers, used by many private defense contractors from Romulus to New Moscow, its also rather heavy and would certainly hurt to get pistol whipped by."
+	force = 22
```
mcmeiler, line 4: This seems very, *very* powerful. Are you sure?

ReturnToZender: I think the revolver should be comparable to the HoS' laser.

SingingSpock:
I would recommend keeping the silver ID, but giving them access to all basic departmental areas as part of the trim. That means that they can then be given a few additional places, such as the office of the current acting captain, without compromising their ability to enter the main part of engineering if/when the CE goes down.

ReturnToZender:
> I would recommend keeping the silver ID, but giving them access to all basic departmental areas as part of the trim. That means that they can then be given a few additional places, such as the office of the current acting captain, without compromising their ability to enter the main part of engineering if/when the CE goes down.

They have basic departmental access normally, as well as Head offices.

SingingSpock:
> > I would recommend keeping the silver ID, but giving them access to all basic departmental areas as part of the trim. That means that they can then be given a few additional places, such as the office of the current acting captain, without compromising their ability to enter the main part of engineering if/when the CE goes down.
> 
> They have basic departmental access normally, as well as Head offices.

Incorrect. Their access has long been broken. They get ACCESS_RESEARCH (the R&D room) but not ACCESS_SCIENCE (the actual department). The only Head office they have access to is the Captain, none of the others (not even HoP).

modular_skyrat/modules/blueshield/code/special.dm
projectkepler-ru: It is comparable and the revolver isn't more powerful

mcmeiler:
> @mcmeiler
I increased it's melee damage as to make it a punishing weapon for anyone who approaches blueshield and as for any other future concerns, this can be addressed by cracking down on people in general, I have been playing blueshield only as an extension of security WHEN the head of security is in immediate danger or when the head of staff is a pacifist( which is often the case in my timezone)

I can get *why* the change was made, but, when an antag shoves over a blueshield and steals their pistol, they suddenly have a melee weapon comparable to an energy sword in power, for no cost. They can't fire it without a mindshield, but, even that can be mitigated with enough prep, or a break-in into the armory (Which is now more feasible given their 22 damage melee weapon).

projectkepler-ru:
@mcmeiler  I'll reduce it a bit :v

modular_skyrat/modules/blueshield/code/special.dm
projectkepler-ru, line 4:
```suggestion
	force = 16
```

PR #143, opened 2023-03-30
Heliostation port

userisinvalid:
## About The Pull Request

Adds Heliostation from Fulpstation, with standard bubber stuff as a port, including but not limited to:
- NT rep office
- Blueshield office
- Cryopods
- Blacksmith's office
- ERT shuttle bay

## Changelog

:cl:
add: Ported Heliostation from Fulpstation
/:cl:


KathrinBailey (changes requested):
This has some definite issues that need to be addressed before merging.
Here are the ones I'm going to be hard on.
1. Heliostation_bubber.dmm needs removing, this is not an upstream map, so you don't need to have two versions of the map.
2. Five-tile wide hallways need narrowing down. They are colossal and are full of completely empty space.
3. Skyrat's dock for the CC ferry needs to be removed or built in better. It's ugly, look at how Deltastation does it.
4. Security needs some sort of rework, there's too many hallways and it's too spread out.
5. Some areas, notably atmos, looks like they won't be fully lit. I'll need you to go around the map and check all areas that should be have AI coverage and adequate lighting **before** a testmerge.
6. The docking area is colossal, too much empty space. It needs condensing as no one will be spending any time here.

**Keep this as a draft until it passes checks, there should be no pipenet or powernet issues. If you don't understand what the checks mean and how to find the issues they point out, ask for help and I'll give it.**

nevimer:
stale

PR #145, opened 2023-03-30
Fixes a bunch of issues + removes planting on ash

NullDagaf:
## About The Pull Request

­­> Fixes ash rune accidentally consuming ashwalkers when a sacrifice is already present
­­> Fixes ash rune consuming the entire stack instead of only one
­­> Fixes ash rune consuming equipment
­­> Fixes the "trill" emote so it can be called via "trill", and not "trills"
­­> Allows replanting fruiting cacti, restricts to one-plant-per-tile
­­> Aaand removes planting things on ash soil. Teshari B-Gone.

I'll think of a replacement for the dirt piles later.

## Changelog

:cl:
balance: removed planting on ash
fix: fixed ash runes being horribly broken
/:cl:


projectkepler-ru:
were these tested? @NullDagaf

NullDagaf:
> were these tested? @NullDagaf

Now they were, yes.

PR #146, opened 2023-03-30
Pixeltilting

NullDagaf:
## About The Pull Request

Didn't test yet. Wait approximately 15 minutes for the code to compile.
## Changelog

:cl:
add: adds pixeltilting
/:cl:


NullDagaf:
Okay, so my version of Bubberstation has broken keybindings, always. Neat.

KathrinBailey (changes requested):
needs specifically testing to see if crit/dead people can do this, ghosts and other mobs that aren't supposed to

This is just a direct copy and paste from Splurt, but with checks removed.

(edit: copy paste from splurt as in it's not been adapted to how skyrat does pixel-shifting I know you did the original thing for splurt)

PR #149, opened 2023-03-30
Makes glass floors removable

NullDagaf:
## Changelog

:cl:
fix: makes glass floors removable
/:cl:


NullDagaf:
Nevermind, this resolves into space and I don't particularly wanna change that.

PR #156, opened 2023-04-07
Ash Walker Ashtongue Fix

KathrinBailey:
## About The Pull Request
Hopefull fixes ashwalkers not having the language granted.
I've examined the code, and they **should** be given the language, but something is missing out somewhere along the chain, specifically for the ash walker species players picking the ghost role.

My methodology here is to go along the chain that involves spawning a player in as an ash-walker, and finding the last few links in that chain, in which I'll add a little link of my own to give the language. So, we're having the ghost role spawners grant the language.

## Changelog

:cl:
fix: Ashwalker species characters that join as the ghost-role should now start with Ashtongue, instead of just Draconic and Common.
/:cl:

### DIFF
--- a/code/modules/mob_spawn/ghost_roles/mining_roles.dm
+++ b/code/modules/mob_spawn/ghost_roles/mining_roles.dm
@@ -255,6 +255,7 @@
 
 	spawned_human.mind.add_antag_datum(/datum/antagonist/ashwalker, team)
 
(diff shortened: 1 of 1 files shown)


projectkepler-ru:
I checked it myself too but from what I can gather
I think it has to do with the character preference loading? more testing will be done

NullDagaf:
Character prefs (which I assume also handle languages) get loaded after the mob gets inited (duh), and more importantly in this case, after special() does. Did you test this works properly?

PR #163, opened 2023-04-13
Steals all of SPLURT's synthliz cosmetics

ReturnToZender:
Add: Literally all the Synthliz cosmetics

Needs testing to make sure their colors aren't all fucky, but I did my best to mimic the way Skyrat handles it.


nevimer:
you can reopen if you get it done

PR #170, opened 2023-04-15
Legalizes nuking from space

NullDagaf:
## About The Pull Request

So apparently if you detonate the nuclear device from outside of Delta's bridge, it doesn't count as "blowing up the station".
This fixes that.

## Changelog

:cl:
fix: legalizes nuking from space
/:cl:


Cyprex:
No, don't just remove the space check wholesale. I need to know more about what specifically the issues is(a screenshot would help) to advise you better.

Nicolai904:
dumb pr fix map instead of making it blow up anywhere

Serxule:
nukies will now plant on a single tile in the middle of space that nobody can reach or find

KathrinBailey:
Draft until you can explain how this won't be a balance issue.
It sounds like it is a problem, but this isn't the correct fix for it.

NullDagaf:
Nevermind, apparently this is svul. Closing.

> nukies will now plant on a single tile in the middle of space that nobody can reach or find

Also. Nuke pinpointers are a thing.

PR #171, opened 2023-04-15
Removes the Otie model

NullDagaf:
## About The Pull Request

Removes the Otie alt borg sprites because I don't like them.
dreammaker_G5Z1zfy3jl

## Changelog

:cl:
fix: removes otie
/:cl:

PR #172, opened 2023-04-15
Revert "Removes the Otie model"

nevimer:
Reverts Bubberstation/Bubberstation#171

Removing content like this, untested, as streamed in discord, because one person doesn't like a thing, 6 minutes after it is made, is pretty silly. That PR is untested and bound to break things. Removing content because someone doesn't like a thing is pretty awful, and isn't a pattern that should be followed or a justified reason for content removal.


SomeDongus:
image

nevimer:
That PR doesn't even follow any modularity guidelines, nor remove it cleanly. It's not a good PR and should not be merged if the code bases sustainability is concerned.

NullDagaf:
> That PR doesn't even follow any modularity guidelines, nor remove it cleanly. It's not a good PR and should not be merged if the code bases sustainability is concerned.

By all means, if you know a way to modularly remove sprites, do let me know how.

PR #174, opened 2023-04-15
Revert "Revert "Removes the Otie model""

NullDagaf:
Reverts #172

Removing content like that, untested, as streamed in discord, because one person doesn't like a thing, 6 minutes after it was made, is pretty based. That PR is untested and bound to break things. Removing content because someone doesn't like a thing is pretty based, and is a pattern that should be followed by all, and a justified reason for content removal.


Cyprex:
The only reason you ever provided was "I don't like this." That is not a valid reason. As such it should have never been merged in the first place & the revert is valid.

You are free to reopen a removal PR with a better reasoning, but we might well decline it anyway- I don't see any reason to remove the sprite.

PR #201, opened 2023-04-19
Fixes an issue with status indicators

nevimer:
## About The Pull Request

Hopefully prevents race conditions because it doesn't need to run so often. I can't recreate these issues locally and it will need admin VV help.


nevimer:
Busted. I will refactor these completely soon.

PR #206, opened 2023-04-22
Fixes disablers having a cell_type way too large (300 shots vs 20 like on TG)

ReturnToZender:
## About The Pull Request
Check the title.

## Changelog

:cl:
Fix: disabler having a cell_type too large
/:cl:


KathrinBailey (approved):
someone removed the special ammo type but forgot to change the cell I presume

PR #208, opened 2023-04-23
Test revert of #190 to try and solve the error-people.

KathrinBailey:
## About The Pull Request
It **might** be the status indicators causing the issue.
This is to be TMed only to test if it is or not.


nevimer:
You could also just untick the human.dm file that adds the component, making nobody have the component to test. This issue cannot be reproduced locally for me, nor can the stuck indicators. Might be a linux thing?? 

with nobody getting the component by default, you can try adding it in t-dome

NullDagaf:
Disregarding for a moment this is probably the worst approach to disabling something.
The issue is primarily in how the status indicator feature is implemented. A subtype of the FoV plane master (???) is used to render the status effects.. for.. some reason. Which, understandably, causes a myriad of odd behavior. The proper solution is re-doing how said feature is implemented, not ticking off a bunch of files.
> with nobody getting the component by default, you can try adding it in t-dome
The issue is the plane master acting aloof, not the component itself. If it was the component's fault, you wouldn't have the error displayed even as an observer/lobby player.

CenkTheRaptor:
as far as i am aware the reason this happens is because of overlays, overlays have a limit and if you go over it it makes you into an error permanently

nevimer:
> Disregarding for a moment this is probably the worst approach to disabling something. The issue is primarily in how the status indicator feature is implemented. A subtype of the FoV plane master (???) is used to render the status effects.. for.. some reason. Which, understandably, causes a myriad of odd behavior. The proper solution is re-doing how said feature is implemented, not ticking off a bunch of files.
> 
> > with nobody getting the component by default, you can try adding it in t-dome
> 
> The issue is the plane master acting aloof, not the component itself. If it was the component's fault, you wouldn't have the error displayed even as an observer/lobby player.

It was only a subtype of that because if I tried making a dedicated plane for it, it would just not work, and create that TGstation logo ever since the plane cube update. Let's not disregard that update being responsible for all sorts of other ongoing render issues to this day.

nevimer:
> 
> > with nobody getting the component by default, you can try adding it in t-dome
> 
> The issue is the plane master acting aloof, not the component itself. If it was the component's fault, you wouldn't have the error displayed even as an observer/lobby player.

There is nothing wrong with it, you are right, but it can't be tested at all this way, because you can't even manually add the component now.

KathrinBailey:
> You could also just untick the living.dm file that adds the component, making nobody have the component to test. This issue cannot be reproduced locally for me, nor can the stuck indicators. Might be a linux thing??

This was originally hastily done before roundend because the error overlays have been a bit much for a while.

> breaking the game, because the authors failed to understand WHY the planes were used the way they are. You can literally untick your ability to see one of the games master planes on the world now, because of this, on master, because someone repathed the plane to not be a subtype. Why??

I take the L on the middle bit but breaking the game? The error overlays had been there for days prior.
Reason it was repathed to not be a subtype was because status indicators should **never** have been merged in the first place, it didn't pass CI and it caused every PR after it to fail CI.

CI fails because the indicators were on the same plane as something else, why you can't have a subtype of something sharing the same plane I have no idea, because I'm not an expert in planes, planemasters and layers. Obviously these planes were not supposed to be shared, otherwise there wouldn't be unit tests to fail checks if they do get shared. I'm guessing the red error signs are because of this.

Status indicators being merged when it wasn't ready yet wasn't your fault of course

Issue #114, opened 2023-03-18
Meta Station lacks a fax machine at the bridge


ronfino:
Title says it all.

PR #210, opened 2023-04-25
Adds a fax machine to the MetaStation Bridge

tolliamlew:
Does what it says on the tin. there's now a fax machine on the bridge, as requested in issue #114 I might be a little late on this one? idk which map actually gets used, the metastation one or metastation bubbers, but either way there's a fax there now!

## About The Pull Request

All i did was add a fax machine to the bridge on the MetaStation Map. I ran my test server and everything to make sure it was set up correctly and everything seems fine.

This one's probably a little late, but I did it anyway just to get a feel for using the mapping tools.


KathrinBailey (changes requested):
needs to be metastation bubber

PR #221, opened 2023-05-07
Adds a BlueShield office to Icebox

tolliamlew:
I've heard a few people complaining about the lack of blueshield offices on the maps, so I decided to start fixing it.

## About The Pull Request
 This adds a blueshield office just south of the NT-Rep's office, in that same little hallway. There is a small attached bedroom, and the windows have working shutters, like the other outward facing offices. This also means that icebox now has a spawn for the blueshield's locker, which is always handy.


KathrinBailey:
image
Hiya, you have some ATs as well as double-stacked pipes.
Contact me if you need help with these!

PR #230, opened 2023-05-16
Borer Changes

Icarus-The-Sun:
## About The Pull Request

Removes the tier 1 ability for all borers to lay eggs, restricting normal egg production to Hivelords. Adds a new ability accessible by Diveworms and Symbiotes at tier 6 with a double cooldown and chemical cost (AND NOW A 5 POINT CHEMICAL EVOLUTION POINT COST) that allows them to lay eggs. Alters the Hivelord path to be BLOOD-SPEED-EGGS rather than EGGS-BLOOD-SPEED. Tested with each path to ensure the correct abilities were locked and available depending on prior paths.

## Changelog

:cl:
add: Adds a new T6 borer ability allowing egg-laying for Symbiotes and Diveworms at a higher cost and cooldown.
balance: Makes egg production a T3 path for Hivelords, forcing borers to commit to a path while limiting their ability to complete objectives on their own. Adds Learning from Blood as Hivelord's T1 ability.
/:cl:


destrucktoid:
I would suggest that the symbiote egg laying not cause damage to the host or at the very least not cause damage that the symbiote cannot fix with mannitol/neurine. In exchange i would propose requiring chemical points to lay an egg instead. As a symbiote you tend to have too many chemical points once you've reached t6 anyway so adding that as a chemical point dump would be good IMO.

Icarus-The-Sun:
Broken as of now, due to update.

PR #234, opened 2023-05-20
Adds Barkbox and collars [WIP]

ReturnToZender:
## About The Pull Request
See title. 
Issues: the new collars cannot be colored, and do not have a worn sprite, even when told to use the overlay as worn, and despite being polychromic. I side-stepped the issue of GAGS incompatibility entirely by simply making the greyscale sprite the real sprite. 
Everything else functions perfectly. Original pet collar can be worn.

## Changelog

:cl:
Add: the barkbox vendor
Add: Barkbox vendor as random
Add: Dog treats
Fix: Pet collars are wearable again
/:cl:


nevimer:
Conflicts! I want this, please fix?

modular_zubbers/code/modules/clothing/neck/_neck.dm
```diff
@@ -0,0 +1,11 @@
+/obj/item/clothing/neck/petcollar/mob_can_equip(mob/M, slot, disable_warning = FALSE, bypass_equip_delay_self = FALSE, ignore_equipped = FALSE)
+	return TRUE
```
Cyprex, line 2: No, remove this proc entirely. This needs to just be a parent call, because its very important that`/obj/item/proc/mob_can_equip` and so on are still called!

Cyprex (changes requested):

PR #236, opened 2023-05-21
Actually gets Bubberstation up to Sync with the Rodent Server

nevimer:
## About The Pull Request

Sync's the codebase with SKYRAT. **DO NOT SQUASH MERGE THIS. DO NOT. YOU NEED THIS COMMIT HISTORY.**
ALSO KEEPS THE BUBBERSTATION EDITS AND CONTRIBUTIONS

**THIS WILL BREAK DONOR ITEMS BECAUSE THEY SHARED THE SAME FILES AS SKYRATS DONOR ITEMS**
I will not fix this, please separate them in a follow up PR. 

Document your edits. Don't edit UPSTREAM FILES if you can avoid it, make new paths. OVERRIDES PLEASE.
**DOCUMENT YOUR EDITS.** ALL OF THIS COULD BE DONE MODULARLY. 

**Please enforce contribution guidelines, and do not circumvent them. This was PAINFUL. **

## Changelog

:cl:
code:sync's the server with the rodent server. 
/:cl:


nevimer:
https://github.com/Skyrat-SS13/Skyrat-tg/actions/runs/5035486788/jobs/9031039692

Skyrat themselves have the same CI issues. So, these CI issues will have to be fixed upstream.

Cyprex:
We can fix CI here, I can look into it. More importantly how did you create this? Manually move changes over? I was planning to rebase our branch. We were getting thousands of merge conflicts in files we never edited, presumably because I reverted an upstream update at one point. I was going to rebase to remove the revert & the commits that introduced the update, which would hopefully allows us to `git merge` without all these bogus merge conflicts

nevimer:
> We can fix CI here, I can look into it. More importantly how did you create this? Manually move changes over? I was planning to rebase our branch. We were getting thousands of merge conflicts in files we never edited, presumably because I reverted an upstream update at one point. I was going to rebase to remove the revert & the commits that introduced the update, which would hopefully allows us to `git merge` without all these bogus merge conflicts.
> 
> Doing it this way has the disadvantage that it rewrites history and thus breaks open PRs & forks of contributors. That said we NEED to be able to painlessly merge Skyrat updates again in the future.

I took a fresh skyrat, and merged bubberstation on top of it. This will preserve the timeline (the way I merged it keeps the commit metadata,  whereas a pull or a squash would not), which created the issue where the revert upstream PR created a few conflicts (it post-dates the line edits), which itself was resolved by reverting that PR aswell. This **should** resolve any further issues. I picked and compared each differing file by hand, totaling about five hours of work. I had local CI running and a VPS running CI aswell, to help me find issues as I corrected them to save time. 

As long as this is not squash merged, all the commit metadata from skyrat should be preserved, minimizing any change of conflicts in the future.

Cyprex:
Let's go

PR #241, opened 2023-05-25
[MODULAR] Progtot goals override

nevimer:
## About The Pull Request
MERGE AFTER THE UPSTREAM UPDATE - This is based off that branch!

Removes
VSCodium_Yufk9WyxDq

## Changelog

:cl:
balance: some traitor goals will not be spawning. 
/:cl:


nevimer:
Note, this adds _modular_zubbers folder to OVER-RIDE skyrat AND TG code before supercalls.

Higgin:
With the possible exception of the comms console hack since it throws in a lot of extra stuff on top of what dynamic is already tuned to do (with pirates/fugitives/sleepers coming in on top of an already-successful traitor short of them getting to their final objective - it's a bit of a death spiral for the crew at that point) why remove these? They're good.

Serxule:
no, not only does it not even make sense, assassinate is a 1 time kill (not a greentext assassination), eyesnatching just isnt a problem (and likely wont be done), hacking comms console is a non-problem (it only increases threat, so high chance of nothing even happening), and removing final obj entirely is a shitty way to deal with certain final objectives

Higgin:
> no, not only does it not even make sense, assassinate is a 1 time kill (not a greentext assassination), eyesnatching just isnt a problem (and likely wont be done), hacking comms console is a non-problem (it only increases threat, so high chance of nothing even happening), and removing final obj entirely is a shitty way to deal with certain final objectives

The comms console hack does try to spawn pirates, fugitives, sleepers, or a random unfavorable event (which can include meteors, rod, and portal storm by default) but the minimum #s on the extra spawns could be raised to account for that (or those additions just ruled out completely in favor of the unfavorable events and fugitives.) Pirates and sleepers among those are the biggest threats that might really lock the station into a negative feedback loop. 

But final objectives are meant to be hammers, I think any of them should stay on the table - even Starfury and the dark matteor.

kiwedespars:
honestly i think keeping all the objectives in is a better idea because it gives rounds more variety. more chaos for "what type of server we are" but also more variety for the primary server antagonist (i have never seen changeling roll naturally and i have only seen heretic a handful of times)

if a traitor has reached their final objective, it's a rarity to see, even now with it enabled, it'll probably be an hour in or beyond that, and the round is close to ending naturally anyways if they're being that chaotic without anyone stopping them.

security should be able to deal with them before it gets to that point, or an ert gets called if security can't handle it, because that's what usually happens. more depth for traitor is kinda epic and having more antag variety from it (space dragon, nukies, romerol zombies, malf, etc.) is nice

mcmeiler:
Let's remove all possible antagonistic actions antags can do against players. Remove them entire while you're at it.

No but seriously. This just turns the server into a hugbox where traitors can't do anything to other players and stealth all round long, never doing anything interesting.

Fragenstien:
Final objectives were removed because the gameplay loop was as follows:
>antag power games with easy objectives for 45 minutes
>antag ends round and interrupts literally everything everyone was doing

We are not that kind of server. If you want final objective back fix those two problems.

BurgerLUA:
There seems to be a metric fuckton of changes (5000+!) in this PR. What happened?

PR #242, opened 2023-05-28
[NON-MODULAR] Mostly reverts Skyrat grab/punch changes.

Higgin:
## About The Pull Request

Reverts some of the changes to grabs and punches made by Skyrat's combat reworks over the last two years.

For **grabs:**
Resisting an aggro grab on tg has a base chance of 60% to escape.

_Previously,_ being prone (for any reason) would apply two penalties to reduce this to 20%. Having stamina damage of over 80 (roughly half your total before stamcrit at 150) or being a pushover would make this 15% (for either one of the aforementioned effects) and then 12% for both. Oversized grabbers would reduce it further to 10%. Every failed resist roll at being grabbed imposes a four-second movement lock and deals 10-15 stamina damage.

_Now,_ being prone alone **does not decrease the chance of escape.** Resting (deliberately laying down) now imposes one penalty (taking you to 60% from 30% and punishing people who deliberately use crawl combat to try to avoid getting knocked down/slipped.) The stamina damage penalty for failing to escape a grab has been increased back to the old tg values of 15-20 per failure. So, if you consistently fail - it can still get you in trouble, but it's less likely to do so in a timeframe allowing somebody to cuff you or otherwise murder you without first having some reasonable chance to escape.

On Skyrat, the old grab mechanics made chain-shoves into aggro-grabs a much-too-reliable fight-ender off of a single slip. Returning towards tg balance means that you should, if you intend to really lock someone down without having cuffs, go for a neck grab. It also weakens quickly aggro-grabbing people out of a fight to try to disarm them, at which point (if they failed the roll previously) they might be entirely movement-locked while standing. It's a lot less goofy and fight-ending with these values.

For **punches:**

They dealt 2.6 of their brute damage value as stamina damage. The tg default value is 1.5. Punches really shouldn't be better than batons, disablers, or any other weapon for a nonlethal takedown.

## Changelog

:cl:
balance: Grab escape chances are now higher and don't rise precipitously on becoming floored.
balance: Punch/kick stamina damage is now lower, making unarmed fighting much less immediately decisive.
/:cl:


code/__DEFINES/~skyrat_defines/combat.dm
```diff
@@ -1,4 +1,22 @@
-#define PUNCH_STAMINA_MULTIPLIER 2.6
+#define TRAIT_ALREADYSTAMINAFLOORED    	"alreadystaminafloored" //Trait to keep track whether someone was floored recently by stamina damage.
+#define STAMINA_KNOCKDOWN_COOLDOWN 10 SECONDS
+
+#define MAX_HUMAN_LIFE 135
```
nevimer, line 4: Something is sussy about this entire files edits.

code/modules/mob/living/living.dm
```diff
@@ -1117,12 +1117,14 @@
 	. = TRUE
 	if(pulledby.grab_state || body_position == LYING_DOWN || HAS_TRAIT(src, TRAIT_GRABWEAKNESS) || staminaloss > STAMINA_THRESHOLD_HARD_RESIST) //SKYRAT EDIT CHANGE: if(pulledby.grab_state || resting || HAS_TRAIT(src, TRAIT_GRABWEAKNESS))
 		var/altered_grab_state = pulledby.grab_state
-		if(body_position == LYING_DOWN || HAS_TRAIT(src, TRAIT_GRABWEAKNESS) && pulledby.grab_state < GRAB_KILL) //If prone, resisting out of a grab is equivalent to 1 grab state higher. won't make the grab state exceed the normal max, however - SKYRAT EDIT CHANGE: if((resting || HAS_TRAIT(src, TRAIT_GRABWEAKNESS)) && pulledby.grab_state < GRAB_KILL) //If resting, resisting out of a grab is equivalent to 1 grab state higher. won't make the grab state exceed the normal max, however
+		if(resting || HAS_TRAIT(src, TRAIT_GRABWEAKNESS) && pulledby.grab_state < GRAB_KILL) //BUBBER EDIT CHANGE: if(body_position == LYING_DOWN || HAS_TRAIT(src, TRAIT_GRABWEAKNESS) && pulledby.grab_state < GRAB_KILL) //If prone, resisting out of a grab is equivalent to 1 grab state higher. won't make the grab state exceed the normal max, however
```
nevimer, line 1120: I'm so confused

Higgin: Resting is deliberately proning as with the rest verb, body_position also applies if you're just knocked down.

code/__DEFINES/~skyrat_defines/combat.dm
Higgin: There was a cleanup that took place that removed a lot of the old defines between the original PR and the upstream merge, I missed it.

Issue #252, opened 2023-06-06
Spines on lizards dont wag with them


Sweettoothart:
Reporting client version: 514.1589

## Round ID:
1373

## Testmerges:
- [Test revert of #190 to try and solve the error-people.](https://github.com/Bubberstation/Bubberstation/pull/208)
- [[MODULAR] Modularizes #134](https://github.com/Bubberstation/Bubberstation/pull/237)
- [Virus Cost Changes](https://github.com/Bubberstation/Bubberstation/pull/218)
- [Adds new donor items, a new public hairstyle, and private donor items for Thedragmeme](https://github.com/Bubberstation/Bubberstation/pull/229)
- [[SEMI-MODULAR] Allows observers the ability to activate pAI cards on their own](https://github.com/Bubberstation/Bubberstation/pull/217)
- [Fixes cargo infinite money exploit](https://github.com/Bubberstation/Bubberstation/pull/209)
- [Hierophant appreciation sprite update.](https://github.com/Bubberstation/Bubberstation/pull/239)
- [[NON-MODULAR] Radioblock revert](https://github.com/Bubberstation/Bubberstation/pull/243)

nevimer:
What tail specifically?

Majkl-J:
> What tail specifically?

Not a tail issue, spines fail to animate with the tail

magatsuchi:
![image](https://github.com/Bubberstation/Bubberstation/assets/88991542/fe15694f-787e-4e6a-9eca-da1c86f60e85)

(please post this issue to either Skyrat-SS13/Skyrat-tg or tgstation/tgstation)

PR #273, opened 2023-06-15
Attempts to band aid spine wagging

Majkl-J:
## About The Pull Request

closes: #252

### Quick rundown of what is broken
For some reason, someone somewhere deleted the Insert proc call for lizard tails. I have looked through about 50 different files, ran through the last 3 months of PRs that touched organ/customization code, and found nothing out of place, be it for my inexperience or influence of a higher power.

Thus I have put together this QoL quickfix: Spines are now checked again when trying to wag, which works.

(Please someone who knows more about this try to point me in the right direction so I don't have to do this)

- [x] Searched for the error
- [x] Found the issue
- [x] Cried a lot trying to fix it
- [x] Tested the fix
- [ ] Make the spines not disappear

## Changelog
:cl:
fix: Lizard spines no longer disconnect from the rest of the body upon wagging
/:cl:

### DIFF
--- a/code/modules/surgery/organs/external/tails.dm
+++ b/code/modules/surgery/organs/external/tails.dm
@@ -117,11 +117,14 @@
 	///A reference to the paired_spines, since for some fucking reason tail spines are tied to the spines themselves.
 	var/obj/item/organ/external/spines/paired_spines
 
-/obj/item/organ/external/tail/lizard/Insert(mob/living/carbon/reciever, special, drop_if_replaced)
+//BUBBER FIX START
+//This edit atempts to bandaid the spine wagging issue and should be overriden the moment someone else figures it out
+/obj/item/organ/external/tail/lizard/Insert(mob/living/carbon/receiver, special, drop_if_replaced)
 	. = ..()
-	if(.)
-		paired_spines = ownerlimb.owner.get_organ_slot(ORGAN_SLOT_EXTERNAL_SPINES)
-		paired_spines?.paired_tail = src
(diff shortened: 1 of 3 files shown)


code/modules/surgery/organs/external/tails.dm
```diff
@@ -117,11 +117,14 @@
...
 
-/obj/item/organ/external/tail/lizard/Insert(mob/living/carbon/reciever, special, drop_if_replaced)
+//BUBBER FIX START
+//This edit atempts to bandaid the spine wagging issue and should be overriden the moment someone else figures it out
+/obj/item/organ/external/tail/lizard/Insert(mob/living/carbon/receiver, special, drop_if_replaced)
 	. = ..()
-	if(.)
-		paired_spines = ownerlimb.owner.get_organ_slot(ORGAN_SLOT_EXTERNAL_SPINES)
-		paired_spines?.paired_tail = src
+	paired_spines = ownerlimb.owner.get_organ_slot(ORGAN_SLOT_EXTERNAL_SPINES)
```
Majkl-J, line 124: The animation on these is still broken but I didn't notice (Spines disappear now), Soothes thankfully brought it up.

Will fix asap

Majkl-J: Mind you this is still better than them not animating

SilvernKell:
Bless you

code/modules/surgery/organs/external/tails.dm
Majkl-J: Ok I looked through this. The `build_icon_state()` now works correctly, updating the icon state... but something still makes the spines disappear????

nevimer:
it does not really work right... might be fixed by an upstream merge

PR #283, opened 2023-06-21
Removes Hand Teleporter Portal Dispelling

BurgerLUA:
## About The Pull Request

You can no longer dispel portals with the Hand Teleporter.

# Why this is good for the game

While hand teleporters should be strong, I feel that giving the ability to immediately close portals after you traveled through them is perhaps too strong for the item in question. Currently, there is nothing stopping a Head of Staff or an antagonist from taking the hand teleporter round start and then using it to teleport away constantly to evade threats.

## Changelog

:cl: BurgerBB
balance: The hand teleporter can no longer dispel portals.
/:cl:

### DIFF
--- a/code/game/objects/items/teleportation.dm
+++ b/code/game/objects/items/teleportation.dm
@@ -134,6 +134,7 @@
 	. = ..()
 	active_portal_pairs = list()
 
+/* Bubberstation change: Removes portal dispelling.
 /obj/item/hand_tele/pre_attack(atom/target, mob/user, params)
 	if(try_dispel_portal(target, user))
 		return TRUE
@@ -149,6 +150,7 @@
 /obj/item/hand_tele/afterattack(atom/target, mob/user)
 	try_dispel_portal(target, user)
 	. = ..()
+*/
 
 /obj/item/hand_tele/pre_attack_secondary(atom/target, mob/user, proximity_flag, click_parameters)
 	var/portal_location = last_portal_location


kiwedespars:
citmain change:
dispel but on a 2-3 second action bar

Draggeru:
Roughly half of command mains dont know there's a way to dispel the hand tele portal in the first place, and if they do oh my God effort if you mess up attacking them in the first place?

This feels really weird and like you don't really know how the hand tele works in practice in the first place, you generally need to have the hand tele set, and then actively pull it out while possibly being attacked. Dispelling the portal isn't an instantaneous act either, it takes human input to dispell it. 

>antagonist from taking the hand teleporter round start and then using it to teleport away constantly to evade threats.

If you want to do this, please nerf the traitor hand tele that they can buy which is functionally much easier to abuse and vastly more instantaneous than the regular hand tele.

BurgerLUA:
> Roughly half of command mains dont know there's a way to dispel the hand tele portal in the first place, and if they do oh my God effort if you mess up attacking them in the first place?
> 
> This feels really weird and like you don't really know how the hand tele works in practice in the first place, you generally need to have the hand tele set, and then actively pull it out while possibly being attacked. Dispelling the portal isn't an instantaneous act either, it takes human input to dispell it.
> 
> > antagonist from taking the hand teleporter round start and then using it to teleport away constantly to evade threats.
> 
> If you want to do this, please nerf the traitor hand tele that they can buy which is functionally much easier to abuse and vastly more instantaneous than the regular hand tele.

That's a next PR.

Draggeru:
I feel like the EMP being able to disable the hand tele to some degree would be better than removing the reward players are given for learning how to operate the hand tele effectively. We need to remember only two central hand-teles exist period, nerfing it into the ground is a bad idea. On the flip side the syndicate version can be bought indefinitely as long as you have the TC for it. On top of that we are an MRP server, so we should reward mechanical skill as long as roleplay isn't being harmed.

TLDR: Being able to disable the hand tele to some degree is a better alternative

Higgin:
> citmain change: dispel but on a 2-3 second action bar

if anything this, but this change is generally unneeded, we don't have people doing constant telejukes here.

PR #285, opened 2023-06-22
22 june 2023 upstream merge

nevimer:
Merging after CI pass.


Cyprex:
This broke META CI. Such breakage should be avoided, and if not possible at least avoided.

nevimer:
Impossible to avoid. Our maps and spider webs are 1:1 right now, so it will be an upstream issue.

PR #291, opened 2023-06-25
Removes CENTCOM channel from NTRep, Blueshield

ReturnToZender:
## About The Pull Request
Comments out CENTCOM access from the NTRep and Blueshield headsets. They serve little purpose, the CENTCOM frequency cannot go down (although it's minimally useful because only these two roles get the channel) and it's pretty weird for the Blueshield to hear Deathsquad chatter.

## Changelog

:cl:
del: headset_cent NTRep
del: headset_cent Blueshield
/:cl:


Draggeru:
I feel like it may be a good idea to give them an alternative radio channel that can go down, having the rep and blueshield be able to communicate any CC orders outside of command or security channels is a roleplay benefit. Especially if the captain or hos needs to be demoted or private orders need to be given

Higgin:
It might honestly be better to give Deathsquads one of their own if that's the primary pain-point. When inspectors or ERTs show up, they have the centcom channel, and they're often there because of something the NTR did - it makes sense for them to be able to talk directly.

ReturnToZender:
> It might honestly be better to give Deathsquads one of their own if that's the primary pain-point. When inspectors or ERTs show up, they have the centcom channel, and they're often there because of something the NTR did - it makes sense for them to be able to talk directly.

Such as through the Command channel or PDAs, both of which are afforded to ERTs and command officers, respectively?

BurgerLUA (changes requested):
You should just remove the lines outright instead of commenting out the changes.

BurgerLUA (approved):
Nevermind, forgot this is a modular coderbase xd

projectkepler-ru:
This seems unnecessary

PR #297, opened 2023-06-26
Buffids

Cyprex:
## About The Pull Request

Regular IDs get more wildcard slots than silver IDs
Silver IDs get infinite slots, making them identical to gold IDs.

Happy to discuss these values and hear what you think!

## Changelog

:cl:
balance: ID cards can hold a lot more accesses now: Silver IDs can hold anything like Gold IDs, and regular IDs have a lot more wildcard slots
/:cl:

### DIFF
--- a/code/__DEFINES/id_cards.dm
+++ b/code/__DEFINES/id_cards.dm
@@ -11,13 +11,13 @@
  */
 
 /// Wildcard slot define for basic grey cards. Only hold 2 common wildcards.
-#define WILDCARD_LIMIT_GREY list(WILDCARD_NAME_COMMON = list(limit = 2, usage = list()))
-/// Wildcard slot define for Head of Staff silver cards. Can hold 3 common, 1 command and 1 private command.
-#define WILDCARD_LIMIT_SILVER list( \
-	WILDCARD_NAME_COMMON = list(limit = 3, usage = list()), \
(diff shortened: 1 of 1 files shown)


Serxule:
i dont think silver IDs should be the same as gold (might aswell revert to old IDs) but more access for normal IDs is great, also could you make agent ID unlimited slots? its pretty useless as is

Fragenstien:
Silver ids shouldn't have access to captain only areas, but I'd be fine with that being the only difference between them and gold

nevimer:
Made off a bad branch

zeroisthebiggay:
i don't know if making all access so easily available when on most maps it's trivial to steal a silver id is such a good idea. especially if you want to avoid heads of staff giving themselves aa to powergame as soon as they get their hands on the captain's spare id before putting it back into its gigahugbox safe

Cyprex:
> i don't know if making all access so easily available when on most maps it's trivial to steal a silver id is such a good idea. especially if you want to avoid heads of staff giving themselves aa to powergame as soon as they get their hands on the captain's spare id before putting it back into its gigahugbox safe
What would you think would be appropriate instead?

zeroisthebiggay:
> What would you think would be appropriate instead?

nothing because id balance is fine as is really
if you really want to buff them then give silver and regular ids more wildcard slots and also let agent ids have all access because holy shit the access stealing part of agent ids has been made so bad

PR #299, opened 2023-06-27
Ticks Charlie Station

ReturnToZender:
## Changelog

:cl:
Add: Charlie Station as a ghost role spawn
/:cl:


ReturnToZender:
Looks like all of the space ruins are unticked, but they clearly still spawn, this is likely not the solution

PR #300, opened 2023-06-28
Disables the Icemoon bases

ReturnToZender:
## About The Pull Request
Temporarily disables Icebox spawns. They're spawning right next to the Skyrat mining base and random miners end up getting shot by the turrets. A similar issue occurs with Ashwalkers.

## Changelog

:cl:
Del: Skyrat's Icemoon Syndie Base
Del: Skyrat's Icewalker Camp
/:cl:


ReturnToZender:
Gonna work on fixing these issues after doing this PR, but I'm not sure how

zeroisthebiggay:
> Gonna work on fixing these issues after doing this PR, but I'm not sure how

you can do the brute force mapping route like how it is on lavaland and forcefully make an area around the mining base physically unable to spawn ruins but that'd be way better off being merged on tg upstream i assume since modular mapping to that scale is generally unfeasible i think

PR #308, opened 2023-07-01
[SEMI-MODULAR] Progtot Final Objective Tweaks, Singularity Fixes

Higgin:
## About The Pull Request

Alternative to #241 

Hard-locks final objectives from appearing before 2hrs30. ~~We have transfer votes at 2hrs25.~~

Raises progression requirements for final objectives (both earned and overall.) 

Fixes our singularity sprites being messed up for both regular singularities and the dark matter singulo.

Also removes eyesnatching.

Makes time-dependent rewards and penalties less dramatic.

Why? Final objectives are cool and give progtots somewhere to go. While we concede that we don't want to hand any one person the power to end a round early, most of our roundend votes pass on the first go-'round, and the amount of work and disruption necessary to get to the final objectives should now ensure that if the crew votes to stick around, they are leaning into the conflict - or that if they vote to get out, the traitor will have to have been very active and have a very limited window to inflict their disruption (making it a sort of 'run for the exits' scenario at a point when everyone already decided to leave.)

This is also proposed as a return of the assassinate, comms console, and eye-snatching objectives overridden in #241 - they're not really any more vicious than the current objectives, and they add extra legs/more interesting threat to what traitors do than killing pets or stealing things. The comms console is generally going to only add unfortunate events - which themselves are currently otherwise very lacking with how our events have been balanced.

## Changelog

:cl:
balance: Progtot final objectives now cannot appear before 2.5 hours into a round and require additional progression.
balance: Progtot rewards are now more heavily reduced by ahead-of-time penalties.
del: No more eyesnatching objectives for progtots.
fix: Singularity sprites at larger sizes, including the dark matter singularity, should work again.
/:cl:


Fragenstien:
Worth a try, though I don't think we should add back in eye snatching

Higgin:
> Worth a try, though I don't think we should add back in eye snatching

fair nuff, can take it or leave it - should be gone now.

zeroisthebiggay:
> We have transfer votes at 2hrs25.

this is factually wrong
transfer votes are at two hours thirty

all this pr does is make traitors able to do final objectives after the end round vote is already called

Higgin:
> > We have transfer votes at 2hrs25.
> 
> this is factually wrong transfer votes are at two hours thirty
> 
> all this pr does is make traitors able to do final objectives after the end round vote is already called

yeah i'm not sure why i put 2:25 - the vote wraps up at 2:30. 

Similar effect, but a pretty big change. I'm going to see about putting this up to a vote.

PR #312, opened 2023-07-02
(NON-MODULAR) Salt PR

Kilmented:
## About The Pull Request
Adds a heft penalty to the Deagle's duel wielding ability, it's a 60 damage gun, if you had two you could essentially one click everyone without armor and anyone with armor wouldn't post much of a threat either. That's like having two Mosins, one for each hand.

Also, fixes a gripe I have had with cigarettes for the LONGEST time. 

## Testing
I have personally tested the cigar(ette) changes, they work as intended. Even the healing the lungs bit. 
I have not personally tested the Deagle changes, but I see no reason why it shouldn't work.

### DIFF
--- a/code/game/objects/items/cigs_lighters.dm
+++ b/code/game/objects/items/cigs_lighters.dm
@@ -166,7 +166,7 @@ CIGARETTE PACKETS ARE IN FANCY.DM
 	/// Should we smoke all of the chems in the cig before it runs out. Splits each puff to take a portion of the overall chems so by the end you'll always have consumed all of the chems inside.
 	var/smoke_all = FALSE
 	/// How much damage this deals to the lungs per drag.
-	var/lung_harm = 1
+	var/lung_harm = 0.1 // Bubberstation edit, makes cigs not give you stage four lung cancer after two. You'd need to smoke ten times the amount of cigarettes as before in order to get the same amount of damage.
 	/// If, when glorf'd, we will choke on this cig forever
 	var/choke_forever = FALSE
 	/// When choking, what is the maximum amount of time we COULD choke for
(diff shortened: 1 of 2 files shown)


yourdoom9898:
Cigarette changes are pretty cook. TIL that the cigarettes produced by the industrial black extract have Krokidil in them.

Serxule:
i dont see why krokodil cigs would heal lungs ngl

Serxule:
> Who ARE you?? You're not in the discord or anything... image

cause im not in the discord, and im taking a break from the serb

yourdoom9898:
> i dont see why krokodil cigs would heal lungs ngl

They're already healing with the regen jelly in them, and are inherently "slime magic shit" with a crossbred core. It's easily one of xenobiology's many "just don't think about it" moments.

Serxule:
> > i dont see why krokodil cigs would heal lungs ngl
> 
> They're already healing with the regen jelly in them, and are inherently "slime magic shit" with a crossbred core. It's easily one of xenobiology's many "just don't think about it" moments.

oh that makes more sense, thought it was only krokodil

Zirok-BYOND:
>60 damage
Wew, but upon checking the .357 has the same. Feel free to nerf any aspect of this weapon, especially if it's less accessible than the .357 which
a) is very loud
b) Has very limited ammo (Until you get to an autolathe, anyways, but even then the magazine capacity sucks)
c) Can only be obtained by traitors.

Nuke ops weapons are a separate thing and need to be very, very powerful since nukies are expected to be massively outnumbered

Nicolai904:
make the prs split tbh making a salt pr with some other change is not that great also i would not nerf the deagle its a weapon that has no way of getting ammo back and mostly admin only unless you are lucky and get it from the safe in the vault which gives you singular deagle

Zirok-BYOND:
Do these modular please

PR #314, opened 2023-07-04
[Modular] Makes all techwebs generate points

ReturnToZender:
## About The Pull Request
Makes it so that all new techwebs (including the station techweb) generate points by DEFAULT. This is to make techwebs compatible with ghost roles making new servers.
Coupled with turning on NO_DEFAULT_TECHWEB_LINK, this will solve ghost roles being connected to the station research network by default, while still allowing them to research.

## Changelog

:cl:
Add: Techwebs generate points by default.
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/research/techweb/_techweb.dm
+++ b/modular_zubbers/code/modules/research/techweb/_techweb.dm
@@ -1,0 +1,2 @@
+/datum/techweb
+	var/should_generate_points = TRUE


Cyprex:
I think this may introduce issues- rather than flip the default this should only be enabled for a separate interdyne techweb.

modular_zubbers/code/modules/research/techweb/_techweb.dm
```diff
@@ -0,0 +1,2 @@
+/datum/techweb
+	var/should_generate_points = TRUE
```
BurgerLUA, line 2: Shouldn't this be just ``should_generate_points`` as the var is already defined elsewhere?

PR #319, opened 2023-07-06
[NON-MODULAR] Buffs Burn Wounds

Higgin:
## About The Pull Request

Makes burns less anemic (haha) by making them slightly easier to inflict, remove a small, flat amount of blood when they inflict and making the first tier of burn wounds still require treatment (they take an incredibly long time to proceed to sepsis.)

## Changelog

:cl:
balance: Burn wounds remove a small flat amount of blood when they inflict, are slightly easier to inflict, and can proceed to develop infections at all tiers.
/:cl:


Higgin:
> i disagree with the idea of moving `more` towards a wounds centric metagame because i'm of the belief that it should be the opposite
> 
> this change doesn't really make lasers compete with ballistics in terms of sheer lethality so long as this `small flat bloodloss` rate is as small and flat as you say it is
> 
> really the worst i can see this change doing is making people that die of fire more bogus to revive because they'll have no blood in addition to being horribly infected and scorched

it's not meant to make them a direct competitor, but it is meant to make them require some attention. It takes about 6 burn wounds or upgrades of any stripe to get to the point of any serious danger from blood loss.

It takes about half an hour leaving a t1 burn wound entirely untended (assuming it doesn't just clear up on its own, which they can) to get to sepsis.

When a t1 slash or puncture can very easily end a run left untreated, assuming you don't very quickly pop your one epipen and have another, it becomes pretty obvious that burns sucked by comparison. With a lot of our massively slippery single antags, at least for sec, this has made ballistics the meta - because most things that will heal or no-sell the flat damage from a ballistic still won't address bloodloss if you're continually dipping in and out of fights.

Lasers have the benefits of recharging and piercing windows. They have the downsides of EMP vulnerability, needing to recharge at static locations (may be revisiting this with MCRs later,) and the relative shittiness of burn wounds.

It shouldn't be decisive in most cases, but it should help address the areas where people otherwise ignore burn wounds as though they don't exist, but bleeds mean everything.

I personally think wounds are a good thing. They make weapon choice more interesting and demand some counterplay or cooldown on fights instead of mindlessly HoTing between encounters. This does make husk/fire victims have a bad time, but then, you've got free orders, saline, and transplants here if you really need them.

zeroisthebiggay:
> it's not meant to make them a direct competitor, but it is meant to make them require some attention. It takes about 6 burn wounds or upgrades of any stripe to get to the point of any serious danger from blood loss.

burn wounds already require immediate attention due to the potentiality of infections and their increased damage taken

even minor wounds inflict a ten percent damage increase against that body part which only means that due to the way wound bonuses are calculated only increases the likelihood of you getting a worse burn.

> It takes about half an hour leaving a t1 burn wound entirely untended (assuming it doesn't just clear up on its own, which they can) to get to sepsis.

what's the point of this change.

> When a t1 slash or puncture can very easily end a run left untreated, assuming you don't very quickly pop your one epipen and have another, it becomes pretty obvious that burns sucked by comparison.

ballistics as a general rule have always been stronger than lasers in direct combat due to wounds such as bonebreaks and bleeding. the only boon lasers have had is the ability to shoot through windows.

regardless of how easy it is to get rid of slash wounds they've always been king since you can't mend them easily mid chase and even if you're fast you can easily lose ten to twenty percent of your blood which mind you you only need to lose `thirty` to be in the threshold to be crit and killed by bloodloss.

> With a lot of our massively slippery single antags, at least for sec, this has made ballistics the meta - because most things that will heal or no-sell the flat damage from a ballistic still won't address bloodloss if you're continually dipping in and out of fights.

see above point. burns are to be addressed unless you don't want to take increasingly severe damage.

> and the relative shittiness of burn wounds.

taking more damage is always bad. hellfire lasers inflicting terrible burns often on their first shot means you're just going to be taking an extra twenty to thirty percent damage from them for the rest of the fight.

> It shouldn't be decisive in most cases, but it should help address the areas where people otherwise ignore burn wounds as though they don't exist, but bleeds mean everything.

just make higher tier burn wounds develop infections faster. the whole bloodloss thing is simultaneously nonsensical and unintuitive.

> This does make husk/fire victims have a bad time, but then, you've got free orders, saline, and transplants here if you really need them.

bloodloss is already a pain in the ass to deal with and i don't think saddling it onto the players who died and also the players who have to deal with it is a very fun change.

kiwedespars:
captain's laser fires hellfire lasers now for some reason which this pr would buff making it literally SSS tier as a gun. tacking on more bloodloss to the combat system that already suffers from too much of it doesn't seem right, especially when their gimmick is damage amp. left untreated? toxins upon thee, is a bonus.

idk i think this is unnecessary and to exacerbate the bleed meta is only going to make the game feel worse. the game already feels like even as a ninja, getting shot by the hos's fuckoff revolver will lead to you losing at least 10% blood minimum outside of immediately penning and just leaving the fight with blinks.

just nerf bleeding a tinge

Higgin:
@zeroisthebiggay I've been looking at this in practice and it's currently TM'd - you can ignore burns a lot of the time since T1's don't infect and T2s take a long-ass time but are very easily beaten with a bit of mesh. This makes T1s worth considering, T2s easier to get to, and taking multiple burns over a wide area in a short period of time something that actually matters.

A lot of the time, shots scatter - they don't necessarily find the same bodypart that received a wound. The lack of a tie-in to something that has lasting consequences (like bleed) that can't be very easily dealt with after the fact also makes them worse. 

This makes them better in a way that accomplishes the goals of wounds (separating the consequences of combat out from just raw damage) while not making them overbearing. So far, I like what I'm seeing in how they're playing out.

projectkepler-ru:
@Higgin  I'm curious as to what's the intended goal of this PR

BurgerLUA (changes requested):
Needs to be more modular.

PR #321, opened 2023-07-06
Nerfs fingerprints + Olfaction 

BurgerLUA:
## About The Pull Request
- Fingerprints are now partial and sometimes corrupted. In most cases it is still easy to tell who the fingerprint belongs to, in rare cases it can't.
- The more someone touches something, the more fingerprints it leaves.
- Fingerprints are no longer sorted by time when the fingerprint was placed.
- Fingerprint count on objects has a soft limit of 5, and a hard limit of 8, with the oldest fingerprint being removed.
- Olfaction now has a 50% chance to work per fingerprint entry.

image

Example image of me touching the same object several times.

## Changelog

:cl: BurgerBB
balance: Nerfs fingerprints and Olfaction.
/:cl:


zeroisthebiggay:
isn't the whole reason of syndicate soap to be used to scrub fingerprints off of crime scenes

https://github.com/Bubberstation/Bubberstation/blob/c76bc9f114cc343ab1556774de99ef66390d2cee/code/__DEFINES/cleaning.dm#L12

why are coders so allergic to intelligent play
it's so easy to not get caught with fingerprints because there's literally two one tc items that completely invalidate them `(soap and toolbox combat gloves)`

Vynzill:
I believe if you leave prints and get caught you 100% deserve it. You can remove prints or make it harder with various things (cleaning items, certain gloves). Don't think they need a nerf, players just need to play smart and avoid getting caught.

Detective-Google:
interesting idea, works pretty well on goonstation. would like to see how it plays out

Majkl-J:
This sounds like a needless overcomplication for something one can bypass with simple gloves or soap

Zirok-BYOND:
I wasn't aware of soap doing this- and gloves leave fibers, right? Also, does anything prevent you from leaving jumpsuit fibers?

Majkl-J:
> I wasn't aware of soap doing this- and gloves leave fibers, right? Also, does anything prevent you from leaving jumpsuit fibers?

Whilst the other two points are true, there are a lot of clothes vendors around for people to abuse.

Cyprex:
> Whilst the other two points are true, there are a lot of clothes vendors around for people to abuse.
It's fundamentally not sanely viable to change most of your attire every time you walk through an airlock. It's a huge hassle. Prints I don't know- prints might be fine. I believe fibers are overpowered her though, due to people actually caring about identity, rather than every engineer wearing the same jumpsuit etc.

code/modules/detectivework/scanner.dm
```diff
@@ -135,8 +135,14 @@
...
 
 	else if(!ismob(scanned_atom))
-
-		det_data[DETSCAN_CATEGORY_FINGERS] = GET_ATOM_FINGERPRINTS(scanned_atom)
+		//Bubberstation change: Support for new fingerprints.
+		var/list/found_fingerprints = GET_ATOM_FINGERPRINTS(scanned_atom)
+		if(length(found_fingerprints))
+			det_data[DETSCAN_CATEGORY_FINGERS] = list()
+			for(var/k in found_fingerprints)
+				det_data[DETSCAN_CATEGORY_FINGERS] += found_fingerprints[k]
```
Cyprex, line 143: As far as I can tell you are clearing the list, then adding things to it one-by-one and then shuffling it? Is there a reason you can't just shuffle it?

BurgerLUA: ``det_data[DETSCAN_CATEGORY_FINGERS]`` doesn't exist as a list before I call ``det_data[DETSCAN_CATEGORY_FINGERS]`` = list()``

GET_ATOM_FINGERPRINTS returns the actual list (not a copy of the list) of the fingerprints. It should not be manipulated via shuffle or else issues will occur.

GET_ATOM_FINGERPRINTS is a key/value pair, so I only need to copy the value and not the key.

Cyprex:
I still want this- cooldown could also be an effective way to nerf prints.

PR #329, opened 2023-07-10
Provocateurs (latejoin revolutionaries) can no longer cause the shuttle to be delayed.

BurgerLUA:
## About The Pull Request

Provocateurs (latejoin revolutionaries) can no longer cause the shuttle to be delayed.

## Why this is good for the game

Latejoin stealth roles shouldn't have an impact on whether or not a shuttle call should be called considering how easy it is to hide as a revolutionary and how easy it is to delay the round to 3+ hours.

## Changelog

:cl: BurgerBB
del: Provocateurs (latejoin revolutionaries) can no longer cause the shuttle to be delayed.
/:cl:


zeroisthebiggay:
are these even enabled i have not once seen a rev round on this accursed codebase

zeroisthebiggay:
> They are.

Huh.

destrucktoid:
If the head revolutionaries are not doing revolutionary things then, as stated by INKVD, that's something the admins online at the time need to deal with, it's exceptionally poor form to just hide and delay the round as both head rev and head of staff. 

There are ways to deal with head revs hiding though, it's called security mindshielding the entire station, which they should be doing on a rev round anyway. If the shuttle has been called and the revolution is completely unknown then, as said before, that's a problem with the head revs and the admins not forcing them to do things/replace them with a new head rev, not a gamemode problem.

PR #330, opened 2023-07-10
Dynamic Config Overhaul

BurgerLUA:
## About The Pull Request

This PR basically includes every current ruleset into the config and assigns them weights, costs, and other stuff per my personal opinion on what is good.

Some gamemodes still have 0 weight (not possible to pick) but still have costs and requirements assigned.

Missing Skyrat + Bubberstation roles, like blueshield, are protected from enemy roles for dynamic. All heads cannot role antag either.

## Changelog

:cl: BurgerBB
config: Redoes the Dynamic config settings
balance: Adds missing Skyrat + Bubberstation roles to protected and enemy roles for dynamic.
/:cl:

### DIFF
--- a/config/dynamic.json
+++ b/config/dynamic.json
@@ -3,164 +3,323 @@
 	"Roundstart": {
 		"Traitors": {
 			"cost": 8,
-			"scaling_cost": 9,
-			"weight": 0,
+			"scaling_cost": 8,
+			"weight": 100,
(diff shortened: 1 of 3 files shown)


BurgerLUA:
I have discovered my PR is ultimately pointless because the config file in the code is not actually the one being used. My changes to jobs is fine.

PR #331, opened 2023-07-10
Removes Birdshot and Northstar from map vote rotation

BurgerLUA:
## About The Pull Request

Removes Birdshot and Northstar from map vote rotation

## Why is this good for the game

While it is nice that there are extra map options, and while there has been some significant effort put into these maps, the general consensus on discord, in ooc, and even in game that people really don't like these two maps.

Some of the problems include clientside lag, very confusing and waypoint-unfriendly layout, excessive use of multi-z, and way too spacious for the pop.

## Changelog

:cl: BurgerBB
del: Removes Birdshot and Northstar from map vote rotation
/:cl:

### DIFF
--- a/config/maps.txt
+++ b/config/maps.txt
@@ -16,7 +16,7 @@ endmap
 map birdshot
 	#default
 	maxplayers 66
-	votable
+	#votable
 endmap
 
 map metastation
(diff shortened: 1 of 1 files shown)


StrangeWeirdKitten:
Personally, I think Northstar would be better with tweaks to what people find annoying. Mostly, it's just engineering and the anti-personnel land mines (why are these even mapped in)

nevimer:
This PR won't change anything except repo defaults.

Issue #336, opened 2023-07-11
Holoparas can't see shit


Majkl-J:
Turns out holoparas dont get nightvision even though they should.

Not entirely sure if this is a bug or just a regular oversight, but it is something they should get, especially scout ones.

## Reproduction:
Stab yourself with holopara as miner
Get holopara
Turns out the poor thing is basically blind and cant help on lavaland

PR #337, opened 2023-07-11
Allows guardians and holoparas to see in the dark

Majkl-J:
## About The Pull Request
Miners are able to get holoparas. The issue? Total blindness, which makes them not useful and literally just drags another player in with you to do nothing due to the lack of sight.
I could split this, but I believe normal holoparas deserve this change too.

This gives them nightvision

closes: #336

## Changelog

:cl:
fix: Holoparas can now naturally see in the dark
/:cl:


Detective-Google:
why not upstream this?

Majkl-J:
> why not upstream this?

Because I am lazy and hate tg github with a burning passion.

Actually maybe skyrat would accept this

nevimer:
merging knowing this will break CI, fixing it in an upstream merge pr

PR #339, opened 2023-07-12
Let Cargo Borgs break bad

Gr33d-y:
All this PR does is add a boxcutter to cargo borgs. Why? Because at the moment they are unable to open wrapped packages at all. This is not a massive deal on its own, however, cargoborgs are required to wrap packages to lift and haul them. This means that they are unable to deliver the packages that they wrap up, open them, and bring back the empty box in one trip, just as an example of one of the reasons they need a boxcutter.

Will this give cargoborgs a means to deal damage to others? Sure, but a lot of borgs have the ability to do damage with one or more of their tools far better than a boxcutter.

This was tested and I even learned that by doing this, borgs can recycle some wrapping paper this way! yippee!


zeroisthebiggay:
nonmobular

Zirok-BYOND:
Not sure what's up with the PR title

Change is totally fine, this has force 10(= damage, as far as I understand)+ sharpness, extinguisher which they already get has force 10 too.

Needs a comment //BUBBER EDIT ADDITION, if you could enable maintainer edits that would be swell!

zeroisthebiggay:
> Not sure what's up with the PR title

401-boxcutter

it's a breaking bad reference

Gr33d-y:
> Not sure what's up with the PR title
> 
> Change is totally fine, this has force 10(= damage, as far as I understand)+ sharpness, extinguisher which they already get has force 10 too.
> 
> Needs a comment //BUBBER EDIT ADDITION, if you could enable maintainer edits that would be swell!

should be enabled, does it work?

zeroisthebiggay:
> should be enabled, does it work?

if you made changes then you didn't push or forgot to commit because they're not here

modular_skyrat/modules/cargoborg/code/robot_modules.dm
```diff
@@ -16,6 +16,7 @@
 		/obj/item/crowbar/cyborg,
 		/obj/item/extinguisher,
 		/obj/item/universal_scanner,
+		/obj/item/boxcutter,
```
projectkepler-ru, line 19:
```suggestion
		/obj/item/boxcutter, //Bubberstation Edit
```

nevimer:
im merging this right now to follow up with an upstream merge

PR #340, opened 2023-07-12
(Non-Modular) Fixes a bug where station podpersons aren't friendly to bees

DerFlammenwerfer:
## About The Pull Request
There are two distinct species of Podperson, the primal one that is found in seed vaults, and the station-playable one with less than half the heal rate of the former. Both are intended to be treated as always friendly by bees, but due to an unreported code oversight, the species ID for station-side podpeople was never added to the proc that decides if someone is friendly to bees.

## Changelog

:cl:
fix: Station-side podpeople will no longer be attacked by bees as though they aren't plant people.
/:cl:


code/modules/hydroponics/beekeeping/beebox.dm
```diff
@@ -16,7 +16,7 @@
 
 
 /mob/living/carbon/human/bee_friendly()
-	if(dna && dna.species && dna.species.id == SPECIES_PODPERSON) //bees pollinate plants, duh.
+	if(dna && dna.species && dna.species.id == (SPECIES_PODPERSON || SPECIES_PODPERSON_WEAK)) //bees pollinate plants, duh.
```
BurgerLUA, line 19: This code will not work.

The proper code should be
``if(dna && dna.species && (dna.species.id == SPECIES_PODPERSON || dna.species.id == SPECIES_PODPERSON_WEAK))``

BurgerLUA (changes requested):

code/modules/hydroponics/beekeeping/beebox.dm
DerFlammenwerfer: I'll make this change

BurgerLUA (approved):

Cyprex:
PR upstream to skyrat or tg, whichever adds primal podpeople

nevimer:
lost to the great rebase...

PR #341, opened 2023-07-13
Blob events can no longer trigger on planet-based maps, (Icebox).

BurgerLUA:
## About The Pull Request

Blob events can no longer trigger on planet-based maps. Currently the only planet-based map is Icebox.

## Why this is good for the game

A recent round lead to an exploit where a blob spawned in a ruin. Fixing the area flags to prevent this would be optimal, but that wouldn't be a good solution to the problem as outlined below in notes.

This PR makes it so that Blob events can no longer trigger on planet-based maps. Blobs are just way too strong on these types of maps, especially if they have multi-z in them, as they are much more harder to find and have much more area to expand in.

## Notes

A proper "fix" that people will suggest is making it so that ruins cannot be possible blob locations, but this will not stop the issue from occurring given that there is a blob event that lets you be a human. It would be possible for a smart player to create an area in a far away exterior custom-area on the planet. A player might suggest "oh, just make it so that custom areas are exempt!", but that would make it possible for players to make areas immune.

This PR is not a fix but rather a balance change so that blobs can't spawn on maps with a metric fuckton of natural area to expand in.

## SALT PR???? SALT PR???? HUH?????

I was observing the round mentioned after I died from security flashing me as a rogue cyborg and killing me with lethals.

## Changelog

:cl: BurgerBB
balance: Blob events can no longer trigger on planet-based maps. 
/:cl:


BurgerLUA:
Also the whole blob being able to spawn in ruins is apparently an intended features as outlined here.

```
/area/ruin
	name = "\improper Unexplored Location"
	icon = 'icons/area/areas_ruins.dmi'
	icon_state = "ruins"
	has_gravity = STANDARD_GRAVITY
	area_flags = HIDDEN_AREA | BLOBS_ALLOWED | UNIQUE_AREA
	ambience_index = AMBIENCE_RUINS
	flags_1 = CAN_BE_DIRTY_1
	sound_environment = SOUND_ENVIRONMENT_STONEROOM
```

StrangeWeirdKitten:
tbh, fixes like these should go upstream. 

> Also the whole blob being able to spawn in ruins is apparently an intended features as outlined here.
> 
> ```
> /area/ruin
> 	name = "\improper Unexplored Location"
> 	icon = 'icons/area/areas_ruins.dmi'
> 	icon_state = "ruins"
> 	has_gravity = STANDARD_GRAVITY
> 	area_flags = HIDDEN_AREA | BLOBS_ALLOWED | UNIQUE_AREA
> 	ambience_index = AMBIENCE_RUINS
> 	flags_1 = CAN_BE_DIRTY_1
> 	sound_environment = SOUND_ENVIRONMENT_STONEROOM
> ```

Lol. To be honest, removing blobs from icebox would be how it goes. I think the other maps are fine. Icebox is problematic because the 2nd z level has very limited areas where stations like tram has multiple areas to transfer down

BurgerLUA:
It's not a fix, it's a balance change.

Plus I'm pretty sure I'm gitbanned upstream so that's not something I want to touch.

Nicolai904:
deffinitely more like oversight and can be easily changed by fixing ruins to not allow blob since it has a flag that allows it on the other note blob infection(when person gets ability to burst into blob) as well wont be allowed to burst in places that there is no flag with blob_allowed flag and if they do they will get teleported to a random space on the station not allowing blobs to spawn on multi Z seems rather nuclear option DO not have allow_blob flag so that is quite invalid point

Zirok-BYOND:
> but that would make it possible for players to make areas immune.
I don't quite follow, sorry!

BurgerLUA:
Disabling blob is on ruins would not fix the issue of blob being super strong on Icebox. There are a lot of areas on icebox that are really isolated from the station that are still considered station, such as some offshoot maint areas.

Serxule:
off-station blob tiles dont count towards total blob count, a small ruin isnt going to do anything

BurgerLUA:
> off-station blob tiles dont count towards total blob count, a small ruin isnt going to do anything

absolutely incorrect.

https://github.com/Bubberstation/Bubberstation/blob/c76bc9f114cc343ab1556774de99ef66390d2cee/code/modules/antagonists/blob/structures/_blob.dm#L192-L200

https://github.com/Bubberstation/Bubberstation/blob/c76bc9f114cc343ab1556774de99ef66390d2cee/code/game/area/areas/ruins/_ruins.dm#L3-L8

Serxule:
my point doesnt change, a small ruin isnt going to be OP or anything, off-station blob tiles still do not count and if anything a ruin will make being a blob harder

BurgerLUA:
> my point doesnt change, a small ruin isnt going to be OP or anything, off-station blob tiles still do not count and if anything a ruin will make being a blob harder

There are are lot of ruin spawns that are close to station that are fairly large. You could easily create a blob in a ruin and then expand to the station, which is usually less than half a screen away.

Serxule:
> > my point doesnt change, a small ruin isnt going to be OP or anything, off-station blob tiles still do not count and if anything a ruin will make being a blob harder
> 
> There are are lot of ruin spawns that are close to station that are fairly large. You could easily create a blob in a ruin and then expand to the station, which is usually less than half a screen away.

while yes, it also means you're going to have to fight the entire station to get onto the station/stay on it and to protect your main mass where ever it is (as they can just trail back the path), rather than just having to protect your main mass like normal

projectkepler-ru:
I appreciate everyone's comments on this one.
and while I agree with @Zirok-BYOND  that the best way to fix this is to disallow blob from forming off-station to begin with

This is still to be considered, thank you regardless. @BurgerLUA

PR #346, opened 2023-07-15
(Modular) Alt recipes for makeshift tools

DerFlammenwerfer:
## About The Pull Request
Another request from BurningCold, this PR adds alternative recipes for makeshift tools, allowing string from weapon-crafting to be used for the tools as an alternative to cable coil.

## Changelog

:cl:
add: Added alternative recipes for the five makeshift tools. This does not include the makeshift MFC cell or cable layer. These recipes use silk string instead of cable coil.
/:cl:


projectkepler-ru (approved):
I like this a lot,  the items are something you will likely have on hands and if not, are something you can still obtain semi-consistently.

PR #347, opened 2023-07-15
(Non-Modular) Blowgun can fire pills into uncovered mouths.

DerFlammenwerfer:
## About The Pull Request
A request from BurningCold for a gimmicky meme feature allowing blowguns to fire a pill into someone's mouth if it's uncovered. I had to add a new projectile and ammo casing to handle this, and modify the blowgun's code to load that projectile. I tried many ways of using the is_mouth_covered proc, but none of them worked because that proc is BROKEN. I instead coded it to check if you're wearing any kind of facewear or not, or if you're wearing a helmet that covers the face. If you're wearing either one, the pill will not transfer its reagents and be lost, otherwise, you will ingest whatever's in the pill. As a fun little side feature, the pill launched from the blowgun will inherit the sprite of the pill that was loaded into it.

## Changelog

:cl:
add: Added ammo casing and bullet support to handle the new blowgun feature
code: Added a function to the blowfun's loading proc to let it load and fire pills like a spitball
/:cl:


projectkepler-ru:
@DerFlammenwerfer  how common are blowgun exactly?

DerFlammenwerfer:
> @DerFlammenwerfer how common are blowgun exactly?

Blowguns aren't exactly common, you need someone to grow bamboo, and the downside of using them is 20 stamina damage and 20 oxygen damage per firing, which is not insignificant.

nevimer:
make this more modular or document your edits

code/modules/projectiles/guns/special/syringe_gun.dm
```diff
@@ -201,3 +201,32 @@
 	user.adjustStaminaLoss(20)
 	user.adjustOxyLoss(20)
 	return ..()
+/obj/item/gun/syringe/blowgun/attackby(obj/item/A, mob/user, params, show_msg = TRUE)
+	..()
```
BurgerLUA, line 205: This should be ``. = ..()``

BurgerLUA (changes requested):

nevimer:
victim of the great rebase, but this is worth it to remake

PR #349, opened 2023-07-16
[MODULAR] gives lone infiltrators microbombs

zeroisthebiggay:
## About The Pull Request

brings lone infiltrators in line with drifting contractors in that they aren't supposed to be a midround loot bag
they have droves of syndicate loot including the fabled redsuit and even a syndicate encryption key that security will otherwise loot and use to eavesdrop on what might be other traitors minding their own business

for god's sake they even have an atropine medipen in their bag and an instructions list detailing that they're able to use the atropine medipen so they don't blow the fuck up all over their friends

this is an oversight

## Changelog

:cl:
balance: lone infiltrators have microbombs
/:cl:


Serxule:
lone infils are more-or-less normal traitors fyi, they just get some extra gear

zeroisthebiggay:
> lone infils are more-or-less normal traitors fyi, they just get some extra gear
> 
> (them spawning with a nukie survival kit is the oversight arguably)

i still think that midrounds shouldn't give loot which could impede upon the main antagonists' romping

there's two forms of midround traitor in code infiltrator deserves to be just a bit different

nevimer:
You'll have to convince me that a person exploding when dying is better than the RP that follows if they don't.

zeroisthebiggay:
> You'll have to convince me that a person exploding when dying is better than the RP that follows if they don't.

nine times out of ten infiltrators are killed instead of arrested because their high profile nature generally ends with them doing a lot of crime really fast and being slippery due to the versatility of their MOD
not to mention their starting kit has a pistol in it that they're probably going to use to shoot someone

nine times out of ten if they're arrested nonlethally or revived after the aforementioned killing they're going to get put into perma whereinupon they will not be roleplaying because nobody chooses to interact with prisoners and all the current prisoner maisn are lrp botany abusers and dentskulls that do nothing but toe the line between breaking out and being annoying because they're a noncrew whose existence revolves around causing chaos and sowing dread amongst the crew so of course security isn't going to charge them with anything less than perma or execution

the whole `but the roleplay factor` argument is illogical when you realize that bubber is closer to tg with furries than it is to servers like bay or aurora when it comes to gameplay and 'roleplay envirornment'
this argument goes double as most of the sec mains are jaded citadel and tg players who use the meta options to eliminate antags as quickly as possible when they turn up

BurgerLUA (approved):
Code looks good.

nevimer:
stale and out of sync

PR #357, opened 2023-07-17
Blob laser resistance buff

Lutowski:
## About The Pull Request
Regular blob tiles take half as much damage from lasers, all others take 25% less. They already have a lot of health and I think anything more than that makes it feel like a slog to kill.

## Why It's Good For The Game

Blobs feel like Swiss cheese right now if you have a laser. Enough lasers makes it completely trivial, so this should help level the playing field a little. This should only impact lasers, so emitters will also get hit but other sources of heat will not.

## Changelog

:cl: Vekter
balance: Increased blob tiles' resistance to lasers to compensate for the recent buff to laser damage.
/:cl:

### DIFF
--- a/code/modules/antagonists/blob/structures/_blob.dm
+++ b/code/modules/antagonists/blob/structures/_blob.dm
@@ -35,6 +35,7 @@
 /datum/armor/structure_blob
 	fire = 80
 	acid = 70
+        laser = 50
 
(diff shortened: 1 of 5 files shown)


BurgerLUA (approved):
Code looks good.

StrangeWeirdKitten:
Upstream is going to make this PR redundent

Cyprex:
Redundant

PR #358, opened 2023-07-17
Adds an additional affection module for normal borg chassis.

Chirynnn:
## About The Pull Request

Yippee, first PR here!
This adds the Alternative Affection Module, essentially allowing all chassis that don't have R_TRAIT_WIDE to use it. Same costs, both available at round start. Why not just merge it into one, you ask? Because the original one has items pertaining to dogborgs, and it didn't feel right to not change them up a bit. I also added a couple extra verbs to the actions.

https://github.com/Bubberstation/Bubberstation/assets/77589317/a08ba442-f869-4628-9ae7-638e8dcc9e5d

## Changelog

:cl: Chiryn
add: Alternative affection module for normal cyborgs
/:cl:


Chirynnn:
what is this conflict what do I do

DuffCreeper:
I personally would've just joined it under one item, the module basically just needs an additional flag and you're golden. And while I can understand having things pertaining to dog borgs on something that isn't. It does prevent having essentially two items that do the same thing

PR #356, opened 2023-07-17
(Non Modular) Replica Pods with Bluespace Activity grant their podperson a permanent Bluespace Prophet trauma

DerFlammenwerfer:
## About The Pull Request
Another request from Burning Cold, which I could see being a really funny feature for any enterprising botanists looking to make money off the Grey Tide

## Changelog

:cl:
code: When a Replica Pod is harvested and makes a pod person, if the plant has the Bluespace Activity trait it gives the resulting pod person the Bluespace Prophet trauma with absolute resilience.
/:cl:

PR #360, opened 2023-07-18
(Non-Modular) Podperson cloning via Replica Pod

DerFlammenwerfer:
## About The Pull Request
This PR adds a feature that was requested by BurningCold. Because the styling function of right-clicking with secateurs is currently non-functional, I've elected to repurpose it. You can now right-click any podperson with secateurs and, so long as you aren't in combat mode, in 2.5 seconds you will create a special graft-like item called a cutting, which will inherit the name, mutations, quirks, and mind (or previous mind, if ghosted) of the body it is taken from. This will cause 10 brute to the podperson. This cutting may then be applied to a replica pod growing in a tray, if that plant does not already have a ckey associated with it. In addition to being reborn as a Primal podperson, the replica will inherit mutations from their cutting, and have their visual appearance restored from the client's preferences. Because of the similarity to the effects of a SAD and the similar lack of safeguards against identity alteration or unintended side effects of such I've included logging for when a podperson is cloned by a Replica pod.

Please note that this will conflict with https://github.com/Bubberstation/Bubberstation/pull/356 , even though I'm not expecting that one will actually get merged.

## Changelog

:cl:
add: Added the graft_podperson item, called a cutting, which is used with replica pods already growing in trays due to the lack of suitable blood for sampling
code: Changed the right-click function of secateurs to collect cuttings from podpeople.
code: Altered the creation of pod people so that it can distinguish between replicas from blood samples and replicas from cuttings. Blood-sampled replicas will function exactly the same as before, while podpeople get to keep their mutations and are visually-actualized when they emerge. 
/:cl:


nevimer:
lost to the rebase

PR #366, opened 2023-07-19
Shortens the Akula Language lore

50Potas50water:
## About The Pull Request
Inconsistent with the lore length of other languages.
Takes up more space in the menu than it needs to
Kept in the more important sounding pieces of it. 
If the PR goes through it could be preserved somewhere in the lore bin 

## Changelog

:cl:
del: Shortened Akula language lore
/:cl:


magatsuchi:
i extremely agree with this but take this pr to skyrat instead

PR #376, opened 2023-07-21
Reworks the Ninja Energy Katana so it isn't as powerful, but still cool.

BurgerLUA:
## About The Pull Request

The energy katana is too strong. I have decided to make it less strong but still make it interesting to use so people don't hate me for doing a simple 1 line change.

Here is what is changed:
- The sword no longer has an innate block chance.
- The sword gives a 100% chance to block unarmed attacks.
- The ninja sword now has two stances; slash stance and stab stance. There is a 4 second cooldown between switching to the two.
- Slash stance (default) gives 30 damage per hit, with 15 armor penetration. The type of wounds given is slashing.
- Stab stance gives 20 damage per hit, with 50 armor penetration. The type of wounds given is piercing.
- For 2 seconds after you switched a stance, you gain 100% chance to block melee attacks, and a 50% chance to block ranged attacks.
- Teleport distance is reduced to 6 tiles. A smart teleport system was added so you will always teleport to the safest area if you try to teleport to a bad area.
- It takes 20 seconds to regenerate a teleport charge, with maximum 3 charges.
- There are two different inventory icons for stances, and a sound plays when you change stances. The sound is specifically designed to indicate when you are blocking, and when you can change stance again.

## Changelog

:cl: BurgerBB
balance: Refactors the Ninja Energy Katana so it is less powerful. Examine it to see what it can do now!
/:cl:


Oscari-pog:
The dangerous teleport thing is dumb. Please, call it a nerf like it actually is, instead of pretending like its helpful.

Alecksohs:
Personally I think the power of a ninja is just fine. He's strong. He's also designed to get completely mobbed by every borg and worker on the station so.

Kraseo:
space isn't a safe tile for ninja to be in since... when exactly?

BurgerLUA:
> The dangerous teleport thing is dumb. Please, call it a nerf like it actually is, instead of pretending like its helpful.

the title literally implies a nerf my dude

zeroisthebiggay:
image
considering this was made right after a round where a ninja was being particularly super slippery
isn't this the definition of a kneejerk pr
the teleport system specifically is nonsensical and stupid

oh yeah let's take away the antag that only fights in melee's escape tools while simultaneously making them substantially worse  and clunky with a goofy ass hyperborean system that nobody will understand

Serxule:
i get that ninjas are slippery, BUT ITS A NINJA, THEY ARE MEANT TO BE SLIPPERY (and removing block is just stupid)

BurgerLUA:
> image considering this was made right after a round where a ninja was being particularly super slippery isn't this the definition of a kneejerk pr the teleport system specifically is nonsensical and stupid
> 
> oh yeah let's take away the antag that only fights in melee's escape tools while simultaneously making them substantially worse and clunky with a goofy ass hyperborean system that nobody will understand

Hi. I was playing lawyer that round. I was basically witness to security not really having any sort of fun fighting against the antagonist because of how evasive it is. The ninja's movement speed is enough evasion as it is.

Cyprex:
For future reference, refactor means this:
> Refactoring is the process of restructuring code, while not changing its original functionality. The goal of refactoring is to improve internal code by making many small changes without altering the code's external behavior.

This is a balance change.

nevimer:
You would probably find more luck taking this to tgstation, but that's not to say we won't accept it here.

zeroisthebiggay:
> The ninja's movement speed is enough evasion as it is.'

ninjas move at default speed???? tg removed the speedboots when they made ninjas use a modsuit?????

BurgerLUA:
> You would probably find more luck taking this to tgstation, but that's not to say we won't accept it here.

I am not comfortable with introducing this to /tg/station considering the anger I've received here over it.

YakumoChen:
A reminder that nerfs do not hamper the top percentage of robust players, you should be balancing for the skill floor instead of the ceiling.

Serxule:
> A reminder that nerfs do not hamper the top percentage of robust players, you should be balancing for the skill floor instead of the ceiling.
> 
> For every one robust ninja you salt PR a nerf for, there's probably 5 unrobust ninjas that got clapped instantly that this PR only hurts more.

this was entirely what i was thinking, this would _at most_ inconvinence the robust players, but for everyone else? this would completely fuck them over and likely lead to death after death

BurgerLUA:
The lower skill level was considered, which is why the teleport mechanic was better coded and makes sure that you will always teleport on use instead of just saying "can't teleport here" or whatever message it used to be.

A sound effect was added to indicate when the block starts, when it ends, and when you can switch stance again so it's not something you have to memorize.

A tooltip text was added to indicate which stance you're currently in so you don't have to constantly focus on the chat to see what stance you're in.

YakumoChen:
> The lower skill level was considered, which is why the teleport mechanic was better coded and makes sure that you will always teleport on use instead of just saying "can't teleport here" or whatever message it used to be.
> 
> A sound effect was added to indicate when the block starts, when it ends, and when you can switch stance again so it's not something you have to memorize.
> 
> A tooltip text was added to indicate which stance you're currently in so you don't have to constantly focus on the chat to see what stance you're in.

Adding context and functionality (READ: forcing the ninja to waste dash charges on illegal moves)(ALSO READ: making combat needlessly comples) is NOT remotely similar to preserving the skill floor for the antag.

zeroisthebiggay:
the `six tile range limit` feels so arbitrarily chosen
the screen is 7x7 seven on the correct screen size setting and 7x`whofuckingknowsidon'tplaywithwidescreen` on widemode
arbitrarily setting it to 6 instead what would otherwise be the corner of your screen just makes no sense

BurgerLUA:
remaking this PR after a few code changes because it's really honestly a bad sign for this PR when someone complains about the 6 tile range because it's apparently arbitrary because it's not 7

yourdoom9898:
> remaking this PR after a few code changes because it's really honestly a bad sign for this PR when someone complains about the 6 tile range because it's apparently arbitrary because it's not 7

Or...because the interactive viewport (widescreen or not) is a 7x7 tile square...

PR #380, opened 2023-07-22
Adds EggRP

Majkl-J:
## About The Pull Request

Ports yogstation's suicide act (eggt) that transfers your soul into an egg.

This is just a silly feature, not much else

## Changelog

:cl:
add: Suiciding with an egg makes you into an egg
/:cl:


Cyprex:
Sorry- but silly meme features like this isn't really what Bubber is for, we do take RP a bit more seriously!

There's heaps of ways things could go wrong here, or tg could refactor suicides or whatever- small as it is it's still a maintenance burden and we really can't justify it for something as memey as this!

PR #382, opened 2023-07-22
fixes oversized synths/species with nohunger getting huge guts

Kraseo:
## About The Pull Request

as stated by the title, makes it so that synths do not get overwhelmingly crippled by the fact that they will never be able to restore their nutrition due to missing a vital component (and the fact they cannot process nutriment)
the latter is more of an aesthetic thing as skeletons and such should probably not have huge fleshy guts where bone should be

## Changelog

:cl:
fix: Synths with the oversized quirk are no longer doomed due to having a fleshy stomach where a power cell should be, and can recharge from power sources again.
fix: Species that have no need to eat no longer spawn with huge guts from the oversized quirk.
/:cl:


nevimer:
I'll accept it but please do this upstream too.

projectkepler-ru:
@Kraseo ?

Kraseo:
rebasing broke my shit so i'm just going to reopen this later in a new pr :D

nevimer:
Go for it! This is basically approved.

PR #385, opened 2023-07-22
Re-adds cat scream

tf-4:
## About The Pull Request

Restores the cat scream from https://github.com/Skyrat-SS13/Skyrat-tg/pull/22050

All tested and such.
image

## Changelog
:cl:
sound: Re-instated the cat scream previously removed.
/:cl:


Cyprex:
I don't find this grating but also- wtf is this supposed to be? This is supposed to be a scream? I'll leave it open for others to comment, but I'm not inclined to add it back, personally.

ReturnToZender:
If it's modular, why not?

nevimer (approved):

PR #387, opened 2023-07-24
Adds a new Advanced Energy Katana to the Uplink

BurgerLUA:
## About The Pull Request

Adds a new advanced energy katana to the uplinks not belonging to nuke ops and clown ops. It can be purchased for 16 TC. It is like the energy katana that ninjas get, but with a few major differences.

- The sword doesn't have an innate block chance.
- The sword gives a 100% chance to block unarmed attacks.
- The sword has two stances; slash stance and stab stance. There is a 4 second cooldown between switching to the two.
- Slash stance (default) gives 30 damage per hit, with 15 armor penetration. The type of wounds given is slashing.
- Stab stance gives 23 damage per hit, with 50 armor penetration. The type of wounds given is piercing.
- For 2 seconds after you switched a stance, you gain 100% chance to block melee attacks, and a 50% chance to block ranged attacks.
- Teleport distance is reduced to 5 tiles. A smart teleport system was added so you will always teleport to the safest area if you try to teleport to a bad area.
- It takes 30 seconds to regenerate a teleport charge, with maximum 3 charges.
- There are two different inventory icons for stances, and a sound plays when you change stances. The sound is specifically designed to indicate when you are blocking, and when you can change stance again.

The katana can be refunded for the full purchase cost just in case people don't like it.

## Changelog

:cl: BurgerBB
add: Adds a new Advanced Energy Katana to the uplink. Similiar to the Ninja katana but with a few added features/limitations.
/:cl:


yourdoom9898:
Once again, playable viewport (widescreen or not, despite widescreen being visibly a 7x11) is a 7x7 square, so shortening the warp to six tiles just creates a weird border for no reason.

BurgerLUA:
> Once again, playable viewport (widescreen or not, despite widescreen being visibly a 7x11) is a 7x7 square, so shortening the warp to six tiles just creates a weird border for no reason.

this is such a trivial nitpick, this makes literally no sense to complain about

when testing solo, the range that I set felt fine. if people think it's too small or too large during testing, then it can be adjusted

BurgerLUA:
The range is 5, by the way.

BurgerLUA:
Okay so this isn't bikeshedded further because this is literally the stupidest thing to argue about.

I thought it over. 6 tiles is good because while the view range is 7 (I don't even know if it's 7 and I can't be bothered to check), HUD elements like the inventory are commonly in the way at some edges. A deadzone is good because of those elements being in the way.

BurgerLUA:
> > Okay so this isn't bikeshedded further because this is literally the stupidest thing to argue about.
> > I thought it over. 6 tiles is good because while the view range is 7 (I don't even know if it's 7 and I can't be bothered to check), HUD elements like the inventory are commonly in the way at some edges. A deadzone is good because of those elements being in the way.
> 
> You clearly haven't played /tg/ in a long time... I mean that in the sincerest way possible.

clearly you don't own an air fryer

nevimer:
I'll testmerge this tonight.

YakumoChen:
What's the perk to buying one of these over a similarly priced d-esword, which isn't burdened by cursed heart-like mechanics and isn't completely countered by wearing armor and a gun?

BurgerLUA:
> What's the perk to buying one of these over a similarly priced d-esword, which isn't burdened by cursed heart-like mechanics and isn't completely countered by wearing armor and a using a gun?

the teleport

YakumoChen:
> > What's the perk to buying one of these over a similarly priced d-esword, which isn't burdened by cursed heart-like mechanics and isn't completely countered by wearing armor and a using a gun?
> 
> the teleport

Why spend 16 on a teleport sword when you can get more consistent results with a 7 tc esword and bluespace crystals, or the red handtele

Or just buy sleeping carp?

BurgerLUA:
There is nothing consistent about using bluespace crystals for teleporting.

YakumoChen:
> There is nothing consistent about using bluespace crystals for teleporting.

They don't rob you of a teleport for going into a wall or into space and you can hold more than three at a time.

The esword does more damage than the katana in either mode and has a block chance of **50**. For **melee and projectiles.**

BurgerLUA:
Half of what you said is wrong and the other half just feels like a conflicting opinion that overestimates how useful bluespace crystals are.

The energy katana allows you to go into space.
The energy katana has a block chance of 100% for unarmed attacks.
The energy katana has a 100% block chance for melee attacks, and a 50% block chance for ranged attacks 2 seconds after switching stances.

The energy katana has a teleport range of 6. Bluespace crystals have a range of 4.
The destination in which you teleport is completely random with bluespace crystals, while the dash with the Energy Katana is controlled.
Bluespace crystals are itemized, thus are annoying to manage and use compared to the infinitely recharging Energy Katana.

YakumoChen:
> Half of what you said is wrong and the other half just feels like a conflicting opinion that overestimates how useful bluespace crystals are.
> 
> The energy katana allows you to go into space. The energy katana has a block chance of 100% for unarmed attacks. The energy katana has a 100% block chance for melee attacks, and a 50% block chance for ranged attacks 2 seconds after switching stances.
> 
> The energy katana has a teleport range of 6. Bluespace crystals have a range of 4. The destination in which you teleport is completely random with bluespace crystals, while the dash with the Energy Katana is controlled. Bluespace crystals are itemized, thus are annoying to manage and use compared to the infinitely recharging Energy Katana.

Having to click a button every four seconds so you get blocking for only half that, assuming you're even getting perfect timing, means it has an effective block rate of 50 melee, 25 ranged, which is lower than an esword.

An esword always has better damage because it does a flat 30 brute, 35 AP

Clicking an action button mid combat is a hostile experience and you're forced to switch modes so it's difficult and pointless to manage modes outside of spamming the switch

You do not define a "safe teleport" and imply that you waste a charge teleporting somewhere "unsafe" which in most cases is where you started.

Bluespace crystals do not care for "safe teleporting" and while random you can simply crush them until you get away, whereas the sword has 3 charges with a huge cooldown

The sword is designed to be frustrating to use and isn't even a good weapon.

BurgerLUA:
I have a solution to your feedback.

Just don't buy it.

Think that a teleport sword is too expensive? Don't buy it. If people don't use it during testmerge, or feel that it is too expensive,. then the price can be decreased.

Detective-Google:
Why? What gap does this fill in the uplink? what does it bring to the table except a knockoff of a fairly iconic weapon. 
also not sure why this has a refund option when most other items don't, save for items like a holoparasite where the holo ghosting can just fuck it completely, or you can get no ghosts, etc. Refunds are usually reserved for items that simply *do not work* if a condition isn't met. 

is there anything to stop someone buying this, cutting up their target, refunding it for all their TC back, and then just locking their uplink, destroying the murder weapon and keeping all their TC safe?

Detective-Google:
as an addendum, what's the thinking behind the selective block against unarmed? shoves, grabs, etc are an important part of managing an escape from someone with a high power melee weapon, I think it's fairly hostile design that if players are trying to push or punch someone who is stabbing them it's just going to be blocked, the amount of people you see fistfighting the russians on void raptor shows that a lot of players don't bother carrying a high melee damage item on them 'just in case'. In addition to having a fairly strong ability to close the gap on fleeing targets, people who run behind doors and close them etc. Seems like something that becomes pretty trivial for sec to deal with, and nightmarish for normal players, who have a lot of their counters to these weapons stripped off. 

Another thing I'd like to know, what happens if you buy two? with progtot traitors can regularly get up to around 40TC, and having two of these would allow you to keep double the uptime on the _100% melee block_ have they been balanced around this eventuality? Does the player have to be holding the item for this to work? what if they ended up with three, and juggled them around?

BurgerLUA:
I actually never considered that traitors could get two of these lmao.

projectkepler-ru:
> I actually never considered that traitors could get two of these lmao
yea

nevimer:
I really wanted this one....

PR #392, opened 2023-07-25
adds a verb to forcibly close tgui windows

Kraseo:
## About The Pull Request

see title. 

## Changelog

:cl:
add: Adds a verb to forcibly close TGUI windows, useful if your PC crashes and the game still believes it's still open when you rejoin the game.
/:cl:


code/controllers/subsystem/tgui.dm
```diff
@@ -286,6 +286,7 @@ SUBSYSTEM_DEF(tgui)
  */
 /datum/controller/subsystem/tgui/proc/on_open(datum/tgui/ui)
 	LAZYOR(ui.src_object.open_uis, ui)
+	ui.user.tgui_open_uis |= ui
```
nevimer, line 289: add a bubber edit comment
also, wont this mess with statpanel and chat? im sure this is already tracked somewhere

nevimer: what was also get duplicate of the same ui, will it close those with that operator?

Kraseo: this was a misdelete caused by https://github.com/tgstation/tgstation/pull/76215
`tgui_open_uis` is used to track every ui currently open by an user while `all_uis` is used to track every ui currently open by every user
i have tested the code and this will not mess with either tgui chat or tgui statpanel

not sure what you meant by the second question, please rephrase

Kraseo: fixed in https://github.com/tgstation/tgstation/pull/77101

Kraseo:
the main issue of windows refusing to open again if you happened to crash your client with them open is basically solved by that one line in tgui code so this whole thing is pointless (i am not a smart person)
but i'd still like it to be merged if tgui ever happens to break again for anyone so i'll keep it open

Cyprex:
Should be PRed to upstream tg

magatsuchi:
you must pr higher

image

nevimer:
On the flip side, I don't see why we can't have it anyways as a modular verb. Do both.

magatsuchi:
> On the flip side, I don't see why we can't have it anyways as a modular verb. Do both.

if modularity can be avoided, it definitely should, this is something that would happily be accepted upstream

PR #395, opened 2023-07-27
Prevent admins from pressing buttons that cause the server to reach out to external services and fail.

nevimer:
AAAAAA.

...

AHHHHHHHHH.

Locks up the server waiting for responses it will never get.

### DIFF
--- a/code/modules/admin/topic.dm
+++ b/code/modules/admin/topic.dm
@@ -1456,6 +1456,9 @@
 		usr << browse(dat.Join("<br>"), "window=related_[C];size=420x300")
 
 	else if(href_list["centcomlookup"])
+		if(href_list) // BUBBER EDIT BEGIN - Crash Bandaid
+			message_admins("a naughty admin was prevented from hanging the server sending an external query.")
+			return // BUBBER EDIT END - Crash Bandaid
 		if(!check_rights(R_ADMIN))
 			return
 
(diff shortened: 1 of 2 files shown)


StrangeWeirdKitten:
I need this

Cyprex:
Why the heck does this crash instead of just silently timing out asynchronously and failing.

Cyprex:
To be clear if the server locks up for a few seconds then that's not great but imo acceptable- admins can figure out NOT to press funny buttons.

We use CC lockups, and these really shouldn't cause issues...

code/modules/admin/topic.dm
```diff
@@ -1456,6 +1456,9 @@
 		usr << browse(dat.Join("<br>"), "window=related_[C];size=420x300")
 
 	else if(href_list["centcomlookup"])
+		if(href_list) // BUBBER EDIT BEGIN - Crash Bandaid
```
Cyprex, line 1459: Also, href_list is guaranteed to have the key "centcomlookup", does that not ensure this condition is always true? I don't know how truthiness works with regards to lists.

nevimer: > Also, href_list is guaranteed to have the key "centcomlookup", does that not ensure this condition is always true? I don't know how truthiness works with regards to lists.

If anything is in the contents of a variable its true. Theres some weird null differences.

nevimer: Also it's intentionally like this as it doesn't make the linter cry like just adding a return does.

magatsuchi (approved):
approved, for the reasons listed in maintainerbus

make it priority that we get this entire subsystem either implemented or removed

code/modules/admin/topic.dm
magatsuchi: nonissue bc of linter context

PR #410, opened 2023-08-02
Removes borer egg from uplink

GrayRachnid:
## About The Pull Request
This removes the ability for traitors to purchase borer eggs from the uplink.

## Why It's Good For The Game
As it stands, borers are too powerful as a 2nd tool to traitors. Unlike Holoparasites which dust you upon death, borers come with no downsides other than their steep cost. A cost which can easily be recouped with the help of the borer in your head helping you do side objectives.

A good traitor with a borer in their head is a terrifying force to be reckoned with. A borer-empowered traitor at their prime cannot be dealt with by any means possible. One such example was 3 weeks ago when Weiss Ii used a borer egg as their first purchase to accumulate a total of 63 TC while running permanently mach 3 around the station. After their TC hoarding, they used it all to buy an elite syndi modsuit, double e-sword and glue alongside bits and bobs. With the assistance of the borer's speed and overpowered wound recovery chems, they were essentially taking no damage and being able to revive. This is on top of their very heavily armored suit and the 75% block chance d-sword that cannot be disarmed.

That is one such example. I will admit that this is by all means a salt PR and I will not shy away from it. But I will also note that Borers are an unbalanced antagonist and factor into the round. They come from a codebase that only uses them for flavor and roleplay, not for actual balanced gameplay.

It's my hope that borers either get completely removed or overhauled to become actually balanced, but I do not see anyone picking up the latter and the former will piss off too many people. So this is my compromise, traitors cannot purchase them anymore and everyone is happy. A round will happen where they both spawn but to that I just say thanks dynamic. 

## Changelog

:cl:
del: Removes borer egg from traitor uplink.
/:cl:

### DIFF
--- a/modular_skyrat/modules/skyrat-uplinks/code/categories/dangerous.dm
+++ b/modular_skyrat/modules/skyrat-uplinks/code/categories/dangerous.dm
@@ -4,13 +4,3 @@
 	item = /obj/item/gun/ballistic/automatic/c20r/toy/unrestricted/riot
 	cost = 4
 	purchasable_from = ~(UPLINK_NUKE_OPS | UPLINK_CLOWN_OPS)
-
-/datum/uplink_item/dangerous/cortical_borer
-	name = "Cortical Borer Egg"
-	desc = "The egg of a cortical borer. The cortical borer is a parasite that can produce chemicals upon command, as well as \
-			learn new chemicals through the blood if old enough. Be careful as there is no way to get the borer to pledge allegiance \
-			to yourself. The egg is extremely fragile, do not crush it in your hand nor throw it. \
-			The egg is required to sit out in the open in order to hatch. (Cannot be hidden in closets, etc.)"
-	progression_minimum = 20 MINUTES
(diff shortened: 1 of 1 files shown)


GrayRachnid:
This is a salt PR.

GrayRachnid:
No actually you don't know what you're talking about, they're not that bad. You just have to consider the fact that security is forced to treat you lethally similar to a holoparasite while you're running around the station at mach 7

SonicGotNuked:
Seems like we should just remove meth from borers

magatsuchi:
(also this PR needs a lot more than just a type path removal but we’ll get there when we get there)

magatsuchi:
average salt pr outcome

Detective-Google:
> average salt pr outcome

borers can copy traitor stims allowing them to maintain 100% stim uptime for the entire round, they can do this while also keeping you dosed on meth that has no downside, (an injection of borer meth just under the OD threshold has no downside and gives you ~10 minutes of meth.), while acting as a mini medbay crammed up your rear where nobody  can remove it without surgery. I could go on, but I think the point is fairly obvious and you should stop posting bait <3

PR #423, opened 2023-08-08
Enables borers spawning.

Nep993:
Enables them, nuff said. It was set at 999 players needed, lowered it to 35... please god why?


nevimer:
I mean, I don't think this was used or where the bug was, but the errors are not your fault.

PR #432, opened 2023-08-10
Essential fixes so the game isn't broken.

nevimer:
~~disables status indicators and fixes a merge conflict that created a logic error~~
okay it does way more. does fix some CI issues tho. Removes screenshot tests as we don't have the bot to fix it.


nevimer:
This is the price we pay

PR #444, opened 2023-08-11
Adds Death Penalty 2

BurgerLUA:
## About The Pull Request

Adds a mood penalty for death. The mood penalty can be removed by having a great sanity, and a good mood.
Adds a quirk that prevents the mood penalty from being applied.

## Why It's Good For The Game

Needs more penalties for death, even if it's minor like mood.

## Changelog

:cl: BurgerBB
add: Adds a mood penalty for death, plus new quirk that negates it. The mood penalty for death can be removed by having a great sanity and a good mood.
/:cl:

### DIFF
--- a/code/__DEFINES/traits.dm
+++ b/code/__DEFINES/traits.dm
@@ -309,6 +309,9 @@ Remember to update _globalvars/traits.dm if you're adding/removing/renaming trai
 #define TRAIT_BLOOD_DEFICIENCY "blood_deficiency"
 #define TRAIT_JOLLY "jolly"
 #define TRAIT_NOCRITDAMAGE "no_crit"
+//Modular zubbers change start.
+#define TRAIT_I_DO_NOT_FEAR_DEATH "i_do_not_fear_death"
+//Modular zubbers change end.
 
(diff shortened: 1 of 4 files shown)


YakumoChen:
-9 is a fairly absurd debuff

nevimer (approved):
I approve of it in the current concept. It's been discussed before we want this. We might even go further in the future with death consequences. Maybe you can add something special for coroners?

BurgerLUA:
> -9 is a fairly absurd debuff to have no timeout at all. That's at the same level of losing a limb (8 minutes) or having a condition that is actively killing you.
> 
> Most positive mood events are +1 aside from special ones and don't last very long, so to counter it you'd have to be spending a ton of time working on mood items to overcome it.

I chose this number because +9 is the same effect as getting high and drunk.

If people find that it is too punishing, I can reduce it by half or 25%.

Majkl-J:
You will probably climb out of medbay with like 2 more negative quirks for surgery, and I really dont want my character becoming a smoking drugged wreck unable to walk fast just because the bad mood from death doesn't go away after a while.

The punishment of having to see the only doctor having the combined brainpower of a clown full of neurotoxin when reviving me is more than enough.

PR #447, opened 2023-08-12
Adds bounties to the blacksmith job

Majkl-J:
## About The Pull Request

Adds a bunch of bounties for the smith, requesting a wide range of items that can be easily expanded upon if needed.

## Why It's Good For The Game

The current way of smith making money is boring (Click two buttons repeatedly to make a glass ball), which also leaves almost no reason to do actual smithing, with its tiny reward of "The clown wanted a sword that one shift and paid me 50cr" as compared to making 8800 credits for 10 glass balls that can be pumped out at this rate every 5 minutes.

The hope here is to give smiths a decision between bounties and glassblowing. This is probably still unbalanced one way or the other, which is why I am asking for a testmerge until the sweet spot can be worked out (I can only test if this works mechanic wise by myself).

## Changelog

:cl:
add: Centcom now might request some smithing items from the station's smith
/:cl:


Cyprex (approved):
Code looks good. The bounties should all be tested but you don't have to do that, we can TM it for that if you'd rather not. No idea if the reward/requirements are appropriate (-ly balanced). I'm not familiar with blacksmithing, so I'm just asking about this on Discord!

code/__DEFINES/economy.dm
```diff
@@ -65,6 +65,9 @@
 #define CIV_JOB_GROW 12
 #define CIV_JOB_ATMOS 13
 #define CIV_JOB_RANDOM 14
+//BUBBER EDIT START
+#define CIV_JOB_SMITH 21 //By making this higher we avoid having to maintain this value if more bounties are added upstream
```
Cyprex, line 69: For future reference by anyone else this works just as well technically if you do it modularly but you should NOT do this modularly, as these defines need to be unique, and if they're not you get awful subtle bugs.

Majkl-J:
> Code looks good. The bounties should all be tested but you don't have to do that, we can TM it for that if you'd rather not. No idea if the reward/requirements are appropriate (-ly balanced). I'm not familiar with blacksmithing, so I'm just asking about this on Discord!

I know the bounties work as they should (they beam stuff up just fine). I want this not to become the new meta, but a fairly balanced alternative so people can actually decide on what they want to do instead of just being forced to blow into the funny tube for half their shift.

PR #451, opened 2023-08-13
Fixes sensitive snout typo

tf-4:
:cl:
spellcheck: Fixed a typo with examine text for the "Sensitive Snout" quirk.
/:cl:

### DIFF
--- a/modular_skyrat/master_files/code/datums/traits/negative.dm
+++ b/modular_skyrat/master_files/code/datums/traits/negative.dm
@@ -79,5 +79,5 @@
 	alert_type = null
 
 /datum/status_effect/sensitivesnout/get_examine_text()
-	return span_warning("[owner.p_their(TRUE)] snout is rather bappable...")
+	return span_warning("[owner.p_Their())] snout is rather bappable...")
 // BUBBER EDIT MOVE END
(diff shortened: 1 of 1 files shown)


nevimer:
Build error

modular_skyrat/master_files/code/datums/traits/negative.dm
```diff
@@ -79,5 +79,5 @@
 	alert_type = null
 
 /datum/status_effect/sensitivesnout/get_examine_text()
-	return span_warning("[owner.p_their(TRUE)] snout is rather bappable...")
+	return span_warning("[owner.p_Their())] snout is rather bappable...")
```
Majkl-J, line 82:
```suggestion
	return span_warning("[owner.p_Their()] snout is rather bappable...")
```

Majkl-J (changes requested):

modular_skyrat/master_files/code/datums/traits/negative.dm
Majkl-J: Also since this comes from srat and is an error there you would be better off pushing this upstream

nevimer: This is an us thing, rodent serv doesn't have the trait do anything at all because of various reasons.

Majkl-J: Either way they still have it in its broken form

PR #462, opened 2023-08-16
Jukeboksss Update

Majkl-J:
## About The Pull Request

Purchasable jukebox for 10k credits

WARNING: YOU WILL NEED BAR ACCESS

## Why It's Good For The Game

We added a shit ton of songs into it only for it to be admin only. There is no reason for it to not be accessible to crew
(And yes I know it spawns in ruins and on birdshot. That isn't easily accessible)

## Changelog

:cl:
add: Jukeboxes can now be purchased via the cargo shuttle
/:cl:


Majkl-J:
> It's laggy and causes people with existing poor connections to DC. 👎

Testmerge it to see if this is true. It appears so little we cannot be sure this is a real issue

PR #473, opened 2023-08-21
Changes the cost of the borg expander to be 1 sheet of iron, not 100

DuffCreeper:
## About The Pull Request

Originally this PR was going to include the fix to the borg expansion exploit but I am very tired and dont wanna stare at the transform function so this will have to do for now

## Why It's Good For The Game

Miners dont have to bust their backs to mine more iron because a single borg decided to become bigger, that plus its a cosmetic upgrade

## Changelog

:cl:
fix: Changed the cost of the Expansion Module to be 1 sheet instead of 100
/:cl:


softcerv:
This really needs to be properly fixed with a size cap or something upstream.

PR #476, opened 2023-08-21
remove unnecessary files

Cyprex:
These laws seem to be identical in tg files they modularly override, so I'm not sure why they exist.

## About The Pull Request
These were apparently added in the upstream merge. These look identical to tg versions, not sure why they exist. @nevimer


StrangeWeirdKitten:
@Cyprex we can probably merge this if it's still something that needs to be refactored.

PR #494, opened 2023-08-29
Adds a bit of QoL to the monophobia quirk

NithaIsTired:
## About The Pull Request
Checks for plushies nearby so they also count as a friend, hug them away!
fixes it not looking at pets when picked up in hands.
Makes the price more in pair with RDS, as it does have hallucinations symptoms and heart attacks included.

## Why It's Good For The Game
Makes the quirk more tolerable to the few poor souls that decides to play with it. Encourages RP while giving it a unique way to
do so.

Now you have an excuse to just be holding a shark plushie all the time!
dreamseeker_j3rDCTCW12
dreamseeker_Y3S9PHfLHI

## Changelog

:cl:
add: Added a bit of QoL to the monophobia quirk
/:cl:


code/datums/brain_damage/severe.dm
```diff
@@ -189,6 +189,16 @@
...
+
+	for(var/obj/item/toy/plush/plush in oview(owner, 1))
+		if(plush.stuffed)
+			return FALSE
+	for(var/obj/item/toy/plush/plush in  owner.held_items) //Now you have an excuse to carry a shark plushie around all the time
+		if(plush.stuffed)
+			return FALSE
+	for(var/mob/M in owner.held_items) //makes sure to check hands for your tiny friends!
+		return FALSE
+
```
Majkl-J, line 201:
```suggestion
	for(var/mob/M in owner.held_items) //makes sure to check hands for your tiny friends!
		if(!isliving(M)) //ghosts ain't people
			continue
		if(istype(M, /mob/living/simple_animal/pet) || istype(M, /mob/living/basic/pet) || M.ckey)
			return FALSE

```

code/game/objects/items/plushes.dm
```diff
@@ -120,8 +120,13 @@
 		if(grenade && !grenade.active)
 			user.log_message("activated a hidden grenade in [src].", LOG_VICTIM)
 			grenade.arm_grenade(user, msg = FALSE, volume = 10)
+		if(HAS_TRAIT(user, TRAIT_MONOPHOBIA))
+			to_chat(user, span_notice("You feel your heart warm up... You don't feel so alone."))
+
 	else
 		to_chat(user, span_notice("You try to pet [src], but it has no stuffing. Aww..."))
+		if(HAS_TRAIT(user, TRAIT_MONOPHOBIA))
+			to_chat(user, span_warning("A reminder that even stitched companions can't stop the solitude building up inside your heart..."))
```
Majkl-J, line 129:
```suggestion
			to_chat(user, span_warning("You remember that even stitched companions can't stop the solitude building up inside your heart..."))
```

Majkl-J (changes requested):

YakumoChen:
Doesn't thus generally just invalidate the quirk?

Anyway these edits aren't modular so they won't work.

NithaIsTired:
> Doesn't thus generally just invalidate the quirk?
> 
> Anyway these edits aren't modular so they won't work.

It doesn't really invalidate the quirk, it depends a lot on rns so this should make it less annoying if you do want/need to do something alone for some time or maybe even afk urgently, given the chance of just having a heart attack because you walked away from people for a minute and you got bad rng.

You were always able to just bring a pet and it would completely invalidade the quirk anyways if you weren't roleplaying it out. I just aim to add a bit more options to the quirk to make it more fun, giving it another unique way to roleplay it.

As for modularity I can take a look at it later today.

YakumoChen:
> > Doesn't thus generally just invalidate the quirk?
> > Anyway these edits aren't modular so they won't work.
> 
> It doesn't really invalidate the quirk, it depends a lot on rns so this should make it less annoying if you do want/need to do something alone for some time or maybe even afk urgently, given the chance of just having a heart attack because you walked away from people for a minute and you got bad rng.
> 
> You were always able to just bring a pet and it would completely invalidade the quirk anyways if you weren't roleplaying it out. I just aim to add a bit more options to the quirk to make it more fun, giving it another unique way to roleplay it.
> 
> As for modularity I can take a look at it later today.

I feel like there is a profound difference between holding an animal that you have to go out of your way to find, is noisy, and dies in a hostile environment compared to a toy you spawn with. If anything, the responsibility of caring for the pet is more interesting using it to bypass the quirk. 

I don't see how replacing it with a boring roundstart item is good for the qame in comparison.

NithaIsTired:
> I feel like there is a profound difference between holding an animal that you have to go out of your way to find, is noisy, and dies in a hostile environment compared to a toy you spawn with. If anything, the responsibility of caring for the pet is more interesting using it to bypass the quirk.
> 
> I don't see how replacing it with a boring roundstart item is good for the qame in comparison.

It is not replacing it. You can leave pets anywhere in your vision and they will count as a companion. With this you need to hold a plushie to make it count as a companion. Plushies are way more easy to steal, get caught on fire or rip their stuffing and you can't really get another one with the same name after you lose the roundstart one, making them more unique than a random generic pet you got from cargo and is there just because you don't want to randomly get a heart attack.
This is aimed at giving the quirk a bit more freedom to be roleplayed in more ways without being punished by death that is simply RNG. I don't think completely relying on mechanical parts of the game is good.

NithaIsTired:
aaauuugh I'll need to do this PR again... Probably should sleep before I do it this time

PR #501, opened 2023-08-30
Clockcult Cogging Changes

destrucktoid:
## About The Pull Request

This update is a balance change around Cogging APC's as clockwork cultist.

Right now when a cog is put into an APC the following happens:
Over the course of 5 minutes the APC loses 100w every tick followed by (once the cog is initialised) 50w every tick until the cog is removed.
The APC can also no longer charge from the grid.

The changes this PR does are the following:
Cogged APC's now drain 100w every 10 ticks pre-initialisation and 50w every 10 ticks after initialisation (effectively 10w and 5w per tick)
APC's can now charge whilst cogged.
Not a change but important context: APC's by default charge at 5w per tick

This means that post initialisation rather than the APC's no longer charging at all and just being drained till empty whilst still being a drain on the station's power net and creating an alert on the network. The APC's instead will now be slowly drained whilst initialising followed by effectively no longer gaining charge%, but also not losing charge% (they'll still charge, it's just we drain the same amount as they charge by. This is easily distinguishable if someone is looking at the APC thanks to the fact we bulk drain)

## Why It's Good For The Game

Clockwork cultist right now is in an awkward spot. They aren't a full and normal station antagonist, only being able to be gained via an opfor. However, usually these opfors are limited in scope and relatively peaceful. By reducing power drain and making that power drain more stealthy we also make engineering less angry at the clockwork cultist as well as making them more stealthy. 

This change also means that a clockwork cultist no longer has to micro their cogged APC's for power generation when they get low (Remove their own cogs to allow an APC to charge only to add it back once it's charged.) since power generation to the ark was stopped when the APC reached 0, yet the cog still drained power from the grid (possible oversight).

We also provide a buff to increase to the Ark's power maximum from cogging APC's, this is because that due to the slower power gain, in instances where you need a lot of power (such as combat or building/using clockwork structures with the Ark's power), the slower power gain means that you can easily run out of power. Even before this change running out of power mid combat was a very easy issue to run into, however it could be offset by having a cogged APC with a full bluespace cell since the draining of that would usually provide enough power to the ark for most of your needs mid-combat. However, with the changes to power generation (with it now being more passive and less active), having a higher power maximum is required.

## Changelog

:cl:
refactor: changed how cogged APC draining works
balance: Cogged APC's can now regenerate power
balance: Cogged APC's now are only drained for 10w/5w per tick rather than 100w/50w
balance: Cogged APC's now give 500w max power to the ark, rather than 250w max power.
/:cl:

Special Thanks:
Thanks to LDip999 for helping (let's be real basically coding it for me) me do this.


destrucktoid:
Made an oopsie and forgot to create new branch before pushing changes. New PR for the same changes is [here](https://github.com/Bubberstation/Bubberstation/pull/502)

PR #503, opened 2023-08-30
Neurine + Holywater cures magic traumas during lobotomies

BurgerLUA:
## About The Pull Request

Performing the lobotomize step with neurine + holywater present in the patient's system will cure magic traumas.

## Why It's Good For The Game

Near-permanent traumas are generally bad (except in traits). Both chemicals are (relatively) difficult enough to obtain to give the reward.

Untested so who knows if it works.

## Changelog

:cl: BurgerBB
balance: Neurine + Holywater cures magic traumas during lobotomies
/:cl:


Majkl-J:
Considering that borers have a great 5% chance to just gimp your game, among some other things that cause permanent trauma, yea why not.

PR #504, opened 2023-08-30
Illegal Borgs are now Hard(er to obtain because they are now their own tech web that requires a bunch of prerequisites and lots of points). 

BurgerLUA:
## About The Pull Request

Illegal Borg Modules are locked behind a 20000 point tech tree, which requires Illegal Tech and many other tech types.

Draft because untested.

## Why It's Good For The Game

People apparently have a huge issue with Illegal Borg Modules being a thing. This PR makes it so that it is harder to obtain.

## Changelog

:cl: BurgerBB
balance: Illegal Borg Modules are locked behind a 20000 point tech tree, which requires Illegal Tech and many other tech types.
/:cl:


Detective-Google:
adding more point requirements doesnt really actually stop this, research points are usually setting at like 60k with nothing but mech guns to research fairly early into the round regardless. Should just be removed entirely, emagged borgs are very powerful and were never meant to be able to just be printed en masse by the crew.

YakumoChen:
I typically see all these nodes purchased before illegal tech anyway, what's the difference?

destrucktoid:
The issue is that on vanilla TG this is fine. You know why? BECAUSE THERE IS NO SECBORGS. How about instead of doing another stupid change that completely gimps all borgs you just acknowledge that secborgs are bad for the game and remove them.

Strazyplus:
Lot of rounds illegal tech is never obtained and most of the time is during last bit of duration of the round. additionally I have not seen anyone complaining about Illegal Borg Modules.

This is essentially is removal. You will likely never see Illegal Borg Modules if this PR gets merged.

destrucktoid:
> Lot of rounds illegal tech is never obtained and most of the time is during last bit of duration of the round. additionally I have not seen anyone complaining about Illegal Borg Modules.
> 
> This is essentially is removal. You will likely never see Illegal Borg Modules if this PR gets merged.

This is all straight up not true. We see illegal borg modules almost every round, i would know because i play robotics most rounds, furthermore all this change does is adds a 15000 point cost to get the modules which nobody cares about since science usually has plenty of points for it and if they don't then the borgs will still get it anyway because they have access to the consoles and they need their unlimited ammo rapid fire laser gun/illegal hypospray/shock hug module so they can hunt valids easier. Also regarding your point about the last bit of the round, the last bit of the round, say the last hour to half an hour is _the most important time for antagonists_ since it's usually where they go "rounds ending soon, i've got all my gamer gear, time to go loud." _This change does nothing for that_ other than add a 15000 point extra cost which would be paid instantly anyway.

StrangeWeirdKitten:
I'd honestly just support its removal.

StrangeWeirdKitten:
This needs discussion and a decision made on it. But there are also some outstanding merge conflicts.

StrangeWeirdKitten:
@BurgerLUA a flat removal would be what gets merged. Change the scope of this PR and I'll discuss it amongst staff if we should keep this or not.

modular_zubbers/code/modules/research/techweb/all_nodes.dm
```diff
@@ -11,6 +11,14 @@
...
 	research_costs = list(TECHWEB_POINT_TYPE_GENERIC = 1000)
 
+/datum/techweb_node/borg_syndicate_module
+	id = "borgsyndicatemodule"
+	display_name = "Illegal Cyborg Modules"
+	description = "Merges Pull Request #4678: \"Re-adds hidden cyborg modules\", with 2 upvotes and 2856 downvotes."
+	design_ids = list("borg_syndicate_module")
+	prereq_ids = list("syndicate_basic", "adv_robotics", "cyborg", "cyborg_upg_util", "cyborg_upg_serv", "cyborg_upg_engiminer", "cyborg_upg_med")
+	research_costs = list(TECHWEB_POINT_TYPE_GENERIC = 15000)
+
```
Cyprex, line 21: This is commented out for some reason? I believe it happened in the upstream merge commit. Odd.

code/modules/research/techweb/all_nodes.dm
```diff
@@ -2226,7 +2226,7 @@
 	design_ids = list(
 		"advanced_camera",
 		"ai_cam_upgrade",
-		"borg_syndicate_module",
+		// "borg_syndicate_module", BUBBERSTATION CHANGE: MOVED TO /datum/techweb_node/borg_syndicate_module.
```
Cyprex, line 2229: If it gets completely removed it'd need an update.

Cyprex:
I can't speak from experience but I wouldn't be surprised if this module was (too) common. It's rather powerful- I'm in favor of an outright removal though I'd prefer making it more difficult to obtain. But I'm not sure just making it expensive in terms of points is sufficient- perhaps some experiments, perhaps some destructive analyzer stuff? Some antag items, or possibly something like a charged yellow extract(the crossbreed).

My reason is that I think it's useful to get borgs familiar with their malf tools outside of malf rounds, which are rare, and rarely giving them fun toys sounds interesting to me!

Eastwardrope:
I'd say rather than removing it perhaps make it something that is antag only such as with hyjacking requiring them to alt click the console something similar like that so things like Heretics or BBs aren't gimped by the lack of illegal modules if they choose to take the AI to their side

PR #505, opened 2023-08-30
Nerfs Fingerprints + Olfaction: Remade

BurgerLUA:
## About The Pull Request

Fingerprints are now partial and sometimes corrupted. In most cases it is still easy to tell who the fingerprint belongs to, in rare cases it can't.
The more someone touches something, the more fingerprints it leaves.
Fingerprints are no longer sorted by time when the fingerprint was placed.
Fingerprint count on objects has a soft limit of 5, and a hard limit of 8, with the oldest fingerprint being removed.
Olfaction now has a 50% chance to work per fingerprint entry.

## Why It's Good For The Game

Fingerprint scanning is very OP, even if the """counter""" is "Wear heavy gloves."

Also is more immersive.

## Changelog

:cl: BurgerBB
balance: Nerfs fingerprints and Olfaction. Fingerprint readings are scrambled.
/:cl:


BurgerLUA:
me when touching an object gives a near perfect fingerprint or nothing every time

YakumoChen:
95% of crimes I see when playing detective have very obvious / loud sources without needing to scan

This just makes the last 5% I actually get to use my scanner for likely to be unsolvable if I even find prints

Trubbl3:
skill issue, just learn how to change clothes and use gloves, this nerf shouldn't exist

ThinkArbies:
Way to nerf a role that hardly ever gets to be anything more than a sec officer in nicer clothes.

BurgerLUA:
> Way to nerf a role that hardly ever gets to be anything more than a sec officer in nicer clothes.

thanks im proud of it

Tupinambis:
Controversial PR apparently but personally I don't find "click a door and know who touched it" to be engaging as a detective or as an antag. You're not doing much detective work in that case.

Gloves and disguises still are advantageous even with the change, so I don't see a problem.

nevimer:
Alright, we tried it as a testmerge for a bit. It didn't really change anything, but I feel like this best to be take upstream. I appreciate the work that went into this, but it's totally better to be done at tgstation.

PR #506, opened 2023-08-31
Remove Haloperidol from Borgs

Alecksohs:
## About The Pull Request

Remove Haloperidol, a sleep chemical from borgs. Haloperidol is a busted chemical for combat and it has to go. This is a bandaid fix to the instant injection issue, but until I can convince more people, this'll have to do. I will be experimenting with the injection timer. I removed Haloperidol from that list. This won't break.

image

## Why It's Good For The Game

Haloperidol is a busted chemical which puts the victim to sleep, allowing any borg to click-combat insta-inject with modsuit penetration any individual they want, taking them out of the fight in which they can then pump them full of more. Sorry but no.

## Changelog

:cl:
del: Removed Haloperidol from Medical Borg Hyposprays
/:cl:

### DIFF
--- a/code/game/objects/items/robot/items/hypo.dm
+++ b/code/game/objects/items/robot/items/hypo.dm
@@ -10,7 +10,6 @@
 		/datum/reagent/medicine/lidocaine\
 	) //SKYRAT EDIT line 10 added Lidocaine
 #define EXPANDED_MEDICAL_REAGENTS list(\
-		/datum/reagent/medicine/haloperidol,\
 		/datum/reagent/medicine/inacusiate,\
 		/datum/reagent/medicine/mannitol,\
 		/datum/reagent/medicine/mutadone,\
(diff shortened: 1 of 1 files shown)


Alecksohs:
> I think this is a hard change, but one that works as bandaid. I think a better solution would be to make borgs injections NOT instant and possibly not MOD-piercing.

Help me argue this in the discord.

chemistrymain2:
> I think this is a hard change, but one that works as bandaid. I think a better solution would be to make borgs injections NOT instant and possibly not MOD-piercing.

The opposite, honestly.

Cyborg injections being instant is fine, so long as the injections are nonharmful.

Insta-injections are important for mediborgs working as combat support, which is the intended role among other things-to be superior to baseline humans in some respects while losing some of the versatility and higher utility floor available to humans, and being much less effective at self defense.


Haloperidol and piercing hypospray are the problems-_they_ need to go, because the piercing hypospray means that the borg can heal people through modsuits or thick material, and haloperidol is a strong sleep sedative you can't counter.

nevimer:
We will go with this over adding delays. Sorry for making you work more, but sometimes the simple solution is the better one :P

PR #510, opened 2023-09-01
Buffs post-tackle grabs because I used tackle once as a Sol Officer and it didn't work out so now I'm making a PR about it.

BurgerLUA:
## About The Pull Request

Buffs post-tackle grabbing. Landing an expert tackle now gives an aggressive grab. Anything less than that (that isn't a botch) is a passive grab.

Removes the guaranteed concussion on hit with a botched tackle.

## Why It's Good For The Game

Tackles are never used on Bubberstation. I've legitimately never seen an antagonist or a security officer ever use a tackle. I tried using a tackle one round and even after landing an "expert" one, it put me at a serious disadvantage compared to the one who was tackled due to Skyrat's version of stamina combat. This PR buffs tackles, making the grabs it gives stronger.

Also removes the guaranteed concussion on hit with a botched tackle against mobs because that doesn't really make sense since it doesn't check for helmets or other factors when giving it.

## Changelog

:cl: BurgerBB
balance: Buffs post-tackle grabbing. Landing an expert tackle now gives an aggressive grab. Anything less that isn't a botch is a passive grab.
/:cl:


nevimer:
Makes sense to me.

PR #512, opened 2023-09-02
Allow cargo borg mail bags to put things in chutes

the-og-gear:
## About The Pull Request
This allows cargo borgs to use their mail bags to put mail into delivery chutes. This also works through plastic flaps.

Testing may not be comprehensive, though I couldn't think of any cases where it should fail due to the explicit type checks and calling the default proc if those fail. Putting things in the chute does work as intended both through flaps and on the chute itself.

## Why It's Good For The Game
Apparently this was a thing cargo borgs couldn't do, so they relied on a cargo tech to do it for them or had to hand-deliver mail. Weird af, and legitimately not unbalanced to let them do that part of the job when y'know. They can put stuff on tables fine?

## Changelog

:cl:
qol: Cargo borgs can now put anything that fits in their mail clamp into mail chutes. Yes, you can even put it into chutes through plastic flaps, no alt+click required!
/:cl:


YakumoChen:
take it upstream?

PR #514, opened 2023-09-03
Hugboxes Heretic Sacrifices

BurgerLUA:
Draft because untested.

## About The Pull Request

Heretic sacrifices have been changed. Here is what happens on sacrifice. It checks if the victim has a heart. If they do not have a heart, then the sacrifice fails. On success:

The victim's heart is removed.
The victim gains the trait that allows them to liv without a heart.
Shadow handcuffs are applied to the victim.
The victim goes unconscious for 20 seconds. Paralyzed for 24 seconds. Immobilized for 24 seconds.

20 seconds later:

The victim teleports to a safe, lit location. Chapel is prioritized.
The victim is revived and healed a little, if they can be revived.
Vocal stuttering to the target is applied for 40 seconds.
The heretic is informed where the victim now is and if they're dead or alive.
A visible heretic influence is placed where the victim teleported.
The victim gains a brain-surgery curable trauma that causes a phantom that only they can see to appear and chase them. If the phantom gets too close, it can do brute damage.

Victims are no longer teleported to Brazil, and victims can remember being sacrificed if they didn't die. The phobia trauma is replaced with something more lethal, but more curable.

## Why It's Good For The Game

While the introduction of Brazil was something pretty clever, the minigame was easily defeatable if you just ran in an intercardinal direction. It is also pretty annoying to effectively be stuck in Brazil for a minute because of the sacrifice.

Heretics can't be touched due to current metaprotection rules as well as the feature that makes you so you forget being sacrificed. With the removal of forgetting being sacrificed, heretics can now get caught by doing sacrifices. This should allow the round to progress and not making the heretic basically untouchable.

This PR also prevents being sacrificed multiple times (unless you get a new heart) so you can't actively help heretics :^).

## Changelog

:cl: BurgerBB
balance: Removes the Heretic Sacrifice Realm. Replaces it with a feature that just teleports the victim to a safe location with a trauma that can harm if not treated via brain surgery.
/:cl:


BurgerLUA:
Keep in mind that this PR is not final and feedback on features is welcome as long as it is constructive. Nothing is set in stone here.

AJK7370:
remembering being sacrificed wouldn't be a good idea, as it would most likely just encourage the person sacrificed to validhunt the heretic if it wasnt a consensual sacrifice

Majkl-J:
All we need is a trauma that doesnt ruin one's game

StrangeWeirdKitten:
There's a reduction in the trauma given to people to be surgery level. Otherwise, there is consequences if you get sacrificed.

PR #529, opened 2023-09-09
Hugboxes Blob

BurgerLUA:
## About The Pull Request

Blob can no longer go to code delta and end the game.
Blob can expand indefinitely. 
Blob can no longer hold the shuttle.

## Why It's Good For The Game

Blob is a weird mid-round ghost gamemode where it is the only gamemode that can effectively (and literally) end the round. It is also the only mid-round ghost gamemode where it requires the entire crew to cooperative or perish. This PR effectively makes blob a choice to fight, like all roundtypes (except nukies), because forcing everyone to be all hands on death or perish is really bad.

## Changelog

:cl: BurgerBB
add: Blob can no longer go to code delta and end the game. Blob can expand indefinitely. Blob can no longer hold the shuttle.
/:cl:

### DIFF
--- a/code/__DEFINES/blob_defines.dm
+++ b/code/__DEFINES/blob_defines.dm
@@ -13,7 +13,7 @@
 
 // Generic blob defines
 
-#define BLOB_BASE_POINT_RATE 2 // Base amount of points per process()
+#define BLOB_BASE_POINT_RATE 2.5 // SKYRAT EDIT CHANGE
 #define BLOB_EXPAND_COST 4 // Price to expand onto a new tile
 #define BLOB_ATTACK_REFUND 2 // Points 'refunded' when the expand attempt actually attacks something instead
 #define BLOB_BRUTE_RESIST 0.5 // Brute damage taken gets multiplied by this value
--- a/code/__DEFINES/blob_defines.dm
+++ b/code/__DEFINES/blob_defines.dm
@@ -5,7 +5,7 @@
 #define OVERMIND_STARTING_REROLLS 1 // Free strain rerolls at the start
 #define OVERMIND_STARTING_MIN_PLACE_TIME (1 MINUTES) // Minimum time before the core can be placed
 #define OVERMIND_STARTING_AUTO_PLACE_TIME (6 MINUTES) // After this time, randomly place the core somewhere viable
-#define OVERMIND_WIN_CONDITION_AMOUNT 400 // Blob structures required to win
+// BUBBERSTATION CHANGE: HUGBOXES BLOB #define OVERMIND_WIN_CONDITION_AMOUNT 400 // Blob structures required to win
 #define OVERMIND_ANNOUNCEMENT_MIN_SIZE 75 // Once the blob has this many structures, announce their presence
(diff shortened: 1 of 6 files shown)


Detective-Google:
does blob still hold the shuttle

BurgerLUA:
> does blob still hold the shuttle

I completely forgot to make it not do this, thanks for reminding me.

Kraseo:
If people fight the blob, they will probably kill it, continuing the round as usual.

If people DON'T fight the blob (which is the path most people will take if this gets merged), they will just call the shuttle or get overwhelmed when the blob starts spamming tiles, ending the round. Not unlike a blob win.

What's the point of this PR?

YakumoChen:
The whole point of holding the shuttle is to prevent the blob from spreading from the station?

If the blob reaches a critical threshold where the crew effectively can't stop it, how is this different from ending the round automatically?

BurgerLUA:
Give that choice to command instead of it being forced by an arbitrary system.

YakumoChen:
Blob reaching critical mass bypasses what's just as likely to be 10 minutes of the blob curbstomping the crew because a blob that's reached the "round end" goal isn't going to be able to give the crew a fair fight, that's the whole point

Command already has an alternate, nuke-based choice to deny the blob the win

Blob holding the shuttle is biohazard containment.

BurgerLUA:
It's not about winning or losing, it's about giving more freedom to players on how they want to handle the threat and allowing an antagonist's round not be ruined by a midround ghost role.

Majkl-J:
From all my blob fights as antag, I cant pick one that wasn't fun.

Heck I even had a DARKHONK vs blob once. No idea how it's meant to suck out fun from antags, it just adds more opportunity for that rare cooperation

YakumoChen:
> and allowing an antagonist's round not be ruined by a midround ghost role.

Chud antags who see blob as round ruining versus Chad antags who see it as an opportunity for chaos

StrangeWeirdKitten:
Good game modes have an ending. If the blob gets stupid big and everyone is dead, then why not let it just end the round? It already does this. You fight or you die. That's how blob works on TG. It's how blob works on Skyrat. It's how blob works on here. People overall enjoy blob anyway

Tupinambis:
The entire point of delta is that by the time it reaches critical mass it's nearly impossible to fight off anyway. Endless expansion would only delay a functional round end which just seems silly to me.

Oscari-pog:
Blob can be handily defeated by crew given the usual lack of other threats which are very often the reason that blob manages to win without dying. Crew has no excuse to lose given that they are able to pick up a gun and click at the blob.

Or, yknow. If there is literally a singular secborg with the illegal module. Or if engineering sets up 2 emitters aimed at the core. Did you know that reflective tiles get damaged by things they reflect, and that you can entirely avoid the backlash damage to the emitter by just using a reflector? 

If the blob wins, they deserve to have a grand ending.

nevimer:
This is conflicting a little too hard ATM, and overwhelmingly a negative change. We should do this differently if we do.

PR #530, opened 2023-09-09
Threat Level is displayed at the join game menu + newscasters

BurgerLUA:
## About The Pull Request

image

## Why It's Good For The Game

Transparency in the system is good. I was the first one to make it so that the alert level is displayed on join, I will also be the first one to make it so dynamic threat value is also displayed.

## Changelog

:cl: BurgerBB
add: Threat Level is displayed at the join game menu, as well as newscasters + command report, if the config option is enabled.
balance: Intercept messages no longer go to code blue.
/:cl:


ReturnToZender:
This is REALLY good, actually, hopefully it will pass tests

BurgerLUA:
> This is REALLY good, actually, hopefully it will pass tests

oh a typo that I added at the last second is preventing it from compiling lmao

Detective-Google:
I don't think it should show the specific number tbh; but that's a minor gripe.

beyond that I think we should re-enable the threat report if we're gonna do this, since as it stands this is another reason not to  ready up and to instead latejoin so you can check the threat.

BurgerLUA:
> I don't think it should show the specific number tbh; but that's a minor gripe.
> 
> beyond that I think we should re-enable the threat report if we're gonna do this, since as it stands this is another reason not to ready up and to instead latejoin so you can check the threat.

I could probably make the specific number show up when it's hovered over so it looks nicer.

Detective-Google:
> > I don't think it should show the specific number tbh; but that's a minor gripe.
> > beyond that I think we should re-enable the threat report if we're gonna do this, since as it stands this is another reason not to ready up and to instead latejoin so you can check the threat.
> 
> I could probably make the specific number show up when it's hovered over so it looks nicer.

I meant more it should be obfuscated to the players. That's just a matter of opinion though

BurgerLUA:
This PR has been ready for testmerging/merging for a bit now.

StrangeWeirdKitten:
> This PR has been ready for testmerging/merging for a bit now.

I've brought it up to the other senior admins. Keep in mind it was a conscious decision to disable the threat report and obfuscate the threat level until the round ends. If this PR is to be considered, then the threat report would have to be enabled on the config again. I'll come back with a decision later.

StrangeWeirdKitten (changes requested):
This PR needs to respect ``no_intercept_report`` on the config.

StrangeWeirdKitten:
Staff disabled the threat level on the config for obfuscation. For it to be merged for an improvement to the threat report, it needs to respect the config file as per above.

BurgerLUA:
There is literally no reason to close this. Just leave it open so I don't have to rewrite the PR.

PR #361, opened 2023-07-18
(Non-Modular) Comments out Skyrat Modsuit Overrides

StrangeWeirdKitten:
## About The Pull Request

This PR aims to comment out `modular_skyrat\modules\modsuit_armour\modsuit_armour.dm`. It reverts it back to TG armor. This is a massive balance change for armor

**This will need to be test merged**

## Why It's Good For The Game

Skyrat has severely buffed crew side armor values while reducing antag armor values. The most effected are the HoS and Captain modsuit armor values which have 40 percent armor reduction. You can face tank a lot of ammo with these values. The only better armor is the elite syndicate modsuit w/ armor boosting. Then again, if you're fighting the captain. You are also fighting all of security, the blueshield, and crew more often than not. 

Skyrat stacks heavily against their antags and this will hopefully reduce some of the pain in the long run. 

## Testing
**Skyrat Armor Values**

image

**TG Armor Values**

image

## Changelog

:cl: Danger Kitten
balance: Reduces modsuit armor values back to TG standard
/:cl:

PR #531, opened 2023-09-09
Removed skyratified armories, reverting them to tg standard gearsets.

Tupinambis:
## About The Pull Request

Removes the skyrat armory overrides and converts Void Raptor's armory to TG standards. Essentially this means:
CMGs, MCRs, Smartguns, Peacekeeper Shotguns are gone. Eguns, Laser Guns, Disablers, and Riot Shotguns are in.

## Why It's Good For The Game

The game is balanced around security having rechargeable energy weapons pretty much across the board, with their only ballistic option taking time to reload, being a bit harder to store backup ammo for, and punishing misses. The current weapon set flies in the face of balance and can end up screwing the crew or antags over. 

CMRs are the worst offender given that they are functionally just lower damage C-20rs that are very effective at wounding, do a ton of damage, and generally make uplink weapons look bad in comparison with a lot more room for error than standard tg ballistic options with EMP immunity to boot. MCRs go in the direction of being nearly useless without upgrades, and being terrible against threats like blobs due to their overheat mechanic. Moving back to standard TG affair feels like a good step in fixing some of the balance problems skyrat has introduced.

## Changelog

:cl: Tupina
balance: The roundstart armory on all maps has had its skyrat weapons replaced with standard TG weapons.
/:cl:


IgiariValkyr:
Please keep the smartrifle? :<

ReturnToZender:
Absolutely AWESOME, been waiting for this forever

Serxule:
peacekeeper shotguns are actually worse than riot shotguns

StrangeWeirdKitten:
#361 NEEDS to be merged with this if it goes through

Tupinambis:
> peacekeeper shotguns are actually worse than riot shotguns

From what I can tell they're just reskins, but if you could point me to where they actually get nerfed I'll look into changing it.

Serxule:
> > peacekeeper shotguns are actually worse than riot shotguns
> 
> From what I can tell they're just reskins, but if you could point me to where they actually get nerfed I'll look into changing it.

peacekeepers have a smaller mag capacity, i think its 2 less than the riot shotty

YakumoChen:
I swore we had a PR that did this already but I think it was overwritten by skyrat code

anyway when are we getting the red security vendors in the armoury again

nevimer:
The problem with editing voidraptor is that it may not stay forever, but, merging anyways.

PR #533, opened 2023-09-10
Small Slappy [ Salt PR ] 

Detective-Google:
## About The Pull Request

no you should not be able to craft an esword because you had 1000 credits and engineering access lying around.
codes out both 'cargo' methods of acquiring slappy parts. If you want your free esword convince the admins you deserve one.

## Why It's Good For The Game

40 damage weapon that does knockdown, knockback, disarms, and has innate block. the supposed downside is "le slowdown" which is easily counterracted by both xenobio and robotics powergay.
firefox_HX0kKttLui

## Changelog

:cl:
balance:big slappy kits are no longer conventionally obtainable.
/:cl:

before you unironically assume this is a salt PR I've not died to this weapon.
https://cdn.discordapp.com/attachments/705679956021215264/1141159579695198218/bigslappy.mp4


Detective-Google:
image
 the stupid heretic rune in the middle of space is erroring biodome, truly one of the maps of all time.

nevimer:
wow, this is a thing?

IgiariValkyr:
Big slappy also has a high chance to backfire and cause damage to you, via wounds and dislocations. Though I agree it could be balanced further.

Detective-Google:
> Big slappy also has a high chance to backfire and cause damage to you, via wounds and dislocations. Though I agree it could be balanced further.

You are welcome to create an alternative pr that balances this further than something that can be entirely ignored with (1) brute medkit and/or a beaker of libital

Detective-Google:
to be more specific, the self damage is more or less inconsequential because libital, which is very easy to come by, heals at 3 brute per tick(every 2 seconds), while the weapon's delay is 1.3, meaning that even assuming you are landing every hit as the click cooldown comes off, the self damage is negligable, 

1.5 healing per second vs 7.6 damage per second; meaning the effective damage you're taking if you hit every time click delay is off cooldown is 6.1, which, at skyrat's buffed health levels, means you're able to hit someone 22 times and do **660** damage before you'd hit 135( and even then, you can stay fighting for longer than that if your mood is good.)

 in short, "but it hurts you too" is not an excuse for a crew sided weapon, because any crew member who breaks their leg chasing down an antagonist can just go to medbay and get it fixed, the antagonist they are validhunting with it has to use up limited healing resources on the go to deal with the damage.

given the fact that players absolutely do not land hits every time click delay is off generally. it should be considered that generally speaking the crew is not given access to any* melee weapons above 30 damage, for good reason. weapons like the fireaxe of which you are limited to two are only 25 damage, chainsaws are 26. 

The only real exception to that is mechs; some of which have higher punch damages; coming with their own downsides which I don't think need listing.

This is also assuming the kind of person who can be arsed to powergame enough to get one of these is above going to chem and mixing helbital (which is three ingredients with almost no pH fuckery), or salicylic which heals 4.  (it heals less under 25 damage but if you're using this chances are you'll be over that threshold most of the time.) if you make a 50/50 beaker of salicylic acid and libital you're healing ~7 damage every 2 seconds; or 4.5 out of the 6.1 you could possibly deal to yourself.

PR #545, opened 2023-09-14
Ion Storm announce chance increased from 33% to 100%.

BurgerLUA:
## About The Pull Request

Ion Storm announce chance increased from 33% to 100%.

## Why It's Good For The Game

Ion laws are kind of insane when you think about it. 

25% chance to replace the entire lawset with a completely different template lawset.
10% chance to remove a random law.
10% chance to replace a random law with something else.
10% chance to shuffle the order of the laws randomly.

This isn't "chance to do one of those" it's literally 4 different dice rolls on changing the AI's laws. In most scenarios, the AI will usually have some sort of serious damage to the lawset that will allow them to basically antag without antag status. Instead of removing that fun feature, I decided it would be just best to make it so that the crew is aware of an ion law so it is more balanced.

## Changelog

:cl: BurgerBB
balance: Ion Storm announce chance increased from 33% to 100%.
/:cl:

### DIFF
--- a/code/modules/events/ion_storm.dm
+++ b/code/modules/events/ion_storm.dm
@@ -16,7 +16,7 @@
 	var/botEmagChance = 1
 	var/ionMessage = null
 	announce_when = 1
-	announce_chance = 33
+	announce_chance = 100 //BUBBERSTATION CHANGE: 33% to 100%.
 
 /datum/round_event/ion_storm/add_law_only // special subtype that adds a law only
 	replaceLawsetChance = 0


Oscari-pog:
Honestly making it not announce is funny, I'd just flip the values. 33% chance not to announce, instead of to announce.

BurgerLUA:
> Honestly making it not announce is funny, I'd just flip the values. 33% chance not to announce, instead of to announce.

I might work on this tomorrow and make 2 different ion storm events: One serious and one funny. The serious one is normal ion law behavior which is announced, and the funny one is adding a randomly generated law which isn't announced.

destrucktoid:
This is in my opinion a bad change. Ion laws are random and having an AI having a weird random Ion law is fun, usually the law is reset pretty quickly if it's a bad one whereas if it's a more quiet or funny one they tend to remain in place as the crew doesn't bother to reset them. Like, what problem is this PR addressing other than making a round less fun?

BurgerLUA:
Reopening with different changes.

PR #549, opened 2023-09-15
[CLOSE AFTER UPSTREAM] Fixes ctf reloads and some other unrelated behavior (Part 1 - upstream edition)

Majkl-J:
## About The Pull Request

Pushing upstream, we can have this testmerged until then
Why was there a random caliber override?

There's more stuff I have in the works after this is merged

## Why It's Good For The Game

A bugfix of shitcode

## Changelog

:cl:
fix: Laser ammo mags now recharge correctly
/:cl:


Majkl-J:
This has now been merged upstream, so it shall be closed when we update

PR #551, opened 2023-09-15
The announce chance of the Ion Law event is based on what is affected.

BurgerLUA:
## About The Pull Request

The base announce chance for ion laws is 0%.
Increases by 0% if the event adds a random ion law.
Increases by 25% if the event changed the lawset to another template lawset.
Increases by 50% if the event removed a random law.
Increases by 75% if the event removed and replaced a random law.
Increases by 75% if the event shuffles all the laws.

## Why It's Good For The Game

Ion laws are kind of insane when you think about it.

25% chance to replace the entire lawset with a completely different template lawset.
10% chance to remove a random law.
10% chance to replace a random law with something else.
10% chance to shuffle the order of the laws randomly.

This isn't "chance to do one of those" it's literally 4 different dice rolls on changing the AI's laws. In most scenarios, the AI will usually have some sort of serious damage to the lawset that will allow them to basically antag without antag status. Instead of removing that fun feature, I decided it would be just best to make it so that the crew is aware of an ion law depending on what it does so it is more balanced.

## Changelog

:cl: BurgerBB
balance: The announce chance of the Ion Law event is based on what is affected.
/:cl:


YakumoChen:
Fun removal
This will make ion announce every single time, which will invariably cause an immediate reset. The AI does not get a chance to do any roleplay or actions with their new laws.

PR #552, opened 2023-09-15
Overhauls Jukebox Sound

BurgerLUA:
## About The Pull Request

I spent an absurd amount of time tweaking the jukebox formula and getting the code to work so please clap.

This PR basically makes it so that the sound Jukeboxes make is more pleasant by doing the following:

- Jukeboxes can go to 100 volume, but the actual volume that players can hear is volume capped to 50. The higher volume just affects how far away the jukebox can be heard from.
- Jukeboxes will play out of both left and right speakers/headphones unless you're far enough away.
- The maximum distance at which Jukeboxes play is significantly reduced.

## Why It's Good For The Game

Jukeboxes currently have a bug where they are 100 volume by default, and can be set to that via pressing the reset button. This fixes that and adds a custom sound system so that music doesn't exclusively play in one ear if you're to the left or right of a jukebox.

## Changelog

:cl: BurgerBB
add: Significantly improves Jukebox sound calculations.
/:cl:


Oscari-pog:
Playing into both ears is a fucking blessing Holy shit

PR #570, opened 2023-09-24
early pull: https://github.com/tgstation/tgstation/pull/78564

nevimer:
Volume pump UI and logging pump rate now scales with seconds_per_tick. I don't think we can access that for atmos, so I'm just doing it manually instead.
https://github.com/tgstation/tgstation/pull/78564

People must see the error of their ways.


StrangeWeirdKitten:
This was closed on TG. If we add it hear, it needs proper bubber edit documentation comments and a changelog.

StrangeWeirdKitten:
Take this upstream

PR #574, opened 2023-09-25
Revert "Fixes vents having "infinite" pressure caps. (#77686)"

nevimer:
TGstation did this thing where they broke atmospheric mechanics by introducing a non communicated mechanic. This revert will allow supermatter setups to actually recover from fires again without having to extinguisher the chamber. Replacing old solidified universal to every codebase mechanics with magical volume pumps that defy the rules is not a good change. 

I feel strongly on having consistent device behavior with NO undefined arbitrary invisible restrictions.


ChakatStormCloud:
Honestly, I thought the vents **always had** that 4500kpa limitation, everything **else** does. The only exceptions have ever been volume pumps and injectors, which are just volume pumps into tiles.
but the effect on enforcing that limitation on the SM is just... bad, and the solution they came up with, making the vent "overclockable" just doesn't fix it, it just destroys the vent and leaves you with at best, 2 vents, and at worst 0.

BurgerLUA (approved):

StrangeWeirdKitten:
This needs bubberstation edit comments. Especially on a system like atmospherics. Plus a changelog

Deadgebert:
```
if(environment_pressure >= 50 * ONE_ATMOSPHERE)
	return FALSE
```
it looks like you didn't remove the pressure caps, you've just disabled the overclocking. This will not allow the SM to recover on its own

PR #579, opened 2023-09-28
Added *chirp, modified *caw

Birdiest-Bird:
## About The Pull Request

Added *chirp cuz it's essential for a birdier communication. Also followed the same structure for emotes as modular skyrat so more custom emotes can be added there in the future without it being messy!

And yes, I've tested it and it works, woo!

## Why It's Good For The Game

Birds are good. Bird communication is also good.

## Changelog

:cl: Caw!
add: Added *chirp, modified *caw and *caw2
/:cl:


nevimer:
I think we might have one or two of those already.

Birdiest-Bird:
> I think we might have one or two of those already.

Yeah! we do have a caw and caw2 ogg files, but I remember giving the source audio back when skyrat rebased to current tg it's a very edited audio of a crow video I got, I can share if you'd like! But now I got new audio files for the caws and, most importantly, added *chirp for all bird chirping purposes

nevimer:
bird up!

PR #585, opened 2023-10-01
[TESTMERGE READY] [FULLY-MODULAR] [NOW WITH 99% LESS MEMORY LEAKS] Gives synths internal computers

Majkl-J:
## About The Pull Request

Part 2 of my synthcentric update series
See name, synths now have internal computers.

### Progress + features

- [x] Added the computer to the brain
- [x] Added action to open the computer
- [x] Made the checks that make the computer accessible work (What do you mean UI_UPDATE)
- [x] ~~Made it consider the synth's ID slot its own ID slot~~ added a way to insert the ID
- [x] Allow other people to open it

### Usage guide

To insert a id, target eyes and hit the synth with it.

To interface with the computer, get your pda, target eyes, and hit the synth with it

## Why It's Good For The Game

They are walking computers but aren't given an internal ntos computer. This rectifies that.

Makes races more diverse and interesting!

Allows you to play minesweeper on your synthetic GF

## Changelog

:cl:
add: Synths are now equipped with an internal computer
/:cl:


Majkl-J:
Turns out I added a memory leak. Hurray!

Eastwardrope:
So I found 2 bugs, 1 the insert IDs works but refuses to let you eject cards.

The second glitch I found was that it won't read gases around the synth

the third is it has connection even when connection is otherwise impossible such as deepspace or Tcomms damage

Majkl-J:
> So I found 2 bugs, 1 the insert IDs works but refuses to let you eject cards.
> 
> The second glitch I found was that it won't read gases around the synth
> 
> the third is it has connection even when connection is otherwise impossible such as deepspace or Tcomms damage

Number one is gonna need more details since I had been told by several people it works, gonna need more details.

Number two is an issue with the programs themselves checking the loc of the computer, which is null, instead of the physical I have no idea how to fix this without major refactoring that would also be nonmodular

Number three is intentional but TComms damage should still take it out

StrangeWeirdKitten:
@Majkl-J no single letter variables. Change ``T`` to ``target`` for example. Otherwise, I can probably test merge this and see how it goes.

StrangeWeirdKitten (changes requested):
No single letter variables. Change ``T`` to ``target`` as an example for better code readability.

StrangeWeirdKitten:
Also fix linters

Majkl-J:
> No single letter variables. Change `T` to `target` as an example for better code readability. @Majkl-J

Not doing that. It's constrained to one proc of at most 10 lines, it's like using i in a for loop and is a standard code practice. Nobody is gonna go "I wonder what this variable means" when they can look 5 lines above to see exactly what it does, with their IDE probably even highlighting it

StrangeWeirdKitten:
> I put a blank mining ID gotten from the conscription package in cargo into the slot cause I was curious if it'd let me use that and my normal ID as IDs it does not. The ID refused to come out even when I clicked the eject ID button. Might be linked to the fact it wasn't my ID

This is a TG bug and I informed Waterpig about this fact. So should be getting fixed on TG.

StrangeWeirdKitten (approved):

PR #597, opened 2023-10-07
Removes Engineering Cyborg Electric Welder (+ modularization)

StrangeWeirdKitten:
## About The Pull Request

**This PR Primarily removes the borg electric welder.** (If this is such a hot debate, I'll atomize the below changes later)

I've also taken the initiative to add some modularization to the codebase by moving all the techweb stuff to ``New()`` inside our all_nodes folder with the borg plasma cutter.

Basically, modularize the tech web and designs.

## Why It's Good For The Game

Basic code organization and modularization put away,

The Electric Welder is powerful. It is extremely powerful. It is an infinite charge almost no delay welder from skyrat with a tool speed of 0.2 given to a machine that can use it to repair itself to full health in one second. I think the ability of an engineering cyborg to run into combat, fight a blob, pull back, and repair itself to full health without any expenditure of welding fuel nearby or time, is a severe balance issue. This should not have been added to the game. It also makes a very decent melee weapon since it's not going to run out of fuel.

## Changelog

:cl: StrangeWeirdKitten
del: Engineering borgs no longer have an electric welder upgrade. 
refactor: Modularized and moved all our techweb edits into one folder. 
/:cl:


Majkl-J:
Ok how about, and hear me out here: We make it use the borg's charge and slow down its self use

StrangeWeirdKitten:
> Ok how about, and hear me out here: We make it use the borg's charge and slow down its self use

That was thought of, but if we made it use the cell charge, then I see no difference between a borg having to run to a recharger versus a welding tank when extensively fighting with it. It just seems redundant versus removing it entirely. The entire point of the electric welding tool is it's basically an infinite welder with a tool speed of 0.2. I can set the tool speed back to 0.5 and add a penalty to the borg's cell and it'll basically be the unupgraded welding tool again. It's also not just the self-use.

Majkl-J:
> > Ok how about, and hear me out here: We make it use the borg's charge and slow down its self use
> 
> That was thought of, but if we made it use the cell charge to a noticeable amount, then I see no difference between a borg having to run to a recharger versus a welding tank when extensively fighting with it. It just seems redundant versus removing it entirely. The entire point of the electric welding tool is it's basically an infinite welder with a tool speed of 0.2. I can set the tool speed back to 0.5 and add a penalty to the borg's cell and it'll basically be the unupgraded welding tool again. It's not just the self-use that I am considering.
> 
> The welding tool they get is already a good welding tool anyway.

Ah fair point, do carry on

PR #601, opened 2023-10-09
Vetted Player Locks

nevimer:
## About The Pull Request

This PR makes it so that players who are vetted receive a message on inspect in game, as to indicate their status within the player vetting system. Staff in game will also be able to know who is and isn't. 
MGJNhWOmc3
dreamseeker_bG2t0qSBZ6
3VbJKWQA4j

Mechanical Organic Interface Activity is also locked behind being vetted now.

## Why It's Good For The Game

Ambiguity is bad in this case, so this adds incentive on both sides to get vetted. 

## Changelog

:cl:
add: Added new player ranks for being vetted.
code: Organic Interface is now locked behind being vetted. 
/:cl:

### DIFF
--- a/code/modules/admin/verbs/admingame.dm
+++ b/code/modules/admin/verbs/admingame.dm
@@ -43,7 +43,10 @@
 
 		if(SSplayer_ranks.is_veteran(M.client, admin_bypass = FALSE))
 			player_ranks += "Veteran"
-
+		//BUBBER ADDITION START
+		if(SSplayer_ranks.is_vetted(M.client, admin_bypass = FALSE))
+			player_ranks |= "Vetted"
+		// BUBBER ADDITION END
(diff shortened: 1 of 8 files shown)


BurgerLUA:
Honestly, nice.

YakumoChen:
I can get like, having it be useful for staff to know who they've vetted, but I don't think I want nor should need a mechanical incentive to hand part of my real life ID to use ERP tools

bunny232:
I will say adding the text to the look closer examine would be nice, the current examine is already hella clogged and honestly lags my game a bit when im trying to quick examine people.

ReturnToZender:
This needs to say "vetted as 18+" or basically explain what vetted is in a brief way in it.
There's a good few people who don't actually know what getting vetted is, or why we do it.

BurgerLUA:
I think it should only show up when examining ooc notes or w/e.

PR #604, opened 2023-10-11
[FULLY MODULAR] New quirk: Hungry

Majkl-J:
## About The Pull Request

Adds a new neutral quirk, hungry, with the current effect of doubling the decay of one's nutrition.
Neutral because Moffy told me to make it so, might make it a -1 or add something positive to it if others have a different opinion.

## Why It's Good For The Game

Even more reason to interact with the kitchen for those who take this quirk. Please take it I need someone to talk to

image

This is not a quirk for those fat fetishists to abuse like some of you claimed, they already have their one

## Changelog

:cl:
add: New quirk: Hungry
/:cl:


BurgerLUA (approved):
Code seems fine, however I personally have an issue traits causing other people to suffer. It's going to suck when 1 person requires the same nutrition as 2 people, but I guess that doesn't really matter since botany exists.

PR #608, opened 2023-10-14
Replaces the Blueshield's Sindano with the WT-550 (and makes it only admin spawned)

StrangeWeirdKitten:
## About The Pull Request

This replaces the blueshield's Sindano with the WT-550. 
I also unticked ``modular_skyrat\modules\sec_haul\code\guns\cargo_stuff.dm`` which gives 3 Sindano's for 4000 credits from cargo.

I also made it a gun case. 
image

## Why It's Good For The Game

I don't need to explain why giving the crew a round start nukie tier weapon is bad for the game. If I edited the values, then I would have to maintain these guns. I honestly don't care enough for Skyrat guns and would rather revert them all back to TG.

## Changelog

:cl: StrangeWeirdKitten
del: Fully removes the ability to order the Sindano
balance: Replaces the Blueshield's Sindano with the WT-550
/:cl:


YakumoChen:
Un-including files does not work. It will be reverted in the next upstream bundle. Just comment out the whole file.

StrangeWeirdKitten:
Weh

PR #613, opened 2023-10-15
[DNM] Moves the claw implant to 6 points

Majkl-J:
## About The Pull Request

This is one of the possible resolutions to the whole claw discussion.

The alternative is nerfing the weapon, as removal is downright just stupid.

## Why It's Good For The Game

We have decided that a roundstart weapon with less power than a kitchen knife is apparently too good to be only a 4 quirk point cost.

See here:
https://discord.com/channels/1059199070016655462/1162423073673322537

## Changelog

:cl:
balance: Razor claw implant point cost 4 -> 6
/:cl:

### DIFF
--- a/modular_zubbers/modules/customization/modules/client/augment/implants.dm
+++ b/modular_zubbers/modules/customization/modules/client/augment/implants.dm
@@ -1,0 +1,5 @@
+/datum/augment_item/implant/l_arm/razor_claws
+	cost = 6
+
+/datum/augment_item/implant/r_arm/razor_claws
+	cost = 6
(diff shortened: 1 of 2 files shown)


nevimer:
I don't know... punching is more robust but doesn't cause bleed...

Majkl-J:
> I don't know... punching is more robust but doesn't cause bleed...

Do chime in on it you're the maintainer here

StrangeWeirdKitten:
> I don't know... punching is more robust but doesn't cause bleed...

The claw implant set to knife mode are a subtype of knives and they have the same attack value. But they're also undroppable. I would rather remove all its attack force and make people to buy the sharpening stone from an emagged cargo console if they want to cause any real damage with them.

StrangeWeirdKitten:
what I said before

PR #615, opened 2023-10-16
Makes paper bins craftable

Majkl-J:
## About The Pull Request

See name.

## Why It's Good For The Game

Good to make something assistants can steal for bounties renewable

## Changelog

:cl:
qol: made paper bins craftable
/:cl:


nevimer:
I think the whole bounty system sucks and I would remove it entirely

Majkl-J:
> I think the whole bounty system sucks and I would remove it entirely

Too bad I like it and it's better than the old one

BurgerLUA (approved):
Looks fine, but I wonder if there would be a better way to just make it spawn an empty paper bin, not requiring paper.

nevimer:
this is the kind of jank ss13 is about

PR #616, opened 2023-10-17
Caps midround budget to current player count.

BurgerLUA:
Draft because untested.

## About The Pull Request

The midround threat allowance is hard capped to current living player count. This means that the midround threat can never be higher than the current threat count, and will increase or decrease as players leave/join/die/revive. This does NOT affect starting threat budget, so nukeops can still roll.

## Why It's Good For The Game

Prevents the game from spawning an absurd amount of antags during lowpop and midpop hours. Highpop would be mostly unaffected unless the game decides to allocate 50+ threat to midround budgets.

## Changelog

:cl: BurgerBB
balance: The midround threat allowance is hard capped to current living player count.
/:cl:


StrangeWeirdKitten:
Dynamic already does this with a linear interpolation if it's set up properly in the config. We just had the low pop threshold set to 10, which is why the threat was going off the rails. Basically at 10 people, the maximum possible threat was already at 100. I've made some major readjustments. If it's still an issue, I can run more dynamic simulations and readjust the curve.

PR #623, opened 2023-10-19
Changes Networked Fibers expansion point bonus from 4 to 3

BurgerLUA:
## About The Pull Request

Changes Networked Fibers expansion point bonus from 4 to 3.

## Why It's Good For The Game

Networked fibers is generally the go-to meta for blob players if they want to rush code delta. The bonus they get from expanding is pretty huge and if the blob is stealthy, it can result in a very easy 2 early juggernauts. This may be fine for /tg/ players who are usually always on the move and doing something, but not for Bubberstation players where a majority of people just stay in one department and roleplay.

## Changelog

:cl: BurgerBB
balance: Changes Networked Fibers expansion point bonus from 4 to 3.
/:cl:


StrangeWeirdKitten:
See #647

PR #627, opened 2023-10-19
Gives secmed back basic medical access

Majkl-J:
## About The Pull Request
Gives secmed morgue and medical accesses.

## Why It's Good For The Game

Please for the love of god they don't even get crew trackers and medbay just lets them inside to get whatever they need.

They dont even get guns anymore and sec would rather steal their medkits for themselves, it's better they can maybe get stuff from med so they can do the bare minimum of their jobs.

## Changelog

:cl:
add: Secmeds got medical access back. This is so they can get refills without bothering everyone and actually makes them able to do their job
/:cl:


StrangeWeirdKitten:
Unironically security medic is just redundant and I don't see why a paramedic can't just do this job.

PR #583, opened 2023-09-30
Ports Gimmick Traitor from Daedalus Dock (and raises starting TC)

nevimer:
## About The Pull Request
mstsc_mBeessMCFA
mstsc_821RI1VnPn

ports the meat from https://github.com/DaedalusDock/daedalusdock/pull/564

## Why It's Good For The Game
I don't think anyone but the player of traitors enjoys progressive traitor. This replaces it with the gimmick traitor goals, which are much more entertaining for everyone. Big QOL.

## Changelog

:cl:
add: Traitor Most traitor goals have been replaced with gimmick traitor goals. 
/:cl:

PR #633, opened 2023-10-21
Removes traitor surplus crates, buffs starting TC cost

BurgerLUA:
PR that added it in question: https://github.com/tgstation/tgstation/pull/71869

It's a good PR for /tg/ but not for Bubbers.

## About The Pull Request

Increases base starting telecrystals from 20 to 30.
Removes surplus crates, shared surplus crates, and shared surplus crate keys from the uplink.

## Why It's Good For The Game

I've had multiple traitor players tell me or admit in general channels that since the change in traitor progression, they have now resorted to purchasing the surplus options every time because they get more bang for their buck. In my opinion, there are several problems with this (as listed below), so they were removed and the starting TC amount buffed.

### "I bought it, thus I must use it."

In most cases, where a traitor gets a random item, they feel the need to use it, regardless of objectives, circumstances, and escalation, even if they wouldn't normally use it. For example, a traitor with two steal objectives gets a syndicate bomb in a surplus crate. What would they do with it? Most players would just use it to straight up bomb a department, either as a distraction or to break into an area, because why wouldn't they? They have it, and if they don't use it that's pretty much giving a free syndicate bomb to security. In most situations, if they didn't have that bomb, they'd never think of buying it to break into an area because there are much smarter purchases to make.

### "Hey, my good friend Joe is a traitor. We should buy a shared surplus crate."

/tg/ added a shared surplus crate to encourage betrayal or cooperation between traitors. This is a very interesting and \~soulful\~ mechanic, but the problem is that it is generally only a purchase made with someone you trust entirely, such as someone you're good friends with. It unfortunately gives an advantage to people who are friends with eachother, as they are less likely to betray you in the end.

### "Hey I should betray this other traitor and get their shared surplus crate loot so I have 80TC worth of items."

I haven't heard of this happening yet, but there is really nothing stopping a traitor from being a dick and stealing 80TC worth of items from another person. If this happens, you're effectively ruining someone's round over greed, you now have 80TC worth of items, and you're essentially removing a traitor's power from the round by taking all their TC.

## Changelog

:cl: BurgerBB
balance: Increases base starting telecrystals from 20 to 30. Removes surplus crates, shared surplus crates, and shared surplus crate keys from the uplink.
/:cl:


YakumoChen:
Good PR. I'd be tempted to say if you get backstabbers over the shared crate, oh well, but we have 3 hour rounds, not 40 minute rounds.

Tupinambis:
Strongly agree with this PR, not sure how it will end up meshing with gimmick traitor TC wise if both get merged but this on its own seems good.

StrangeWeirdKitten:
#583 already makes traitors have 25 TC which is extremely generous. If you need more that that, then you're probably doing a gimmick that should be opfored

projectkepler-ru:
Why do you need 30 telecrystal we already get 25 @BurgerLUA

BurgerLUA:
> Why do you need 30 telecrystal we already get 25 @BurgerLUA

Because of the removal of surplus crates. If you think 25 is enough, then I can revert that change and just make the PR remove surplus crates.

code/__DEFINES/antagonists.dm
```diff
@@ -275,7 +275,7 @@ GLOBAL_LIST_INIT(human_invader_antagonists, list(
 #define MAROON_PROB 30
 
 /// How many telecrystals a normal traitor starts with
-#define TELECRYSTALS_DEFAULT 20
+#define TELECRYSTALS_DEFAULT 30 //BUBBERSTATION CHANGE: 20 TO 30.
```
StrangeWeirdKitten, line 278: Discard this and set it back down to 20. We have another test merge PR right now that sets it to 25.

StrangeWeirdKitten (changes requested):

StrangeWeirdKitten:
I'll ask the other admins if they've noticed anything with the surplus crates and if they feel like it needs a removal.

projectkepler-ru:
> > Why do you need 30 telecrystal we already get 25 @BurgerLUA
> 
> Because of the removal of surplus crates. If you think 25 is enough, then I can revert that change and just make the PR remove surplus crates.

I'm not sure about other traitor and how they play but atm 25 TC is enough for me to get more than what I need (provided security arent powergaming and armed to hell less than 1 hours in)

30 TC is fine if security are getting a buff anytime soon- but I'd prefer to becareful with how it's handled because that's one way to turn it into a nuclear arm race

BurgerLUA:
> > > Why do you need 30 telecrystal we already get 25 @BurgerLUA
> > 
> > 
> > Because of the removal of surplus crates. If you think 25 is enough, then I can revert that change and just make the PR remove surplus crates.
> 
> I'm not sure about other traitor and how they play but atm 25 TC is enough for me to get more than what I need (provided security arent powergaming and armed to hell less than 1 hours in)
> 
> 30 TC is fine if security are getting a buff anytime soon- but I'd prefer to becareful with how it's handled because that's one way to turn it into a nuclear arm race

The PR can be testmerged, and if people feel that 25TC isn't enough, it can be upped to 30TC.

projectkepler-ru:
> > I'm not sure about other traitor and how they play but atm 25 TC is enough for me to get more than what I need (provided security arent powergaming and armed to hell less than 1 hours in)
> > 30 TC is fine if security are getting a buff anytime soon- but I'd prefer to becareful with how it's handled because that's one way to turn it into a nuclear arm race
> 
> The PR can be testmerged, and if people feel that 25TC isn't enough, it can be upped to 30TC.

yeah understandable.

StrangeWeirdKitten:
After some discussion, the surplus crate isn't really something that's an issue in need for a full removal. Plus we are eventually going to most likely drop TC back down to 20. Both versions are a gamble, realistically. I got staff input and they don't see player issues significant enough to justify a full removal.

BurgerLUA:
If you drop TC cost back to 20, people are just going to buy more surplus crates.

PR #638, opened 2023-10-23
Nerfs Blob's ability to space things based on temperature

BurgerLUA:
## About The Pull Request

Blob atmos only passes if the temperature is above 15.56 degrees Celsius.

## Why It's Good For The Game

Atmos processing has evolved over the years while blob balance (with respect to that) has vastly remained untouched. Some servers I've played on have handled this by making blob tiles block atmos, but I think I can add that and make it better.

## Changelog

:cl: BurgerBB
balance: Blob atmos only passes if the temperature is above 15.56 degrees celcius.
/:cl:


projectkepler-ru (approved):

StrangeWeirdKitten:
See #647

PR #642, opened 2023-10-24
removes the hos beacon entirely

ReturnToZender:
## About The Pull Request
This makes it so the HoS gun beacon doesn't show up, as intended

## Why It's Good For The Game
The HoS beacon was meant to not show up, it looks like it doesn't, but just to be sure I've fixed this code

## Changelog

:cl:
Del: commented out gun beacon
/:cl:


YakumoChen:
The assumption was that someone would fix it later, but you might as well remove the modular file too if you're doing that

BurgerLUA (approved):

projectkepler-ru (changes requested):
Remove all associated files with it, I have no intention of returning to fix it.

PR #647, opened 2023-10-25
Nerfs Blob

BurgerLUA:
## Notice to Maintainers

These are some pretty sweeping changes that should be tested before a merge (of course). I recommend asking players in the middle of a round if they want a chaotic antag to spawn, and if the poll passes you just spawn a blob event :^).

## Notice to Players

Everything is subject to change. Maybe something that worked well in paper doesn't really work well at all in this PR. If that happens, leave feedback here and describe the situation. Posts that just say "This stat edit is shit." without explanation is useless to me.

## About The Pull Request

Nerfs a lot of blob defines to how they were previously back in the day, or what I feel is best given the current state of what the crew has to deal with the blob.

Here are all the changes.

| What was Changed | Old Value | New Value | Reason |
| - | - | - | - |
| Blob starting points | 60 | 0 | Puts a bigger emphasis on not getting caught out early, as well as balances it for the fact that people on Bubberstation don't move around too much compared to upstreams due to the nature of roleplay. |
| Blob Expand Cost | 4 | 3 |  Since spawned blobs start out weaker, they are now also cheaper. |
| Blob Attack Refund | 2 | 0 |  Making attacks a little more expensive (basically from 2 to 3 ) will prevent easy breaching of areas and make open areas a little more rewarding compared to cramped areas. |
| Blob Node/Core Reinforce Chance | 2.5 | 0 | The chance is so low, good or bad rng shouldn't determine whether or not a blob wins. Tile reinforcement is cheaper too, so it isn't as bad of a change. |
| Blob Reagent Inject | 25 | 10 | Injecting 25u of a blob reagent per attack is legitimately insane. |
| Blob core automatic expansion range | 3 | 4 | Makes it easier for blobs that start out. |
| Blob node automatic expansion range | 2 | 3 | Makes it more rewarding to invest into nodes first. |
| Blob node automatic reflector range | 0 | 1 | Nodes given automatic reflector creation makes it so that smart node placements are well rewarded. |
| Blob Factory Max Spores | 3 | 5 | Spore HP was nerfed, so to compensate, the numbers were increased. |
| Blob Resource Max HP | 60 | 100 | A buff to make the nerf to resource nodes not so bad. |
| Blob Resource HP Regen | 15 | 3 | 15 HP regeneration is INSANE. Reduced to something more reasonable. |
| Regular Blob Turf Starting HP | 24 | 10 | Blob tiles should start out very weak as to not make it seem like you're fighting something endlessly when the blob can just shit out a 24 HP structure every time it falls. |
| Blob Reflector HP | 150 | 100 |  150 for a reflector blob seems a little strong given that security doesn't really have ballistics anymore. |
| Strong Blob Upgrade Cost | 15 | 10 | Since blobs turfs spawn weaker, you're going to need a lot of these. |
| Reflector Blob Upgrade Cost | 15 | 10 | Matches the upgrade cost of strong blobs. 
| Strong Blob Refund Amount | 4 | 0 | Prevents an exploit where you refund a strong blob before it is destroyed. |
| Reflective Blob Refund Amount | 8 | 0 | Prevents an exploit where you refund a reflective blob before it is destroyed. |
| Resource Node Refund Cost | 15 | 40 | Resource nodes contain resources, yes? |
| Factory Node Refund Cost | 25 | 30 | 30 was chosen because it is half of the factory cost of 60 |
| Blob Relocate Core Cost | 80 | 100 | Relocating your core is extremely strong and basically negates the tactic of having your core rushed. It should be very expensive to do. |
| Strain Reroll Cost | 40 | 50 | Rerolling strain should be costly because it fundamentally changes your weakness, voiding anything the crew prepared for you. |
| Blob Spore Health | 40 | 20 | Blob spores should be pretty weak since they're filled with gas, and they are quite strong. |
| Blob spore cooldown | 8 seconds | 4 seconds | Compensates for the above. |
| Blob zombie player control | Yes | Only possible if done by a player-controlled blob spore | Player-controlled zombies are a little crazy on top of the blobbernauts. Do the distributed neurons strain if you want that. | 
| Blobbernaut Cost | 40 | 70 | Blobbernauts are player controlled tanks. 40 way too cheap for the craziness they can do. |
| Blobbernaut health | 200 | 250 | To compensate for the added cost. |
| Blobbernaut reagent inject amount | 20 | 5 | Injecting 20 reagents per hit is insane.  |

## Why It's Good For The Game

Security (and generally most crew weapons) have received several nerfs over the years compared to /tg/, meanwhile blob has received straight buffs. Given that blob is extremely rare, I don't think people submitting PRs to remove/nerf weapons/equipment are thinking of blob when they make the PRs.

## Changelog

:cl: BurgerBB
balance: Nerfs Blob
/:cl:

### DIFF
--- a/code/__DEFINES/blob_defines.dm
+++ b/code/__DEFINES/blob_defines.dm
@@ -5,23 +5,23 @@
 #define OVERMIND_STARTING_REROLLS 1 // Free strain rerolls at the start
 #define OVERMIND_STARTING_MIN_PLACE_TIME (1 MINUTES) // Minimum time before the core can be placed
 #define OVERMIND_STARTING_AUTO_PLACE_TIME (6 MINUTES) // After this time, randomly place the core somewhere viable
-#define OVERMIND_WIN_CONDITION_AMOUNT 500 // Blob structures required to win BUBBERSTATION CHANGE: 400 TO 500.
-#define OVERMIND_ANNOUNCEMENT_MIN_SIZE 50 // Once the blob has this many structures, announce their presence BUBBERSTATION CHANGE 75 TO 50.
+#define OVERMIND_WIN_CONDITION_AMOUNT 400 // Blob structures required to win
+#define OVERMIND_ANNOUNCEMENT_MIN_SIZE 75 // Once the blob has this many structures, announce their presence
 #define OVERMIND_ANNOUNCEMENT_MAX_TIME (10 MINUTES) // If the blob hasn't reached the minimum size before this time, announce their presence
-#define OVERMIND_MAX_CAMERA_STRAY "5x5" // How far the overmind camera is allowed to stray from blob tiles. 3x3 is 1 tile away, 5x5 2 tiles etc BUBBERSTATION CHANGE: 3x3 TO 5x5
+#define OVERMIND_MAX_CAMERA_STRAY "3x3" // How far the overmind camera is allowed to stray from blob tiles. 3x3 is 1 tile away, 5x5 2 tiles etc
 
 
 // Generic blob defines
 
 //#define BLOB_BASE_POINT_RATE 2 // Base amount of points per process()
 #define BLOB_BASE_POINT_RATE 2.5 // SKYRAT EDIT CHANGE
-#define BLOB_EXPAND_COST 4 // Price to expand onto a new tile
-#define BLOB_ATTACK_REFUND 0 // Points 'refunded' when the expand attempt actually attacks something instead BUBBERSTATION CHANGE: 4 TO 0.
+#define BLOB_EXPAND_COST 3 // Price to expand onto a new tile. BUBBERSTATION CHANGE: 4 TO 3.
+#define BLOB_ATTACK_REFUND 0 // Points 'refunded' when the expand attempt actually attacks something instead BUBBERSTATION CHANGE: 2 TO 0.
 #define BLOB_BRUTE_RESIST 1 // Brute damage taken gets multiplied by this value BUBBERSTATION CHANGE: 0.5 TO 1.
 #define BLOB_FIRE_RESIST 1 // Burn damage taken gets multiplied by this value
 #define BLOB_EXPAND_CHANCE_MULTIPLIER 1 // Increase this value to make blobs naturally expand faster
-#define BLOB_REINFORCE_CHANCE 100 // The seconds_per_tick chance for cores/nodes to reinforce their surroundings BUBBERSTATION CHANGE: 2.5 TO 100.
-#define BLOB_REAGENTATK_VOL 5 // Amount of strain-reagents that get injected when the blob attacks: main source of blob damage BUBBERSTATION CHANGE: 25 TO 5.
+#define BLOB_REINFORCE_CHANCE 0 // The seconds_per_tick chance for cores/nodes to reinforce their surroundings BUBBERSTATION CHANGE: 2.5 TO 0.
+#define BLOB_REAGENTATK_VOL 10 // Amount of strain-reagents that get injected when the blob attacks: main source of blob damage BUBBERSTATION CHANGE: 25 TO 10.
 
 
 // Structure properties
@@ -30,7 +30,7 @@
 #define BLOB_CORE_HP_REGEN 2 // Bases health regeneration rate every process(), can be added on by strains
 #define BLOB_CORE_CLAIM_RANGE 12 // Range in which blob tiles are 'claimed' (converted from dead to alive, rarely useful)
 #define BLOB_CORE_PULSE_RANGE 4 // The radius up to which the core activates structures, and up to which structures can be built
-#define BLOB_CORE_EXPAND_RANGE 3 // Radius of automatic expansion
+#define BLOB_CORE_EXPAND_RANGE 4 // Radius of automatic expansion BUBBERSTATION CHANGE: 3 TO 4.
 #define BLOB_CORE_STRONG_REINFORCE_RANGE 1 // The radius of tiles surrounding the core that get upgraded
 #define BLOB_CORE_REFLECTOR_REINFORCE_RANGE 0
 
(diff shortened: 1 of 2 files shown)


Majkl-J:
Allow me to toss in my thoughts:

1. Most of this is fine and genuinely seems like not culling blob too much
2. The new blob health is way too low, making them die from basically a sneeze. Would rather see ~10-12 HP instead of 5
3. The rerolls make the fights more dynamic and allow blob to adapt to situations they can't predict. Borgs leading the fight? EMP. Emitter spam? Fire blob.. etc

StrangeWeirdKitten:
> Security (and generally most crew weapons) have received several nerfs over the years compared to /tg/, meanwhile blob has received straight buffs. Given that blob is extremely rare, I don't think people submitting PRs to remove/nerf weapons/equipment are thinking of blob when they make the PRs.

This is not correct, specifically, TG actually gave Blob slightly more laser resistance a few months back because they buffed lasers and didn't account for blob. There's multiple ways to tackle a blob and it isn't an antag that you can just instantly pick apart without a little bit of thought behind it. I've pretty much exhausted my long list of strategies that hard counter it and I do not support neutering an antag that runs once every week to the ground. Here's the recent PR I'm talking about.

https://github.com/tgstation/tgstation/pull/76458

Imo the last two wins were staff manually spawning a blob and of course, that would ignore things like population while the last natural spawning blob (over a week ago) lost after very basic crew organization. Already one or two assistants with an SC-1, softsuit, and a recharger can do some serious kneecapping of the blob's economy.

BurgerLUA:
> Allow me to toss in my thoughts:
> 
> 1. Most of this is fine and genuinely seems like not culling blob too much
> 2. The new blob health is way too low, making them die from basically a sneeze. Would rather see ~10-12 HP instead of 5
> 3. The rerolls make the fights more dynamic and allow blob to adapt to situations they can't predict. Borgs leading the fight? EMP. Emitter spam? Fire blob.. etc

I think I agree with all these points. I will change starting hp to 10 and re-enable free blob rerolling tokens given over time.

Majkl-J (approved):
This approve has no value but let's see how it goes

mcmeiler:
Removing the blob starting points completely cucks the blob if it gets found early. The idea is to give it a chance *even if* a random greytider finds it 5 seconds after spawning, not to let that greytider kill it while it's nearly or completely defenseless depending on the starting strain. If you roll blazing oil? GG within 4 seconds if there's a fire extinguisher nearby.

BurgerLUA:
> Any antag that can be killed several screens away by 2 emitters pointing at a focuser with no reasonable counter doesn't need a nerf at all. Using equipment that engineering gets at roundstart any blob can be killed basically effortlessly. Upping the core relocate cost, lowering reflector health, and kneecapping the starting resources is killing the only things that'll actually keep you temporarily safe from an emitter array, or even just some random guy with a laser carbine.
> 
> Between this and the other blob nerf PR you have open, this feels really excessive, especially considering just how hard the hard counter for blob is.

With this PR, I want to make it so that emitters aren't THE solution to take down a blob, but rather other mechanics that exist in SS13, which is why I nerfed the cost of strong blobs and reflector blobs.

StrangeWeirdKitten:
Emitters are not the solution to take down blob, but they're easy to set up if you break into engineering. There's an entire host of options already available and five different methods to get EVA protection variating from screaming at command to open EVA towards opening a fire locker and drinking a cup of coffee. One assistant with a laser gun, a soft suit, and recharger shooting at the blob where it is most exposed will kneecap its economy and kill blobbernaughts. That's one of the many ways to fight a blob that the most greyshirts of greytide can achieve the moment they get a gun.

StrangeWeirdKitten:
I did some testing on a local world and the blob dies to a sneeze. These are numbers that are designed to make the blob pick a spot further into the station and be found early, with no real way to defend itself. Already if it's found before the 50 size announcement, it's fighting a very steep hill.

- 70 points on blobbernaughts alone is pretty insane on top of the 60-point factory cost even if it's given 50 more health. Factories can only be placed 7 tiles apart, and a certain distance away from a node. To get two or three blobbernaughts, the blob already has to be big. 

- Core relocation should not cost 100 points. Especially on most maps where the most secluded of areas are only a few tiles away from space. Sometimes you need to relocate your core, 80 is a steep cost, but 100 is absurd

- Most blob strains take 1 or 2 hits of a single laser to break a basic tile and 10 health is way too low. Having to constantly 20 points on something that will realistically be broken quickly just to stop someone from duel-wielding SC-2's from digging straight through you is a bit steep. Plus on top of reflective blobs having a health decrease. Don't forget that blob IS balanced around being taken down with lasers. 

- Emitters still exist, and will realistically be even more powerful. Especially when you can sit next to your emitter with a welder to repair any reflective blob shots. This isn't going to skew the emitter gameplay towards more "favorable" tactics. It's just going to make an already very hard counter much more powerful.

- ```#define BLOB_RESOURCE_HP_REGEN 3 //BUBBERSTATION CHANGE: 15 TO 3.``` is a very steep nerf and it means the blob will have to constantly be expanding resources, on top of its already higher resource costs. 

- ```#define BLOB_REAGENTATK_VOL 10.``` and ```#define BLOBMOB_BLOBBERNAUT_REAGENTATK_VOL 5.``` makes multiple strains practically useless when this is their main source of incapacitation or damage.

The crux of the issue with blob winning is people failing to do even the most basic things in a fight, like dragging corpses away. If you don't do this, you kind of get what's coming to you.

YakumoChen:
The health values alone punish blob for just trying to fight the crew

It has to expend a ton of extra resources to expand towards a crew defended position since it will have to constantly replenish killed blobs

Completely neuters the favoured method of surrounding someone to trap and attack them because its too easy to kill blob tiles and escape

YakumoChen:
Pretty sure blob resources cap at 99 so its literally impossible to core relocate

BurgerLUA:
> Pretty sure blob resources cap at 99 so its literally impossible to core relocate

I'll test this out.

StrangeWeirdKitten:
I asked Ozzo about all the blob PRs. He agreed that we're not making any blob balance changes for at least the next month- if then. Blob is not running enough to get realistic numbers to make drastic balancing changes.

I will also close #638 and #623

BurgerLUA:
> I asked Ozzo about all the blob PRs. He agreed that we're not making any blob balance changes for at least the next month- if then. Blob is not running enough to get realistic numbers to make drastic balancing changes.
> 
> I will also close #638 and #623

Legitimately one of the worst takes I''ve seen on here.

PR #667, opened 2023-10-30
[Not Modular] Drifting contractor adjustments

ReturnToZender:
## About The Pull Request
Removes the duplicate contractor event, and makes it so there can only be one contractor spawn at a time.

## Why It's Good For The Game
Contractors are easily able to fight off the entire security team, especially when left to fester. This removes the duplicate event so spawns can actually be controlled via config, and makes it so only one contractor can spawn at a given time.

## Changelog

:cl:
del: extra contractors
/:cl:


destrucktoid:
> Contractors are easily able to fight off the entire security team

uhhh, no, not really. If there is like 1-2 seccies maybe? But in our usual med-highpop 4-8 sec shifts they really cannot.

StrangeWeirdKitten:
Dynamic is going to handle this antag, not a random event. I'm merging this soon

PR #673, opened 2023-10-31
Fixes the GMM cargo order

NithaIsTired:
## About The Pull Request
Just removes the galactic materials market from comments so it can be actually be bought from cargo.
Was broken because of the big slappy removal somehow.

## Why It's Good For The Game
The GMM is great for lowpop rounds where we lack miners, so fixing that is nice. No more constant suffering for materials.
image

## Changelog

:cl:
fix: The GMM is back on cargo.
/:cl:


StrangeWeirdKitten:
It looks like this is a bug. The lines merged inside the comment of the slappy being commented out.

PR #690, opened 2023-11-06
restaurant portal can be moved w/ wrench

Daved27hundred:
## About The Pull Request

The portals that spawn the tourist bots can be unbolted from the floor, pulled around, and rebolted somewhere else.

## Why It's Good For The Game

These things are immovable and it gets annoying when you have to build around them because of that, especially if you don't even use them.

I initially made them destructible but this is a far better fix

## Changelog

:cl:
:qol: restaurant portal can bolted and unbolted
/:cl:


StrangeWeirdKitten:
> Alternatively, I could keep the flag and make the thing boltable/unboltable if shown how

Just make it unboltable.

StrangeWeirdKitten (changes requested):
Just make it unboltable since it's not an item you can ever reconstruct.

Daved27hundred:
ok i think this should work, but Im gonna need to look at it ingame. Let's see if I remember how to boot up my own world again.

Daved27hundred:
image
Success! I'll change the thing.

PR #698, opened 2023-11-08
Central Command Sword & Sword Sheath

ChromeFoxxity:
Adds a CentCom themed sword and sword sheath. Quite expensive one too.

## About The Pull Request

Adds sprites, and working code for a whole new sword sheath, and sword. It is the same as Captain's sword in damage, and effects.

The sword has been tested, and works just fine from what I've tried.

## Why It's Good For The Game

As a Representative that usually has a sword on his hip, it's quite annoying to see that it's red. It doesn't fit, does it? So, I added a much more expensive, and effective sword for Central Command. It is the same as the Captain's, just renamed. So it won't effect much if one were to use it.

## Changelog

:cl:
add: Sword sheath, and sword
image: Added sword, and sheath icons for it's use in-game.
/:cl:

image
image
image
image


ReturnToZender:
This is super cool, but
L + not modular + not added to the CentCom Commander's outfit, so it won't show up on their hip when spawned + no inhand sprites, so this fails tests + if you want this on the NTRep and not just admin-spawned Central personnel, you'll probably need to make a weak ceremonial version instead, as the Captain's saber by default is a very powerful weapon

ChromeFoxxity:
> This is super cool, but L + not modular + not added to the CentCom Commander's outfit, so it won't show up on their hip when spawned + no inhand sprites, so this fails tests + if you want this on the NTRep and not just admin-spawned Central personnel, you'll probably need to make a weak ceremonial version instead, as the Captain's saber by default is a very powerful weapon
> 
> If you need any advice on how to do this, you can DM me on Discord, I'd be glad to help

I sincerely do apologize for somehow having it fail every check, I have no idea how to fix it, or what I'm doing wrong. I'm really new to this whole Github thing. I WOULD have added it to the commander kit, but they already had the Mateba on their belt. And if I was just making an example since I always take the captain's sabre anyway since they never really want it.

ReturnToZender:
> > This is super cool, but L + not modular + not added to the CentCom Commander's outfit, so it won't show up on their hip when spawned + no inhand sprites, so this fails tests + if you want this on the NTRep and not just admin-spawned Central personnel, you'll probably need to make a weak ceremonial version instead, as the Captain's saber by default is a very powerful weapon
> > If you need any advice on how to do this, you can DM me on Discord, I'd be glad to help
> 
> I sincerely do apologize for somehow having it fail every check, I have no idea how to fix it, or what I'm doing wrong. I'm really new to this whole Github thing. I WOULD have added it to the commander kit, but they already had the Mateba on their belt. And if I was just making an example since I always take the captain's sabre anyway since they never really want it.

Also you don't need to add every individual variable from the captain's saber to a subtype, only ones that are different from the original weapon, subtypes inherit the properties of their master

PR #712, opened 2023-11-13
Ascii Healthhud

nevimer:
## About The Pull Request

Bit of an experiment. I wanted to add numbers and make the healthhud look better.

https://github.com/Bubberstation/Bubberstation/assets/77420409/7f44418e-cc46-4915-92a3-a0f811fa7d6b

YES. I know it's kinda fugly with the gap. Yeah, I should tweak it more. 

## Why It's Good For The Game

Honestly, I might keep the numbers only. But I think it adds some uniqueness!

## Changelog

:cl:
code: Changed the healthhud to be entirely maptext.
add: You can now see a simple health readout on healthhud. 
/:cl:


Majkl-J:
This will terribly clog the screen and is not as readable as a single bar, I'd rather see this added to the healthud examine.

nevimer:
> This will terribly clog the screen and is not as readable as a single bar, I'd rather see this added to the healthud examine.

kwFImkIY4m
numbers alone don't look bad

nevimer:
dreamseeker_xCbB5zBXpw
this is with the normal healthbar

Majkl-J:
> > This will terribly clog the screen and is not as readable as a single bar, I'd rather see this added to the healthud examine.
> 
> kwFImkIY4m numbers alone don't look bad

They don't but they also aren't as readable as a bar which tells you exactly how critical that person is. The numbers usually aren't important, just the damage types and the overall amount of damage visualised by the bar. 

I would rather see small horizontal bars atop the regular bar to tell you the type of damage

The-Sun-In-Splendour:
Make it a toggle using one of those buttons in the top left so you can choose wether you use it or not. If it's on all the time it's going to be a chore wearing healthHUDs.

zeroisthebiggay:
i don't like it. that is my feedback
it's just not natural or intuitive to read

nevimer:
Current todo:
Make it a medhuds mode, somehow. Hud code is too simple right now, so I may have to make this an alternative hud that combines with the base medhuds object, like engineering scanner goggles.

code/game/data_huds.dm
```diff
@@ -171,15 +171,67 @@ Medical HUD! Basic mode needs suit sensors on.
 	var/image/holder = hud_list?[HEALTH_HUD]
 	if (isnull(holder))
 		return
-
-	holder.icon_state = "hud[RoundHealth(src)]"
+// BUBBER EDIT BEGIN
```
Cyprex, line 174:
```suggestion
// BUBBER EDIT BEGIN- Health Hood with numbers
```

Cyprex: Or something like that

Cyprex:
How performant is this? I understand Runechat isn't very performant.

As for the bars- I think they're unnecessary & better served by the current HUD. Being less in your face is the point there. I think the individual damage numbers do have a point though, but I don't know if this is better to play with or not. Also, how does this interact with runechat, does the text overlay or something?

StrangeWeirdKitten (changes requested):
I've given my feedback. TD;LR make is a mode you can switch to off the basic default one. Use #717 for feedback please. I'll be looking at that too.

nevimer:
> How performant is this? I understand Runechat isn't very performant.
> 
> As for the bars- I think they're unnecessary & better served by the current HUD. Being less in your face is the point there. I think the individual damage numbers do have a point though, but I don't know if this is better to play with or not. Also, how does this interact with runechat, does the text overlay or something?

Runechat is not performance because it's overly complicated with a ton of expensive proc calls (measuretext), and it creates a new individualized, personalized, preference respecting, image object... for every single client in the game. So every single chat, emote, etc, can cause 50+ datums to be generated.

This just applies and sets variables on a single global image. Maptext is basically free as it's a clientside variable and rendered entirely client side.

Shroopy:
Doesn't work correctly on small models, the text is illegible. Also, it's very difficult to use on horizontal characters.

YakumoChen:
image
in-game example (you can see run-off numbers and sideways text and large/varied damage types taking increasing amounts of screen real estate)

Majkl-J:
This unironically makes me not wanna play medbay. It has terrible readability and provides significantly less knowledge on how critical a patient is than the good old bar

ReturnToZender:
Medical doesn't need more information than just "how close is X to dying," and the existing system works more than well enough for it, while ALSO being small and intuitive. This system has its bugs, but more than that-- it's really unnecessary. Medical already receives more than enough information to know at a glance what is needed from the existing HUD, and the specific damage type can be inferred by a simple scan or by examining the patient.

Either way, this is really just unnecessary bloat. I appreciate the PRs for Medical but they don't need everything spoonfed to them to do their jobs, they have everything they need and then some.

nevimer:
use the forum to discuss and not the pr body.

StrangeWeirdKitten:
The testing of this PR is finished. The feedback thread can be found [here](https://github.com/Bubberstation/Bubberstation/discussions/723) I'll be unlocking the PR regarding code standards / bugs. But I've already provided my input and the necessary changes needed.

code/game/data_huds.dm
```diff
@@ -169,17 +169,38 @@ Medical HUD! Basic mode needs suit sensors on.
...
+	var/oxy = "<span style=color:cyan>[ROUND_UP(human.oxyloss)]</span>"
+	string += "<span class='center'>[brute]|[burn]|[tox]|[oxy]</span>"
+	if(medhud.icon_state == "hudhealth100")
+		holder.maptext = ""
+		return .
+	else
+		holder.maptext = MAPTEXT_SPESSFONT("<span style='font-size: 6pt; line-height: 1.0'>[string]</span>\n")
+		holder.maptext_x = -48
+		holder.maptext_y = 30
+		holder.maptext_width = 128
```
BurgerLUA, line 202: I recommend using the RESET_TRANSFORM appearance flag to ignore all size/rotation changes from the health hud so it's readable for larger/smaller mobs.

If that doesn't work, use KEEP_APART along with it.

BurgerLUA: ``holder.appearance_flags = RESET_TRANSFORM | KEEP_APART``

Majkl-J:
Stale for a month, reopen when work continues

PR #726, opened 2023-11-15
[MODULAR GALORE] Makes empty synth bodies printable on the exofab

Majkl-J:
## About The Pull Request

Part 4 of my synthcentric update series

See name

## Why It's Good For The Game

Getting gibbed as synth means you are basically RR currently. Maaybe you could get borged, but that's about it. So I fixed that, and you can now print the synth bodies at robotics

## Changelog

:cl:
add: Synths can now be printed in exofabs
/:cl:


nevimer:
good idea. we should be able to print human torsos too.

Majkl-J:
> good idea. we should be able to print human torsos too.

You can. It's called humanising a monkey

StrangeWeirdKitten (approved):

nevimer:
> > good idea. we should be able to print human torsos too.
> 
> You can. It's called humanising a monkey

thats sketchy though, and it leaves people with odd bugs as they remain subtypes of monkeys even when "fixed"

StrangeWeirdKitten:
> > > good idea. we should be able to print human torsos too.
> > 
> > 
> > You can. It's called humanising a monkey
> 
> thats sketchy though, and it leaves people with odd bugs as they remain subtypes of monkeys even when "fixed"

the SAD has always fixed this *decently enough*. But I feel like this is out of scope of this PR. There needs to be a way to print synth torsos because the SAD doesn't fix this.

modular_zubbers/code/modules/synths/code/synth_prefab.dm
```diff
@@ -0,0 +1,17 @@
+/mob/living/carbon/human/species/synth
+	race = /datum/species/synthetic
+
+// Used for printing dead synth bodies
+/mob/living/carbon/human/species/synth/empty
+
+/mob/living/carbon/human/species/synth/empty/Initialize(mapload)
+	. = ..()
+	var/old_death_sound
+	death_sound = null // We don't need them screaming
```
nevimer, line 10: reset this to initial after

modular_zubbers/code/modules/synths/code/synth_prefab.dm
```diff
@@ -0,0 +1,17 @@
...
+	death()
+
+	for(var/obj/item/organ/internal/brain/B in organs)
```
nevimer, line 13: should do var = locate (type) in organs instead

modular_zubbers/code/modules/synths/code/synth_prefab.dm
```diff
@@ -0,0 +1,17 @@
+/mob/living/carbon/human/species/synth
+	race = /datum/species/synthetic
+
+// Used for printing dead synth bodies
+/mob/living/carbon/human/species/synth/empty
+
+/mob/living/carbon/human/species/synth/empty/Initialize(mapload)
+	. = ..()
+	var/old_death_sound
```
nevimer, line 9: this is never set to anything other than null

modular_zubbers/code/modules/synths/code/synth_prefab.dm
```diff
@@ -0,0 +1,17 @@
...
+	death_sound = null // We don't need them screaming
+	death()
+
+	for(var/obj/item/organ/internal/brain/B in organs)
+		B.Remove(src)
+		QDEL_NULL(B)
+
+	death_sound = old_death_sound // Now they can die loudly again
```
nevimer, line 17:
```suggestion
	death_sound = initial(death_sound) // Now they can die loudly again
```

nevimer:
also make sure this costs more than just fixing the body. fixing the body should be much cheaper resource wise.

modular_zubbers/code/modules/synths/code/synth_prefab.dm
```diff
@@ -0,0 +1,17 @@
...
+
+/mob/living/carbon/human/species/synth/empty/Initialize(mapload)
+	. = ..()
+	var/old_death_sound
+	death_sound = null // We don't need them screaming
+	death()
+
+	for(var/obj/item/organ/internal/brain/B in organs)
+		B.Remove(src)
+		QDEL_NULL(B)
```
nevimer, line 15:
```suggestion
	var/obj/item/organ/internal/brain/B = get_organ_slot(ORGAN_SLOT_BRAIN) 
	if(B)
		B.Remove(src)
		QDEL_NULL(B)
```

StrangeWeirdKitten:
> Member

It doesn't cost anything now to revive a synth / fix the body except time, wires, and welding fuel.

modular_zubbers/code/modules/synths/code/synth_prefab.dm
Majkl-J: The initial is `null` as death_sound is set by the species post compilation

modular_zubbers/code/modules/synths/code/synth_prefab.dm
Majkl-J: This is also null

modular_zubbers/code/modules/synths/code/synth_prefab.dm
Majkl-J: wilco

modular_zubbers/code/modules/synths/code/synth_prefab.dm
Majkl-J: Ok this was actually missed as I was making this PR on 3 different codebases with IPCs. Woops

Majkl-J:
Tested again after the minor changes and works fine

StrangeWeirdKitten:
Looks fine. No runtimes when I ran it on a test prod of the server. I'm merging it

PR #741, opened 2023-11-19
Ghosting, suiciding, and DNR makes your body ever so slightly transparent

Majkl-J:
## About The Pull Request

See name

## Why It's Good For The Game

Removes the need to examine in order to see whether the person is cataconic. Heard we are moving away from the chat window from nevimer, so this should be an easy extension of that

## Changelog

:cl:
add: Ghosting or going DNR makes you appear slightly transparent
/:cl:


nevimer:
This is a bit MMO. Weird solution to an issue, but not a bad one.

yourdoom9898:
> This is a bit MMO. Weird solution to an issue, but not a bad one.

It's really no more MMO than being able to verify someone's soul has detached from their body

StrangeWeirdKitten:
I'm having a stroke right now. Only thing is Linters, which should be an easy fix. It's just tick enforcement.

StrangeWeirdKitten:
image

What the actual fuck. I swear to god it's wrong, but this is what Linters wants. I think Linters is on acid. Why does it want G P G and not the objectively correct G G P. But it's deciding the wrong version is a green check mark, so I guess we're going with this then I'll end up fixing it later down the line when tick enforcement decides to become sane again

PR #751, opened 2023-11-22
Makes roundending significantly harder by making the brain (Almost) indestructible

Majkl-J:
## About The Pull Request

See name. The only way to completely glass a brain now is cremation and the gibber (Soon to be changed probably hopefully)

## Why It's Good For The Game

Pretty sure we are not meant to roundend people. But there's so many hazards that might just end up removing you from the game, namely stuff like bombs gibbing you so hard your brain gets dusted in the explosion (Might be a synth only issue, to be checked). This makes it need much more deliberate effort to actually remove someone out of the game permanently.

## Changelog

:cl:
tweak: Brains are no longer easily destructible with fire, explosives, or acid. The gibber and crematorium still work
/:cl:


Majkl-J:
This might not resolve the actual issue that made me create this, but I still believe this is a step in the right direction

Cyprex:
what about deep frying?

Cyprex:
Also selling via the cargo shuttle- anything that can destroy the blackbox(traitor objective) can probably destroy brains too

modular_zubbers/code/modules/surgery/organs/internal/lungs/_brain.dm
```diff
@@ -0,0 +1,2 @@
+/obj/item/organ/internal/brain
+	resistance_flags = INDESTRUCTIBLE
```
Cyprex, line 2: this overrides the ones defined originally- are any defined? Not sure if it should be added in New()... or not.

Majkl-J: This flag is above all the others, in my testing this includes lava and acid resistance. And there are none

Majkl-J:
> what about deep frying?

This is to mainly stop hazards like explosions and other stuff that can happen without deliberation. Deliberate RR is still possible, as it probably should be considering stuff like changelings exist.

PR #765, opened 2023-11-24
Maintenance Loot Rebalancing

BurgerLUA:
## About The Pull Request

Overall Changes:
- Halves the amount of maint loot available (x5 multiplier for elevated structures to x3)
- Rebalances maintenance loot weights (Trash and common objects are only half as common).

Notable changes:
- Removes head stamps from spawning.
- Moves soap to the common loot pool instead of the trash loot pool.
- Adds a one use die of fate as an incredibly rare oddity spawn, like it used to be.

## Why It's Good For The Game

My new change to trash piles creating trash significantly increased the amount of trash in maint, thus reductions needed to be applied so maint isn't just 99% trash.

Also some objects were too common or causing weird problems.

## Changelog

:cl: BurgerBB
balance: Rebalances maintenance loot.
/:cl:


BurgerLUA:
I now got to find out why this crashes after pollution (usually when the map needs to load).

BurgerLUA:
wow it's the first thing I guessed

PR #778, opened 2023-11-27
[MODULAR] Freedom Crate!! An early celebration of that American holiday thing.

theselfish:
## About The Pull Request

Title. The crate includes everything an American needs, except guns which are sold separately.

## Why It's Good For The Game

It's contraband, it also provides a way for people to get an American flag, if they want it.

If the Russians are allowed to have a [loser of WW2 ideology] flag, why not Americans?

image
image

## Changelog

:cl: Cursor, as dictated by George Washington & Martin Van Buren, 
add: The Freedom Crate is now available to sovereign citizens who refuse to respect Cargo's laws.
/:cl:


Majkl-J (approved):
Looks solid to me codewise, and contentwise. All this stuff is in the game, why not make it buyable like this?

Also the check fail is not caused by this, I would rerun the checks but I either can't or don't know how

theselfish:
> Looks solid to me codewise, and contentwise. All this stuff is in the game, why not make it buyable like this?

I know some people wanted it in the loadout, the flag. I disagree with that as...I mean, getting to put USA flags everywhere would be weird. But if people want them, they can buy them via hacking a console.
Which other silly stuff like ATVs and Crabs (IIRC) are kept.

PR #787, opened 2023-11-29
Remove bow and arrow crafting recipe*

projectkepler-ru:
## About The Pull Request
As it turns out, 7 months ago, TG had 50 damages.
What does this mean?
Skyrat was never made in mind that arrows deal straight up 50 damages, and neither was TG( to a degree, given you can't really just craft arrow there as far as my VSC searching skill goes) 

Let's [Review](https://github.com/tgstation/tgstation/pull/74811) this is where the gitblame was linked to so that's as good as it get
What in god name is this shit
image

image

I never thought I'd have to nerf down a feature I didn't even know exist until someone did something funny with it

and who do we blame for this?
Well skyrat of coruse
image

This PR removes* the recipe to craft bow and arrows for most  easily obtainable on station materials, making it more or less a blacksmithing only item or restricted to ashwalkers/icecat entirely
The goal here is not to remove the bow and arrows, but make it harder to obtain and utilise as station-side. 

Wooden Bow recipe is now disabled
Pipe Bow recipe is now disabled
Arrow* recipe is now disabled and can only be obtained via blacksmithing
You are still able to make the bone bow and arrows/ashen arrows, as it was always intended as a lavaland item, and I don't see an issue with if the crew went down and put the efforts to obtain those item that they should be punished for it.

Standard arrow damage reduced to 30
Standard arrow embed reduced to 50% instead of 90%
This is a really half-hazard solution and I know someone will come up with a better one later, but leaving it like this does not help anyone except the people who uses it

## Why It's Good For The Game
Skyrat and TG literally could not had anticipated a combination of both of their PR coming to fuck each other up,  and it's now our problem, In a perfect world, bubberstation does not have issues with ballistic weapon because most of them don't deal enough damage to send you back to the lobby screen. 
Unfortunately, this is not a perfect world, we're going to be repeating that mistake a few more time and I'm sure of it. 
However, everytime, someone has to come up and deal with it. This is one of those time

### DIFF
--- a/code/modules/projectiles/guns/ballistic/bows/bow_arrows.dm
+++ b/code/modules/projectiles/guns/ballistic/bows/bow_arrows.dm
@@ -29,11 +29,11 @@
 	desc = "Ow! Get it out of me!"
 	icon = 'icons/obj/weapons/bows/arrows.dmi'
 	icon_state = "arrow_projectile"
-	damage = 50
+	damage = 30 //BUBBER EDIT ORIGINAL 50
 	speed = 1
 	range = 25
 	embedding = list(
-		embed_chance = 90,
+		embed_chance = 50, //BUBBER EDIT ORIGINAL 90
 		fall_chance = 2,
 		jostle_chance = 2,
 		ignore_throwspeed_threshold = TRUE,
--- a/code/modules/projectiles/guns/ballistic/bows/bow_arrows.dm
+++ b/code/modules/projectiles/guns/ballistic/bows/bow_arrows.dm
@@ -33,7 +33,7 @@
 	speed = 1
 	range = 25
 	embedding = list(
-		embed_chance = 50, //BUBBER EDIT ORIGINAL 90
+		embed_chance = 90
 		fall_chance = 2,
 		jostle_chance = 2,
 		ignore_throwspeed_threshold = TRUE,
--- a/code/modules/projectiles/guns/ballistic/bows/bow_arrows.dm
+++ b/code/modules/projectiles/guns/ballistic/bows/bow_arrows.dm
@@ -29,7 +29,7 @@
 	desc = "Ow! Get it out of me!"
 	icon = 'icons/obj/weapons/bows/arrows.dmi'
 	icon_state = "arrow_projectile"
-	damage = 30 //BUBBER EDIT ORIGINAL 50
+	damage = 50
 	speed = 1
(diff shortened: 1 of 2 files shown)


projectkepler-ru:
I pray to god someone has a better solution.

Majkl-J:
> I don't remember IRL bow arrows shooting faster than any projectile bullets.

I don't remember lasers not being the speed of light

In complete honesty, we can also just make the projectile get a damage buff on lavaland, like half the items that are used on there do anyhow. It is the only place where it's consistently not used on carbons.

projectkepler-ru:
> In complete honesty, we can also just make the projectile get a damage buff on lavaland, like half the items that are used on there do anyhow. It is the only place where it's consistently not used on carbons.

ya

atoshiko:
You know, given how rough it is to use bows, giving them high damage is a fair trade. Don't think this is a justified nerf for this.
Also making Ashwalker gaming much harder than it have to be. With damage being 50, it kills legion with 2 arrows, watchers with somewhat...5, if I am to be not mistaken. If you bring it down to 30...guess _what_ happens.
It has never been an issue and you just making the game unfun for the bow and arrow users. I am however voting for removing them from the crafting menu, just do not touch the numbers.

StrangeWeirdKitten (changes requested):

code/modules/projectiles/guns/ballistic/bows/bow_arrows.dm
```diff
@@ -29,11 +29,11 @@
 	desc = "Ow! Get it out of me!"
 	icon = 'icons/obj/weapons/bows/arrows.dmi'
 	icon_state = "arrow_projectile"
-	damage = 50
+	damage = 30 //BUBBER EDIT ORIGINAL 50
```
projectkepler-ru, line 32:
```suggestion
	damage = 50
```

code/modules/projectiles/guns/ballistic/bows/bow_arrows.dm
```diff
@@ -29,11 +29,11 @@
...
 	speed = 1
 	range = 25
 	embedding = list(
-		embed_chance = 90,
+		embed_chance = 50, //BUBBER EDIT ORIGINAL 90
```
projectkepler-ru, line 36:
```suggestion
		embed_chance = 90
```

PR #796, opened 2023-11-30
Removes Explosive Lemon Grudgecode because I died to it once.

BurgerLUA:
## About The Pull Request

Explosive lemon detonation time changed from 1 to 6 seconds to 4 to 12 seconds.

## Why It's Good For The Game

Absolutely insane decision by the person who decided to make the minimum time 1 second considering that activating a lemon grenade does not enable throw mode (despite literally every other grenade doing this) and expecting people to take less than 1 second to realize this and fight with the controls with an average ping of 100ms.

PR makes the change much more reasonable by adding a minimum time of 4 to 12 seconds so they're not too powerful in combat but can still be used viably without c*ders fucking over players.

## Changelog

:cl: BurgerBB
balance: Explosive lemon detonation time changed from 1 to 6 seconds to 4 to 12 seconds.
/:cl:

### DIFF
--- a/code/modules/hydroponics/unique_plant_genes.dm
+++ b/code/modules/hydroponics/unique_plant_genes.dm
@@ -609,7 +609,7 @@
 		our_plant.color = COLOR_RED
 
 	playsound(our_plant.drop_location(), 'sound/weapons/armbomb.ogg', 75, TRUE, -3)
-	addtimer(CALLBACK(src, PROC_REF(detonate), our_plant), rand(4 SECONDS, 12 SECONDS)) //BUBBERSTATION CHANGE: 1 TO 6 INTO 4 TO 12.
+	addtimer(CALLBACK(src, PROC_REF(detonate), our_plant), rand(4 SECONDS, 10 SECONDS)) //BUBBERSTATION CHANGE: 1 TO 6 INTO 4 TO 10.
 
 /datum/plant_gene/trait/bomb_plant/potency_based/detonate(obj/item/our_plant)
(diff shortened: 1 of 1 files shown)


YakumoChen:
not supposed to be a reliable weapon, like IEDs. 

4-12 seconds just means its practically unusable as an actual weapon because you can't prime the throw time and nobody is going to smartly stand near an explosive weapon for the entire 8 second time window it could explode for.

fix the throw mode if its not automatic?

code/modules/hydroponics/unique_plant_genes.dm
```diff
@@ -609,7 +609,7 @@
 		our_plant.color = COLOR_RED
 
 	playsound(our_plant.drop_location(), 'sound/weapons/armbomb.ogg', 75, TRUE, -3)
-	addtimer(CALLBACK(src, PROC_REF(detonate), our_plant), rand(1 SECONDS, 6 SECONDS))
+	addtimer(CALLBACK(src, PROC_REF(detonate), our_plant), rand(4 SECONDS, 12 SECONDS)) //BUBBERSTATION CHANGE: 1 TO 6 INTO 4 TO 12.
```
Majkl-J, line 612: The time range should probably be the same

```suggestion
	addtimer(CALLBACK(src, PROC_REF(detonate), our_plant), rand(4 SECONDS, 10 SECONDS)) //BUBBERSTATION CHANGE: 1 TO 6 INTO 4 TO 10.
```

Majkl-J (changes requested):

PR #801, opened 2023-12-02
Suit Sensors EMP Overhaul (Because I died in maint after getting EMP'd once)

BurgerLUA:
## About The Pull Request

Changes how EMPs affect suit sensors.

Light EMPs will adjust suit sensors up or down one level (instead of it being completely random).
Heavy EMPs will break the suit sensors, causing it to forcibly set to broadcast everything, however the data may not be entirely... accurate.

image

(Health and location values are completely randomized, with the location data taken directly from the ion file strings list).

## Why It's Good For The Game

(Mostly) Prevents people from effectively being round removed and made irrelevant in cases where people suffer an EMP and are also killed in hidden areas.

## Changelog

:cl: BurgerBB
balance: Getting EMP'd now affects your suit sensors differently.
/:cl:


nevimer:
How do I fix the issues an EMP would cause after this?

BurgerLUA:
> How do I fix the issues an EMP would cause after this?

The previous method would work; cable coil

PR #808, opened 2023-12-04
Reverts the Expand module to Skyrat Default

theselfish:
## About The Pull Request

Why.
Whywhywhywhywhy would you comment this out?

I'm one of the first past the post to call Skyrat edits/additions stupid but this is there for a reason.

## Why It's Good For The Game

Borgs should take up two tiles.
Max.
This, in addition to the Raptor Borg sprites which are another can of worms is not good.

If your sprite is already larger than a normal cyborg, it doesn't need to be even larger. 

## Changelog

(Do we really need a CL for this?)
:cl:
fix: You can no longer expand non-32x32 cyborg sprites. Again.
/:cl:


theselfish:
I disagree with your reasoning, but, I don't exactly have a leg to stand on.
Choice's been made.

PR #754, opened 2023-11-23
Sol-Fed Guns Return to Cargo

IgiariValkyr:
## About The Pull Request

After quite a bit of begging from myself and other crew, Zender has agreed that we can re-add the rest of the Sol-Fed Guns back to cargo. However, the guns of course had to be nerfed down. As such, the bullet projectiles (which are what matters) are edited to have immensely lowered wound chance.

Wound_Bonus used to be 10. Bare_Wound_Bonus used to be 20.
Wound_Bonus is now -5. Bare_Wound_Bonus is now 0.

Damage values have been edited as well. See the replies below.

The following sheet was the original TTK testing with no damage edits. They were stronger than lasers, but I do believe that we have the damage scaled down now enough to be on par or slightly weaker.
<details>
  <summary>TTK Tests</summary>
(The following numbers are based on Lethal ammo at their original damage values, and not the Flathead or Hollowhead rounds. The ratio of x-y-z is based on x = soft-crit, y = hard-crit, z = death.)

________________

Carwo "Renoster" Shotgun
(Dependent on shotgun shell rather than ammo)

Carwo 'Kiboko' Grenade Launcher
(Grenade physics make it difficult to judge, as if the person rolls "heroicly jumps on the grenade to reduce shrapnel", it reduces the damage heavily. Far more useful for crowd control with the different grenade types.)

Weapons firing .35 Sol Short
[Trappiste "Eland" Revolver, Trappiste "Wespe" Pistol, Carwo "Sindano" Submachine Gun]
Standard Crew - 6-7-10(ish) 
Decked Cargo - 9-11-15 
Decked HoS - 10-11-16 
Syndie Mod - 8-10-14 
Boosted Elite - 13-16-22 

Weapons firing .40 Sol Long
[Carwo "d'Inanteria" Rifle, Carwo "d'Elite" Marksman Rifle, Carwo "d'Outomaties" Machinegun (very high rate of fire and large magazine)]
Standard Crew - 4-5-7 
Decked Cargo - 7-8-12 
Decked HoS - 7-8-11
Syndie Mod - 6-7-9/10
Boosted Elite - 10-12-16

Weapons firing .585 Trappiste (holy shit that's a big bullet)
[Trappiste 'Skild' Pistol (slow firing, 7 rounds), Trappiste "Takbok" Revolver" (slow firing, 5 rounds)]
Standard Crew - 3-4-6
Decked Cargo - 6-7-9
Decked HoS - 5-6-9
Syndie Mod - 5-6-8
[...]

PR #655, opened 2023-10-26
Adds the WT-551, Unskyrats the WT-550 ammo

BurgerLUA:
## The WT-551
image

This adds the WT-551. A remade version of the WT-550 that is worse in every way. Fortunately, that means that it is balanced enough to be put in NanoTrasen armories.

Compared tot he WT-551, it is bulkier and slightly slower (0.3 second fire delay compared to 0.2). Additionally, it is commonly used with rubber-tipped rounds or FlatHead rounds, which are a special surplus of ammo that deals less damage and has no wounding, embedding, or penetrative power. Regular ammo can be purchased from cargo or researched later, with special ammo also being available later.

Note that this does not replace the WT-550.

## FlatHead ammo

image

Flathead ammo deals 18 brute damage (compared to the original 20), and 5 stamina damage per hit. It is extremely weak against armor, has no embed chance, has virtually no wounding chance. It's perfect for cheap corporate companies dealing with cheaper personnel. This is the type of lethal ammo that security will use for the gun, unless someone speedruns weapon research.

## Research Progression

image

### Basic WT-550/WT-551 Ammunition.
Flathead rounds and Rubber rounds for the WT-550/WT-551 can be researched for 2500 points after unlocking the "Weapon Development Technology" node.

### Advanced WT-550/WT-551 Ammunition.
Regular rounds and AP rounds for the WT-550/WT-551 can be researched can be researched for 5000 points after unlocking the "Advanced Weapon Development Technology" and "Basic WT-550/WT-551 Ammunition" nodes.

### Illegal WT-550/WT-551 Ammunition.
Incendiary rounds for the WT-550/WT-551 can be researched can be researched for 7500 points after unlocking the "Illegal Technology", "Exotic Ammo" , and "Advanced WT-550/WT-551 Ammunition" nodes.

### Syndicate Research

Removes the WT-550 ammo from syndicate research since it is now redundant.

## Cargo

image

WT-551 rifles can be ordered in pairs (2) for the cost of a parrot, a grilling starter pack, or a crab rocket (1600 credits). This value was chosen because it is slightly higher than the thermal pistols, and the traitor-ordered WT-550 rifle pack (which contains lethal ammo + spare lethal ammo).

Additional FlatHead, Rubber, and Regular ammo can be ordered from cargo as well.

Cargo techs no longer get WT-550s in the mail, but instead WT-551s (why was this a thing holy shit).

## Armory

image

2 WT-551s can be found in the armory.
For balance purposes one (1) laser rifle was removed.

[...]

PR #618, opened 2023-10-17
Skyrat Armory Removal

BurgerLUA:
## About The Pull Request

Other than mapping overrides and direct item removals, Skyrat can no longer apply armory spawn changes.

The new armory spawn for all maps is now the following:
3-6 Barrier Grenades (Depends on Map)
3-6 Riot Shields (Depends on Map)
3-6 Riot Armor  (Depends on Map)
3-6 Riot Helmets  (Depends on Map)
3-6 Breaching Hammers (Depends on Map)
x2 Rubbershot Ammo Boxes
x2 Beanbag Ammo Boxes
x1 Breaching Slugs Boxes
x4 Disablers
x3 MCRs
x2 Energy Guns
x2 Laser Guns
x3 Riot Shotguns
3-6 Bulletproof Helmets  (Depends on Map)
3-6 Bulletproof Armor  (Depends on Map)

Fixes Burgerstation having incorrect armory spawning types.

## Why It's Good For The Game

/tg/ good, skyrat bad, upvotes to the left.

## Changelog

:cl: BurgerBB
balance: Skyrat no longer has an effect on armory spawns.
fix: Fixes Burgerstation having incorrect armory spawning types.
/:cl:

PR #831, opened 2023-12-08
Rebalances Sol-Fed Guns

ReturnToZender:
## About The Pull Request
This rebalances Sol-Fed weapons to be more distinct from each other, as well as more balanced in terms of damage and wounding.
Update: Now has basic Sol-Fed sidearms behind weapons permit.

This also makes each of the Sol-Fed weapons distinct from each other in purpose, rather than all of them being all-around.
All weapons have been tested by hand!
image
image
image

## Why It's Good For The Game
This prevents antagonists from being shut down in 5 seconds (hyperbole) like before, and allows for longer engagements and more opportunities to escape. Which is always better.

## Changelog

:cl: ReturnToZender, The Sharkening
add: Department ordered SolFed guns
add: Sol-Fed Firearms now require a weapons permit to order
fix: Adjusted Sol weapon balance to be more fair
/:cl:

PR #835, opened 2023-12-09
Revert "Sol-Fed Guns Return to Cargo"

BurgerLUA:
Reverts Bubberstation/Bubberstation#754

# Why

Not only is half of this non-modular, it solves a problem that was already fixed by #655 and #618

# Template??

For some reason there is no template for revert PRs


IgiariValkyr:
Reverting this will also affect #768 and currently testmerged #831

PR #838, opened 2023-12-11
Waterpig grows more insane with modularity: The massive PR that probably breaks shit

Majkl-J:
## About The Pull Request

I have circuitry theory finals in 1 hour. And as such, to distract myself from my impending doom and to stop thinking about loop currents, I have decided to start working on this. (Update: I passed)

This isn't even close to how I wish our modularity to look

But that's future Waterpig's problem (Note to self: Fix the no_antag button)

In another news, this probably breaks stuff.

## Why It's Good For The Game

Modularity is good. The more stuff we can modularize the better, and incase it gets removed upstream it's as simple as... removing our modular override.

## Changelog

:cl:
refactor: Refactors modularity significantly
/:cl:


nevimer:
We're basically trading merge conflicts for compile and potential runtime errors later on. Which is fine.

Majkl-J:
> We're basically trading merge conflicts for compile and potential runtime errors later on. Which is fine.

The law of equivalent exchange

Cyprex:
Note that as far as I understand removing a base edit can still cause it to trigger a merge conflict later down the line- you have to rewrite history to get rid of the merge conflict potential.

Majkl-J:
> Note that as far as I understand removing a base edit can still cause it to trigger a merge conflict later down the line- you have to rewrite history to get rid of the merge conflict potential.

With most of these you can just pick to use the master file and be done with it. Hopefully it will be one large burst of conflicts next upstream (Which will happen anyhow, seeing all the refactors)

Majkl-J:
This is far past what can be easily reviewed. I suggest we TM this for like a week and if nothing breaks we can merge this

code/modules/vending/clothesmate.dm
```diff
@@ -209,7 +209,7 @@
 		/obj/item/clothing/under/costume/swagoutfit = 1,
 		/obj/item/clothing/shoes/swagshoes = 1,
 		/obj/item/instrument/piano_synth/headphones/spacepods = 1,
-		/obj/item/clothing/head/helmet/toggleable/pinwheel = 1, //BUBBER EDIT: Pinwheel hat
+		/obj/item/clothing/head/helmet/toggleable/pinwheel = 1, //BUBBER EDIT: Pinwheel hat // BUBBER TODO: Modularize
```
StrangeWeirdKitten, line 212: Skyrat wrote its own modularization system for the vendors btw

modular_zubbers/master_files/code/controllers/subsystem/tts.dm
```diff
@@ -0,0 +1,7 @@
+// Note: This is terrible and permanently disables tts without fixing the underlying issue
```
StrangeWeirdKitten, line 1: We don't have TTS anyway and this can be fixed if we do get TTS. This is a bandaid solution, yes. But this is a button that crashes the server.

StrangeWeirdKitten:
We'll test merge this, obviously. Do not merge until we are sure nothing is broken.

code/modules/vending/clothesmate.dm
Majkl-J: I know, I am gonna include it right after this is finished, individually

modular_zubbers/master_files/code/controllers/subsystem/tts.dm
Majkl-J: Fair point

StrangeWeirdKitten:
Seems to work

PR #842, opened 2023-12-12
[EARLY MIRROR] Holiday Supermatter

lessthnthree:
## About The Pull Request

Early mirror of improved upstream Christmas SM. https://github.com/tgstation/tgstation/pull/80254
Removes the modular override.

Thanks to @Majkl-J for the original concept.

dreamseeker_OR0HVPqRxF

## Changelog

None required, it'll create a duplicate when the mirror makes it through someday


Majkl-J (approved):
I am biased. We need this

(Also the sprites and code are prettier, so)

PR #844, opened 2023-12-13
Adds layer adaptors to kilo atmos

Majkl-J:
## About The Pull Request

I was complained to about this by an unnamed third party, which is usually how I get PR ideas.

This is still probably not perfect but better than what we have now

(Also visually the pipes seem to connect. I haven't actually tested IF they flow but they probably should
image

## Why It's Good For The Game

It's like the rest of the maps' atmos now

## Changelog

:cl:
qol: Kilo: Layer adaptors in atmos
/:cl:


StrangeWeirdKitten:
Fix CI

StrangeWeirdKitten (changes requested):
Fix CI

PR #847, opened 2023-12-14
Security reskin update. Red Sec edition

Deek-Za:
## About The Pull Request

Alter's the following security clothing items to have an additional red sec option when reskinning.
Uniform
Boots
Gloves
Cap
Backpack/Satchel/Duffel

image

## Why It's Good For The Game

Some people prefer the classic(And much cooler) red security outfit. We have a vendor, but that means every time you start a shift you have to spend extra time to replace all those pieces and also spend 10 credits per item. This just speeds up the process considerably.

## Changelog

:cl:
qol: Red sec options added to existing reskinnable security equipment. 
/:cl:


theselfish:
While I'm not an expert on modularization, that would be @Majkl-J , I'm pretty sure you could easily modulerize this.

Deek-Za:
> While I'm not an expert on modularization, that would be @Majkl-J , I'm pretty sure you could easily modulerize this.

Tried a crack at that earlier but it was causin' some issues. But maybe they have a better functioning way of handling it. Then yeah, for sure.

modular_skyrat/modules/goofsec/code/sec_clothing_overrides.dm
```diff
@@ -63,6 +63,15 @@
...
 		),
+		//BUBBER EDIT ADDITION - RED SEC
+		"Red Variant" = list(
+			RESKIN_ICON = 'icons/obj/storage/backpack.dmi',
+			RESKIN_ICON_STATE = "backpack-security",
+			RESKIN_WORN_ICON = 'icons/mob/clothing/back/backpack.dmi',
+			RESKIN_INHAND_L = 'icons/mob/inhands/equipment/backpack_lefthand.dmi',
+			RESKIN_INHAND_R = 'icons/mob/inhands/equipment/backpack_righthand.dmi',
+			RESKIN_INHAND_STATE = "securitypack"
+		),
```
Majkl-J, line 74: Gonna use this snippet as an example. Going from the top of my head so this might not be exact but either way gives a good example of how to do modularization of lists

```suggestion
/obj/item/storage/backpack/security/Initialize(mapload)
	. = ..()
	unique_reskin["Red Variant"] = list(
			RESKIN_ICON = 'icons/obj/storage/backpack.dmi',
			RESKIN_ICON_STATE = "backpack-security",
			RESKIN_WORN_ICON = 'icons/mob/clothing/back/backpack.dmi',
			RESKIN_INHAND_L = 'icons/mob/inhands/equipment/backpack_lefthand.dmi',
			RESKIN_INHAND_R = 'icons/mob/inhands/equipment/backpack_righthand.dmi',
			RESKIN_INHAND_STATE = "securitypack"
		),
```

Majkl-J: (Note: If this doesn't work it's probably due to some pre-initialization of the reskins, in which case I'll look into it proper)

Deek-Za: Looks like it's working fine! I now see what I did wrong with my original Modu attempt. Thanks, chief. Will get this updated soon.

modular_zubbers/code/modules/clothing_reskins/redsec.dm
```diff
@@ -0,0 +1,65 @@
...
+			RESKIN_WORN_ICON = 'icons/mob/clothing/hands.dmi'
+		)
+	. = ..()
+
+/obj/item/clothing/under/rank/security/officer/Initialize(mapload)
+	unique_reskin["RedSec Variant"] = list(
+			RESKIN_ICON_STATE = "rsecurity",
+			RESKIN_WORN_ICON_STATE = "rsecurity"
+		)
+	. = ..()
```
Majkl-J, line 58: Ok, problem here: We can't add reskinning on this specific piece of clothing, because altclick already handles changing how you wear it.
Also this might break subtypes like /obj/item/clothing/under/rank/security/officer/grey

Deek-Za: All the items listed, including _that_ one already have reskinning on them. I'm just adding an additional option to them(Red Sec) You're right, it does cause you to set the outfit's adjustment to casual while also changing the look of it, but you can just alt click it again to adjust it back since you can only reskin once per clothing item. So it does work even though you'd effectively have to alt-click it twice(Reskin + Adjust, Adjust back). 
I really do wish the original Reskin code was not on Alt+click though. Part of the reason why I didn't add Winter coats to this list is because it's got a different alt+click adjustment code that makes it impossible to use with the reskin code as is.

Deek-Za: Alright. After testing some things out.
So that uniform is already reskinnable, so I am not adding that particular functionality. As for the subtypes, they are already setup with flags to prevent from "inheriting" that reskin feature(Except for gray, which is just a gray variant that's unobtainable? but seems to offer the reskin option currently, likely intentionally). So the orderly for example can't reskin his outfit. 
However doing it in that particular modular method definitely breaks a lot of the subtypes.
So I think I might have to go semi-modular for the sake of the uniform since it's so high up the hierarchy. This allows it to function and keeps all the subtypes working appropriately.

projectkepler-ru:
> It still doesn't sit with me that one runtimed for some reason... I'll try to see what sort of issue was it and if it can be modular, but otherwise looks good codewise.

probably can't be helped

Majkl-J:
> > It still doesn't sit with me that one runtimed for some reason... I'll try to see what sort of issue was it and if it can be modular, but otherwise looks good codewise.
> 
> probably can't be helped

Like that ever stopped me

Deek-Za:
Right-o. I think I got it. It should now not attempt to apply that customization with the subtypes that have their unique_reskin value set to null. Which is stopping all those indexing error runtime things we were gettin'.
So we're back to modular, baby!

Majkl-J (approved):
Heck yea. Ready to merge

PR #852, opened 2023-12-15
[Modular] Fixes the autopunctuation regex

ReturnToZender:
## About The Pull Request
This fixes the autopunctuation regex so it SHOULD be compatible with a server running 1515.1620 now
I dunno why it broke but I was told to change it by Zirok and so I did

## Why It's Good For The Game
This makes it so the server can actually update I think

## Changelog

:cl: ReturnToZender, distributivgesetz
fix: the broken regex for autopunctuation
/:cl:


ReturnToZender:
Despite this being an improved regex, the problem is still not solved, the radio still does not work on 515.1620

ReturnToZender:
I made an improved regex, but the kind folks over at [Skyrat](https://github.com/Skyrat-SS13/Skyrat-tg/pull/25698) went ahead and made a better version of my autopunctuation regex that even capitalizes the letter I if you forget to.

So, thanks to them over there for being my inspiration! I guess in that sense, this PR is now an early pull, though I'll be updating it soon to make it work with formatting (+_|)

PR #865, opened 2023-12-18
Nerfs Crab17 because I lost my life savings once

BurgerLUA:
## About The Pull Request

The Crab17 uplink item cannot spawn in surplus crates, nor can it be discounted.

## Why It's Good For The Game

The Crab17, while a fun addition, is kind of getting annoying fast (nearly purchased every round to the point where people unironically metagame it if they know who is traitor). To discourage overusage, it cannot spawn in surplus bundles, nor can it be discounted.

## Changelog

:cl: BurgerBB
balance: The Crab17 uplink item cannot spawn in surplus crates, nor can it be discounted.
/:cl:


Deek-Za:
Much better imo. I hate that machine to death. And you tend to get like 2-3 in a round when it shows up.
I think moving it to non-restricted areas is great because realistically you *shouldn't* let people into restricted areas, But if you don't then you're gonna look like a dickhead.

Detective-Google:
the fact it spawns in a random locked area is kinda the point, it's supposed to be a distraction item to get people to go somewhere to get at an objective. the other changes are kinda whatever but I don't see why they're worth breaking with upstream for, is there a reason this couldn't go to SR/TG?

radarseas:
as for the **restricted rooms** the entire point of the crab is a random division that other antagonists and non antags can capitalise on by the fact the way will be restricted, it also makes people pile up which is perfect for hypnoflash bombings (making that 12TC actually pay off requires a crab almost) or other stuff. If it rolls a hallway, bad luck, if it rolls the AI's chamber, that's funny and extremely rare, let it happen.

besides, there's a reasoning we have the greytide virus

theselfish:
I agree with everything else, but I think it should still be able to spawn wherever it desires.

StrangeWeirdKitten:
I've did a lot of thinking on it, and it's realistically a very harmless inconveniance for an antagonist to antagonize the station. That's really what it comes down to, an inconveniance you are often able to solve in a few minutes. It's random for a reason, and we're not the server where we put inconvenient stuff in the hall because it's slightly annoying.

BurgerLUA:
> I've did a lot of thinking on it, and it's realistically a very harmless inconveniance for an antagonist to antagonize the station. That's really what it comes down to, an inconveniance you are often able to solve in a few minutes. It's random for a reason, and we're not the server where we put inconvenient stuff in the hall because it's slightly annoying. There's going to be trends. Yesterday it was the Crab phone, today it's the fake blob announcements. Then eventually people will move on

have you considered that maybe both of these things are shit

radarseas:
I will add to this as well, the person that crabbed, **unless they don't have a bank account, which means they're a contractor that REALLY put time into this, or a lone infil in which case they ditched most of their useful TC... for something almost useless to them...** will have a _**REDICIOUS AMOUNT**_ of cash which can be immedately traced with the banking accounting console printable by the entirety of service with absolutely no plausable denability unless they keep micromanaging their cash intake till the crab ends, its not even that stealthy if you know the game well.

Just gotta play the strength of this item to its weakness. there's no metaprotections that it wasn't a traitor's doing so just get creative and you'll quickly realise its not a get-out-of-jail card.. hell you could probably get a cut of that cash in exchange for not outing them, RP-wise and be richer than you were pre-crab.

nevimer:
re-open requested to make changes and re-present the PR

PR #867, opened 2023-12-18
Loadout stuff, Include all future loadout edits in the respective module folder from now on

Majkl-J:
## About The Pull Request

All loadout Bubber overrides and Bubber exclusive loadout items now have their own module.

Also cyborg visor mask is now available in the loadout menu because it looks cool

## Why It's Good For The Game

Better code organizing + a cool mask added as a bonus

## Changelog

:cl:
add: Cyborg visor mask is now available in the loadout menu
code: Loadout items modular folder
/:cl:


theselfish:
Here was me planning to make a 100% cosmetic varient of the welding goggles for myself.

Thank you.
Next, hopefully we can sort the fact that we have
code
master_files which has it's own code section
and modules instead of just.
One or two file structures.

Majkl-J:
> Here was me planning to make a 100% cosmetic varient of the welding goggles for myself.
> 
> Thank you. Next, hopefully we can sort the fact that we have code master_files which has it's own code section and modules instead of just. One or two file structures.

I am slowly chewing my way through all this and refactoring it. Rather would go in smaller increments than to break stuff

Majkl-J:
This is so behind I don't wanna touch it with a mile long stick

PR #868, opened 2023-12-18
Removes Burgerstation from being votable

BurgerLUA:
## About The Pull Request

Removes Burgerstation from being votable. Yes I know that this should just be a config option but this PR is basically written and official permission for me to disable the map.

## Why It's Good For The Game

I hate playing on my own map because it means there is a high chance that people will complain about it when something bad happens. It basically sours my mood for the round and makes it difficult to play. There are rounds where I just want to chill and relax but that isn't possible when people ICly and OOCly badmouth the map.

Burgerstation 2 is in development anyways and that will fix most issues that people have with the map and people can wait until then.

## Changelog

:cl: BurgerBB
del: Removes Burgerstation from being votable.
/:cl:

### DIFF
--- a/config/maps.txt
+++ b/config/maps.txt
@@ -77,7 +77,6 @@ endmap
 ##BUBBERSTATION MAPS##
 
 map burgerstation
-	votable
 endmap
 
 map ss13_construct


Cyprex:
To be clear we use a different config serverside- your PR only updates the config someone checking out the codebase on localhost would see, so it doesn't matter much whether or not this PR is merged or not- we need to change it in the live server config, which kitten has!

PR #874, opened 2023-12-21
Makes the Changeling Horror Form scream quieter, but still horrifying.

BurgerLUA:
I recommend testmerging this and not merging this for now.

## About The Pull Request

Makes the Changeling Horror Form scream quieter.

Note that this isn't modular. Applying a modular edit to all instances where this is used in files would likely create more conflicts than editing the sound file that hasn't changed since forever.

## Why It's Good For The Game

Horrorform was too loud. This makes it more quieter.

## Changelog

:cl: BurgerBB
sound: Makes the Changeling Horror Form scream quieter, but still horrifying.
/:cl:


KazooBard:
I love this! People heard me through 50 walls hahah

PR #896, opened 2023-12-25
reverts the map vote to actually being democratic and does the same with the transfer vote

zeroisthebiggay:
## About The Pull Request

undoes the tg change that makes it so map votes are completely and utterly random
~~this could hypothetically be done modularly since it's a var and i will do it if the urge strikes me but i'm lazy so~~ i made it modular

~~also does the same to transfer votes~~ transfer votes may also be fine but it's good to be safe or so i say

## Why It's Good For The Game

the number one pop killer is maps people don't like being randomly voted in because there was chaos on the shuttle or something that prevented people from actually picking out the map they liked

right now it's quite literally a gamble if you will get delta which ten people voted for or get lima which one dude voted for as a bit and it rolled the ten percent chance to kill you instantly

## Changelog

:cl:
del: random map vote chances
del: random transfer vote chances
server: the map vote no longer has a modifier that makes it inconsistent
/:cl:


projectkepler-ru (approved):

nevimer:
Shouldn't have been merged due to the next upstream mirror already removing it

PR #898, opened 2023-12-25
Nerfs and Buffs botany because I played it once

BurgerLUA:
Draft because untested, uncompiled, and unhinged.

## About The Pull Request

I can't fairly access if this is an overall nerf pr or a buff pr, so people should be the judge. I can, however, tell you the changes.

- Decreases the botany life tick delay from 20 seconds to 15 seconds (growing and everything is faster).
- Perennial Growth received a heavy nerf. Grafts only give "Artifical Perennial Growth" which still give Perennial Growth, but cause penalties to be applied on harvest (plant is given 25% reduced yield and potency after harvest, and takes damage equivalent to shearing the plant with the botany shears). Regular Perennial Growth only exists on plants that normally have it, and it cannot be obtained.
- Decreases the instability requirement for species mutations from 60% to 30%.
- The percent chance for a plant to change species modifiers now ~0.8 points per instability instead of 0.5 points per instability.
- Decreases the instability requirement for "hard" mutations (drastic change in stats) from 40% to 30%.
- The maximum possible potency value for a plant is now based on instability. The baseline potency cap is 100, but increases by an additional point per instability. 

## Why It's Good For The Game

I tried out botany a bunch and I found it was incredibly slow compared to other mechanics. It feels like more of a slot machine than virology if you want a certain plant combination, and there are a lot of funny things you can do with botany that just feel weird, such as near infinite wood generation because of tower cap perennial growth. Hopefully this PR is sane enough for people to appreciate it.

## Changelog

:cl: BurgerBB
balance: Makes botany processing more frequent (faster), increases species mutation chances, nerfs Perennial Growth, and makes the maximum potency possible based on instability.
/:cl:

### DIFF
--- a/code/__DEFINES/botany.dm
+++ b/code/__DEFINES/botany.dm
@@ -75,7 +75,7 @@
 // obj/machinery/hydroponics/var/plant_status defines
 
 /// How long to wait between plant age ticks, by default. See [/obj/machinery/hydroponics/var/cycledelay]
-#define HYDROTRAY_CYCLE_DELAY 20 SECONDS
+#define HYDROTRAY_CYCLE_DELAY 15 SECONDS //BUBBERSTATION CHANGE, 20 SECONDS TO 15 SECONDS.
 
 #define HYDROTRAY_NO_PLANT "missing"
 #define HYDROTRAY_PLANT_DEAD "dead"
--- a/code/modules/hydroponics/grafts.dm
+++ b/code/modules/hydroponics/grafts.dm
@@ -36,7 +36,7 @@
 	if(trait_path)
 		stored_trait = new trait_path
 	else
-		stored_trait = new /datum/plant_gene/trait/repeated_harvest
+		stored_trait = new /datum/plant_gene/trait/repeated_harvest/lesser //BUBBERSTATION CHANGE: ADDS LESSER REPEATED HARVEST.
 	icon_state = pick(
 		10 ; "graft_plant" , \
 		5 ; "graft_flower" , \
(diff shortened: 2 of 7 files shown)


xXPawnStarrXx:
Due to the recently added food buffs, this could be just a net negative. You could end up with worse quality food by literally doing nothing wrong. If you lose 25% potency and yield, you're literally dropping an entire quality tier. 

Lowering instability requirements is also a negative, since many crops can get to 30-40 before they're even ready for harvest, so if you need a mutation or want multiple offsprings from one plant, you'll have to grow more and hope you don't get shafted. Some plants like tomatoes already start with something like 25 instability and have multiple offsprings, You don't have many ways to even counter instability besides harassing medical for a plumbing dispenser ASAP.

BurgerLUA:
> Due to the recently added food buffs, this could be just a net negative. You could end up with worse quality food by literally doing nothing wrong. If you lose 25% potency and yield, you're literally dropping an entire quality tier.
> 
> Lowering instability requirements is also a negative, since many crops can get to 30-40 before they're even ready for harvest, so if you need a mutation or want multiple offsprings from one plant, you'll have to grow more and hope you don't get shafted. Some plants like tomatoes already start with something like 25 instability and have multiple offsprings, You don't have many ways to even counter instability besides harassing medical for a plumbing dispenser ASAP.

The 25% penalty is added AFTER the harvest, so you can still get 100% potency and 10 yield.

xXPawnStarrXx:
If you want to make this faster; You'd be better off adding maturation speed as a modifiable variable.

PR #901, opened 2023-12-26
ALLOW ROCK THE VOTE!

yorii:
## About The Pull Request

I don't actually know what it does..

## Why It's Good For The Game

Allows people to revote when the 1/100 vote wins for a shitty map nobody wants to play?

## Changelog

:cl:
config: allowed rock the vote
/:cl:


yorii:
I have been informed it lets you reroll the vote one time, using the same weighted values.

YakumoChen:
all that really happened with this was that rock the vote was voted for every time
there are fixes being made for the random vote weighting anyway

projectkepler-ru:
Thoughts? @StrangeWeirdKitten

YakumoChen:
This is also config and won't actually change the server so there's that

Majkl-J:
Not needed with #896

BurgerLUA:
Rock the vote should be enabled, I think, because sometimes a bad map is a wakeup call.

The ideal system is rock the vote being called automatically when the shuttle leaves, and the map vote being called 1 hour after the round starts.

yorii:
> Not needed with #896

Yeah but that was a shitty change and I don't understand why it was speedmerged...

Majkl-J:
> > Not needed with #896
> 
> Yeah but that was a shitty change and I don't understand why it was speedmerged...

It was around for 2 days. Speedmerging is generally any change merged under 24 hours

You had time to comment on it, or name it as an alternative in your PR, prompting further discussion.

Calling it shitty also doesn't exactly help us, try actually providing a point as to why not have the regular vote back

Majkl-J:
> Rock the vote should be enabled, I think, because sometimes a bad map is a wakeup call.
> 
> The ideal system is rock the vote being called automatically when the shuttle leaves, and the map vote being called 1 hour after the round starts.

I'll bring this up in maintainer discussion as I agree with this point

nevimer:
Wrong repo. We don't use this ones configs. This is only for people who freshly pull.

PR #906, opened 2023-12-26
Nerfs Ling Horror Form

StrangeWeirdKitten:
## About The Pull Request

Brute damage: 40 -> 25
Speed: 0.5 -> 1
Health: 750 -> 500

They can keep their regeneration

## Why It's Good For The Game

A very easy to get ability. I have two options. I can make it much harder to get, or I can nerf it. But I'd rather do the second option.

First of all, you're a tank that hits hard. You should not be running around the station at more than normal simple mob speed.

Second of all, you have a 3 brute regen per second. You can live with 500 health. You can also devour people to regen health anyway.

Third of all, you have other abilities, including range. You can afford to loose some melee damage in compensation.
## Changelog

:cl:
balance: Nerfs Horror Form
/:cl:


Majkl-J (changes requested):
Please make it modular

projectkepler-ru:
I think we can override this on bubber modular folder actually? @StrangeWeirdKitten

StrangeWeirdKitten:
Sure, I'll do that later

destrucktoid:
Suggestions: 
If we are nerfing move speed then they need a new escape tool as they will end up with literally no escape tools when they begin to run low on health. As such i would recommend either giving them back SOME move speed or allow them to vent crawl once again. Alternatively you can look into giving them additional armor or health regen, maybe even a teleport/jaunt or an ability to become invisible temporarily.

i think almost halving the damage output to 25 is a bit excessive, We'll see how it plays out though in game. 

I definitely don't think the health changes should go ahead packaged with all the other changes.

The thing is with horrorform is once you become it there is no going back into your normal form. As such it is almost always used as a worst case scenario panic button. If we are removing all of it's mobility options as well as nerfing it's health then i think it becomes almost useless, especially with the reduction to it's damage output.

StrangeWeirdKitten:
I'll let them have 30 force.

Majkl-J (approved):
Love you. Ready to merge in my book.

PR #919, opened 2023-12-28
[TG UPSTREAM MIRROR] (#80593) (#80594) (Fixes CI on biodome + other CI fixes)

Majkl-J:
Mirrors 
- https://github.com/tgstation/tgstation/pull/80594
- https://github.com/tgstation/tgstation/pull/80593

New ability unlocked: PR Mirroring

## About The Pull Request
Immerse was causing harddels due to it having references to mobs. Makes it use weakrefs for mobs instead.

The delete the world test was showing that the turbine's non-core parts, the compressor and outlet were causing runtimes, so I made the deactivation of the core remove the references from the secondary turbine parts

## Why It's Good For The Game
Less harddels means less lag, hopefully

Immerse would cause harddels if a mob was deleted while it was in it's list. It could probably also happen if a turf was deleted too, but doing that here would be much harder.

no CL since nothing playerfacing


Majkl-J:
Ran the check like 4 times and it is no longer failing, so

PR #925, opened 2023-12-30
Adjusts officer position scaling to approach TG values

ReturnToZender:
## About The Pull Request
This makes it so officer positions hover between 6 and 10 depending on population. Skyrat set their security population to 8, with no scaling, but Skyrat has a lot less conflict, and a lot more individual power for Security. This helps remedy the issue.

## Why It's Good For The Game
Security has been properly balanced to approach TG, now this will make sure we have the highpop scaling we need to match. Set to 10 as opposed to twelve as I'd prefer to not overshoot.

## Changelog

:cl: ReturnToZender
balance: partially reverts the Skyrat limit on Security population
/:cl:


projectkepler-ru:
I've always wanted to make this PR but I've just


never gone around to actually doing it
@StrangeWeirdKitten  it's something I really want but I want to hear your take on this

projectkepler-ru:
@ReturnToZender  and as for the extra equipment to compensate- currently due to my PR, security still starts with a sec belt so this likely won't be an issue but they won't start with a sechud, so it'll be an item that they have to source in the game but it isn't **hard** to get

You could do me a favour and PR to give them a sechud (not sunglasses) to solve that issue

PR #935, opened 2024-01-01
bubber always message staffchat ahelps

nevimer:
internal change test, we might get sick of it


ReturnToZender:
This should also send faxes to the Discord and let us respond to them from the Discord like tickets. Or prayers!

Shadow-Quill:
looks like the contents of ahelps are HTML encoded -- can they be decoded before being broadcasted to Discord?
also, it'd be nice if tags like `<a href='?priv_msg=ckey_here'>...</a>` were stripped from the output

PR #946, opened 2024-01-04
teshari accessory sprites

aKromatopzia:
## About The Pull Request

so far, just the department arm bands and allergy tag. I'm probably making a second pr for all the other clothing sprites when I get to them Soon™. Although since SR is read-only now, I'm not sure where to really put this. I can't exactly make a new file either, since the teshari loadout stuff points at the modular_skyrat accessory dmi for overrides.
## Why It's Good For The Game

it was missing and that annoyed me, electric boogaloo. the allergy dogtags clipping about five tiles away into the aether was particularly egregious.

## Proof Of Testing

teshside
teshband
deputy

## Changelog

:cl:
add: teshari accessory sprites
/:cl:


nevimer:
hmm. we'll have to devise a way to make these files be another file not in the skyrat one.

aKromatopzia:
> hmm. we'll have to devise a way to make these files be another file not in the skyrat one.

my only idea is simply to change where teshari_accessories_override points to but that's also hacky

PR #853, opened 2023-12-15
Grants the ability to open Clown slots once more.

theselfish:
## About The Pull Request

Title. Skyrat disabled it because they hate fun. We don't.

## Why It's Good For The Game

Say a Clown dies. Is arrested, or is just a bit shit.
What can you do?
Jackshit.
Pray, fax Clown Planet, but why wait on God or the Clown in the High Castle when you can do it yourself?
This also let's Antags, or Clowns summon more friends. Assuming people in the lobby wanna be a clown.

P.S. I commented out and unticked the Skyrat file for double safety. Tried to just override it, didn't work. If you have a better idea for this. You have the floor.

## Changelog

:cl:
add: Clown slots are re-openable by royal decree. The incident between Nanotrasen and King Honkington has been resolved.
/:cl:

PR #960, opened 2024-01-06
#853 but it works.

theselfish:
## About The Pull Request

I'm a dumbass.

## Why It's Good For The Game

Stuff should work.
## Proof Of Testing

image

## Changelog

:cl:
add: ACTUALLY REALLY Lets Clown and Mime slots be opened again.
/:cl:


code/modules/modular_computers/file_system/programs/jobmanagement.dm
```diff
@@ -36,7 +36,7 @@ GLOBAL_VAR_INIT(time_last_changed_position, 0)
 		return FALSE
 	// SKYRAT EDIT ADDITION START
 	if(job.veteran_only)
```
nevimer, line 38: one merely cannot create an empty followup to the if condition. this will surely provoke linters.

modular_zubbers/code/modules/jobs/job_types/clown.dm
```diff
@@ -1,3 +0,0 @@
-/datum/job/clown/New()
```
nevimer, line 1: ill need to check why this existed

theselfish: I added it to try and get around the bitflag. Just commented it.

code/modules/modular_computers/file_system/programs/jobmanagement.dm
theselfish, line 38:
```suggestion
	//if(job.veteran_only)
```

Majkl-J (approved):
Bitwise operations are a bitch. I'll look into why the original workaround didn't work later, as of current, ready to merge

theselfish:
> Bitwise operations are a bitch. I'll look into why the original workaround didn't work later, as of current, ready to merge

The original method didn't work because of the check I'm commenting out.
While everyone is technically a vet, the job, mechanically is a vet job. So the code is saying 'they can't open a slot no matter what'.

Majkl-J:
> The original method didn't work because of the check I'm commenting out. While everyone is technically a vet, the job, mechanically is a vet job. So the code is saying 'they can't open a slot no matter what'.

Do we need to remove the rest then?

theselfish:
> > The original method didn't work because of the check I'm commenting out. While everyone is technically a vet, the job, mechanically is a vet job. So the code is saying 'they can't open a slot no matter what'.
> 
> Do we need to remove the rest (Flags and such) then?

I'm unsure, all I know is this works for what I'm trying to do.
I mean, I would say yes if we're really cutting the rope with skyrat, but, you're the maintainer.
I think this is the only example of it.

code/modules/modular_computers/file_system/programs/jobmanagement.dm
theselfish: Commented out.

theselfish:
@Majkl-J @nevimer @StrangeWeirdKitten Waiting on the worms, here.

Majkl-J (changes requested):
Since breaking off from skyrat isn't certain yet, put back the modular flag override, then I'll merge it

theselfish:
> Since breaking off from skyrat isn't certain yet, put back the modular flag override, then I'll merge it

The override doesn't work, to my knowledge. I think it was written wrong?

Majkl-J (approved):
image

PR #969, opened 2024-01-08
Gives medical borgs back their crew monitor

StrangeWeirdKitten:
## About The Pull Request

Simply allows medical borgs to get the crew monitor upgrade again

## Why It's Good For The Game

We have a ton of medical chaos and they need a little help. This was removed on Skyrat because of worries that borgs encroach on their job. Are some of these worries valid? Yes. But medical here always has plenty of work to go around.

## Proof Of Testing

image

## Changelog

:cl:
balance: gives medical borgs back their crew monitors
/:cl:

### DIFF
--- a/modular_skyrat/master_files/code/modules/research/techweb/all_nodes.dm
+++ b/modular_skyrat/master_files/code/modules/research/techweb/all_nodes.dm
@@ -245,11 +245,11 @@
 	design_ids += list(
 		"borg_upgrade_surgicaltools",
 	)
-
+	return ..() // Bubber Edit - Moves this up because it doesn't compile otherwise
 /*	design_ids -= list(
 		"borg_upgrade_pinpointer",
-	)*/ //BUBBERSTATION REMOVAL
-	return ..()
+	)
+	return ..() */ //BUBBERSTATION REMOVAL
 
 /datum/techweb_node/basic_mining/New()
 	design_ids += list(
--- a/modular_skyrat/master_files/code/modules/research/techweb/all_nodes.dm
+++ b/modular_skyrat/master_files/code/modules/research/techweb/all_nodes.dm
@@ -248,8 +248,8 @@
 
 /*	design_ids -= list(
 		"borg_upgrade_pinpointer",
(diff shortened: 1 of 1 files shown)


nevimer:
This might seem like a bad idea on paper. In reality, however, medical players need all the help they can get.

modular_skyrat/master_files/code/modules/research/techweb/all_nodes.dm
```diff
@@ -246,10 +246,10 @@
 		"borg_upgrade_surgicaltools",
 	)
 
-	design_ids -= list(
+/*	design_ids -= list(
 		"borg_upgrade_pinpointer",
 	)
-	return ..()
+	return ..()*/ //BUBBERSTATION REMOVAL
```
nevimer, line 252: You can't just remove the return ..() part. You will cause unforseen consequences here.

nevimer (changes requested):

Strazyplus:
Great to have that back... but... @StrangeWeirdKitten Which one is it?
Crew monitor which tells you crew vitals or...
The Crew tracker which is a pinpointer.

the icon you showed is the Crew Pinpointer.

Detective-Google:
> This might seem like a bad idea on paper. In reality, however, medical players need all the help they can get.

disagree, still only takes one medical borg to almost entirely invalidate the medical department and making the mediborgs take the extra step of actually going and checking a console or asking the AI is pretty much the only thing that stops paramedics being 100% invalid. why even have paramedics when an unslowable unspaceable fireproof all access machine with a rollerbed can do the job more efficiently.

StrangeWeirdKitten:
> Great to have that back... but... @StrangeWeirdKitten Which one is it? Crew monitor which tells you crew vitals or... The Crew tracker which guides you to the crewmember.
> 
> the icon you showed is the Crew Pinpointer. Cyborgs technically have a pinpointer called lifeline while slightly a tedious thing to obtain.

Both of these items are connected together. This is the upgrade TG provides.

PR #975, opened 2024-01-09
Update dynamic event population threshold

lessthnthree:
## About The Pull Request

Updates the event population threshold setting to reflect server population cap. With the old numbers catering to the other server and lack of population cap, things like level 2 moldies and other things that use this threshold don't spawn because Bubber doesn't ever get 120-130 alive population at any given time, in fact it can't due to the cap.

This tightens the range of what's defined as minpop to maxpop.

Events that use a specific defined number of minimum players for an event have no change.

## Why It's Good For The Game

Increased variety in certain random events

## Proof Of Testing

image

## Changelog

:cl: LT3
config: Adjusted population thresholds for certain random events
/:cl:


StrangeWeirdKitten:
yeah I was meaning to do this at some point, but I keep forgetting lol. Thanks!

PR #976, opened 2024-01-09
[EARLY PULL] Replace natural beheading with cranial fissures

StrangeWeirdKitten:
…our skull) [MDB IGNORE] (#26020)

* Replace natural beheading with cranial fissures (splitting your skull) (#80703)

## About The Pull Request

Replaces natural beheading (doing a lot of damage to the head) with cranial fissures, which split your skull in half.

dreamseeker_2023-12-31T19-38-16

While you have this wound, your eyes can be pulled out of your head with bare hands, and slipping will spill your brain out of your head.

Removes beheading objective from traitor.

Zombies can be beheaded all the same.

Any other way of beheading, such as surgery or amputation shears, is still possible.

PR #978, opened 2024-01-09
Dismemberment Alternative

BurgerLUA:
Alternative to #976

## About The Pull Request

Heads can only be dismembered if they have a critical-tier (Compound) fracture.
If a head was supposed to be dismembered from a blow, a compound fracture will be applied.

## Why It's Good For The Game

This is a replacement to the up and coming dismemberment rework that makes dismemberments harder to obtain through accidents. It is functionally the same, however with the reduced grossness for it.

The reason why I am so against that PR is that I think it's too gruesome for SS13. Yes, we have disemboweling, delimbing, gibing, etc etc but personally I draw the line at still being alive and aware with your brain exposed. It legitimately gives me nausea and headaches thinking about it because of a condition I have.

## Proof Of Testing

## Changelog

:cl: BurgerBB
balance: Heads can only be dismembered if they have a critical-tier (Compound) fracture. If a head was supposed to be dismembered from a blow, a compound fracture will be applied.
/:cl:


StrangeWeirdKitten:
When it comes down, there will need to be changes to ensure this PR is compatable. But I'm 100 percent fine with this.

BurgerLUA:
That's something I can do when there is an upstream merge.

Majkl-J:
image

Considering you will already be dead by the point you can even be considered for dismemberment, this just adds tediousness where there isn't any. Personally feel that the problem lays elsewhere, that being that head and brain sprites are terrible to find.

But sure still better than whatever tg is trying

Anyhow to stop fissures from happening, all you should need to do is
```
/obj/item/bodypart/head
	can_dismember = TRUE
```
See this:
https://github.com/Bubberstation/Bubberstation/pull/976/files#diff-bdc3322ee50db9d7162b55de252b5b9520abfdd3b46de8963ecad5eabc8fc5beR97

Majkl-J (approved):
To be tested and merged together with https://github.com/Bubberstation/Bubberstation/pull/976

modular_zubbers/code/modules/surgery/bodyparts/dismemberment.dm
```diff
@@ -0,0 +1,14 @@
...
+
+/obj/item/bodypart/head/try_dismember(wounding_type, wounding_dmg, wound_bonus, bare_wound_bonus)
+
+	if(!can_dismember(src)) //No point in running the below if we can't dismember in the first place.
+		return FALSE
+
+	if(!src.get_wound_type(/datum/wound/blunt/bone/critical)) //If we don't have this wound...
+		var/datum/wound/blunt/bone/critical/wound_to_add = new //Give it to them...
+		if(wound_to_add.apply_wound(src,wound_source = "near dismemberment")) //But in case we can't add it for some reason, just do the fallback code!
+			return FALSE
```
nikothedude, line 12: dont do this! wounds are way more modular now, this wont work on synths

seeing this made me realize we really need a limb.cause_wound_of_type_and_severity

```suggestion
	if (!owner.cause_wound_of_type_and_severity(WOUND_BLUNT, src, WOUND_SEVERITY_CRITICAL, wound_source = "near dismemberment"))
	    return FALSE //But in case we can't add it for some reason, just do the fallback code!
```

BurgerLUA:
Tested and works.

PR #982, opened 2024-01-10
Reworks or outright removes some of the problem causing Wizard spells so it can possibly be re-added to rotation

BurgerLUA:
### Summon Soulstones
**Description:** A belt filled with 6 purple soulstone shards that can can be used to trap and enslave the spirits of your friends! You also gain the spell Artificer, which lets you spawn a shell every 60 seconds. Unlike other artifacts, this cannot be used by the crew. Unlike cult constructs, wizard constructs are colored purple rather than red, so don't expect to dupe the crew into thinking there's a cult.
**Reason for Removal:** Soulstones were effectively round removal. They basically dusted your character and forced you to be in a rock. If you were lucky, you could get saved by the Chaplain or get put into a construct, but either of these option result in unsatisfying gameplay. If you want minions, just get another spell that gives minions.

### Curse of Madness
**Description:** Using this curses the station, warping the minds of everyone inside, causing lasting traumas (you can send a message with it too as the catalyst). These traumas are usually pulled from the regular list, but there are a couple of unique magical traumas only found by using this spell.
**Reason for Removal:** Curse of Madness, while funny, can absolutely straight up fuck you over for the entirety of the round if you rolled a crippling trauma, such as blindness, paralysis, or deafness. Bubberstation rounds typically last an upwards of 2 hours so having this is just not a good idea.

### Staff of Change
**Description:** Allows the wizard to randomly change anyone and anything into an alien, slime, lizardman, cyborg, Syndicate cyborg, monkey, or more! Holds up to 6 charges, and recharges one every 4 seconds. Using this staff as a non-wizard has a backfire effect, instantly shapeshifting the wielder. Transformations into cyborgs and xenos are disabled on backfires.
**Reason for Removal:** Very easy to effectively get round removed because you got transformed into a dog and the only way to revive you is an inducer. Also can't be used by crew so you literally cannot get saved under any circumstances unless the wizard themselves uses it as using it as crew will cause it to backfire. Also a favorite to use by griefers because it means that they can dick around as a xenomorph or a slime.

### Barnyard Curse
**Description:** This spell dooms an unlucky soul to possess the speech and facial attributes of a barnyard animal. Anyone unfortunate enough to be targeted with this spell cannot remove the mask without the help of [chemistry](https://tgstation13.org/wiki/Guide_to_chemistry#Fluorosulfuric_Acid) and can only communicate with obnoxious animal noises. For when you just don't give a fuck.
**Change:** The curse of the mask only lasts 3 minutes after being applied.
**Reason for Change:** A spell that effectively makes you mute and prevents you from wearing another mask. Lasted forever unless you used a heavy dose of acid (which I think is currently bugged at the moment) or straight up decapitation.

### Mind Transfer
**Description:** Swaps the minds of the wizard and victim, best used if the wizard is being beaten in a physical battle. Mindswap targets will be unconscious for quite some time. Only works at melee range.
Does not work on cultists, revolutionaries, changelings, petrified people or summoned demons.
**Change:** Now only works on soulless beings, which means you can mindswap with pets or monkeys.
**Reason for Change:** Forcing people out of their character is not very good especially when in 99% of cases it results in round removal as people (rightfully) mistake you for the wizard.

### Smite
**Description:** This spell will charge your hand with a powerful energy that can blow up the next person or cyborg that you touch. Their brain will still be recoverable. It is seen as a cheap way to kill.
**Change:** No longer gibs, but instead instantly kills you and flings you an absurd distance, almost always resulting in fractures or bone breaks.
**Reason for Change:** Insta-gib spells, while funny, were a bit too strong and annoying to fix in SS13. This should be a better alternative.

## Why It's Good For The Game

More gamemodes are good. I think wizard rounds can be fun with the right abilities changed/reworked/etc. Hopefully with this PR, wizard can get enabled again.

## Proof Of Testing

image

Pictured: A wizard who mind transfered into Pun Pun Smiting a naked assistant after removing a mask where they accidentally casted barnyard curse on themselves.

## Changelog

:cl: BurgerBB
balance: Balances or outright removes some wizard spells.
/:cl:


radarseas:
idea.. why not make smite deal like 60-90 in a mix of brute and burns, as well as making it delimb most of someone's limbs? It gives it some of its former, put down power without making it a boring instant-kill move under most circumstances.

It gives a window to react and call for help instead of it being deathwand, but melee. Nearly none of the other wiz spells quite instant-kill like smite does and for them to let people know where their limbs are, plus if they wind up dead, it still leaves medical/robotics with a decent workload for smite victims (its also funnier when the wiz uses it on a carpfu/CQC user and gets the defect uno reverse, with a small window to scream for mercy, instead of outright killed.)

to be honest, I think TG might accept an upstream of smite changes under a similar change, I hard agree on the smite nerf, smite in its current state **_frankly just, sucks._**, its an instant kill gib move in the least interesting way possible.

radarseas:
> Soulstones were effectively round removal. They basically dusted your character and forced you to be in a rock. If you were lucky, you could get saved by the Chaplain or get put into a construct, but either of these option result in unsatisfying gameplay. If you want minions, just get another spell that gives minions.

honestly, I think soul shards having their entire function being reworked to ping dchat for players instead of sharding bodies would be the right way to go if we want to steer from the fears of round removal.

It'll both keep in the soul-belt and add potental for a cult rework on sacrificing without a loss of the use of soulshards axing players from the round, while keeping a consistant threat in (since they can produce constructs easier.

plus it'll prevent chaplains using soulshards on corpses of people that don't really consent to it, like a random person who died in a hallway.

_buuuuut that's probably a little off-topic and for another PR._

BurgerLUA:
Everything is tested and works.

Majkl-J (approved):
Would be funny to see disintegrate home the dead body at another bystander.

Anyhow yea the changes are good. I see no reason why not have them if they will allow wizard rounds to happen, adding more variation.

Majkl-J:
Same reason for merge as #983

PR #990, opened 2024-01-11
[Do not merge] removes headsets from ghost cafe

StrangeWeirdKitten:
upstream bug, will close when they fix it. This is a very temporary solution, bandaid solution. I just edited the headset amounts of each of the vendors to zero.

Basically this bug.
https://github.com/Skyrat-SS13/Skyrat-tg/issues/26077


StrangeWeirdKitten:
I found the bug and I'll look into fixing it later. Basically the headsets are a skyrat_product and exclude_product doesn't also exclude skyrat ones.

PR #1018, opened 2024-01-16
Moves the emergancy shuttle off the interlink

StrangeWeirdKitten:
## About The Pull Request

Moves the emergancy shuttle from the interlink to the normal centcomm dock.

## Why It's Good For The Game

This will allow the EORG to be easier to understand rule wise and you would not have to worry about Interlink dorms

## Proof Of Testing

image

## Changelog

:cl:
code: Switches the emergancy shuttle to Centcom instead of the Interlink
/:cl:


theselfish:
Good.

Every day we cut down the monolith of Skyrat changes.

:+1:

StrangeWeirdKitten:
It works. We have EORG, it is now contained to the admin centcomm.

PR #1038, opened 2024-01-19
Fixes the RBMK having an infinite meltdown state + balance.

BurgerLUA:
## About The Pull Request

image

## Proof Of Testing

I did not test this but it will work. This is my burger seal of approval.

## Changelog

:cl: BurgerBB
balance: The RBMK can now handle more power generation before going into meltdown with upgraded parts, and also generates less power with upgraded parts.
fix: Fixes the RBMK having an infinite meltdown state when safeties trigger at the same time as a meltdown.
/:cl:


nevimer:
ready?

BurgerLUA:
Going to do a little more work with it first to absolutely be sure that there aren't any new issues from the new upgrade maths then it will be ready.

PR #1039, opened 2024-01-19
Adds a new map (Tilestation)

Majkl-J:
## About The Pull Request

The people have spoken

image

## Why It's Good For The Game

image

Also I guess it loads really fast when you set offmap levels to 0

## Changelog

:cl:
add: Newest Nanotrasen budget cuts result in a minimized station layout for maximum work efficiency.
/:cl:


YakumoChen:
Boxstation next

4 tiles
12 walls
4 large cardboard boxes filled with boxes of monkey cubes

YakumoChen:
Actually just make the station a boxing ring

StrangeWeirdKitten:
It's cool but tbh after ss13 construct, I'm not the biggest fan of adding even more map bloat to our files. Also this will 100 percent fail a lot of mapping CI and you need to exclude way more than what is in the current file. But I'll also let @nevimer consider this too.

Majkl-J:
> It's cool but tbh after ss13 construct, I'm not the biggest fan of adding even more map bloat to our files. Also this will 100 percent fail a lot of mapping CI and you need to exclude way more than what is in the current file. But I'll also let @nevimer consider this too.

Honestly it's mostly here for the meme of one silly vote, no need to actually have it, just nice to have the PR around for some insane event (Maybe an april fools one?)

Anyhow, as for CI, I can just make every check that fails excluded in `ignored_unit_tests` in the config

StrangeWeirdKitten:
> > It's cool but tbh after ss13 construct, I'm not the biggest fan of adding even more map bloat to our files. Also this will 100 percent fail a lot of mapping CI and you need to exclude way more than what is in the current file. But I'll also let @nevimer consider this too.
> 
> Honestly it's mostly here for the meme of one silly vote, no need to actually have it, just nice to have the PR around for some insane event (Maybe an april fools one?)
> 
> Anyhow, as for CI, I can just make every check that fails excluded in `ignored_unit_tests` in the config

if we do a silly event map, feel free to reopen with a better one. (I think it's fine to have like a 9x9 station with various departments that's "functional") where it can be TM'd

PR #1053, opened 2024-01-22
De forest refiller

xXPawnStarrXx:
## About The Pull Request
The DeForest Medpens were cool when they were first released, but due to only being able to get them in kits and their one use nature made them somewhat fall quickly out of fashion, when they should be a paramedic's best friend. I wanted to change this so I've added the ability to refill them with the medpen refiller, requiring 10u of the specific chem. Also adding the groundwork for anyone to add any other pens they wish to refill, such as mining ones.

occuisate - oculine or inacusiate.
morpital - morphine.
lipital - probital.
lepoturi - lenturi.
coagulants - saline-glucose solution.
calopine - atropine.
synephrine - epinephrine.
meridine - potassium iodide.
halobinin - antihol.

## Why It's Good For The Game
Medpens are useful, never hurts to have a medkit in your pocket if you know you can replenish the pens. Makes you less likely to 'conserve it' incase you need it (which is never, I always do this.)

## Proof Of Testing
image

## Changelog

:cl:
qol: made DeForest medipens refillable.
/:cl:


Majkl-J (changes requested):
I wonder if it would be better to attach the refill chemical datum type to the pens directly, allowing it to be added to the examine description, and making the refiller work with that.

xXPawnStarrXx:
> I wonder if it would be better to attach the refill chemical datum type to the pens directly, allowing it to be added to the examine description, and making the refiller work with that.

I considered that, but I felt it was easier this way because it's less likely for me to screw something up when it's essentially a clone of something /TG/ already has.

Majkl-J:
I keep telling myself I'll merge this then I subsequently forget

PR #1060, opened 2024-01-24
I BROUGHT YOU INTO THIS WORLD, I CAN TAKE YOU OUT (Removes Slaughter Demons from wizards, limits laughter demons to one purchase)

BurgerLUA:
## About The Pull Request

Removes Slaughter Demons from the wizard spellbook.
Laughter Demons cost 2 points (instead of 1), and are limited to 1 purchase per wizard.

## Why It's Good For The Game

Slaughter Demons are a little too crazy for bubberstation. They're extremely tanky, robust, etc etc and their only drawback is "They could kill the wizard" which never really happens. I'm 100% against ghost roles that have this much power to take out half the crew, but I still think they have some sort of place in Bubberstation, so laughter demons are being kept.

## Proof Of Testing

## Changelog

:cl: BurgerBB
del: Removes Slaughter Demons from the wizard spellbook.
balance: Laughter Demons cost 2 points (instead of 1), and are limited to 1 purchase per wizard.
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/spells/spell_types/disabled_spells.dm
+++ b/modular_zubbers/code/modules/spells/spell_types/disabled_spells.dm
@@ -12,3 +12,8 @@
 // Very easy way to get round removed, and a very easy way to grief the crew as you are now a funny xenomorph that can lay eggs.
 /datum/spellbook_entry/item/staffchange
 	category = null
+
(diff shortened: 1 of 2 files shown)


Majkl-J (approved):
Personally think the limit on laughter demons is all they need as I don't want to overcompensate too much with the cost.

Otherwise looks good, slaughters are only good at slaughtering people's fun

PR #1003, opened 2024-01-14
Remaps NTC offices for higher quality

ReturnToZender:
## About The Pull Request

This is an NTC QOL PR!

This re-maps some of the NT Consultant offices that are so often neglected on various maps, to bring them up to snuff. And by up to snuff, I mean in some cases the bare minimum of usability. KiloStation doesn't even have a fax machine!!

## Why It's Good For The Game

NTC is the number 26 most played job, but most of its offices don't really hold up to the fidelity of the ones made on maps like Void Raptor or Blueshift. This makes most of the NTC offices have a bedroom as well as some special amenities!

## Proof Of Testing
image
image
image
image
image

image

## Changelog

:cl: ReturnToZender
balance: NTR's ACCESS_CREMATORIUM replaced with ACCESS_SERVICE
add: Assistants on-station may be pleased to hear that they can now wear Head of Staff pet collars again. ...Again.
add: Nanotrasen has deemed it fit to order an upgrade to all Consultant offices on stations in the Spinward Sector. Check them out!
/:cl:

PR #1075, opened 2024-01-27
Interlink Remap

ReturnToZender:
## About The Pull Request

This remaps the Interlink to have all latejoins spawn outside the shuttle, adjusts all the dorm rooms, and adds a new, modifiable plaza area to the right side of the shuttle room where players can build shops.

Also includes a barber shop for Assistants, due to #1003, but this does not need to be merged alongside that PR to function. 

This project started with "hm, I could make the dorms a bit smaller" and ended with an entirely different Interlink than what I started with.

## Why It's Good For The Game
The Interlink now has uniform dorms, as well as offering a small social area for players to interact there. It will improve roleplay by adding modifiable shop stalls that can be taken over by players.

## Proof Of Testing
In Map Editor:
image

some screenshots:
image
image
image
image

Original area for reference:
image

## Changelog

:cl: ReturnToZender
add: A new interlink to sink your teeth into.
del: Interlink docking port
/:cl:


theselfish:
image
You should probably add another door here so people can be let in.

lilBlovy:
I do see that there is no Glass PDA to taunt people while they are ghosted. It would be nice if that was added back in somewhere.

lilBlovy:
And, another note, The rooms seem to be really really really REALLY small compared to what they use to be. Would it be possible to maybe make a bigger area where the lower 4 rooms are, move and rotate the higher 4 rooms over to the right of the hall that lets you get down there in the first place, then expand those rooms so there is alot more room to work with?

Majkl-J (changes requested):
Not a big fan of the cutting of the whole lower area, would be much happier if it was somehow made accessible.

Also doesnt this override the current cafe?

ReturnToZender:
> Not a big fan of the cutting of the whole lower area, would be much happier if it was somehow made accessible.
> 
> Also doesnt this override the current cafe?

It sure does! The ghost cafe remains the same, though. With that said, I can restore the lower area and make it accessible to anyone.

ReturnToZender:
Closed due to stale, need to sit down and do a serious refresh on this

PR #1078, opened 2024-01-27
Nerfs Lichdom because someone messaged me on discord about it once

BurgerLUA:
## About The Pull Request

After 3 Lichdom resurrections, announcements will be made about your general phylactery location (the area) when you die.
Phylactery binding now requires the object to be at least bulky in size, and it cannot be fireproof, lava proof, or indestructible.

## Why It's Good For The Game

Lichdom is insanely strong, even with the drawbacks that make you generally weaker. Items can be hidden away in space (skeletons are immune to space!) and can be small or hard to see objects such as pens or cigarette butts. Some items can be fireproof and acid proof as well, requiring usage of mechanics that delete objects entirely (very limited and easy to sabotage, maybe the supermatter is reliable but that can be sabotaged too).

Regardless this PR should make lichdom less crazy.

## Proof Of Testing

Untested and in a draft because feedback is needed.

## Changelog

:cl: BurgerBB
balance: After 3 Lichdom resurrections, announcements will be made about your general phylactery location (the area) when you die. Phylactery binding now requires the object to be at least bulky in size, and it cannot be fireproof, lava proof, or indestructible.
/:cl:


Majkl-J (commented):
I have an iffy feeling about the limited revives thing and would rather see something else, like giving the chaplain the ability to sense the magic when he's in the same general area. 

As for the item indestructibility, yea that makes sense. Why wasnt it there before? (Imagine the captain forgetting the disk)

BurgerLUA:
> I have an iffy feeling about the limited revives thing and would rather see something else, like giving the chaplain the ability to sense the magic when he's in the same general area.
> 
> As for the item indestructibility, yea that makes sense. Why wasnt it there before? (Imagine the captain forgetting the diskie only for the wizard to be linked to it)

Actually the code specifically forbids the nuke disk from being used but nothing else.
Just to be clear, revives aren't limited, but after 3 revives your location gets revealed each time after. Lichdom is EXTREMELY powerful and effectively gives you extra lives to an already evasive and powerful antagonist.

radarseas:
indestructable doesn't mean you can't just toss it into the supermatter (which destroys everything but the NAD, disk, hense its marked exclusively to not be ensoulable.)

PR #1087, opened 2024-01-30
Makes Moon Station actually a moon because people made fun of me for the map not actually being a moon once

BurgerLUA:
image

## About The Pull Request

Moon Station is now a "planetary" map!

### Atmospherics

The atmos is a very cool 60 Celsius with a custom mixture of gas that is still breathable. Special precautions should be taken with the supermatter as cooling it is now harder (but still possible). The default setup with one emitter and no altered pipes (and just using the passive cooling) generates 1MW and doesn't overheat the supermatter. Supematter events do cause the cooling to fail, which is why 3 powered coolers are provided in case that happens.

### Ruins and Lavaland

Given the planetary nature of the map, there are no longer any space ruins. There also isn't a lower level yet, so in the meantime Miners can fly down to lavaland. Soon I will be adding a custom cave jungle system that miners will use for mining, with custom bosses and other cool content.

### Lighting

I added Planetary lighting exclusive to Moon Station from my other coderbase. The lightning does not have a day/night cycle (too performance heavy!) but it still looks pretty cool and lights up the entire outside area. This also means that solars basically has a generous boost of power 24/7.

## Current Issues

Non discovered... so far.

## Changelog

:cl: BurgerBB
add: Makes Moon Station actually a moon now because people made fun of me.
/:cl:


Kepteyn:
Hello I was the one who made fun of you for it not being a moon. Thanks for making it a moon.

PR #1093, opened 2024-01-30
[MIRROR] Fixes the tram's schrodinger-like power consumption

lessthnthree:
## About The Pull Request

Fixes https://github.com/tgstation/tgstation/issues/81106. The tram's dynamic power use does not get recalculated when leaving an area because the powernet between areas is made for static loads, This leads to APCs being drained quickly consuming too much power, or consuming negative energy leading to an APC being charged.

The amount of times use_power is called on tram crossing signals during processing also leads to unexpected power consumption.

This moves tram power to a rectifier machinery object in each area, controlled by signals from the tram. Tram crossing signals are fixed in how they consume power, changing between power use states instead of consuming during processing, and moved from the transport subsystem to fastprocess because they don't need to be running at 20fps.

## Why It's Good For The Game

Tram draining an APC in a matter of seconds is bad. Tram having negative power draw at other times is also bad.

image


lessthnthree:
there we go

PR #1098, opened 2024-02-02
[Non-Modular] Removes Spectral blade because ghosts pointed at my shoes and called them wack

Odairu:
## About The Pull Request
Removes spectral blade from the ash drake loot pool

## Why It's Good For The Game
This item allows ghosts to be visible around a player and they can point and communicate, leading to some metagaming issues, solving this by removing the item rather than administrative action as its intended use is not conducive to our rules. Its not easy to remove their ability to communicate, so removal is just better.

## Proof Of Testing
I opened like 30 chests on local host, no sword

## Changelog

:cl:
Removes Spectral Blade
/:cl:


Swiftfeather:
Spectral blade is fun, why not just lower drop chance?

YakumoChen:
The actual mistake here is relying on the Ghost Court for anything other than spelling out slurs on ouiji boards

imedial:
> The actual mistake here is relying on the Ghost Court for anything other than spelling out slurs on ouiji boards
> 
> Ghosts can't do anything but point and relying on them to point at anything useful is usually a horrible idea
> 
> It's not the only item with ghost interaction anyway

It resulted in an antag and all their thralls being called out through information they had as a ghost, we ruled that because its the intended use of the item its not actionable and that we shpuld remove the item instead

nevimer:
This is a PR from your master branch and thus cannot be used, sorry. you will need to make a PR from our master branch onto a new branch of your own.

CloverIsLucky:
👻  ⬇️ 
AB[]CDEFGH
IJKLMNOP
Q[R]STUVWX
[Y]Z_YES_NO

PR #1054, opened 2024-01-22
Better airlock sounds QOL

pixelkitty286:
## About The Pull Request
Ports Citrp's airlock sounds and makes all airlocks have better sounds than they used to.

## Why It's Good For The Game

Every door sound is the same boring door sound and all sound the exactly same even though each door is different. In other words the door sounds I've got sound more aesthetically pleasing compared to "*SHHH*" "*SHHH CLUNK*".

## Proof Of Testing

https://github.com/Bubberstation/Bubberstation/assets/78038207/246abdf4-35c3-43de-a3a6-074a67865a44

https://github.com/Bubberstation/Bubberstation/assets/78038207/5a40e67c-cada-44f2-882f-474a4d60de0f

https://github.com/Bubberstation/Bubberstation/assets/78038207/a1d9e0df-f29c-48de-bbff-89c9896b7a66

https://github.com/Bubberstation/Bubberstation/assets/78038207/d8d13c8d-14c3-432f-b33c-a07b98c747e0

## Changelog

:cl:
add: Door sounds from Citrp
qol: Better door sounds
sound: added door sound files in modular sounds folder
code: Disabled skyrat's door sound over rides in airlock.dm
/:cl:

PR #1101, opened 2024-02-02
Revert "Better airlock sounds QOL"

Swiftfeather:
Reverts Bubberstation/Bubberstation#1054

These are some of the most grating sounds I've ever heard added to the game. If it was something we only heard a few times for round it'd be fine, but it's EVERY AIRLOCK, one of the devices we interact with THE MOST as players. Some sounds are extremely oud, some are quiet, some are fine and some are grating like nails on a chalkboard. There's also a lot of ones that are obscenely long for how simple and quick of an action airlock opening and closing is. 

It's an extremely inconsistent PR  and should have been TM'd longer before changing stuff players touch so often. None of the feedback from the one TM was allowed to be acted on. 

___________________________________________________________________________________________

:cl:
del: Door sounds from Citrp
qol: better door sounds
sound: changed sound back to better original airlock sound
code: re enables skyrat's door sound override
/:cl:


pixelkitty286:
hey uh I'm terribly sorry about how the current state they were merged I was actually going fix the sounds on some of the doors / remove some since I do agree some are a bit... bad and I was going to be putting better ones in their place that I was going to be pulling from some games

Majkl-J:
I agree with, this should not have gotten in in this state.

PR #1102, opened 2024-02-02
[PORT] [TM Only] ARC Mining - Granite Fixes it Edition

lessthnthree:
## About The Pull Request

Port of https://github.com/NovaSector/NovaSector/pull/670 for ARCmining


StrangeWeirdKitten:
I'll keep this PR cooking in the background while we look at the effects of arc mining in their stock TG forms.

Overall I do like the proximity ore spawning. It makes a great indicator of vent locations. I realize icebox has its own set of issues that are being worked on upstream. But generally once you tap a few vents, I've found you are drowning in points and ores so it's not entirely needed and it allows more interaction with cargo to set up the proper equipment and feast off the rewards. 

This would be more on the basis for people to be allowed to bypass the system all together. But yeah, if you don't check what you are about to activate, you might find a colossus spawning on top of you.

lessthnthree:
Okay.

PR #1103, opened 2024-02-02
Fixes double deathgasp and most simplemobs not deathgasping

NithaIsTired:
## About The Pull Request

Attempt to fix the double deathgasping when mobs with soundeffects die, might cause something unexpected but I didn't find anything out of the usual while testing.

Also brings back simplemobs having a deathgasp so you can actually tell when they are dead.

## Why It's Good For The Game
bugfix good

## Proof Of Testing
dreamseeker_1706907502

## Changelog

:cl:
fix: mobs don't deathgasp twice and simplemobs also deathgasp again.
/:cl:


Swiftfeather:
Seeing as this is severely out of date and should be taken upstream, I'm closing the PR here.

PR #1108, opened 2024-02-04
force atmos hack

nevimer:
Forces the atmos MC to run in two different ways. Very bad, but the MC doesn't work and this is a byond native fallback to keep rounds at least going.


Majkl-J (approved):
image

Cyprex:
The MC doesn't work? How?

PR #1110, opened 2024-02-04
[DNM] Revert "Makes roundending significantly harder by making the brain (Almost) indestructible"

nevimer:
Reverts Bubberstation/Bubberstation#751

Lots of bugs. So many bugs.

### DIFF
--- a/modular_zubbers/code/modules/surgery/organs/internal/lungs/_brain.dm
+++ b/modular_zubbers/code/modules/surgery/organs/internal/lungs/_brain.dm
@@ -1,2 +1,0 @@
-/obj/item/organ/internal/brain
-	resistance_flags = INDESTRUCTIBLE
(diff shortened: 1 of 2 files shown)


Arturlang:
Why was the brain indestructibility even defined in modular_zubbers/code/modules/surgery/organs/internal/lungs/_brain.dm
...in lungs

Majkl-J:
> Why was the brain indestructibility even defined in modular_zubbers/code/modules/surgery/organs/internal/lungs/_brain.dm ...in lungs

It being late and vscode compacting folders. Anyhow if this over the surgery issues when the brain is removed, these propped up way before this was merged, most likely as a result of the organ refactor

StrangeWeirdKitten:
Jank, and the rare time the brain burns. Admins can just hit the respawn.

PR #1116, opened 2024-02-06
Stop spamming logs when engineers lovingly feed the supermatter pipes by the hundreds.

Swiftfeather:
## About The Pull Request

Title 

Also the method I went about this adds a flag to the game, which means in the future if people find another way to spew even more items into the supermatter that spam the log, we can just add the flag to those items.

## Why It's Good For The Game

I got bwoinked this one time because I was spamming the admin log and I told them I'd look for a solution.

This might not be _the_ solution but it is **a** solution. 

## Proof Of Testing

this is what my IC chat tab looked like
image

This is what my logs tab looked like

image

It still logs pipes for initial startup so we good there.
I also verified that other items still set off the log function as expected.

## Changelog

:cl:
admin: Will no longer receive supermatter ate a pipe spam in the log.
/:cl:


code/datums/components/supermatter_crystal.dm
```diff
@@ -314,6 +314,11 @@
...
 	else if(consumed_object.flags_1 & SUPERMATTER_IGNORES_1)
 		return
+	//BUBBER edit start | Addition to ignore items engineers like to feed the SM :3
+	else if(consumed_object.flags_1 & SUPERMATTER_NOLOGS_1)
+		qdel(consumed_object)
+		return
+	//BUBBER edit end
 	else if(isobj(consumed_object))
 		if(!iseffect(consumed_object))
 			var/suspicion = ""
```
Majkl-J, line 324:
```suggestion
	
	else if(consumed_object.flags_1 & SUPERMATTER_NOLOGS_1)

	else if(isobj(consumed_object))
		if(!iseffect(consumed_object) && !(consumed_object.flags_1 & SUPERMATTER_NOLOGS_1)) //BUBBER EDIT | NOLOGS flag Addition to ignore items engineers like to feed the SM :3
			var/suspicion = ""
```

code/__DEFINES/_flags.dm
```diff
@@ -55,6 +55,10 @@ GLOBAL_LIST_INIT(bitflags, list(1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 204
 /// Yes I know this is a stupid flag, no you can't take him from me
 #define DECAL_INIT_UPDATE_EXPERIENCED_1 (1<<17)
 
+/* BUBBER FLAGS */
+//Does item send log notices to admin log when consumed by supermatter?
+#define SUPERMATTER_NOLOGS_1 (1<<18)
+
```
Majkl-J, line 61: Always make some room for any future defines

```suggestion
//BUBBER EDIT START - Flags
//Does item send log notices to admin log when consumed by supermatter?
#define SUPERMATTER_NOLOGS_1 (1<<20)

//BUBBER EDIT END
```

code/game/machinery/pipe/construction.dm
```diff
@@ -18,6 +18,9 @@ Buildable meters
 	icon_state = "simple"
 	icon_state_preview = "manifold4w"
 	inhand_icon_state = "buildpipe"
+	//BUBBER ADDITION START
+	flags_1 = SUPERMATTER_NOLOGS_1
+	//BUBBER ADDITION END
```
Majkl-J, line 23: This one should be made modular with

```
/obj/item/pipe/Initialize(mapload)
	flags_1 |= SUPERMATTER_NOLOGS_1
	. = ..()
```

Majkl-J (changes requested):

lessthnthree:
Was this rejected upstream? It seems like a no brainer. (It spams admin logs too)

ReturnToZender:
So full of pipes yum

Swiftfeather:
> Was this rejected upstream? It seems like a no brainer.

Screenshot_20240206_200835_Discord

code/game/machinery/pipe/construction.dm
Majkl-J: `modular_zubbers/code/game/machinery/pipe/construction.dm`

Arturlang:
Better way to fix this would be to keep a counter of recent dustings, and only show the number of items sacrificed with the name of each unique name and maybe even add a click option for admins to make the SM stop adminlogging

Swiftfeather:
After putting in requested changes it no longer seems to work. hm

PR #1121, opened 2024-02-09
Branding

nevimer:
## About The Pull Request

Adds branding. Now in the HOS locker.

## Proof Of Testing

https://github.com/Bubberstation/Bubberstation/assets/77420409/34595b0f-46a7-41b3-b4e0-7dc897b41931

## Changelog

:cl:
add: HOS gets a sizzling branding knife, for the worst criminals NT has to offer.
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/branding/brand.dm
+++ b/modular_zubbers/code/modules/branding/brand.dm
@@ -1,0 +1,66 @@
+/obj/item/knife/brand
+/* 	icon = 'icons/obj/stack_objects.dmi'
+	icon_state = "rods-1" */
+	name = "branding knife"
+	icon = 'modular_skyrat/modules/modular_ert/icons/pizza/hotknife.dmi'
+	icon_state = "hotknife"
+	inhand_icon_state = "hotknife"
+	desc = "Once known as Lightbringer, this sword has been demoted to a simple pizza cutting knife... It may still have its fire attack powers."
+	righthand_file = 'modular_skyrat/modules/modular_ert/icons/pizza/righthand.dmi'
+	lefthand_file = 'modular_skyrat/modules/modular_ert/icons/pizza/lefthand.dmi'
+	var/list/static/possible_crimes = list()
+	var/current_branding = ""
+
+/obj/item/knife/brand/attack(mob/living/victim, mob/living/attacker, params)
+	var/list/branding_with = list(typesof(/datum/status_effect/branded))
+	branding_with -= /datum/status_effect/branded
+/* 	var/list/preset_brandable_crimes = list()
+	for(var/i in branding_with)
+		preset_brandable_crimes |= i.initial(id) */
+
+
+	var/current_branding = tgui_input_list(attacker, "What crime are you branding them with?", "Branding", branding_with) // todo, make names show
+
+	if(!attacker.combat_mode && do_after(attacker, 6 SECONDS, victim))
+		victim.apply_status_effect(text2path("[branding_with]"))
+		victim.adjustFireLoss(60)
+	return ..()
+
+
+/datum/status_effect/branded
+	id = "generic_branding"
+	duration = -1
+	alert_type = null
+
+/datum/status_effect/branded/get_examine_text()
+	return span_warning("[owner.p_They()] bear branding makes for the crime of [id]!")
+
+/datum/status_effect/branded/murder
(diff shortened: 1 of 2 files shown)


modular_zubbers/code/modules/branding/brand.dm
```diff
@@ -0,0 +1,73 @@
...
+	if(!istype(victim, /mob/living/carbon))
+		return
+	var/datum/status_effect/branded/current_branding = tgui_input_list(attacker, "What crime are you branding them with?", "Branding", branding_with) // todo, make names show
+
+	if(!attacker.combat_mode && do_after(attacker, 6 SECONDS, victim))
+		victim.apply_status_effect(current_branding)
+		to_chat(attacker, span_userdanger("You brand [victim] with the mark of [initial(current_branding.id)]"))
+
+		var/obj/item/bodypart/branded_limb = victim.get_bodypart(deprecise_zone(BODY_ZONE_CHEST))
+		branded_limb.force_wound_upwards(/datum/wound/burn/flesh/critical, wound_source = "Branding")
```
nikothedude, line 28:
```
victim.cause_wound_of_type_and_severity(WOUND_BURN, branded_limb, WOUND_SEVERITY_CRITICAL))
```

same effect except it supports synthetic wounds

modular_zubbers/code/modules/branding/brand.dm
```diff
@@ -0,0 +1,73 @@
...
+/* 	var/list/preset_brandable_crimes = list()
+	for(var/i in branding_with)
+		preset_brandable_crimes |= i.initial(id) */
+	if(!istype(victim, /mob/living/carbon))
+		return
+	var/datum/status_effect/branded/current_branding = tgui_input_list(attacker, "What crime are you branding them with?", "Branding", branding_with) // todo, make names show
+
+	if(!attacker.combat_mode && do_after(attacker, 6 SECONDS, victim))
+		victim.apply_status_effect(current_branding)
+		to_chat(attacker, span_userdanger("You brand [victim] with the mark of [initial(current_branding.id)]"))
```
nikothedude, line 25: missing a period

modular_zubbers/code/modules/branding/brand.dm
nikothedude: you should use this anyway, this is the defacto "give a limb a wound" proc

source: i refactored wounds and added t his proc

nikothedude:
also, what happened to the feature freeze? isnt this a feature?

nikothedude:
theres also no justification for why this is good for the game

YakumoChen:
> also, what happened to the feature freeze? isnt this a feature?

We aren't in a freeze

Kepteyn:
While not _inherently_ terrible, it seems to be a thing that exists to make antag's lives harder for the sake of making it harder. Hard to try to escape from perma when anyone who sees you can see that you're a prisoner, even if you changed your clothes and got a fake ID. If it can be removed - which, being as it is a wound, it may - that isn't quite as bad, but its still another thing that just exists for the sake of making the lives of antags harder.

My biggest concern would be lore-wise: _**WHY**_ does the HOS and security have access to a cattlebrand? Is it universal procedure amongst the GalFed to brand prisoners for their crimes? That seems a particularly barbaric practice if so, and I'd hope some justification for why that's the case would exist. Does GalFed not care how corporations handle security affairs in the frontier? I would expect security to be far less restricted in its use of violence if so. As it stands, the brand just seems like a random, out-of-place tool with little justification for _**why**_ it exists.

Being able to permanently brand someone for live as a criminal because 'he disrespected me' is pretty based though, gotta give that much.

Cyprex:
This might not make you uncomfortable, but it makes me very uncomfortable. It's not a good idea to give sec ideas to torture prisoners(who are generally mechanically largely unable to fight back). This will absolutely end up used in a way that will make people uncomfortable. It's a thing that is largely there for RP reasons, it's excessively cruel and so on and so forth. Tracking implants are a better option mechanically- this is just cruel. To use this item is to consciously decide to torture another player's character. Some might be fine with this. Others not so much, and by existing this item encourages this behavior and essentially communicates to security players that this is a fine thing to do.

Cyprex:
A stamp or similiar non-torturous mark would be fine, feel free to reopen this with that

PR #1128, opened 2024-02-10
Ports BoxStation (NSSJourney) from FluffySTG

shayoki:
## About The Pull Request
A nostalgic return of Boxstation with updates.

https://www.youtube.com/watch?v=guBsPkUkPQg
box3

Derived from FluffySTG, includes minimal Bubber changes.

## Why It's Good For The Game

Gives an alternative option for a Box-type station without the presence of ice.

## Proof Of Testing

Compiles. Runs. Full round tested.
Currently gathering feedback and maintaining.

## Changelog

:cl: Shayoki
add: Adds Boxstation
/:cl:


ReturnToZender:
what da heck Jaeger Calisavry doin here

BlackMajor:
AI Satellite's cables are wired to the station, meaning at shift start, the satellite SMES unit will be sending power to the station and not charge any of the satellite APC's.

YakumoChen:
Love box but Ice has kinda outclassed it at this point, not really keen on having Box on the rotation twice.

iskawhiskers:
Bitrunner does not have access to the Mining Bay in order to pass through to the Bitrunning Den. You may want to slightly rework that particular room in order to separate the two, or let bitrunners access that room.

Glyphee:
2 small things, there aren't any magboots in engi (that I could find), and there are 2 windoors stacked on top of each other at the engineering desk, one of which is called "Head of Personnel's Office" or something to that sort.

Glyphee:
Oh this might be a bigger issue, the grid seems to be hotwired by default to the SM?

Glyphee:
Okay so I think i fixed the hotwiring thing, the cable here needs to be cut, 
cut
, and cables need to be put here, 
cable

Glyphee:
pipe
this too

BurgerLUA:
I think there are too many loot piles in maint. These should be used sparingly tbh and shouldn't be the majority of maint spawns. This just might be a random spawner issue (if it is, tell me) but if it's not you should instead use the maint closet spawn that most maps use.

Yanziko:
image
SMES is not actually connected to maintenance! Because of this, station is not powered.

BurgerLUA:
The "Necromancy Wizard Cosplay Kit" that spawns in the ashwalker room is actually real wizard robes and has very good armor stats on them.

ReturnToZender:
@shayoki The disposals chute in cargo seems to randomly route back into itself, all I hear anytime I play NTC is the loud alarm through the wall as the same three items repeatedly circle 😭

BurgerLUA:
image

The ancient toolbox here deals a lot of damage (7 hits to takedown) and always contains insulated gloves.

BurgerLUA:
There are not enough assistant spawns on the map, causing people to spawn on the interlink.

PR #1132, opened 2024-02-11
Small borg stuff, adds picking up and removes buckling. (holds you like the door bell cat)

pixelkitty286:
## About The Pull Request

Adds the the ability to be scoop up small cyborgs and hold them as a item.

Removes the ability to buckle on / ride small cyborgs.

## Why It's Good For The Game

I've had people try to interact with my as a smaller cyborg on a normal basis and hold them like a catto only to find out they can not be picked up and held.

These changes are only specific to small cyborg chassis full sized cyborgs can not mimic this "weight change" with the shape shifting module this also is the same case with small cyborgs the other way around.

When a small cyborg model is reset they lose the ability to be picked up.

Small cyborgs can not be bagged by normal means and be only put in a bag of holding but if your doing something like that I really feel that would be power gaming at that point.

The removal of riding small cyborgs was added because it always confused me on why you were able to ride a cyborg that evidently shouldn't be able to support the weight of a person.

## Proof Of Testing

dreamseeker_etEDYutoL5

dreamseeker_fN1uDOYmEQ

dreamseeker_F0iq3BNWYZ

dreamseeker_No896zIirT

## Changelog

:cl:
add: Small cyborgs being holdable 
del: Small cyborgs being rideable
code: a few modular changes, and robot_model.dm
/:cl:


nikothedude:
absolutely balance changes bro

buckling is ESSENTIAL to the gameplay as a paramedic borg, removing that will just piss off small borg players and make others not pick the submodule

theres nothing saying the borg cant pick up a human - have you seen a small kobold ffireman carrying a oversized 150% person? they can absolutely hold their weight, and "it doesnt make sense" doesnt work, especially with the outstanding balance concenrs

StrangeWeirdKitten:
Picking up your friendly borg has sweeping implications and there's a reason why fundimentally one player can not pick up another player as an item unless it's something very specific like an AI inside a much more restrictive card / modsuit. While I think the ideas are overall sound, and it makes a little sense. I also don't want sweeping mechanical diffrences between what should just be the visual flavour of your silicon.

nikothedude:
> This only applies to the smaller borg chassis because do you **really** think a cyborg of smaller type could support a high weight class of a person? Because personally I don't think it would. Now I personally wanted to make a new trait that applied this to specific models however, the call I was in wile making this disagreed on making it such and I'm more than willing to do this still.

yes actually, if a small kobold can pick someone else up then of course

the argument "it just makes sense" fails to actually argue as to why a change is good for the game. does seeing a small borg haul someone actively damage your enjoyment of the game? i again ask you to consider the consequences of removing this on those small borg players

pixelkitty286:
> > This only applies to the smaller borg chassis because do you **really** think a cyborg of smaller type could support a high weight class of a person? Because personally I don't think it would. Now I personally wanted to make a new trait that applied this to specific models however, the call I was in wile making this disagreed on making it such and I'm more than willing to do this still.
> 
> yes actually, if a small kobold can pick someone else up then of course
> 
> the argument "it just makes sense" fails to actually argue as to why a change is good for the game. does seeing a small borg haul someone actively damage your enjoyment of the game? i again ask you to consider the consequences of removing this on those small borg players

Again this was not made out of spite it was made out of agreement from when I implemented cat borgs and everyone lost their shit just because they are a small cyborg I said I would rebalance small cyborgs if waterpig didn't get around to it speaking of which I'm going to be looking into possibly a few more things to rebalance them.

pixelkitty286:
wa wa reworking this to be something different

PR #1140, opened 2024-02-12
Bloodsucker additional levelling system.

destrucktoid:
## About The Pull Request

This system adds a new way for bloodsuckers to level in addition to their ordinary one, thus rewarding bloodsuckers for actually engaging in combat and doing antagonistic behaviours rather than playing extremely passive.

Bloodsuckers that drink blood from creatures with a mind (has been or is controlled by a player) get that blood tracked and added to a new system, this blood can be used to purchase level ups.

The blood consumed is taken out of your actual blood volume, meaning that you still need to spend your blood for this level up. Furthermore, whilst the blood per level starts at 30% of your max blood, it increases by 5% per level you gain with this feature, capping out at 90%. 

Before you ask, no you can't farm vassals for unlimited levels. I've set the points you get towards a level to be quartered from drinking from a vassal and if you remember, you cannot (currently) drink past your blood cap. Thus if i drink myself from 0 blood to 1000 blood on vassals alone, that's not enough for a single level. Furthermore, the usual blood drinking penalties apply (dead, non-human, mindless)

No, you can't drink a bunch of blood, buy levels with the points (because it's theoretically cheaper to buy them whilst you are a lower level bloodsucker) and then spend them. The system is designed to auto spend normal levels first before later offering you the ability to buy a level in your coffin.

## Why It's Good For The Game

Bloodsucker right now is an antagonist that is played extremely passively due to all clans except tremere and ventrue having no way to scale that involves combat. Thus you are better to just sit on your laurels and do nothing, waiting for levels to come to you rather than actually doing antagonist things

## Proof Of Testing

Tested with @Arturlang, the code works as intended.
https://www.youtube.com/watch?v=RWxp5fozlvc

## Changelog

:cl:
add: Added a new way for bloodsuckers to earn levels
removed: Blood sucker passive levels no longer cost blood to attain.
/:cl:


theselfish:
Accurate to VTM as far as I can tell.

Everyone ignores the "certain bloodtypes to level up disciplines" rule anyway.

Arturlang:
> Accurate to VTM as far as I can tell.
> 
> Everyone ignores the "certain bloodtypes to level up disciplines" rule anyway.

You know that's actually a REALLY INTERESTING IDEA

Arturlang:
Honestly it may be a good idea to replace the current leveling system with this, but different numbers

modular_zubbers/code/modules/antagonists/bloodsucker/bloodsuckers/procs.dm
```diff
@@ -165,6 +165,26 @@
...
+	var/level_cost = get_level_cost()
+	if(blood_level_gain >= level_cost && bloodsucker_blood_volume >= level_cost) // Checks if we have drunk enough blood from the living to allow us to gain a level up as well as checking if we have enough blood to actually use on the level up
+		switch(tgui_alert(owner.current, "You have drunk enough blood from the living to thicken your blood, this will cost you [level_cost] blood and give you another level",  "Thicken your blood?.", list("Yes", "No"))) //asks user if they want to spend their blood on a level
+			if("Yes")
+				bloodsucker_level_unspent++ // gives level
+				blood_level_gain -= level_cost // Subtracts the cost from the pool of drunk blood
+				AdjustBloodVolume(-level_cost) // Subtracts the cost from the bloodsucker's actual blood
+				blood_level_gain_amount += 1 // Increments the variable that makes future levels more expensive
+			if("No")
+				return
```
JohnFulpWillard, line 180: This can just be an early return if it isn't "Yes" because switch can be x'ed out which would return neither yes or no, and there's no reason to specifically call return anyway.

modular_zubbers/code/modules/antagonists/bloodsucker/structures/coffin.dm
```diff
@@ -221,10 +221,14 @@
 		if(!bloodsuckerdatum.my_clan)
 			user.balloon_alert("enter a clan!")
 			to_chat(user, span_notice("You must enter a Clan to rank up. Do it in the antag menu, which you can see by pressing the action button in the top left."))
-		else
+		if(bloodsuckerdatum.bloodsucker_level_unspent > 0)
```
JohnFulpWillard, line 224: :P

modular_zubbers/code/modules/antagonists/bloodsucker/bloodsuckers/procs.dm
destrucktoid: Should be fixed.

modular_zubbers/code/modules/antagonists/bloodsucker/structures/coffin.dm
destrucktoid: Code completely refactored. If you want to give it a look over, feel free to!

Swiftfeather (approved):
Haven't heard complaints after being tmed for a while.

PR #1151, opened 2024-02-15
Gives robos a little pamphlet that should guide them in treating synth wounds

nikothedude:
## About The Pull Request

Title.

UNTESTED BECAUSE THE FUCKING GAME WONT COMPILE 

## Why It's Good For The Game

Watching robos treat synth wounds is PAINFUL. They NEED to know the tools they have at their disposal, because from my observations its always fumbling clumsily.

## Proof Of Testing

I LITERALLY CANNOT MY GAME DOES NOT COMPILE

image

## Changelog

:cl:
add: Robos now have a pamphlet in their bag that tells them tips and tricks on how treat synth wounds
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/jobs/job_types/roboticist.dm
+++ b/modular_zubbers/code/modules/jobs/job_types/roboticist.dm
@@ -1,0 +1,13 @@
+/datum/outfit/job/roboticist/New()
+	. = ..()
+
+	backpack_contents += /obj/item/paper/pamphlet/roboticist_reminder
+
+/obj/item/paper/pamphlet/roboticist_reminder
+	name = "pamphlet - 'Reminder of roboticist duties'"
+	default_raw_text = "Don't forget that your duties include treating the synthetic lifeforms aboard the station! Heres some tips and tricks to get you started... \n\
+	You can buy <b>health analyzers</b> from your vendor, as well as <b>burn wound treatment chems</b>! Generally, you'll be using the coolant bottles. \n\
+	Make sure to <b>use the health analyzers in hand</b> to activate <b>wound mode</b> if you're ever confused about a wound! <b>Blunt wounds</b> even show \
+	the current treatment step!\n\
+	Your <b>departmental order console</b> includes things like synthetic medicine crates!\n\
+	<b>Epipens</b> are an effective treatment method for synthetic slash/pierce wounds!"
--- a/tgstation.dme
+++ b/tgstation.dme
@@ -8380,6 +8380,7 @@
 #include "modular_zubbers\code\modules\jobs\job_types\clown.dm"
 #include "modular_zubbers\code\modules\jobs\job_types\curator.dm"
 #include "modular_zubbers\code\modules\jobs\job_types\nanotrasen_consultant.dm"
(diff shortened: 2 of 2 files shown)


nevimer:
is BYOND INSTALLED?

YakumoChen:
We should just have a dedicated wiki resource for this or something.
If we need to add a literal item telling them how to do mechanics, something is wrong.

It's more likely that wounds just are badly coded if the treatments are too obscure

nikothedude:
> We should just have a dedicated wiki resource for this or something. Or give them roundstart health analyzers to do this for them (probably the latter instead of a note)
> 
> If the health analyzer isn't helpful enough that needs to be changed asap
> 
> It's more likely that wounds just are not well designed if the treatments are also too obscure

There is, in fact, a dedicated wiki resource on the nova wiki, in the guide to synth surgery. Noone uses it

They also have rojndstart health analyzers in their vendor

Using the wound mode in the analyzers tells you more about the treatments, its just nobody uses it, so I want people to use it

YakumoChen:
> > We should just have a dedicated wiki resource for this or something. Or give them roundstart health analyzers to do this for them (probably the latter instead of a note)
> > If the health analyzer isn't helpful enough that needs to be changed asap
> > It's more likely that wounds just are not well designed if the treatments are also too obscure
> 
> There is, in fact, a dedicated wiki resource on the nova wiki, in the guide to synth surgery. Noone uses it
> 
> They also have rojndstart health analyzers in their vendor
> 
> Using the wound mode in the analyzers tells you more about the treatments, its just nobody uses it, so I want people to use it

Give them a roundstart wound analyzer then

Majkl-J:
> Give them a roundstart wound analyzer then In their bag

They get several in their equipment vendor, but this could be added together with this

> Maybe also make a wikilink book item and automap it in robotics if you really want to make it easy

These are broken thanks to config settings and we do not have the manpower to maintain that as well

> I just think a paper note is a really lazy solution, nothing against you

In-game solutions are always better than having to alt tab, letalone having to scour recent PRs because synths do take an alternative route from nova on here, with their surgeries to be revamped

YakumoChen:
> > Give them a roundstart wound analyzer then In their bag
> 
> They get several in their equipment vendor, but this could be added together with this

I mean the one for only wounds, like miners get in their box, or in medkits

nikothedude:
> > > Give them a roundstart wound analyzer then In their bag
> > 
> > 
> > They get several in their equipment vendor, but this could be added together with this
> 
> I mean the one for only wounds, like miners get in their box, or in medkits

Yeah thats also in the vendor

modular_zubbers/code/modules/jobs/job_types/roboticist.dm
```diff
@@ -0,0 +1,14 @@
+/datum/outfit/job/roboticist/New()
+	. = ..()
+
+	LAZYINITLIST(backpack_contents)
+	backpack_contents[/obj/item/paper/pamphlet/roboticist_reminder] = 1
```
Majkl-J, line 5: Add wound analyzers to them too

Majkl-J (changes requested):

Majkl-J (approved):

PR #1167, opened 2024-02-18
Messages admins when someone is type-baited

ReturnToZender:
## About The Pull Request

Normally, when someone's typing is interrupted, it logs that to speech indicators. This PR makes it so that it messages admins as well.

Part of the RP update!

## Why It's Good For The Game

If someone's type-baited, admins need to know sooner rather than later. 

## Proof Of Testing
image

## Changelog

:cl: ReturnToZender
admin: Interrupted typing now messages admins about it on top of logging it
/:cl:


lessthnthree:
You should adapt it to include the source of who/what made them force_say, otherwise it's not very useful on its own.

ReturnToZender:
> You should add an arg to include the source of who/what made them force_say, otherwise it's not very useful on its own.

agree, let me see what I can do

projectkepler-ru:
Good idea, a false positive is still good as it alert admin that something is happening and they should start paying attention to the round if they were observing and not already looking around.

nikothedude:
id suggest also adding how much time the victim was typing for before the interruption, since most proper type baiting happens in the form of asking someone a random question and attacking as soon as you see the indicator

seeing a typing duration of 20 seconds rather than 0.4 could be a good clue

ReturnToZender:
Some really good suggestions here, but ones that will need more time in the oven from me, so I'm drafting this PR as I figure out how to make these updates

ReturnToZender:
Closed due to stale, I'll continue work on this later

PR #1182, opened 2024-02-20
[MIRROR] Synths update: The ID reckoning (#26332)

Majkl-J:
See https://github.com/Skyrat-SS13/Skyrat-tg/pull/26332

## About The Pull Request
Dejanks the id slot in synth computers by making the computer read directly from your id slot. Cleans up synth code from all the now useless code.

Introduces COMSIG_HUMAN_UNEQUIPPED_ITEM due to some signal related jank which would've made this unable to work otherwise

## How This Contributes To The Skyrat Roleplay Experience

Synths can no longer do silly ID hiding, also the ID slot now makes their computers much easier to use.

## Proof of Testing

image


IgiariValkyr:
Explain it better. I'm taking this three different ways.

- You're removing the ID slot from synth computers?
- You're making it so synth computers can read the ID from the actual ID slot on the person?
- IDs finally have access when inserted to a synth computer?

PR #1190, opened 2024-02-22
Added a Map Whitelist for which the blob event can spawn.

BurgerLUA:
## About The Pull Request

Added a Map Whitelist for which the blob event can spawn. Current maps that are allowed are: Delta Station, MetaStation, Moon Station, and Kilo Station, Birdshot.

## Why It's Good For The Game

There are a lot of maps in rotation that are bad for blob. Many of these are just because they have z-levels and blob is really good on z-levels but there are so maps that don't have z-levels that also have serious problems. Here are the maps that aren't whitelisted and their reasons why

### Void Raptor
While not a multi-z map, the maint section is absolutely unbearable to navigate with the amount of rwalls used in maintenance, the amount of doors used, and a TON of dead ends in loctions. Most maps have a circular maintenance section, but on Void Raptor, maintenance is a legitimate afterthought. 

### Blueshift
Multi-z map, and this map has an insane amount of hidden and restricted maintenance space that not many people know about. Blob almost always spawns in these locations and it can be a good while until someone actually finds it, even after being announced.

### Biodome
Multi-z map.

### Icebox
Multi-z map, with a lot of areas that just have one entry way. Blobs almost always spawn in these areas to destroy stairs/ladders and it is way too powerful especially with an icy atmosphere.

### Lima Station
Multi-z map.

### Northstar
Multi-z map.

### Tramstation
Multi-z map.

## Why not Moonstation?

Moonstation is designed to prevent blobs from spawning on the lower levels, thus doesn't have the multi-z problem.

## Proof Of Testing

Untested.

## Changelog

:cl: BurgerBB
del: Added a Map Whitelist for which the blob event can spawn. Current maps that are allowed are: Delta Station, MetaStation, Moon Station, and Kilo Station.
/:cl:


projectkepler-ru:
@MrMelbert how exactly did TG handle multiz level blob?

lessthnthree:
Given the map rotation it's realistically it's just MetaStation and Moon Station that would allow blobs. This seems like an unnecessary removal for what would be better handled at the server rules level regarding where a blob is allowed to spawn.

Formerxboxgamer:
> @MrMelbert how exactly did TG handle multiz level blob?

I’m not Melbert, but tg has no special rules or code for blobs in multi-Z maps. Crew just plays around it.

Majkl-J:
Feels like this isn't really needed. Sometimes crew simply fumble and fail to defeat the blob, no matter the map. Since blob can't chew above/below, it's the crew's fault if floor breaches cause it to be harder to fight

PR #1202, opened 2024-02-24
Buffs the amount of lethal lasers energy weapons can put out.

Tupinambis:
## About The Pull Request

Lets basically all laser weapons put out roughly 1.3 more shots.

## Why It's Good For The Game

You can't reload energy weapons, and energy weapons struggle with the increase in health coupled with the frequency at which EMPs are used. Ballistics are very strongly preferred for this reason. This should help lasers seem a bit more appealing over all and give sec a bit more power against high threat antagonists, which is something they currently feel a bit ill equipped to deal with.

## Proof Of Testing

It compiles.

## Changelog

:cl: Tupina
balance: Energy weapons can now fire roughly 1.3x more lasers per charge across the board.
/:cl:


YakumoChen:
@Tupinambis this will have some knock-on effects for Blob, which is a mild concern

projectkepler-ru:
I like this

Tupinambis:
> @Tupinambis this will have some knock-on effects for Blob, which is a mild concern

This will have knock-on effects for all simplemobs, but I don't think it's going to be particularly severe or make enough of a difference to drastically change the odds of any of them winning or losing. If the crew is able to coordinate getting tons of rechargers and lasers to the blob they're probably winning anyway.

Majkl-J (approved):
Judging from maintainer chats the unanimous consensus is that this is a better alternative to lowering HP. I'll TM it and see how it goes

Majkl-J (approved):
Been testmerged for 2 weeks with, surprisingly, 0 complaints or suggestions. I'll be merging this today just gotta wait on the runners to actually pick the CI job

PR #1214, opened 2024-02-27
Map Voting now occurs when the shuttle is called, instead of when it leaves.

BurgerLUA:
## About The Pull Request

Map Voting now occurs when the shuttle is called, instead of when it leaves.

It is recommended for server owners to increase the default voting time from 1 minute to 3 minutes.

## Why It's Good For The Game

It is usually chaos when the emergency shuttle leaves because that is usually when antagonists/crew/whatever start chucklefucking, thus it is more difficult to get a vote in when this is happening. A better time I think is when the shuttle is called.

Yes, people can chucklefuck by calling the shuttle early for no reason early in the round, however that is against the rules and admins can just rock the vote if they feel that people are dicking with the system.

## Proof Of Testing

image

## Changelog

:cl: BurgerBB
qol: Map Voting now occurs when the shuttle is called, instead of when it leaves.
/:cl:


destrucktoid:
I feel like moving map votes to happening before a round starts along with the storyteller vote would be better. Alternatively you could make it so that the map vote from shuttle being called only happens when the shuttle call becomes un-recallable

Swiftfeather:
I'd say make the vote start when the shuttle reaches the station. Don't need to worry about shuttle recalling at that point. 

Ideally have it run until round end if you can tie it to that. 

Otherwise might need to factor in early departure so 3.5-4 minutes oughta do it

ReturnToZender:
Starting when the shuttle reaches the station would be good, as it takes a good couple minutes for the shuttle to depart and reach Central

Elli-Skala:
Another issue is blob existing and having to vote mid blob fight

lessthnthree:
Have the vote start when the 3 minute departure timer starts. So if there's a blob, it won't start the vote until the dangerous environment check passes.

Swiftfeather (changes requested):
Have it call the vote when the shuttle arrives at station, and run for a few minutes.

BurgerLUA:
> Have it call the vote when the shuttle arrives at station.

Similiar issue would occur, though. People would get distracted of trying to get on the shuttle.

Swiftfeather:
> > Have it call the vote when the shuttle arrives at station.
> 
> Similiar issue would occur, though. People would get distracted of trying to get on the shuttle.

oh well. The shuttle can't be recalled once it's at station.

BurgerLUA:
> > > Have it call the vote when the shuttle arrives at station.
> > 
> > 
> > Similiar issue would occur, though. People would get distracted of trying to get on the shuttle.
> 
> oh well. The shuttle can't be recalled once it's at station.

True but I rarely see early shuttle calls happen.

AllCorruptingWyrm:
and furthermore this pr must be merged.

PR #1219, opened 2024-02-27
Limits Powersinks + Crab-17 purchases to once each per traitor per round.

BurgerLUA:
## About The Pull Request

Limits Powersinks and Crab-17 purchases to once each per traitor per round.

## Why It's Good For The Game

Powersink and Crab spam is pretty awful. A lot of people recently have been spamming these two objectives and then doing nothing of substance with them. Traitors should be using their TC to further the round and create meaningful conflict instead of what can basically be described as being in a constant state of soft grief.

## Proof Of Testing

If it compiles, it werks.

## Changelog

:cl: BurgerBB
del: Limits Powersinks + Crab-17 purchases to once each per traitor per round.
/:cl:


destrucktoid:
I don't think limiting them is the correct answer here. If people are abusing the shit out of them then that means they are too cost effective. I'd instead propose increasing their TC costs instead. 

People here struggle to find powersinks because unlike TG where you have people constantly roaming and looking for them so they will almost never go boom in a useful place since it'll be found. Here you simply don't. Ergo i think this is more of a balance issue in terms of them being underpriced for this server. As for the crab protocol, there is ways to find out who is crabbing such as the account lookup console and it IS balanced around that. But again, people here don't know this and don't do it. Ergo a price increase is justifiable.

TL:DR i think that you should nerf their price rather than directly limiting their quantity.

BurgerLUA:
I don't want to punish people for purchasing 1 Crab-17 or Powersink. The reason why they're used is because of the meme, which is why when 1 is bought, in 60% of cases 2 are purchased.

projectkepler-ru:
Do note that a lot of our map are just massive and a single powersink has the chance to tank the powernet for abit

Arturlang:
Okay but does it actually work if you've not tested it?

ReturnToZender:
The fake announcement should also be limited to once per traitor per round.

Majkl-J:
> The fake announcement should also be limited to once per traitor per round.

Does it really? All it can do is distract the crew, not actually cause them to lose money repeatedly having to run into the atmos gas  tanks

BurgerLUA:
> The fake announcement should also be limited to once per traitor per round.

I'm conflcited about this because removing fake announcements means people are limited to announcement-enforced gimmicks. I know it can get out of hand with fake blob announcements but once again, I'm very conflicted.

Majkl-J (approved):

PR #1221, opened 2024-02-28
Code refactoring for ninjaDrainAct to allow for the implementation of a maximum amount of total cyborg subversion per Space Ninja

aKromatopzia:
## About The Pull Request

This PR should make it so that when someone with a Spider Clan MOD attempts to hack a sillycone, it fails if they have already cleared that objective or if they did not have the objective entirely.

Tested. Functional. This does make it so that ninjas can't shove an unhacked borg if they've already hacked, but we can chalk that up do a do-nothing bug anyways?

## Why It's Good For The Game

Ninjas do actually tend to be a menace when silicons are about, especially if they keep getting hacked. This code refactoring is just an implementation for possibly capping a ninja's hack amounts.
Set var/datum/antagonist/ninja/borg_hacks_left to a negative number for infinite uses.

## Changelog

:cl:
code: ninjaDrainAct's borg section got edited to allow for a maximum amount of borg hacks per ninja to be set.
/:cl:


AdamElTablawy:
Tg closed a similar PR 

You may want to rename this

StrangeWeirdKitten:
I'd be down to have a 2-3 borg limit.

Swiftfeather:
>Changing the code to match the wiki
vs
>Changing the wiki to match the code

aKromatopzia:
The PR's name is mostly self-criticizing, but this does just refactor the code so that the possibility of a borg limit can be implemented.

aKromatopzia:
There. I've set the default to inf so functionally no change has happened. But, this allows admins to configure the maximum amount a ninja can subvert per spawn. However - and this is more of a policy thing - should ninjas just be forced to have a low amount at once but recharge over time? That way it prevents situations of like 5 ninjaborgs at once, but ninja can still be a continuous presence,

aKromatopzia:
oh that deletes the branch. Oops.

PR #1227, opened 2024-02-29
Ninja code update

aKromatopzia:
## About The Pull Request

The previous PR got deleted because I changed the branch name. Oops. Anyways, a variable was implemented to allow admins to configure the maximum amount of silicons a ninja can turn spider clan. Default is -1, which negative values are treated as infinite. Player-side, unless the variable is changed to any positive number nothing different is felt other than new chat prints which tell the ninja that they successfully hacked a borg. 

## Why It's Good For The Game

Personally I think that ninja's hacking capabilities are a bit too explosive with higher silicon populations. Especially when people start joining posibrains in response to a ninja. That coupled with their EMP makes subversions pretty much guaranteed. I'd make it so that these charges refill over time but that's a bit more work.

## Changelog

:cl:
qol:  Space Ninjas get a message which tells them they successfully hacked a silicon. If hack scarcity is enabled it also tells them their remaining amount. The borg also receives a similar message without information on remaining charges.
code: new variable which caps Space Ninja's ability to hack cyborgs. Default is infinite.
/:cl:


code/modules/antagonists/ninja/ninjaDrainAct.dm
```diff
@@ -296,8 +295,23 @@
 	modelselected["Medical"] = "/obj/item/robot_model/ninja/ninja_medical"
 	modelselected["Saboteur"] = "/obj/item/robot_model/ninja_saboteur"
 	//SKYRAT EDIT: ADDITION END
+
 	if(!do_after(ninja, 6 SECONDS, target = src))
 		return
+	//ZUBBER EDIT START - Puts a cap on ninja's borg hacking
+	//Set to negative numbers for infinite uses
+	var/datum/antagonist/ninja/borg_hacks_left = -1
```
nevimer, line 303: wait. you're making an uninitialized datum reference, and assigning it's object to be -1? that doesn't seem right. you shouldn't need to typecast anything here.

code/modules/antagonists/ninja/ninjaDrainAct.dm
```diff
@@ -296,8 +295,23 @@
...
+	if(borg_hacks_left == 0)
+		to_chat(src, span_danger("UPLOAD ABORTED. SUSPICIOUS ACTIVITY DETECTED."))
+		to_chat(ninja, span_danger("UPLOAD FAILURE. TOKEN DENIED."))
+		return
+	else if(borg_hacks_left > 0)
+		borg_hacks_left -= 1
```
nevimer, line 309:
```suggestion
		borg_hacks_left--
```

code/modules/antagonists/ninja/ninjaDrainAct.dm
```diff
@@ -296,8 +295,23 @@
...
+		to_chat(ninja, span_danger("UPLOAD COMPLETE. AVAILABLE TOKENS REMAINING: [borg_hacks_left]"))
+	else if(borg_hacks_left < 0)
```
nevimer, line 311:
```suggestion
	if(borg_hacks_left)
```

code/modules/antagonists/ninja/ninjaDrainAct.dm
```diff
@@ -296,8 +295,23 @@
...
 	if(!do_after(ninja, 6 SECONDS, target = src))
 		return
+	//ZUBBER EDIT START - Puts a cap on ninja's borg hacking
+	//Set to negative numbers for infinite uses
+	var/datum/antagonist/ninja/borg_hacks_left = -1
+	if(borg_hacks_left == 0)
+		to_chat(src, span_danger("UPLOAD ABORTED. SUSPICIOUS ACTIVITY DETECTED."))
+		to_chat(ninja, span_danger("UPLOAD FAILURE. TOKEN DENIED."))
+		return
+	else if(borg_hacks_left > 0)
```
nevimer, line 308:
```suggestion
	if(!borg_hacks_left)
```

code/modules/antagonists/ninja/ninjaDrainAct.dm
```diff
@@ -296,8 +295,23 @@
...
+
 	if(!do_after(ninja, 6 SECONDS, target = src))
 		return
+	//ZUBBER EDIT START - Puts a cap on ninja's borg hacking
+	//Set to negative numbers for infinite uses
+	var/datum/antagonist/ninja/borg_hacks_left = -1
+	if(borg_hacks_left == 0)
+		to_chat(src, span_danger("UPLOAD ABORTED. SUSPICIOUS ACTIVITY DETECTED."))
+		to_chat(ninja, span_danger("UPLOAD FAILURE. TOKEN DENIED."))
+		return
```
nevimer, line 307: this could be moved into a negative check

code/modules/antagonists/ninja/ninjaDrainAct.dm
```diff
@@ -296,8 +295,23 @@
...
+	else if(borg_hacks_left > 0)
+		borg_hacks_left -= 1
+		to_chat(ninja, span_danger("UPLOAD COMPLETE. AVAILABLE TOKENS REMAINING: [borg_hacks_left]"))
+	else if(borg_hacks_left < 0)
+		to_chat(ninja, span_danger("UPLOAD COMPLETE."))
+	//ZUBBER EDIT END - It works
```
nevimer, line 313:
```suggestion
	var/borg_hacks_left = -1
	switch(borg_hacks_left)
	  if(!borg_hacks_left)
		  to_chat(src, span_danger("UPLOAD ABORTED. SUSPICIOUS ACTIVITY DETECTED."))
		  to_chat(ninja, span_danger("UPLOAD FAILURE. TOKEN DENIED."))
		  return
	  if(borg_hacks_left)
		borg_hacks_left -= 1
		to_chat(ninja, span_danger("UPLOAD COMPLETE. AVAILABLE TOKENS REMAINING: [borg_hacks_left]"))
	//ZUBBER EDIT END - It works
```

nevimer (changes requested):

nevimer:
overall I don't think this works as intended

StrangeWeirdKitten:
> overall I don't think this works as intended

This won't do anything obviously, but it'll mean the code exists. I'm not entirely against giving ninja borgs a maximum of three tokens (or later a 15-20 minute token recharge if it's needed)

These things are extremely powerful and it makes sense to give them a cap.

code/modules/antagonists/ninja/ninjaDrainAct.dm
aKromatopzia: > 

does it not be a numeric value in cases where the var is set to a number?

aKromatopzia:
> overall I don't think this works as intended

The idea is for it to default to 0 and have it so that the base charges value can be edited. Also, "to_chat(ninja, span_danger("UPLOAD COMPLETE."))" is for when it's set to infinite (which is base -1, but if -- works too then nice)

PR #1231, opened 2024-03-02
System Cleaner buff

Superhats:
## About The Pull Request
Simple PR. Doubles the purge rate of system cleaner.

## Why It's Good For The Game
Reasoning: Unlike organics who can benefit from blood filter surgery, system cleaner is, beyond esoteric methods like potassium + water, the only way to purge chems from a synths and generally deal with toxins. It is currently about four times slower than pentetic acid, which means you often have to produce very large amounts of the medicine, and also go through several reagent processing units in the case of something like alcohol poisoning (which can occur very easily due to backpressure surges).

This PR puts system cleaner in line with pentetic for a dedicated purging chem, for a medicine that isn't available at roundstart at all, and which has no other alternative in most cases.

## Proof Of Testing
This was a web PR it's literally just tweaking a single number.

## Changelog
:cl:
qol: System Cleaner now purges chems twice as fast.
/:cl:


Superhats:
Closing because I didn't understand what modular means when I made this.

PR #1242, opened 2024-03-06
[MODULAR I HOPE] Adds persistence to NIF huds

Superhats:
## About The Pull Request
Scrying Lens NIFsoft is now able to be kept persistently between rounds. Accordingly, players may now receive a scrying lens adapter through their loadout.
See #1243 for an alternative which preserves the job restrictions.

## Why It's Good For The Game
Reasoning: HUD glasses are already able to be picked from the loadout, so the precedent of roundstart HUDs is already there. 
This still requires players to first acquire the HUD once through R&D before it can be saved in their NIF, and it still requires them to wear one of a specific selection of eyewear, with the refit giving it a very obvious "HUD-Adapted [Glasses]", making it evident they're using a scrying lens.

## Proof Of Testing
It runs chief. This is like five lines of code.

## Changelog
:cl:
qol: Thanks to Centcom updating to the premium version of Scrying Lens NIFsoft, it may now be persistently kept between rounds.
/:cl:

### DIFF
--- a/modular_zubbers/modules/loadouts/loadout_items/loadout_datum_pocket.dm
+++ b/modular_zubbers/modules/loadouts/loadout_items/loadout_datum_pocket.dm
@@ -26,3 +26,7 @@
 /datum/loadout_item/pocket_items/transpride
 	name = "Trans Flag"
 	item_path = /obj/item/sign/flag/pride/trans
+
+/datum/loadout_item/pocket_items/nif_hud_adapter
+	name = "Scrying Lens Adapter"
+	item_path = /obj/item/nif_hud_adapter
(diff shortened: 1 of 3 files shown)


YakumoChen:
HUD glasses are only available in the loadout to specific jobs, how would persistence aid in keeping this to be the case?

Superhats:
> HUD glasses are only available in the loadout to specific jobs, how would persistence aid in keeping this to be the case?
> 
> image

As I mentioned in the description, addressing that is why I made #1243 as an alternative.

Superhats:
In fact I should probably close this one and keep the other one tbh

PR #1282, opened 2024-03-17
[Nova Port] [TM ONLY; upstream pull] Frontier Tech finishes coming to Spinward!

IgiariValkyr:
## About The Pull Request

When Nova split off from Skyrat, we unfortunately didn't get the rest of the Akhter Frontier machines. It's been a few months, and after playing on Nova for a bit, I need my plastic machines!

This PR adds a lot of flatpacked machines that can let your little hiding hole in the wastelands be livable.

dreamseeker_2024-03-17_04-52-47

- A Portable CO2 cracker, which makes Oxygen from CO2 rich environments with electricity.
- A whole spread of kitchen appliances, like a microwave, range, griddle, "sustenance" dispenser.
- Power generation methods, like a wind turbine, flatpacked solar panels, a sterling generator, even a flatpacked Isotope thing.
- A manual airlock because fuck the AI
- A flatpacked ore silo and material recycler
- An ore thumper, which works similarly to the thumper for ore vents, but works anywhere.
- A GPS beacon
- A Water Synthesizer, because water is hard to find in the middle of nowhere.
- A Botany Synthesizer, for your weed killer and plant steroids.

## Why It's Good For The Game

Frontier equipment is good for creating your own little homes, or places to hide from security. Frontier equipment also felt incomplete since we never saw more come to the server after the Big Split.

## Proof Of Testing

The picture above is running on Bubber code!

## Changelog

:cl:
add: Akhter Frontier Technologies has imported more goods to the sector, allowing many more homemade creations!
/:cl:


StrangeWeirdKitten:
100 percent we'll add this, but I need to know if this PR is a mirror directly from skyrat? Otherwise it should be given an upstream PR. But we can still test merge this all while waiting for the code to trickle down.

IgiariValkyr:
> 100 percent we'll add this, but I need to know if this PR is a mirror directly from skyrat? Otherwise it should be given an upstream PR. But we can still test merge this all while waiting for the code to trickle down.

Fully original port from the Nova lands. I can push it upstream if you want, but we really need to stop relying about that place and switch upstreams imo.

Majkl-J:
> > 100 percent we'll add this, but I need to know if this PR is a mirror directly from skyrat? Otherwise it should be given an upstream PR. But we can still test merge this all while waiting for the code to trickle down.
> 
> Fully original port from the Nova lands. I can push it upstream if you want, but we really need to stop relying about that place and switch upstreams imo.

We are not switching to nova on the simple basis that they are a different design philosophy from ours and also that I and the rest of the surviving maintainer team would go crazy bugfixing their featurecode hell

IgiariValkyr:
> We are not switching to nova on the simple basis that they are a different design philosophy from ours and also that I and the rest of the surviving maintainer team would go crazy bugfixing their featurecode hell

Fair enough. Upstream made.

IgiariValkyr:
Upstream has merged this, so keep testmerged until next upstream pull is made.

lessthnthree:
All these items are great, but it should **very much** so be restricted so that they can only be built off the station z-level, as the intended purpose to be for colony building.

Without that restriction in reality it's just giving people a way to get around cooperating and actually working with other departments to do things. Why work with medical when you can print your own? Engineers or Sci won't give you a jaws? Fuck them print your own!

nevimer:
This was merged upstream, and I pulled that into live. This PR is now in the game.

PR #1295, opened 2024-03-22
Nerfs Heretic Moon Ascension because holy fuck

BurgerLUA:
image

## About The Pull Request

1/5/th of all humans in the game will no longer get converted, but instead just everyone who is viewing the ritual, with no limit.

## Why It's Good For The Game

The code is really bad.
- Assumes humanoid on a station z-level is crew. This includes ghost roles, such as the syndicate.
- No randomness. There is no rng at all involved in picking who is chosen for this, and as far as I know, picks anyone added to the human list first (early joiners). Could be wrong and the list may be shuffled elsewhere.
- Does not filter out those who are antag banned.
- Does not have fallbacks for when it fails to add the antag role for those who already have it.
- Does not filter out people who don't want to be antagonist.
- Possibly a race condition that adds even more heretics present.

The balance is really bad.
- Picks 1/5th of the crew to be converted to antag. 1/5th of the crew. We have like 60 players on average. That's 15 antagonists right there.
- Bubberstation is very generous towards antags and it is very easy to stealth this.
- Cult and Rev are disabled for a reason.

## Proof Of Testing

## Changelog

:cl: BurgerBB
balance: 1/5/th of all humans in the game will no longer get converted, but instead just everyone who is viewing the ritual, with no limit. Go kidnap people and force them to watch the ritual to get converted.
/:cl:

### DIFF
--- a/code/modules/antagonists/heretic/knowledge/moon_lore.dm
+++ b/code/modules/antagonists/heretic/knowledge/moon_lore.dm
@@ -172,7 +172,7 @@
 		Bring 3 corpses with more than 50 brain damage to a transmutation rune to complete the ritual. \
 		When completed, you become a harbinger of madness gaining and aura of passive sanity decrease \
 		, confusion increase and if their sanity is low enough brain damage and blindness. \
-		1/5th of the crew will turn into acolytes and follow your command, they will all recieve moonlight amulettes."
+		Anyone witnessing the ritual will turn into acolytes and follow your command, and they will all recieve moonlight amulettes." //BUBBERSTATION CHANGE: 1/5TH OF EVERYONE TO VIEWERS.
 	gain_text = "We dived down towards the crowd, his soul splitting off in search of greater venture \
 		for where the Ringleader had started the parade, I shall continue it unto the suns demise \
 		WITNESS MY ASCENSION, THE MOON SMILES ONCE MORE AND FOREVER MORE IT SHALL!"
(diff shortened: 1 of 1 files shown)


ReturnToZender:
Normally, I'd say "salt PR" but _holy fuck_

BurgerLUA:
Pretending that this isn't glaring, I literally latejoined as AI and just saw this all going down in realtime. I was basically a glorified observer. Didn't do anything or die except call the shuttle because security was immediately murdehobo'd.

BurgerLUA:
I was told that there were more than 1/5th heretic minions at the time which makes me suspect that a race condition occurred and added more heretics than it was supposed to. Not entirely sure on this because I am not willing to check if any of the procs make anything async.

destrucktoid:
If you are going to do a change like this you should give compensatory buffs to the person who ascends in addition or perhaps a new way to gain lunarites post ascension. The change you've outlined basically makes the ascension completely worthless, especially when compared to other heretic ascensions.

AllCorruptingWyrm:
Personally I think reintroducing side points in place of ascension being gone is a much more happy medium.

Arturlang:
Maybe if someone dies of brain damage in your range, instead they become a lunarite?

Elli-Skala:
> 1/5th of the crew. We have like 60 players on average. That's 15 antagonists right there. 

No.

Swiftfeather:
This is over a month stale, if you get back to it feel free to re-open.

PR #1298, opened 2024-03-22
Unrestricts colony fabricator and makes prybar a one time use item.

xXPawnStarrXx:
## About The Pull Request
This PR aims to remove the insanely overpowered nature of the prybar. Getting access to infinite use door forcers via cargo and without tech is just sad. So you gotta go back to basics or sacrifice space to carry a few!

## Why It's Good For The Game
Having all that content locked behind the cargo console, which most crew can't circumvent without science or cargo assistance for getting a console. The prybar, the bane of balance now balanced, a dirt cheap crowbar that can pry open doors? Only one time now! Hack doors and die to shocks like a real greytider or just rob from engineering.

## Proof Of Testing
image
image
image

## Changelog

:cl:
balance: removed the prybar's infinite uses, unrestricted the fabricator for access to the cool stuff there.
/:cl:


Majkl-J (changes requested):
Don't comment out the actual item itself. Also replace it with a regular crowbar in the design datum, instead of just wiping it as it has no alternative

xXPawnStarrXx:
> Don't comment out the actual item itself. Also replace it with a regular crowbar in the design datum, instead of just wiping it as it has no alternative

The reason it's commented out is because I intend to find a way to make them break after a single use, the two versions of the code I tried didn't work and I'll try again later. I can add a crowbar as a replacement, that's a good point.

Majkl-J:
> The reason it's commented out is because I intend to find a way to make them break after a single use

You can just keep it in the code without it being obtainable, it's still a fairly useful tool for admins to spawn. Up to you though, but the final version should have it not uncommented

modular_skyrat/modules/colony_fabricator/code/design_datums/tools.dm
```diff
@@ -4,7 +4,8 @@
 	description = "Contains all of the colony fabricator's tool designs."
 	design_ids = list(
 		"colony_power_drive",
-		"colony_prybar",
+		//"colony_prybar",
```
StrangeWeirdKitten, line 7:
```suggestion
		//"colony_prybar", BUBBER REMOVAL
```

modular_skyrat/modules/company_imports/code/armament_datums/akh_frontier.dm
```diff
@@ -10,11 +10,11 @@
 
 /datum/armament_entry/company_import/akh_frontier/basic/omni_drill
 	item_type = /obj/item/screwdriver/omni_drill
-
+/*
```
StrangeWeirdKitten, line 13:
```suggestion
/* BUBBER REMOVAL START
```

modular_skyrat/modules/company_imports/code/armament_datums/akh_frontier.dm
```diff
@@ -10,11 +10,11 @@
...
 /datum/armament_entry/company_import/akh_frontier/basic/prybar
 	item_type = /obj/item/crowbar/large/doorforcer
 	restricted = TRUE
-
+*/
```
StrangeWeirdKitten, line 17:
```suggestion
BUBBER REMOVAL END*/
```

modular_skyrat/modules/company_imports/code/armament_datums/akh_frontier.dm
```diff
@@ -29,7 +29,6 @@
 /datum/armament_entry/company_import/akh_frontier/deployables_fab/rapid_construction_fabricator
 	item_type = /obj/item/flatpacked_machine
 	cost = CARGO_CRATE_VALUE * 6
-	restricted = TRUE
```
StrangeWeirdKitten, line 32: Just comment this out with a bubber removal.

modular_skyrat/modules/colony_fabricator/code/design_datums/tools.dm
```diff
@@ -31,7 +32,19 @@
...
+	name = "Crowbar"
+	id = "colony_crowbar"
+	build_type = COLONY_FABRICATOR
+	build_path = /obj/item/crowbar/red
+	materials = list(
+		/datum/material/iron = SHEET_MATERIAL_AMOUNT * 1.75,
+	)
+	category = list(
+		RND_CATEGORY_INITIAL,
+	)
```
StrangeWeirdKitten: This confuses me. Is it the replacement prybar that's supposed to break?

StrangeWeirdKitten (changes requested):

modular_skyrat/modules/colony_fabricator/code/design_datums/tools.dm
xXPawnStarrXx: I forgot to adjust the material price, currently, that's a standard crowbar that costs "*0.5" not "*1.75" and it'll be until I get around to making the prybar break (which may not be until next weekend.)

modular_skyrat/modules/colony_fabricator/code/design_datums/tools.dm
```diff
@@ -30,9 +31,22 @@
...
+	build_type = COLONY_FABRICATOR
+	build_path = /obj/item/crowbar/red
+	materials = list(
+		/datum/material/iron = SHEET_MATERIAL_AMOUNT *0.5,
+	)
+	category = list(
+		RND_CATEGORY_INITIAL,
+		RND_CATEGORY_TOOLS + RND_SUBCATEGORY_TOOLS_ENGINEERING_ADVANCED,
+	)
+
```
StrangeWeirdKitten, line 46: Oh I see, it's just giving a red crowbar as a replacement. It would honestly be better if you created a file inside modular_zubbers specifically to add overrides and the new design path.

modular_skyrat/modules/colony_fabricator/code/design_datums/tools.dm
```diff
@@ -45,7 +59,7 @@
 		RND_CATEGORY_INITIAL,
 		RND_CATEGORY_TOOLS + RND_SUBCATEGORY_TOOLS_ENGINEERING_ADVANCED,
 	)
-
+*/
```
StrangeWeirdKitten, line 62:
```suggestion
BUBBER REMOVAL END*/
```

modular_skyrat/modules/colony_fabricator/code/design_datums/tools.dm
```diff
@@ -30,9 +31,22 @@
...
+	)
+	category = list(
+		RND_CATEGORY_INITIAL,
+		RND_CATEGORY_TOOLS + RND_SUBCATEGORY_TOOLS_ENGINEERING_ADVANCED,
+	)
+
 // Crowbar that is completely normal except it can force doors
 
-/datum/design/colony_door_crowbar
+/*/datum/design/colony_door_crowbar
```
StrangeWeirdKitten, line 49:
```suggestion
/*/datum/design/colony_door_crowbar - BUBBER REMOVAL START
```

modular_skyrat/modules/colony_fabricator/code/design_datums/tools.dm
StrangeWeirdKitten: Basically you can copy all this code into an entirely new .dm and tick it, and it'll add this design to the fabricator. Then when you remove it, you can just delete the file.

modular_zubbers/code/game/machinery/doors/airlock.dm
Majkl-J, line 1: You're better off adding this condition to the core proc with bubber edit comments instead of replacing the whole proc

modular_skyrat/modules/company_imports/code/armament_datums/akh_frontier.dm
```diff
@@ -10,11 +10,11 @@
 
 /datum/armament_entry/company_import/akh_frontier/basic/omni_drill
 	item_type = /obj/item/screwdriver/omni_drill
-
+/* BUBBER REMOVAL START
 /datum/armament_entry/company_import/akh_frontier/basic/prybar
 	item_type = /obj/item/crowbar/large/doorforcer
 	restricted = TRUE
-
+BUBBER REMOVAL END*/
```
Majkl-J, line 17: With the nerf below, does this need to be commented out?

Majkl-J (changes requested):

modular_skyrat/modules/company_imports/code/armament_datums/akh_frontier.dm
xXPawnStarrXx: I personally don't like the fact that you can get a doorjack for 100cr, even if it's one use.

modular_zubbers/code/game/machinery/doors/airlock.dm
xXPawnStarrXx: I keep getting told to modularize things, so I'm making sure it's done. It works as it is currently.

YakumoChen:
Add a span warning while you force the door open to warn the user the prybar will break.

Should be a message like "The prybar groans and bends as it struggles to force open the airlock!"

IgiariValkyr:
> Add a span warning while you force the door open to warn the user the prybar will break.
> 
> Should be a message like "The prybar groans and bends as it struggles to force open the airlock!"

Random chance to do 10 brute to yourself when the crowbar snaps too, lol. Cause if you’re putting your weight into that and it snaps, something is going to hit you. (Don’t though. The fact you made it one use somehow is awesome.)

xXPawnStarrXx:
> Add a span warning while you force the door open to warn the user the prybar will break.
> 
> Should be a message like "The prybar groans and bends as it struggles to force open the airlock!"

I'm probably going to just make it clear in the description that it is one use.

YakumoChen:
Given it can easily lock people into rooms and they're used to it being multiple use, I don't think its really a high enough bar just to change the description (implying that people read item descriptions anyway)

xXPawnStarrXx:
> Given it can easily lock people into rooms and they're used to it being multiple use, I don't think its really a high enough bar just to change the description (implying that people read item descriptions anyway)

That honestly sounds like a ~~skill issue~~ IC issue, if they can't talk their way out of the situation or resolve it another way, they're relying too heavily on the prybar.

modular_zubbers/code/game/machinery/doors/airlock.dm
Majkl-J: If you need to copypaste some tgcode to add it, you and the codebase are better off just adding it into the base file with bubber edit comments

YakumoChen:
This is a game readability issue, not an IC issue, we should be making the game inclusive to the people playing it, not creating "gotcha" moments because the player doesn't read every github PR or examine an item they've used dozens of times before

modular_skyrat/modules/colony_fabricator/code/tools/tools.dm
```diff
@@ -98,7 +98,7 @@
 /obj/item/crowbar/large/doorforcer
 	name = "prybar"
 	desc = "A large, sturdy crowbar, painted orange. This one just happens to be tough enough to \
-		survive <b>forcing doors open</b>."
+		<b>force a door open</b>." //Bubber EDIT
```
YakumoChen, line 101: This is also not clear enough that the item breaks after one use, it can _very_ easily be interpreted as the exact same function as before (if they even notice the slight wording difference)

xXPawnStarrXx: I haven't had a chance to change it yet. I am currently away from home for work but I am going to make it resoundingly clear it breaks after one use.

xXPawnStarrXx:
> The fact you're insistent on being hostile to the players on an obstinately simple change, isn't a good thing. Don't make things a mystery when we can always present clear information to the players clearly and unobtrusively.

I'm not being hostile, I don't see a reason that someone's entire plan should hinge on one item. No back up plan or anything; failure to plan is planning to fail.

modular_zubbers/code/game/machinery/doors/airlock.dm
xXPawnStarrXx: So you 100% want me to place the edit in the main file, because if someone tells me to modularise it after that, I'll send them your direction in that case.

Majkl-J: Yes

xXPawnStarrXx:
> Should be a message like "The prybar groans and bends as it struggles to force open the airlock!"

Trying to get it working cleanly.
image

coffeecoffeecoffee:
how about making it a couple-times-use item? Maybe after 1-2 uses it runs a *chance* to break, and that chance incrementally increases? Gives people a shot at being very, very lucky, and sometimes having a prybar that just absolutely sucks?

xXPawnStarrXx:
> how about making it a couple-times-use item? Maybe after 1-2 uses it runs a _chance_ to break, and that chance incrementally increases? Gives people a shot at being very, very lucky, and sometimes having a prybar that just absolutely sucks?

I think that'd weight in more against what Chen said, if it's a random chance it's obscure. However one use, making it one use and making that very clear... Might be wiser.

Majkl-J:
We have decided that this is an overcomplicated solution. The problem isn't the tool itself but it being available to print so easily, and as such the printing recipe was removed. Further adjustments to its order cost can be made in an atomised PR

IgiariValkyr:
I'd honestly mix both, because it's still very cheap to buy too.

PR #1315, opened 2024-03-28
Nerfs powerator a lot 

Erol509:
## About The Pull Request

Powerator makes a lot of free money from just energy, this makes it a lot worse.

## Why It's Good For The Game

Its too easy to make 100k+ money with it an teg making 20 MW, this nerfs it a lot. 

## Proof Of Testing

Works, its a simple var change nothing serious 

## Changelog

:cl:
balance: Powerator makes too much money, so it was nerfed.
/:cl:


Cyprex:
Do this upstream

Valerioninithis:
So, while I agree some nerfing is probably needed, is the problem with the powerator or how the TEG is balanced...?

projectkepler-ru:
> So, while I agree some nerfing is probably needed, is the problem with the powerator or how the TEG is balanced...?

both

upstream ,powerator make sense. cargo outta make a lotta money to fund the warfare and whatnot 

downstream, cargo being able to do that is incredibly horrifying and we would not appreciate it


TEG is also funny because you can just use thermo machine to change the temperature

Erol509:
https://github.com/Skyrat-SS13/Skyrat-tg/pull/27165

PR #1318, opened 2024-03-29
de-snowflakes basic floor tiles. syncs them to the better tg default.

axietheaxolotl:
## About The Pull Request

see title. currently, basic floor tiles are running an old old snowflaked tau ceti port. SR didn't want to update then, maybe bubber wants to update now.

image

## Why It's Good For The Game

looks significantly better, imo. they make take some getting used to, but I think if you look at it objectively, you can see that they're better.

## Proof Of Testing

see body.

## Changelog

:cl:
imageadd: brings default floor tiles up to /tg/ default.
/:cl:


ReturnToZender:
These look cool, I never thought it was any different elsewhere. Cool!

StrangeWeirdKitten:
Just deleting a file will not be permanent. You need to comment out the entirety of ``modular_skyrat\modules\aesthetics\floors\code\floors.dm``

StrangeWeirdKitten (changes requested):
You need to comment out the entirety of ``modular_skyrat\modules\aesthetics\floors\code\floors.dm`` as deleting files will only show their return next upstream

Swiftfeather:
If you get back to this, feel free to re-open.

PR #1319, opened 2024-03-29
Ashtongue necklaces are now fireproof.

Superhats:
## About The Pull Request
What it says on the tin. The ashen necklaces don't burn anymore. This extends to the loadout's cursed ashen necklace.

## Why It's Good For The Game
It seems strange that an item from the necropolis would burn so easily, especially since it's worn above all other clothing and can't be stopped from catching on fire, in a land that's usually full of fire.

This is particularly directed towards the loadout's cursed necklace, which is unremovable and as such can't be kept safe in a container.

## Proof Of Testing
It's one line of code it should be fine.

## Changelog

:cl:VrazzleDazzle
qol: The ashwalkers' ashen necklaces are now fireproof.
/:cl:


modular_skyrat/modules/modular_items/code/necklace.dm
```diff
@@ -12,6 +12,7 @@
 	worn_icon = 'modular_skyrat/master_files/icons/mob/clothing/neck.dmi'
 	icon_state = "ashnecklace"
 	w_class = WEIGHT_CLASS_SMALL //allows this to fit inside of pockets.
+	resistance_flags = FIRE_PROOF
```
StrangeWeirdKitten, line 15: Needs modularization. Just make an override folder in modular_zubber and have this bit of code in it.

```
/obj/item/clothing/neck/necklace/ashwalker
	resistance_flags = FIRE_PROOF
```

StrangeWeirdKitten (changes requested):

modular_skyrat/modules/modular_items/code/necklace.dm
StrangeWeirdKitten: Or alternatively PR it upstream

Superhats: That's done 👍

PR #1320, opened 2024-03-29
[SALT PR] Death bolts no longer dust corpses

Superhats:
## About The Pull Request
What it says on the tin, again. 

## Why It's Good For The Game
Dying to a death bolt, or getting hit with one while dead (Which isn't hard given they're specifically coded to hit corpses) normally dusts you, which is the one remaining instance of megafauna being a hard round removal. Some fauna, such as vileworm or legion, eat you, but those still allow recovery if the mob is killed, whereas even if a daring miner went to kill the wendigo that downed you, and you didn't get dusted already, you're almost guaranteed to be gone from a death bolt by the time the fight's over. 

Add to that the complete lack of feedback that it was in fact a death bolt that did it, and it can lead to people being dusted with no explanation when fighting colossus/wendigo. 

## Proof Of Testing
Spawned 3 wendigos and a colossus, got nuked by death bolt, corpse still here. 👍 

## Changelog

:cl:VrazzleDazzle
del: Death bolts no longer dust you on death.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/hostile/megafauna/colossus.dm
+++ b/code/modules/mob/living/simple_animal/hostile/megafauna/colossus.dm
@@ -193,12 +193,14 @@
 
 /obj/projectile/colossus/on_hit(atom/target, blocked = 0, pierce_hit)
 	. = ..()
+/**
 	if(isliving(target))
 		var/mob/living/dust_mob = target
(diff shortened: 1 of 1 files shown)


Majkl-J (approved):
dumb_even

PR #1322, opened 2024-03-29
Slight Moonstation Atmos Performance Optimization

BurgerLUA:
## About The Pull Request

Optimizes atmos so that the planetary air doesn't run into float precision errors and creates useless concentrations of air (less than 0.01 moles) when next to airlocks.

This basically caused atmos to process while near airlocks that have been opened at least once, causing maybe 2-5% time dilation by the end of the 1 hour mark.

## Why It's Good For The Game

Optimizations good.

## Proof Of Testing

Compiled and tested.

## Changelog

:cl: BurgerBB
fix: Atmos optimization.
/:cl:


IgiariValkyr:
GigaChad

Eastwardrope:
why do I feel this is related to april fools

Majkl-J (approved):
Minor enough to where I feel comfortable letting this happen each year

PR #1334, opened 2024-03-31
Adds Toothpaste, Tooth Decay, Dentures, and more!

nevimer:
Clandestine PR, made with NO description in the body. Prime material for a speedmerge with no review, certainly!


Superhats:
What a nice PR, we should merge it as soon as possible.

PR #1335, opened 2024-04-01
[MIRROR/PORT] Expanded Interaction Preferences

lessthnthree:
## About The Pull Request

Mirror of https://github.com/Skyrat-SS13/Skyrat-tg/pull/27043 / https://github.com/NovaSector/NovaSector/pull/1527 with Bubber edits included (character directory.)

## Original PR

Port:
Original PR: https://github.com/NovaSector/NovaSector/pull/1527/
Additional port: https://github.com/effigy-se/effigy-se/pull/790

Hi.
This PR adds some preferences for illicit meetups.
Let's go over them.
For clarification, verse refers to 'okay with either bottoming or topping' and switch corresponds with domming and subbing. Jsyk.

For midnight meetings,
Top - Dom
Top - Switch
Top - Sub

Verse-Top - Dom
Verse-Top - Switch
Verse-Top - Sub

Verse - Dom
Verse - Switch
Verse - Sub

Verse-Bottom - Dom
Verse-Bottom - Switch
Verse-Bottom - Sub

Bottom - Dom
Bottom - Switch
Bottom - Sub

I am willing to put 'Yes -' infront of each of those, but I don't really think it's needed? It's sort of implied.
I also changed some of the other ones.
'Ask' is now 'Ask (L)OOC.' Of course you ask ICly, it came free with your 'playing a social interaction game,' and I don't know why we'd have a preference just for Ask ICly? So instead, ask in LOOC if it's a go or not.

As for assault and consumption, I have made the default be 'no,' and slimmed down the options for assault. This sort of assault is implicitly tied with conception, you should go off the preference readout for that.

I have added another preference for hypnosis.
Always/Whenever - Corresponding to just 'yeah whenever's fine even if it's just for trivial RP stuff.'
Gameplay Only - For stuff like hypnobangs or traitor stuff, not for "woohoo."
Ask (L)OOC - We've been over this.
Check OOC - Check the notes.
No - Means no.

image

image

This allows you to get a more granular experience and allows for a wider array of display of identity, and there have been situations where, say, a sub person being engaged with and being expected to bottom when they don't want to.
Additionally, it really, really shouldn't default to 'ask' for extreme content like the other two.

As to the hypnosis preference, I am self-serving; but more importantly, this stuff is surprisingly accessible nowadays (thru the NIFsoft and Hypno Goggles) and there's been more people around with it as kind of a central theme.

<details>

image
  
</details>

## Changelog

🆑 Nerev4r
qol: Interaction preferences have been expanded to make a wider array of characters more comfortable.
/:cl:

### DIFF
--- a/modular_skyrat/master_files/code/modules/client/preferences/erp_preferences.dm
+++ b/modular_skyrat/master_files/code/modules/client/preferences/erp_preferences.dm
@@ -134,10 +134,30 @@
 	savefile_key = "erp_status_pref"
 
 /datum/preference/choiced/erp_status/init_possible_values()
-	return list("Yes - Switch", "Yes - Sub", "Yes - Dom", "Check OOC", "Ask", "No", "Yes")
+	return list(
+		"Top - Dom",
+		"Top - Switch",
+		"Top - Sub",
+		"Verse-Top - Dom",
+		"Verse-Top - Switch",
+		"Verse-Top - Sub",
+		"Verse - Dom",
+		"Verse - Switch",
+		"Verse - Sub",
+		"Verse-Bottom - Dom",
+		"Verse-Bottom - Switch",
+		"Verse-Bottom - Sub",
+		"Bottom - Dom",
+		"Bottom - Switch",
+		"Bottom - Sub",
+		"Check OOC Notes",
+		"Ask (L)OOC",
+		"No",
+		"Yes",
+	)
 
 /datum/preference/choiced/erp_status/create_default_value()
-	return "Ask"
+	return "Ask (L)OOC"
 
 /datum/preference/choiced/erp_status/is_accessible(datum/preferences/preferences)
 	if (!..(preferences))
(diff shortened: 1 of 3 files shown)


lessthnthree:
> I love extra stuff but I do not understand what half of these preferences are meant to communicate, you can just do "Switch - top leaning" "Switch - bottom leaning" and get about the same effect, no?
> 
> I also don't understand the functional difference between being "verse," and being a "switch," as both are okay with both bottoming and topping and there's no functional difference listed here... Every other misc. change is really good though and the inclusion with character directory is PEAK

Yeah I get what you mean. Top/bottom is one scale, and dom/switch/sub is another. A character can be a submissive top, for example. Or someone's a switch and can either be dom or sub, so long as they're on the bottom.

ReturnToZender:
> Yeah I get what you mean. Top/bottom is one scale, and dom/switch/sub is another. A character can be a submissive top, for example. Or someone's a switch and can either be dom or sub, so long as they're on the bottom.
> 
> Unless there's a big glaring issue, I think it's best for the three codebases to share consistent preferences.

I suppose so!

Cyprex:
I have no clue what "verse" means here- my best guess would be "versatile". I could (and will) read more about this, but this ought to be clear from the start

Cyprex:
verse-bottom- dom
What does this mean?

As for Hypno whilst we don't have headrevs in rotation, admins might still run it, and I am of the opinion that it and other conversion antags should *function*. In my opinion this isn't the case as-is if you require headrevs to check someones prefs before flashing them is pretty unreasonable. We do have vampires in rotation, which can make thralls, flavored differently but the end result is the same. 

Imo if we really wanted to allow people to opt out of conversion this ought to be enforced mechanically, and would need to be clearly communicated to headrevs, vampires etc. An overlay seems effective for this.

Cyprex:
To add to this a simple mechanical enforcement of no gameplay-only hypno wouldn't work well- Imagine a headrev running up to someone that's not mindshielded, the flash not working, and the headrev subsequently being outed.

projectkepler-ru (approved):

PR #1339, opened 2024-04-02
Removes uploads in a simpler way than the other PR

Odairu:
## About The Pull Request
Removes upload boards from being printable, discussions are open on another PR, this one just does it in a simpler way

## Why It's Good For The Game
Law ping pong bad

AI CORE IS A VALID UPLOAD YOU CAN NEVER GET RID OF THEM ALL!!!

## Proof Of Testing
Its the same code as an old PR that was closed

## Changelog

:cl:
del: Deletes upload boards from research making them finite on station without admin intervention
/:cl:


nevimer:
id like to see them be part of something else other than printing

YakumoChen:
As in the other PR there are better ways to solve the problem (per your "why its good") that are less disruptive, notably making boards GPS tracked at all times.

aKromatopzia:
>     * making it nearly impossible to subvert the AI quietly under common circumstances (like the AI using multicam to watch the spare board)

Doesnt the camera in secure tag give an alert regardless if its emp'd, walked close to, or etc

StrangeWeirdKitten:
Did more discussions with other maintainers, and played some purged borg a little to understand why this would be awful. If you make the AI unfixable because it took the two uploads, then you have to turn its core into a wasteland which is extremely disruptive. Then if it moves its core? Good luck after that. This is just exchanging one issue with more issues and was part of the reason why I closed the cargo law board PR I tested myself since making it apart of something for the crew to order adds a massive one-sidedness.

StrangeWeirdKitten:
> As in the other PR there are better ways to solve the problem (per your "why its good") that are less disruptive, notably making boards GPS tracked at all times.

Someone can do this. This is a great idea.

aKromatopzia:
> Did more discussions with other maintainers, and played some purged borg a little to understand why this would be awful. If you make the AI unfixable because it took the two uploads, then you have to turn its core into a wasteland which is extremely disruptive. Then if it moves its core? Good luck after that. This is just exchanging one issue with more issues and was part of the reason why I closed the cargo law board PR I tested myself since making it apart of something for the crew to order adds a massive one-sidedness.

there are already rules against moving the core though? You can't do it on purged and perhaps even malf without being slapped with a powergaming charge, per the policy.

PR #1363, opened 2024-04-08
[not modular] Removes death whispers from a random trauma you can get after getting brain surgery

Swiftfeather:
## About The Pull Request
You can't get death whispers after geting surgery. in fact you can't get it at all. Because it doesn't fit. Wow you got brain damage from a surgery or something else that's now causing you to hear the voices of the dead who can tell you anything about the game that's so cool. Sure it's intermittent and not fully reliable but it's also still stupidly broken once a group of ghosts find out you can hear them.

## Why It's Good For The Game
I've gotten this twice before and it's honestly just really unfair and kind of immersion breaking. This is also a salt pr but the salt is just what got me to finally move on this. It don't fit in with the whole "medium" level of roleplay we are striving for and is generally stinky and kinda goes against like all the metagaming principles we've got.

## Proof Of Testing
image

## Changelog

:cl:Swiftfeather
del: Removes death whispers trauma.
/:cl:

### DIFF
--- a/code/datums/brain_damage/special.dm
+++ b/code/datums/brain_damage/special.dm
@@ -273,7 +273,7 @@
 /datum/brain_trauma/special/tenacity/on_lose()
 	owner.remove_traits(list(TRAIT_NOSOFTCRIT, TRAIT_NOHARDCRIT, TRAIT_ANALGESIA), TRAUMA_TRAIT)
 	..()
-
+/*BUBBERSTATION EDIT REMOVAL START
 /datum/brain_trauma/special/death_whispers
 	name = "Functional Cerebral Necrosis"
 	desc = "Patient's brain is stuck in a functional near-death state, causing occasional moments of lucid hallucinations, which are often interpreted as the voices of the dead."
@@ -300,7 +300,7 @@
 /datum/brain_trauma/special/death_whispers/proc/cease_whispering()
 	REMOVE_TRAIT(owner, TRAIT_SIXTHSENSE, TRAUMA_TRAIT)
 	active = FALSE
-
+*///BUBBERSTATION EDIT REMOVAL END
 /datum/brain_trauma/special/existential_crisis
 	name = "Existential Crisis"
 	desc = "Patient's hold on reality becomes faint, causing occasional bouts of non-existence."


Majkl-J (commented):
If I read this correctly, this gives you a random stream of death whispers on average every 100s lasts 5-30s. Doesn't sound extremely impactful, if we _really_ need this, I think I'd rather see the time shortened to where it isn't practical (5-10 maybe), or the chances lowered in half, instead of straight up removal.

It's a very niche thing that probably isn't as impactful as it seems with its random chances

Swiftfeather:
> If I read this correctly, this gives you a random stream of death whispers on average every 100s that lasts 5-30s. Doesn't sound extremely impactful, if we _really_ need this, I think I'd rather see the time shortened to where it isn't practical (5-10 maybe), or the chances lowered in half, instead of straight up removal.
> 
> It's a very niche thing that probably isn't as impactful as it seems with its random chances

This is a salt PR

Majkl-J:
> This is a salt PR. I've come up with the best solution to the issue _I_ have

Well I suggested other ideas on how to tackle this

Swiftfeather:
> > This is a salt PR. I've come up with the best solution to the issue _I_ have
> 
> Well I suggested other ideas on how to tackle this

I also just thought about changing the path to a magical trauma and making it always hear dchat

Majkl-J:
> I also just thought about changing the path to a magical trauma and making it always hear dchat

We can give that a shot. Could be interesting

YakumoChen:
Imagine being able to hear the voices of the dead and taking what literal ghosts tell you as 100% irrefutable fact in character and letting it dictate your round

Doom-external:
> I'm going to need a solid, gameplay concern or justification outside of general disagreement with a extremely niche mechanic. 

I saw a member of security (I believe it was the HOS) roll this during a round with a ninja, once dchat figured out they could be heard they spammed the ninjas location until they were found and killed. The ninja had no idea this was happening and could do nothing to stop it as dchat was flooded with callouts, enough that the majority or all of the messages HOS would hear are callouts. While it was incredibly LRP for the HOS to do, it is something that happened.

Swiftfeather:
This was a salt pr and now I don't care

PR #1369, opened 2024-04-09
You can now research ammo, alongside some other sec changes

projectkepler-ru:
## About The Pull Request
Due to our lord and saviour giving security only 1 gun per mag, I've just remembered after 4 months that I should probably give them more mag per crate and also allow you to buy the mag yourself
This applies to currently every gun crate we have.. alongside name correction as we forgot to de-french all of the guns crate

Alternatively I could just make the magazine require tech research instead but I want to see what people think

## Why It's Good For The Game
The point of ballistic weapon has always been
1) the emp immunity. While most antags are actually weaker to energy weapon
2) being able to reload in combat versus retreating and recharging is a huge deal
And our ballistic option that do exist are almost always going to be the base one that we already trust rather than the one SR given us because it's not something we can reliably source ammo for
People don't like the half the gun because they can't really get more mag for it. That's why the WT-551, and revolvers are still the best ballistic option thus far
So this is meant to address that issue

## Proof Of Testing
N/A

## Changelog

:cl:
balance: Our contract with carwo has been re-written, every gun crate that start with mag now actually start with the amount of mag they should have. Bless the star!
text: fixed some weird naming inconsistencies given that our gun arent french anymore... begone
/:cl:


ReturnToZender:
I'd prefer they be weapons tech but otherwise 👍

projectkepler-ru:
> I'd prefer they be weapons tech but otherwise 👍


I did say
alternatively they could be locked behind weapon tech

ReturnToZender:
> > I'd prefer they be weapons tech but otherwise 👍
> 
> I did say alternatively they could be locked behind weapon tech

That's why I mentioned it, yeah!

xXPawnStarrXx:
The main point of ballistic weapons is they do brute instead of burn and inflict bleed wounds instead of burns. While I support the idea of equalising the usages; trading limited ammo supply for no rechargability and emp resistance but sheer accessability to firearms and more ammo isn't going to help the gameplay loop much from my perspective.

projectkepler-ru:
> The main point of ballistic weapons is they do brute instead of burn and inflict bleed wounds instead of burns. While I support the idea of equalising the usages; trading limited ammo supply for no rechargability and emp resistance but sheer accessability to firearms and more ammo isn't going to help the gameplay loop much from my perspective.

Not really, we can always pushes the ammo tech behind even harder to get research. The point of ballistic to me wasn't a sidegrade but rather a mid to late game tier weaponry security could get. Every department has some sort of critical mass where you can no longer just expect to run in head on with your standard antag gear without getting gamered on, and I feel that should be the case for security too, where their end game is that you can no longer just run in with EMP and expect it to work as well when they were still in their early phase.


I still need to replace the bogseo with something more sane to actually balance. With how rediculous it is

projectkepler-ru:
> The main point of ballistic weapons is they do brute instead of burn and inflict bleed wounds instead of burns. While I support the idea of equalising the usages; trading limited ammo supply for no rechargability and emp resistance but sheer accessability to firearms and more ammo isn't going to help the gameplay loop much from my perspective.


do feel free to point out any further concerns, I want to make sure they're all addressed incase I messed up something along the way or if the research tech given is too easy to achieve

AllCorruptingWyrm:
Security has never been more tolerable and actively enjoyable since this PR.

IgiariValkyr:
Sol Rifle normal mag prints the short mag instead

ReturnToZender:
Noticed a couple drum designs in the research web, they should be taken off. Drum magazines offer way too much power, meaning that Security would never have to retreat from a fight even in cases where they waste copious amounts of ammo.

With that said, though, I love the move towards science being able to research and print magazines for these weapons. Security can have a _little_ power-creep, as a treat.

projectkepler-ru:
> Noticed a couple drum designs in the research web, under illegal tech: they should be taken off. Drum magazines offer way too much power, meaning that Security would never have to retreat from a fight even in cases where they waste copious amounts of ammo.
> 
> Also, don't see much point in offering both the standard and short magazines even if the Standard is twice as much iron, chiefly because iron can be infinitely ordered by cargo, meaning any difference in cost isn't really significant enough to create a choice. 30 shots is a _significant_ upgrade from 15 for a not really significant increase in cost. I'd personally argue that only the Short should be left in, especially due to the type of weapons that fire it.
> 
> With that said, though, I love the move towards science being able to research and print magazines for these weapons. Security can have a _little_ power-creep, as a treat.

i could make it require illegal and alien tech if its that big of a concern, since illegal tech is still not too terribly common

projectkepler-ru:
> Sol Rifle normal mag prints the short mag instead

its been fixed since yesterday but we have testmerge conflict

Majkl-J (approved):

PR #1365, opened 2024-04-09
Pod ruin changes

kaylexis:
## About The Pull Request

Similar to #1338, but fixes the issue with the colony fab not existing by spawning a flatpacked item version in a crate.
also adds scrubbers and some more space

## Why It's Good For The Game

Podpeople haven't been updated in.. never

## Proof Of Testing

Screenshot 2024-04-08 213924

## Changelog

:cl:
add: Added more stuff to the seed vault.
/:cl:

PR #1372, opened 2024-04-09
replaces icemoon interdyne with dauntless

kaylexis:
## About The Pull Request

Atomized from PR #1365, does what the title says.

## Why It's Good For The Game

The current icemoon interdyne (from skyrat) is a mediocre map, and a bunch of people dont like it. 
ISSUES WITH THIS BAD MAP: overuse of glitter, mining markers as lights, normal lights not connected to walls, weird layout, overuse of forcefields, looks like an ERP dungeon, and encourages exploration into the mining level which has a history of leading to conflict

## Changelog

:cl:
del: Replaced icebox interdyne with an ice-ified dauntless
/:cl:


NargaInDisguise:
Instead of doing this, just replace it with the old interdyne that people actually liked? Why are we putting the dauntless there?

PR #1014, opened 2024-01-15
Lavaland Dauntless / Removes cameras

StrangeWeirdKitten:
## About The Pull Request

Removes cameras.

## Why It's Good For The Game

The cameras were a mistake and they don't need them outside of xenobiology. There's a reason why I didn't map Dauntless to have cameras and I've let you all see why. (The station AI)

## Proof Of Testing

image

## Changelog

:cl:
del: Removes the Dauntless cameras
/:cl:

PR #1375, opened 2024-04-10
Removes Dauntless cameras: Again

StrangeWeirdKitten:
## About The Pull Request

Per title. I didn't realize cameras got readded at some point.

## Why It's Good For The Game

#1014

Do you guys actually need these? I really don't think you don't, but I'm willing to listen, and discuss. But it's a 50x50 map and you exist two seconds away from the other end of the ship.

You allow an inventation of more issues when cameras exist. Yes, it's shitty for an AI to start screwing with you, but AIs should not have any interaction with Dauntless, and that should not be possible round start. The other option is to leave it open to player agency. Yes, it's not really an issue right now, But when it inevitabally does become one, because Dauntless doesn't live in a void, then it's on the same agency to use wirecutters and a screwdriver.

## Proof Of Testing

compiles


StrangeWeirdKitten:
You guys can cut cameras. Sure

PR #1395, opened 2024-04-18
Fixes the dual inducers, opting to keep the tg one (#27227)

Majkl-J:
* removes remnants of the skyrat borg inducer

* begone


Majkl-J:
My master stash got showed into it 💀

PR #1398, opened 2024-04-18
Security now always get deathrattle implants

BurgerLUA:
## About The Pull Request

Security now always get deathrattle implants. Note that only security can hear their own deathrattles. Also note that this doesn't roll if the entire station has deathrattle implants.

Mechanically, this just forces the deathrattle security event.

## Why It's Good For The Game

Adds an extra layer of protection for security officers killed in the line of duty. Penalizes random green-shift assassinations of security players just because they're security. This should not affect traitors normally/"fairly" killing security considering:
- Security constantly broadcasts their location anyways.
- Suit sensors exist, and a paramedic/warden/AI usually snitches.
- An AI usually watches most conflicts.

## Proof Of Testing

image

## Changelog

:cl: BurgerBB
balance: Security now always get deathrattle implants. Note that only security can hear their own deathrattles.
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/security_deathrattle/security_deathrattle.dm
+++ b/modular_zubbers/code/modules/security_deathrattle/security_deathrattle.dm
@@ -1,0 +1,18 @@
+//This file makes it so that security will always get a deathrattle.
+
+/datum/station_trait/deathrattle_department/security
+	weight = 0
+
+/datum/controller/subsystem/processing/station/SetupTraits()
+
+	. = ..()
+
+	var/datum/station_trait/deathrattle_department/security/existing_security_trait = locate() in station_traits
+	if(existing_security_trait)
+		return .
+
+	var/datum/station_trait/deathrattle_all/existing_all_trait = locate() in station_traits
+	if(existing_all_trait)
+		return .
+
+	setup_trait(/datum/station_trait/deathrattle_department/security)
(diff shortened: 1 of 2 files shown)


modular_zubbers/code/modules/security_deathrattle/security_deathrattle.dm
```diff
@@ -0,0 +1,18 @@
...
+/datum/station_trait/deathrattle_department/security
+	weight = 0
+
+/datum/controller/subsystem/processing/station/SetupTraits()
+
+	. = ..()
+
+	var/datum/station_trait/deathrattle_department/security/existing_security_trait = locate() in station_traits
+	if(existing_security_trait)
+		return .
```
Majkl-J, line 12: Is this really needed? With the weight set to 0 this should never prop, no?

Majkl-J (changes requested):

modular_zubbers/code/modules/security_deathrattle/security_deathrattle.dm
BurgerLUA: true, but it's just in case:tm:

BurgerLUA:
> Maybe a research item that can be disabled if comms get disabled? Could put it under tracking tech but a full blown round start deathrattle implant is kind of a lot. Especially since death rattle announces go the entire station.

Deathrattle implants only go to people with a deathrattle implant.

The-Sun-In-Splendour:
Can we EMP deathrattles? This feels like a significant buff to security?

PR #1401, opened 2024-04-18
Security 911 (Backup Requested!)

StrangeWeirdKitten:
alternative to #1398 
closes #1398 

## About The Pull Request

Sechailers can now request backup. 

- Requests Backup to the security channel. 1 minute cooldown
- EMPs make the function inoperatable for two minutes.
- It whistles. It puts a massive red blurb of text in the chat. It has balloon alerts. It logs to admins.

## Why It's Good For The Game

I do agree that there should be a quick button to request backup. But I don't think forcing deathrattle implants into all of security is the solution

## Proof Of Testing

- [x] EMP functionality works
- [x] Properly screams in security channel
- [x] It is loud, everyone knows you called for backup, and it logs to admins

image

## Changelog

:cl:
add: 911 Backup Requested to the security hailer
/:cl:


BurgerLUA:
I don't know if this is a good alternative. The reason why I didn't add yog's system (which is similiar to this one) is because I do not want security officers to press that backup button in the middle of conversation. I rather people be punished for ambushing security.

Majkl-J:
> I don't know if this is a good alternative. The reason why I didn't add yog's system (which is similiar to this one) is because I do not want security officers to press that backup button in the middle of conversation. I rather people be punished for ambushing security.

If they are in the middle of talking with a suspect they can also just type `:s need assistance by [location]`. It doesn't make a loud sound unlike the single press button, so really the only time you would want to use it when you either are in combat or wish to escalate into it.

StrangeWeirdKitten:
> I don't know if this is a good alternative. The reason why I didn't add yog's system (which is similiar to this one) is because I do not want security officers to press that backup button in the middle of conversation. I rather people be punished for ambushing security.

I tried to make it as loud and visible as possible for this specific purpose with a clear combat log so it properly communicates why someone might have been immediately shot versus the much more silent :s backup here when people are mid conversation. This will in turn be something to trigger with easy possibilities to disable temporarily. I don't really enjoy the idea of silent implants going off. Yes, security works very much in teams. But sensors are also / should be widely used in conjunction.

Moostard:
Love this thing, but...
I think the backup request should be more visible on sec comms. It can be easily missed. Additional "!" might do it.
It is not needed, but I think it would make it much better, especially for those that cant pay attention to chat all the time to notice it.

StrangeWeirdKitten:
Have been busy, but I need to fix something before this gets merged

StrangeWeirdKitten:
Another silly bug, and I asked for the TM to be pulled until I fix it. I'll probably do it Monday.

KeRSedChaplain:
I foresee this just making stealth antagonism even less common

projectkepler-ru:
> I foresee this just making stealth antagonism even less common

How so?
they still function the same, it's empable, affected by radio jammer

and in any other case y ou may mention it's not any different from normal in-game occurence where a medbay staff notice someone suit get emp'd, or a diligent warden actually watching the console, or AI watching their crew monitoring

StrangeWeirdKitten:
image

ReturnToZender:
What does this PR need fixed to get in? I can handle it, I think this would be a major boon to Security, more so than any weapon could do

Majkl-J:
> What does this PR need fixed to get in? I can handle it, I think this would be a major boon to Security, more so than any weapon could do

The radio inside keeps blaring out comms even when muted, or so I've been told

Majkl-J:
Yea this is basically staled out until the underlying bug is fixed

PR #1415, opened 2024-04-22
unapologetically lowers the body size minimum by .1

ReturnToZender:
## About The Pull Request
This PR lowers the minimum body size you can select by 1. See proof of testing image to see what that looks like.

## Why It's Good For The Game
My bottoms need to be as close to half my height as possible, 0.8 looks like your average 5'2 person, which is good, but we need better.

## Proof Of Testing
Please note that 0.6 body height is provided in frame for reference, the minimum this is set to is 0.7. 
image

## Changelog

:cl: ReturnToZender
balance: Dubious creatures got a buff and can now go down to .7 sprite size.
/:cl:


The-Sun-In-Splendour:
does this decrease max hp like on other servers? if it does not, then why not?

YakumoChen:
It would be nice if there was some kind of drawback to making your sprite incredibly hard to click

Iamgoofball:
:-1: small sprite people already suck ass to click on, this just makes the problem worse

legumesan:
epic

PR #1433, opened 2024-04-26
Not L4D3 - Valve Toggling Event

nevimer:
## About The Pull Request

This event allows valves to be toggled at random as a part of the event pool. Categorized under engineering, makes valves have a small potential to toggle at random. 

Variety!

## Proof Of Testing

https://github.com/Bubberstation/Bubberstation/assets/77420409/363cb6de-f6b6-4e16-bc81-314aa592cc2c

## Changelog

:cl:
add: Valves can trip via Events
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/events/valve.dm
+++ b/modular_zubbers/code/modules/events/valve.dm
@@ -1,0 +1,30 @@
+/datum/round_event_control/valve
+	name = "Release Half-Life 3"
+	wizardevent = FALSE
+	max_occurrences = 5
+	min_players = 5
+	category = EVENT_CATEGORY_ENGINEERING
+	description = "Opens or closes random valves on station."
+	min_wizard_trigger_potency = 0
+	max_wizard_trigger_potency = 7
+	typepath = /datum/round_event/valve
+
+
+//Called when the tick is equal to the start_when variable.
+//Allows you to start before announcing or vice versa.
+//Only called once.
+/datum/round_event/valve
+	announce_when = 50
+	end_when = 20
+/datum/round_event/valve/start()
+	var/list/world_valve_list = list()
+	var/list/station_z = SSmapping.levels_by_trait(ZTRAIT_STATION)
+	for(var/obj/machinery/atmospherics/components/binary/valve/valve in world)
+		if(valve.z == locate(valve.z) in station_z)
+			world_valve_list |= valve
+	var/list/valves_to_trip = list()
+	for(var/i in 1 to 5)
+		valves_to_trip |= pick_n_take(world_valve_list)
+	for(var/obj/machinery/atmospherics/components/binary/valve/current_valve in valves_to_trip)
+		current_valve.interact()
+		announce_to_ghosts(current_valve)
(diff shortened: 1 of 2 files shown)


Swiftfeather (changes requested):
Don't add this

nevimer:
> Don't add this

Engineering needs more events, this goes similar to supermatter surge.

Swiftfeather:
as an atmos main this is just an event that is going to ruin people's entire rounds

nevimer:
> as an atmos main this is just an event that is going to ruin people's entire rounds

It won't, I promise you, theres a large variety of better devices to use which are both more creative and provide better yields.

NargaInDisguise:
can this just, add plasma straight into distro with no warning?

aKromatopzia:
> can this just, add plasma straight into distro with no warning?

no because this event only affects binary valves? this only would effect SM setups by people who prefer binary valves by default; and pure->canister valves on uh... deltastation I think?

Majkl-J:
> > can this just, add plasma straight into distro with no warning?
> 
> no because this event only affects binary valves? this only would effect SM setups by people who prefer binary valves by default; and pure->canister valves on uh... deltastation I think?

Correct, but yes this needs some indication because we almost had a delam already and engineering and by extent security was confused about sabotage for the next 1 hour

StrangeWeirdKitten (changes requested):
Yeah, it needs an indication, and ideally one that lingers so people who are not in the room can immedietly identify what happened versus needing to spot a 1 pixel green sprite

SovietSpaghetti:
in what way is this engaging or fun at all? Just seems like a nuisance of an event even if there is an indicator. Suddenly the whole distro is spaced and nobody can figure out why because they don't see a single green pixel. The SM delams because its cooling loop has been immediately vented to space. 

At best, the event is a mild annoyance that takes less than a second to fix if you manage to see it, at worst it ruins the round because you don't see a single pixel change color.

Redrover1760:
All this pr will people just leave pipes disconnected instead of using valves...

sunofang:
Not really that engaging or fun for engi as many other people have said. Only thing this does is ruin elaborate setups for no real reason. And it can be easily circumvented by just replacing the valves with pumps ordisconnected pipes.

nikothedude:
this is just a really really bad idea, you cant actually really balance something like this because valves can be either completely useless or completely gamebreaking depending on where in a pipenet theyre placed

plus, without a LOCATION in the announcement, how are people supposed to actually know whats been changed?

events like these generally suck ass regardless. "random thing goes wrong at random time" is never a good event in any game that isnt leaning into it, its just tedium

atmos will never go out of their way to seek out these valves and fix them unless its massively fucking with them

R3dtail:
As a long term engineer player yeah this just means don't ever use valves ever.  This isn't a very pleasant thing to work with and is such a hard nerf to using manual valves to basically render them useless.  The whole point of valves is to have aa toggle for gas flow.  Making it basically a straight pipe waiting to happen is ... no.    What does this add to make more interesting rounds?  All I can see this doing is making engineer players frustrated or wholesale ceasing to use manual valves unless they're deliberately trolling.

StrangeWeirdKitten:
this PR is being reworked.

PR #1378, opened 2024-04-10
Removes prybars from the RCF

Superhats:
## About The Pull Request
Band-aid fix since https://github.com/Bubberstation/Bubberstation/pull/1298 is in dev hell. Prybars are no longer available in the RCF, and have been replaced with pocket crowbars.

I want to keep the scope fairly narrow so currently you can still order them standalone for 100 creds each.

## Why It's Good For The Game
I've seen a resurgence of departments just ordering a RCF at roundstart with the dept budget. Every department being able to mass-produce what's essentially jaws of life with zero research involved and only a tiny bit of titanium is quite bad.

## Proof Of Testing
image
But if you close your eyes...

## Changelog

:cl:VrazzleDazzle
del: Prybars are no longer available in the Rapid Construction Fabricator.
/:cl:

PR #1437, opened 2024-04-27
Makes the Prybar Bulky (and adds it back to the RCF)

StrangeWeirdKitten:
Reverts Bubberstation/Bubberstation#1378

## About this PR

Prybar changes. Simply adds bulk to them.

## Why it's good for the game

I think an overall thing that people don't consider in regards to item balance is that bulk massively changes the utility of something. I'm not against it being able to pry open doors, but it can be the size of the robotics mech crowbar in game.

## Proof of Testing

It has girth and compiled and it was too big for my bag.

## Changelog

🆑
balance: Prybars are added back to the colony fabricator, but they are bulky items now.
/:cl:

### DIFF
--- a/modular_skyrat/modules/colony_fabricator/code/design_datums/tools.dm
+++ b/modular_skyrat/modules/colony_fabricator/code/design_datums/tools.dm
@@ -35,7 +35,7 @@
 /datum/design/colony_door_crowbar
 	name = "Prybar"
 	id = "colony_prybar"
-	// build_type = COLONY_FABRICATOR // Zubber Edit: Takes it out of the all-too-easy to acquire RCF
+	build_type = COLONY_FABRICATOR
 	build_path = /obj/item/crowbar/large/doorforcer
 	materials = list(
 		/datum/material/iron = SHEET_MATERIAL_AMOUNT * 1.75,
--- a/modular_skyrat/modules/colony_fabricator/code/tools/tools.dm
+++ b/modular_skyrat/modules/colony_fabricator/code/tools/tools.dm
@@ -101,7 +101,7 @@
 		survive <b>forcing doors open</b>."
 	icon = 'modular_skyrat/modules/colony_fabricator/icons/tools.dmi'
 	icon_state = "prybar"
-	toolspeed = 1.3
+	//toolspeed = 1.3 BUBBER REMOVAL - Why? Jaws are 0.7 themselves.
 	force_opens = TRUE
 	custom_materials = list(
 		/datum/material/iron = SHEET_MATERIAL_AMOUNT * 1.75,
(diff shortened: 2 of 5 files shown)


NargaInDisguise:
While I think making it bulky is a good balance change, the RCF continues to be an incredibly strong printer that cargo buys immediately on round start. I just don't think it should be able to print breaching tools period.

SonicGotNuked:
> While I think making it bulky is a good balance change, the RCF continues to be an incredibly strong printer that cargo buys immediately on round start. I just don't think it should be able to print breaching tools period.

I agree. I’m debating on if the prying needs to be removed completely or just given from the cargo shuttle at a 300-400 price tag

NargaInDisguise:
> I agree. I’m debating on if the prying needs to be removed completely or just given from the cargo shuttle at a 300-400 price tag

Make it only a restricted import that requires the cargo-side to order it, that way it can atleast be cancelled or you can figure out who bought it from the manifests. Theres also the potential of just, not letting cargo print/buy breaching tools for no reason, but its a can of worms no matter what way it goes.

SonicGotNuked:
> > I agree. I’m debating on if the prying needs to be removed completely or just given from the cargo shuttle at a 300-400 price tag
> 
> Make it only a restricted import that requires the cargo-side to order it, that way it can atleast be cancelled or you can figure out who bought it from the manifests. Theres also the potential of just, not letting cargo print/buy breaching tools for no reason, but its a can of worms no matter what way it goes.

It’s already a restricted import at 100 cr. But I’m not entirely comfortable of having easily accessible round start breaching tools in cargo.

Swiftfeather:
I'm still of the opinion it should be flat out removed from player acquisition.

NargaInDisguise:
I also think completely removing it is an entirely valid option. If you need a breaching tool as a job that requires one, jaws of life still exist anyways.

Superhats:
I'm also of the opinion that the RCF needs _less_ utility given to it. It's already a very, very popular purchase you see almost every round in places like cargo and the medbay.

StrangeWeirdKitten:
I've had some more thought to this and I will be removing the door forcer functionality entirely.

Superhats:
> I've had some more thought to this and I will be removing the door forcer functionality entirely.

I'm all for it, but what differentiates them from crowbars if that goes away? I don't think prybars do anything else special?

Majkl-J:
Seeing the current meta stabilize, it seems as if this isn't really needed? The need to actually ask cargo to buy it seems to work, making people either actually seek it out, or simply not carry it on them due to it being an actual hassle to obtain. We can also just role restrict it to engies if it shows up as a major issue again.

PR #1447, opened 2024-04-28
Absolute Pain (Doubles the SMES input / output

StrangeWeirdKitten:
I can not state how heavily fucked our powergrid is right now. So here's a bandaid solution.

## Changelog

:cl:
balance: doubles SMES input / output
/:cl:


destrucktoid:
Have you also made it so that APC arcing's power requirement is doubled?

StrangeWeirdKitten:
> Have you also made it so that APC arcing's power requirement is doubled?

It was already done by LT3

lessthnthree:
Increasing SMES input/output was something I was intending to propose upstream at TG but didn't manage before vacation.

modular_zubbers/code/game/machinery/smes.dm
```diff
@@ -0,0 +1,3 @@
+/obj/machinery/power/smes
```
lessthnthree, line 1: However, don't do it like that.

Instead, edit code\modules\power\smes.dm Line 67
```suggestion
/obj/machinery/power/smes/RefreshParts()
	SHOULD_CALL_PARENT(FALSE)
	var/power_coefficient = 1 // BUBBER EDIT CHANGE 
```

StrangeWeirdKitten: 👍

Redrover1760:
Could you double SMES starting power as well? It runs out so fast and its been agony for every single shift ever without an immediate competent engineering spawning roundstart and rushing the engine.

Majkl-J (approved):

PR #1451, opened 2024-04-29
Fixes bitrunner glitches not spawning if the reference list has a invalid ref

Arturlang:
…has a null ref, somehow

## About The Pull Request
See title

## Why It's Good For The Game
Most of the time, due to this, a glitch won't even spawn

## Proof Of Testing
Went ingame, tested it, spawned as a glitch

## Changelog

:cl:
fix: Bitrunner glitches should now spawn properly more often
/:cl:

### DIFF
--- a/code/modules/bitrunning/server/threats.dm
+++ b/code/modules/bitrunning/server/threats.dm
@@ -155,7 +155,7 @@
 /// Removes any invalid candidates from the list
 /obj/machinery/quantum_server/proc/validate_mutation_candidates()
 	for(var/datum/weakref/creature_ref as anything in mutation_candidate_refs)
-		var/mob/living/creature = creature_ref.resolve()
+		var/mob/living/creature = creature_ref?.resolve()
 		if(isnull(creature) || creature.mind)
 			mutation_candidate_refs.Remove(creature_ref)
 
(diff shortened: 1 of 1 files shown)


Arturlang:
Going to test if this issue is present on /TG/

Arturlang:
Can't reproduce on TG as they are missing a bitrunner map

Majkl-J (commented):
Is this bound to a specific map?

Arturlang:
> Is this bound to a specific map?

Issue was found on a specific map, but its a good fix honestly

Majkl-J:
> Issue was found on a specific map, but its a good fix honestly, as for the actual reason its caused uhhh, probably some mapping issue, though im wondering why it even attempts to iterate through a null object and how that even gets added

Yea I'd rather look for the source than to obfuscate it via nullcheck in this case, if you tell me which map it is I'll look into it with debug tools

code/modules/bitrunning/server/threats.dm
```diff
@@ -155,7 +155,7 @@
 /// Removes any invalid candidates from the list
 /obj/machinery/quantum_server/proc/validate_mutation_candidates()
 	for(var/datum/weakref/creature_ref as anything in mutation_candidate_refs)
-		var/mob/living/creature = creature_ref.resolve()
+		var/mob/living/creature = creature_ref?.resolve()
```
nevimer, line 158:
```suggestion
		var/mob/living/creature = creature_ref?.resolve() // BUBBER EDIT - REF CHECK for RUNTIME AVOIDANCE
```

nevimer (approved):

PR #1453, opened 2024-04-29
Click First Combat (Part 2 | The Baton Knockdown Wallshove Combo)

StrangeWeirdKitten:
Early Pull from upstream. It isn't merged at all on TG. I'll modularize it if they abandon the idea.

## About The Pull Request
This removes the ability for people that were recently batoned to be shoved into a wall. This doesn't affect shove stun, only the initial wall shove, although https://github.com/tgstation/tgstation/pull/81886 is doing things with that if you're interested.

This will basically only be affecting stunbaton/stunprod, because telebaton/contractor baton knock you down on the first hit anyways, meaning that shoving them while they are already on the ground will still stun them.

Video demonstration:
https://github.com/tgstation/tgstation/assets/53777086/aa00a00b-7d94-4d82-ba3a-5dd3d6d1f883

## Why It's Good For The Game
Currently the "I was batoned" trait is meant to prevent double batoning by the same person, and give some reaction time to the person being batoned (at least until they fall to the ground), but it has a massive loophole in that you can shove them into a wall to prevent them from being able to do anything in said reaction time. It's a lame strategy that goes against the baton's 2 hit to stun mechanic, so I think we should patch this up.

### DIFF
--- a/code/modules/mob/living/living_defense.dm
+++ b/code/modules/mob/living/living_defense.dm
@@ -707,15 +707,19 @@
 
 	if(shove_flags & SHOVE_CAN_HIT_SOMETHING)
 		//Don't hit people through windows, ok?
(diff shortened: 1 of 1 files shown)


StrangeWeirdKitten:
Upstream closed this PR and I need to throw on bubber commits later. If no one gets around to that, expect like a month before I'm capable of swinging around and doing that.

PR #1454, opened 2024-04-30
Large skyrat upstream part 4

Majkl-J:
## About The Pull Request

Skyrat finished up their repo, and should finally have stuff working. Time to do the same here.

## Why It's Good For The Game

This should include a shit ton of qol and bugfixes


projectkepler-ru:
me when the map power net arent connected because tg has the electrical storage that was connected but has no power and has to be disconnected to avoid it charging and discharging power

Majkl-J:
> me when the map power net arent connected because tg has the electrical storage that was connected but has no power and has to be disconnected to avoid it charging and discharging power

Have you considered it being the automapper

Majkl-J:
We gotta merge this then fix the map individually because I swear I pulled everything

PR #1455, opened 2024-04-30
Reduces nightmare's burn weakness

Arturlang:
## About The Pull Request
[IDED PR]
Currently nightmare has a 1.5 burn mod, and TG has recently buffed laser damage. This means that nightmares really suffer as a result.
This PR reduces the burn mod to 1.2, so that nightmares wont did in less than a second to lasers.

## Why It's Good For The Game
Lasers deal enough damage to crit a nightmare in 2 volleys if you have lasers in both hands, and the fire speed is fast enough that there's not enough time to react. Usually it's not so much issue for most other antags, as they can wear actual armor. However, nightmares cannot, and lasers can hit them even in the dark, as they generate light sources. That's not to mention the other high damage sources of burns that security has. (Laser shells, custom lasers)
Main issue here is a nightmare has barely enough chance to react to the damage, and is a fragile antag as they're very unlikely to revive if dead, and their main benefit to surviving is the self-healing in the dark, but cannot wear any armor.
This gives the nightmare a chance to react to burn damage, at the least
Note: Scattershot lasers take 2 shells to down a nightmare from point blank, which is likely given they are a melee only and fight in 1-tile maint corridors

## Proof Of Testing
Went ingame, shot at a nightmare, noticed they didn't fold near-instantly

## Changelog

:cl:
balance: Reduced nightmare's burn weakness to 1.2, from 1.5.
/:cl:


Arturlang:
Fun fact, the Lance variant of the microfusion lasers deals 50 damager per hit, which means the nightmare takes 70 damage per hit, which while slower to kill due to the fire rate, is still a critical in 2 laser hits.

PR #1464, opened 2024-05-02
Nerfs Changeling Absorb because I died to it once

BurgerLUA:
### For Changelings
Changeling absorptions:
- Take 15 seconds (previously 45 seconds).
- Now only require an aggressive grab (previously a full neck grab).

### For Changeling Victims
Changeling absorptions:
- No longer husk, but instead cause disfigurement which can be cured by cryoxadone.
- Now causes 20 brain damage, up to 100.
- Now only drains half your current blood volume, instead of 100% of it.
- Causes a severe piercing headwound.

## Why It's Good For The Game

Being absorbed by a changeling is generally not a fun experience. Not because you've died, but rather because your fate lies in medical's hands, and sometimes those hands are... not very skilled. Depending on the amount of medical players and their experience, it could take an absurd amount of time for them to make synthflesh, treat all your wounds, fill you back up with blood, fix/replace your organs. I've had rounds where this process has taken 5 minutes, and I've had rounds where this has taken 40 minutes.

Previously to cure changeling victims, doctors had to:
- Know how to make blood.
- Know how to fix bloodloss.
- Know how to make synthflesh.
- Know how much synthflesh to apply.
- Know how stasis beds work and how they can stop dehusking.

Now, to cure changeling victims, doctors have to:
- Know how to fix bloodloss.
- Know how to put people in cryotubes.
- Know how to make mannitol.
- Know how to fix a piercing wound.

Now with changes to husking and changes to how fast/easy it is to absorb, changelings should now feel encouraged to use it more.

## Proof Of Testing

Untested.

## Changelog

:cl: BurgerBB
balance: Changeling Absorption no longer husks, but instead causes disfigurement. This can be cured via the cryo tubes.
balance: Changeling Absorptions are now significantly faster, and now only require an aggressive grab to perform.
/:cl:

### DIFF
--- a/code/modules/antagonists/changeling/powers/absorb.dm
+++ b/code/modules/antagonists/changeling/powers/absorb.dm
@@ -1,6 +1,6 @@
 /datum/action/changeling/absorb_dna
 	name = "Absorb DNA"
-	desc = "Absorb the DNA of our victim. Requires us to strangle them."
+	desc = "Absorb the DNA of our victim. Requires a firm grip." //BUBBERSTATION CHANGE: NO NECKGRAB REQUIRED.
 	button_icon_state = "absorb_dna"
 	chemical_cost = 0
 	dna_cost = CHANGELING_POWER_INNATE
@@ -19,7 +19,7 @@
 	if(!owner.pulling || !iscarbon(owner.pulling))
 		owner.balloon_alert(owner, "needs grab!")
 		return
-	if(owner.grab_state <= GRAB_NECK)
+	if(owner.grab_state <= GRAB_AGGRESSIVE) //BUBBERSTATION CHANGE: NO NECKGRAB REQUIRED.
 		owner.balloon_alert(owner, "needs tighter grip!")
 		return
 
@@ -145,7 +145,7 @@
 				target.take_overall_damage(40)
 
 		SSblackbox.record_feedback("nested tally", "changeling_powers", 1, list("Absorb DNA", "[absorbing_iteration]"))
-		if(!do_after(owner, 15 SECONDS, target, hidden = TRUE))
+		if(!do_after(owner, 5 SECONDS, target, hidden = TRUE)) //BUBBERSATION CHANGE: 15 SECONDS TO 3 SECONDS.
 			owner.balloon_alert(owner, "interrupted!")
 			is_absorbing = FALSE
 			return FALSE
--- a/code/modules/antagonists/changeling/powers/absorb.dm
+++ b/code/modules/antagonists/changeling/powers/absorb.dm
@@ -145,7 +145,7 @@
 				target.take_overall_damage(40)
 
 		SSblackbox.record_feedback("nested tally", "changeling_powers", 1, list("Absorb DNA", "[absorbing_iteration]"))
-		if(!do_after(owner, 5 SECONDS, target, hidden = TRUE)) //BUBBERSATION CHANGE: 15 SECONDS TO 3 SECONDS.
+		if(!do_after(owner, 5 SECONDS, target, hidden = TRUE)) //BUBBERSATION CHANGE: 15 SECONDS TO 5 SECONDS.
 			owner.balloon_alert(owner, "interrupted!")
 			is_absorbing = FALSE
 			return FALSE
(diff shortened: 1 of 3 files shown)


Odairu:
it says nerfs, but I feel like its a strict buff? assuming you get the same benefits via DNA emporium

BurgerLUA:
> it says nerfs, but I feel like its a strict buff? assuming you get the same benefits via DNA emporium

The changes to how fast it now takes for medical to revive you is much more significant than how speedy it is to absorb.

Wolf-751:
Would in theory also stop a quick cling meta game since blood loss with husking usually signs of a changeling around.

Formerxboxgamer:
SAD exists for quick brain transplants, and apparently ling husk can actually be cured here by normal medical stuff too. Could we just tell med players to do better?

projectkepler-ru:
no

Majkl-J:
So I'll actually address the main points of the discussion as to "Why this isn't needed"

Firstly, whilst our medbay is generally less experienced than most playerbases', those who are are already fairly used to standard ling husking and how to fix it. Changing this also throws off the experienced medbay staff. We'd rather see more information as to how to fix these. 

The med scanner could be of more use this way, just mentioning application of synthflesh could help as there's also plenty of confusion with people from TG who arent used to ling dehusking being a thing.


Secondly, information as to how to make synthflesh is readily available on several wikis. There shouldn't be much trouble, especially once the machines are upgraded. Sure, it can slow medbay down if they aren't ready, but anyone can follow the few steps mentioned there.

As for last, a small note from myself: Whilst a common misconception, **Stasis doesn't stop expose_mob effects, like applying synthflesh**, it only stops reagent and organ processing

BurgerLUA:
The problem I have with changeling is that it is the only antagonist type that requires a significant amount of special treatment to deal with, compared to something like heretic or traitor. Combine that with the fact that it is a stealth/ambush role, making it one of the least fun antagonist types to fight against.

If I get killed with a traitor, worst case scenario I get my limb removed and decapitated, with some wounds. No chemistry is required here and medical can have me up quickly even if they don't know what they're doing.

If I get killed by a changeling, the most common scenario is that I'm disemboweled (the blade almost always does this), husked, and without any blood. This requires knowledge of blood duplication, how synthflesh is made and used, and an extensive amount of time re-adding organs, removing the brute damage, and removing wounds.

Again, no other antagonist type requires the use of chemistry to get you back into the game. Maybe heretic if you get a trauma, but you can still actually live with a trauma and there is plenty of trauma suppressants in the psychologist's locker to make this not needed.

Think of getting killed like getting a brig sentence. If a security officer arrests you for 5 minutes, that's not a big deal, like how if an antagonist causes you to remain dead for 5 minutes, that's not a big deal. If a security officer gives you a 15 minute sentence for no reason, that's a pretty serious deal like how if a security officer gave you a 15 minute sentence for no reason.

BurgerLUA:
Basically, tl;dr, don't close this PR immediately after a day. There is no reason to have an antagonist type that is so exceptionally dogshit when it comes to how much it relies on medical to be competent.

Majkl-J:
Adding information to `code/game/objects/items/devices/scanners/health_analyzer.dm#L164` could be nice

BurgerLUA:
> I think it's an issue that there isn't a communication aspect to players on how to fix major wounds like changeling absorbtion. There is means to get synthflesh without chemistry (from the vendors) and Rezadone is an alternative solution. But it's a lot rarer to be dealing with a lot of absorbed people. I think overall this game would benifit a lot on a little more ingame treatment guides then forcing people to rely on the wiki.

The amount of synthflesh required to dehusk and the amount actually provided in vendors (not to mention how expensive it is) is not reasonable.

StrangeWeirdKitten:
> > I think it's an issue that there isn't a communication aspect to players on how to fix major wounds like changeling absorbtion. There is means to get synthflesh without chemistry (from the vendors) and Rezadone is an alternative solution. But it's a lot rarer to be dealing with a lot of absorbed people. I think overall this game would benifit a lot on a little more ingame treatment guides then forcing people to rely on the wiki.
> 
> The amount of synthflesh required to dehusk and the amount actually provided in vendors (not to mention how expensive it is) is not reasonable.

I'm not against lowering the dehusk requirement.

RadiantFlash7:
Personally I don’t think it’s particularly hard to deal with changeling victims. The worst part is getting ahold of enough synthflesh. Blood is easy enough to fix, and the starting supply doesn’t run out that quickly.

Newer players might get confused, but you treat chsngeling husking the same way you do due to burns, no? Making it a bit clearer what to do for husking in the health analyzer is probably the right way to improve it, rather then flat out changing it to a new way of fixing it. Especially since it’s currently in line with fire-caused husking.

PR #1469, opened 2024-05-02
Curing Husks now only require 40u of Synthflesh, Buffs Changeling Absorption Speed

BurgerLUA:
## About The Pull Request

Changeling Absorb now only requires an aggressive grab (previously a neck grab).
Changeling Absorb now takes 15 seconds (instead of 45).
Changeling Absorb now only drains half your current blood count (as opposed to all of it).

Changeling Dehusking now requires just 40u of synthflesh (previously 200).

## Why It's Good For The Game

Changeling Absorb was slow. This was bad for the victim and bad for the changeling since it means it's less time playing.

Changeling Absorb was too detrimental to the victim. It required an absurd amount of Synthflesh to cure, as well as blood bags. 

One might argue that "It's easy to make a lot of synthflesh", so if that's the case, why does it matter if it's 200 vs 40?

## Changelog

:cl: BurgerBB
balance: Curing Husks now only require 40u of Synthflesh, Buffs Changeling Absorption Speed
/:cl:


BurgerLUA:
burger

StrangeWeirdKitten:
https://github.com/Skyrat-SS13/Skyrat-tg/pull/27568 it's honestly just a communication problem.

BurgerLUA:
SYNTHFLESH_LING_UNHUSK_AMOUNT is 200 on here, by the way. Even if you tell them that you need 200u of Synthflesh, they still need to make 200u of Synthflesh.

Majkl-J:
It was in fact a communication problem, as there seems to be much less confusion about revival now.

PR #1480, opened 2024-05-04
[FIX] Removes the ability to adjust the command stripper outfit

Majkl-J:
## About The Pull Request

No more needing to install CS:S mid ERP

## Changelog

:cl:
fix: Command members no longer have to install CS:S mid erp because they altclicked their stripper outfit
/:cl:


ReturnToZender:
So, how did you find out about this bug Waterpig? Huh? How'd you find it?

Majkl-J:
> So, how did you find out about this bug Waterpig? Huh? How'd you find it?

I plead the fifth

PR #1486, opened 2024-05-06
fixes linters and entombed

StrangeWeirdKitten:
entomed was broken apparently, so was linters. So let's fix that


projectkepler-ru:
so true

MDIPurple:
@StrangeWeirdKitten @projectkepler-ru  hey, this PR broke the Entombed dropdown list selection. It doesn't save anymore and makes a new list every time. Is there an another way to fix this?

PR #1496, opened 2024-05-10
Destroying a closet (including stealth implant boxes) will knockdown anyone inside it for 1 second.

BurgerLUA:
## About The Pull Request

Destroying a closet will knockdown anyone inside it for 1 second.

Reduces the cost of the stealth implant from 8 TC to 6 TC.

## Why It's Good For The Game

Stealth implant boxes are legitimately insane. Previously, they only had one use, but now they have unlimited use and you could basically just conjure up another box after getting shot. Not only that, but they negate the damage and effects of that shot, granting an absurd amount of survivability. 

Stealth implant boxes should be used for stealth and not for combat. If you still want to do box-fu, then get some stun mitigation.

## Proof Of Testing

It just werks.

## Changelog

:cl: BurgerBB
balance: Destroying a closet will knockdown anyone inside it for 1 second.
balance: Reduces the cost of the stealth implant from 8 TC to 6 TC.
/:cl:


IgiariValkyr:
Could just… make them one use again. Or take time to recharge if destroyed

Majkl-J:
The idea of giving it a cooldown (Can be registered with the `COMSIG_ATOM_BREAK` signal on the box) is probably a better idea, yea

BurgerLUA:
I think the stun is a way better implementation considering that the stealth box is meant for STEALTH, not combat.

Just adding a cooldown will still make it useful in combat.

Majkl-J:
> Just adding a cooldown will still make it useful in combat.

I'd rather a one minute cooldown upon being discovered and shot open than a cheap 2 second stun, which is more than enough to get significantly wounded due to being an easy target

rintherat:
2 seconds _might_ be a bit long, but this is a very good step towards balancing out how powerful they are.

Arturlang:
2 second hardstun is a pretty much guaranteed game over in a combat situation and waaay too much, either make it a knockdown or a cooldown

BurgerLUA:
The stun is now 1 second.

Majkl-J (changes requested):
Give it a cooldown instead.

BurgerLUA:
> Give it a cooldown instead.

Cooldown isn't the solution here in my serious opinion. The 1 second stun is fine since that can be mitigated by additional chems.

MajorBugReports:
Requiring more chem-fu to play antag is out of the question imo. This doesn't solve the issue of sec just firing down hallways with lethals either. Which I see far too often an "MRP" and "LRP" servers. Bola's will prevent someone from moving fast inside the box. Giving anything stuns when TG went through the process of making stuns harder to get is no bueno. There is a massive cooldown when the box is broken as well already. There is a more than long enough delay between uses due to that. It's a player skill issue here.

Majkl-J:
Turns out, there is a very visible 2 second window for getting back in the box, and a 10 second cooldown activated upon getting into the box, preventing you from boxing again. Meaning that you'd need to somehow escape, whilst slowed down, and giving a 2 second window showing exactly where you are. Aside from extra survivability it doesn't seem that viable, I see no real reason to add this nerf. Feel free to give me a reason for why we should choose this approach, and why the stealth implant even needs a nerf currently, and this will get reopened.

PR #1499, opened 2024-05-11
Disables the "Spy" antagonist type because I played a round of it once.

BurgerLUA:
## About The Pull Request

Disables the "Spy" antagonist type. It cannot role unless an admin forces it.

## Why It's Good For The Game

So what I can understand from what people told me, Spy is basically progtot. Progtot is not a good system because it is extremely abusable especially considering we have 2.5 hour rounds. When progtot occurs, the following happens:
- People pick xenobiology/miner/chemist/whatever role that is basically "You get stronger throughout the round."
- People do literally nothing for 1-2 hours. Basically greenshift.
- Security/Crew/Whatever get bumrushed with the broken mechanics xenobiology/miner/chemist/whatever has.
- Effect is a million times worse because they now have a lot of telecrystal gear on top of whatever nonsense they got in their 2 hours of farming.

image

## Proof Of Testing

If it compiles, it werks.

## Changelog

:cl: BurgerBB
del: Disables the "Spy" antagonist type. It cannot role unless an admin forces it.
/:cl:


AllCorruptingWyrm:
Spy is prog tot reborn, no one can convince me other wise. We have Bubber balance, not TG balance, we use TG as a root but we are our unique server, not what TG is, so with that said? Every encounter I've had with spy encourages a silent and stealthy playstyle and metafriending to get into departments and I've witnessed it repeatedly, firsthand. its not fun, its just miserable, please, play sec, you will understand.

R3dtail:
It's... neat.  Sure, the ones who do stealth it OR do something interesting with it are neat but most of the times people just  break in, get their shit and fuck off.  Or they're like "Um uh can I have x, y, or z" and then rebuild it.    The latter is an interaction yes, but idk, I want interesting interactions, not none.

Redrover1760:
Spy would be fun if the tc item aspect was just entirely removed.

Or if it was placed with dedicated spy items that you could buy for stealing shit that primarily help with only stealing shit instead of nukie gear and shit.

Swiftfeather:
This ain't it. Do something else to them instead of turning them off.

PR #1502, opened 2024-05-11
Removes air alarm heating and atmos forcer

Majkl-J:
## About The Pull Request

Remove not improve, this would require a rewriting of a bit of atmos handling to fix, and unironically kills every round with a spaced air alarm. Zirok has the exact details on why it is broken so ask them and the TG LINDA gods.

## Changelog

:cl:
fix: Fixes atmos dying
/:cl:


Cyprex:
If this isn't just TM only you should add a comment, but as TM this is gucci!

PR #1517, opened 2024-05-15
[TM ONLY] Removes the funny smithing bugs until upstream fixes them

Majkl-J:
Some people can't be trusted


KeRSedChaplain:
wait why not the bonuses for simplemobs

Majkl-J:
> wait why remove the bonuses for simplemobs, are those applying to carbons or something

Butchering interrupts and cancels the attack chain in the on_attack point. The bonuses are applied in pre_attack, and removed in post_attack. The chain goes `pre_attack -> on_attack -> attack -> post_attack` Meaning you can stack bonuses by butchering and cancelling the action repeatedly

Majkl-J:
> Rather then remove the simple mob buff, would it be possible to give it a timer? Like a 1 or 2 second boost so that subsequent attacks still have the boost or get stronger, but don't let you do stupid shit like semi permanent damage boosts.
> 
> I acknowledge that non capped damage boost is bad, but is it so horrible is a blacksmith can put time into a weapon and enhance the damage beyond the base line? If it's scaled properly and has a cap, say each perfect hit beyond the amount needed for base damage add 0.1 damage you'd need to work the shit out of it, but cap damage at 40. That isn't terribly game breaking and would take 200 swings and that's if you don't finish it on accident, plus if we add a sort of cooling system where if the metal has.... call it a 5 minute time where if it doesn't get reheated or hammered in then it will become unusable and you'll need to start again. So you have to constantly work it to get enhancement, but you can get it, and it's not world shattering

>until upstream fixes them
It's in the title. I'm not putting this much effort into something the original creator promised to fix

KeRSedChaplain:
> > wait why remove the bonuses for simplemobs, are those applying to carbons or something
> 
> Butchering interrupts and cancels the attack chain in the on_attack point. The bonuses are applied in pre_attack, and removed in post_attack. The chain goes `pre_attack -> on_attack -> attack_by -> post_attack` Meaning you can stack bonuses by butchering and cancelling the action repeatedly

incredible :kekw:

Majkl-J:
upstream did fix? I think?

PR #1518, opened 2024-05-16
Makes Ashwalker/Lizard names use a special name generator

BurgerLUA:
## About The Pull Request

Makes Ashwalker/Lizard names use a special generator, which either makes its own name based on consonant/vowel pairs, or with the infamous Verbs-The-Noun format.

Consonant/vowel generation will take up 80% of random names.
Verbs-The-Noun will take up 20% of random names.

Uses the ic/slur filter as well in case it generates Sus-Amongus or something hilarious.

## Why It's Good For The Game

This is basically a test PR that I will eventually(tm) make for /tg/. I just want to know what people think of it first.

The reason why I made this is because I made the mistake of using the standard /tg/ generator for one of my static characters and it's entirely possible for someone else to get the same name since there are limited pickings (1 in 26569 for a female lizardperson).

## Proof Of Testing

Using the new debug verb provided:

image

Example in the character creator:

image

## Changelog

:cl: BurgerBB
add: Makes Ashwalker/Lizard names use a special name generator
/:cl:


KeRSedChaplain:
cringe

code/__HELPERS/names.dm
```diff
@@ -1,8 +1,10 @@
+/* BUBBERSTATION CHANGE: REMOVED IN FAVOR OF ashwalker_names.dm
```
nevimer, line 1: I don't think you need to comment this out...

modular_zubbers/code/modules/ashwalker_names/ashwalker_names.dm
```diff
@@ -0,0 +1,80 @@
...
+	var/list/returning_data = ""
+	for(var/i=1,i<=50,i++)
+		returning_data += "[generate_ashwalker_name(i % 2)]\n"
+
+	user << browse(returning_data, "window=random_ashwalker_names")
+
+/obj/effect/mob_spawn/ghost_role/human/pirate/silverscale/generate_pirate_name(spawn_gender)
+	return generate_ashwalker_name(TRUE)
+
+/proc/lizard_name(gender)
```
nevimer, line 79: because this will just run anyways as it doesn't call parent, you retained the gender argument anyways so it should just work

modular_zubbers/code/modules/ashwalker_names/ashwalker_names.dm
```diff
@@ -0,0 +1,80 @@
...
+
+	if(!.)
+		return generate_ashwalker_name(TRUE)
+
+	return .
+
+ADMIN_VERB(generate_ashwalker_names, R_DEBUG, "Generate Ashwalker Names", "Generate a list of 50 random ashwalker names.", ADMIN_CATEGORY_DEBUG)
+
+	var/list/returning_data = ""
+	for(var/i=1,i<=50,i++)
```
nevimer, line 71:
```suggestion
	for(var/i in 1 to 50)
```

modular_zubbers/code/modules/ashwalker_names/ashwalker_names.dm
```diff
@@ -0,0 +1,80 @@
...
+	"v",
+	"b",
+	"n",
+	"m"
+))
+
+/proc/generate_ashwalker_word(min_chars=3,max_chars=5)
+	. = ""
+	var/is_vowel = !prob(80) //Atmos optimization
+	for(var/i=1,i<=rand(min_chars,max_chars),i++)
```
nevimer, line 50:
```suggestion
	for(var/i in 1 to rand(min_chars,max_chars))
```

modular_zubbers/code/modules/ashwalker_names/ashwalker_names.dm
```diff
@@ -0,0 +1,80 @@
...
+
+	. = list()
+
+	var/regex/regex_inclusion = regex(@'^(\w+)$',"i") //limits to things with one word
+
+	for(var/obj/object as anything in subtypesof(/obj))
+		var/object_name = initial(object.name)
+		if(!object_name || !regex_inclusion.Find(object_name))
+			continue
+		.[capitalize(lowertext(object_name))] = TRUE
```
nevimer, line 11: Someone's going to look at this in the future and go "wtf"

code/__HELPERS/names.dm
BurgerLUA: you do because you can't override global procs

modular_zubbers/code/modules/ashwalker_names/ashwalker_names.dm
BurgerLUA: The assoc list is so duplicate entries don't get added.

code/__HELPERS/names.dm
nevimer: > you do because you can't override global procs

yes you can just do

/myproc(args)

BurgerLUA: ye just confirmed it

PR #1526, opened 2024-05-19
Various Interdyne/Dauntless map and spawn ruin tweaks

tmyqlfpir:
## About The Pull Request

This PR adds an additional check for the space edition of Dauntless to the hack used to stop the bitrunner server from blurting out the cool down across supply radio comms.

It also tweaks a few things with Dauntless/Interdyne, such as giving them rad storm immunity and experi-scanners in the R&D labs.

Interdyne is now successfully placed on the lower Z levels thanks to the `ZTRAIT_ICE_RUINS_UNDERGROUND` flag.

## Why It's Good For The Game

This is fixing an existing bug.

## Proof Of Testing

I have tested this on my server and confirm it works as described as the PR.

## Changelog

:cl:
fix: Dauntless space ruin bitrunner server no longer announces cool down over NT supply comms
fix: Dauntless and Interdyne are now immune to rad storms
add: Added Syndie borg access cards to Interdyne
qol: Replaced the polarized windows in the Interdyne sauna area with blast door shutters
add: Added experi-scanners to Interdyne/Dauntless R&D
fix: Interdyne never spawns on top Z level
qol: Interdyne dorm rooms now have crowbars in closets
qol: Interdyne deck officer does not leave empty sleeper on ghost role spawn
/:cl:


code/modules/bitrunning/server/util.dm
```diff
@@ -7,9 +7,10 @@
 	//BUBBER ADDITION BEGIN - This is a HORRIBLE HACK to stop the radio from blurting out on cargo channel for dauntless prisoners
 	//If the dauntless map is removed, remove this hack!
 	//If the server is moved into a different area for dauntless, update this hack!
-	var/turf/curr = get_turf(radio)
+	var/area/curr = get_area(src)
 	var/area/dauntless_prison = /area/ruin/space/has_grav/bubbers/dauntless/sec/prison
-	if(get_area_name(curr, TRUE) == dauntless_prison.name)
+	var/area/dauntless_space_prison = /area/ruin/space/has_grav/bubbers/dauntless_space/sec/prison
+	if(istype(curr, dauntless_prison) || istype(curr, dauntless_space_prison))
```
Majkl-J, line 13: We're probably better off checking either against `!is_station_level(src.z)` (This might not work on icebox and moon (?)), or against `/area/ruin/space/has_grav/bubbers/dauntless_space` and `/area/ruin/space/has_grav/bubbers/dauntless` directly if the above doesn't work. No need to use vars either, just check against the type directly in this case

Also could be nice to still have the machine at least make the statement locally with the say proc

Majkl-J (changes requested):
Looks okay otherwise

Majkl-J (approved):

PR #1528, opened 2024-05-20
Nerfs No-Talk Borg Emags

ReturnToZender:
## About The Pull Request
I love Silicon

This PR makes it so that you need to complete a 2 second timer in order to emag a borg. The cover lock destruction is instant, like with an authorized ID, but the actual process of emagging takes 2 seconds.

## Why It's Good For The Game
This prevents traitors who walk up and immediately emag a borg while they're talking, without them having the opportunity to react or even speak. Annoying!

## Changelog

:cl: ReturnToZender
balance: Borgs now take 2 seconds to emag. Cover lock functionality unchanged.
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot_defense.dm
+++ b/code/modules/mob/living/silicon/robot/robot_defense.dm
@@ -341,36 +341,39 @@ GLOBAL_LIST_INIT(blacklisted_borg_hats, typecacheof(list( //Hats that don't real
 		balloon_alert(user, "expose the fires first!")
 		return FALSE
 
-	balloon_alert(user, "interface hacked")
-	emag_cooldown = world.time + 100
+	balloon_alert(user, "Hacking interface...") //BUBBER EDIT: CHANGES THIS DESCRIPTION
+	//BUBBER EDIT BEGIN: NERF NO-TALK EMAGS
+	if(do_after(user, 2 SECONDS))
+		emag_cooldown = world.time + 100
 
-	if(connected_ai && connected_ai.mind && connected_ai.mind.has_antag_datum(/datum/antagonist/malf_ai))
-		to_chat(src, span_danger("ALERT: Foreign software execution prevented."))
-		logevent("ALERT: Foreign software execution prevented.")
-		to_chat(connected_ai, span_danger("ALERT: Cyborg unit \[[src]\] successfully defended against subversion."))
-		log_silicon("EMAG: [key_name(user)] attempted to emag cyborg [key_name(src)], but they were slaved to traitor AI [connected_ai].")
-		return TRUE // emag succeeded, it was just counteracted
+		if(connected_ai && connected_ai.mind && connected_ai.mind.has_antag_datum(/datum/antagonist/malf_ai))
+			to_chat(src, span_danger("ALERT: Foreign software execution prevented."))
+			logevent("ALERT: Foreign software execution prevented.")
+			to_chat(connected_ai, span_danger("ALERT: Cyborg unit \[[src]\] successfully defended against subversion."))
+			log_silicon("EMAG: [key_name(user)] attempted to emag cyborg [key_name(src)], but they were slaved to traitor AI [connected_ai].")
+			return TRUE // emag succeeded, it was just counteracted
 
-	if(shell) //AI shells cannot be emagged, so we try to make it look like a standard reset. Smart players may see through this, however.
-		to_chat(user, span_danger("[src] is remotely controlled! Your emag attempt has triggered a system reset instead!"))
-		log_silicon("EMAG: [key_name(user)] attempted to emag an AI shell belonging to [key_name(src) ? key_name(src) : connected_ai]. The shell has been reset as a result.")
-		ResetModel()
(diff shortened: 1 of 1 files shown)


NargaInDisguise:
THANK FUCKING GOD WE'RE SO FUCKING BACK!!!!!!

projectkepler-ru:
terrible

Erol509:
I wonder why :)

ReturnToZender:
I have very few silicon hours, this is the opposite of a salt PR. I can just use my eyes and tell it's no good as a mechanic.

It's just apparent that being emagged instantly with zero counterplay isn't good for anyone. See #1453 for why we don't like this.
(However I will be accepting borg sloppy in the future for this one)

Erol509:
> I have very few silicon hours, and none recently-- this is the opposite of a salt PR. I can just use my eyes and tell it's no good as a mechanic.
> 
> It's just apparent that being emagged instantly with zero counterplay isn't good for anyone. See #1453 for why we don't like this. (However I will be accepting borg sloppy in the future for this one)

I mean its a very good PR. At least it will stop no brain tripple click, and moments where typebait occurs, which itself is hard to prove and find proof, without recording each round.

ReturnToZender:
> I mean its a very good PR. At least it will stop no brain tripple click, and moments where typebait occurs, which itself is hard to prove and find proof, without recording each round.

:)

StrangeWeirdKitten:
👍 add a 3-5 sec doafter() timer to direct law board uploads.

projectkepler-ru (approved):
Final thoughts on this

Yeah sure

code/modules/mob/living/silicon/robot/robot_defense.dm
```diff
@@ -341,36 +341,39 @@ GLOBAL_LIST_INIT(blacklisted_borg_hats, typecacheof(list( //Hats that don't real
...
+			logevent("ALERT: Foreign software execution prevented.")
+			to_chat(connected_ai, span_danger("ALERT: Cyborg unit \[[src]\] successfully defended against subversion."))
+			log_silicon("EMAG: [key_name(user)] attempted to emag cyborg [key_name(src)], but they were slaved to traitor AI [connected_ai].")
+			return TRUE // emag succeeded, it was just counteracted
+
+		if(shell) //AI shells cannot be emagged, so we try to make it look like a standard reset. Smart players may see through this, however.
+			to_chat(user, span_danger("[src] is remotely controlled! Your emag attempt has triggered a system reset instead!"))
+			log_silicon("EMAG: [key_name(user)] attempted to emag an AI shell belonging to [key_name(src) ? key_name(src) : connected_ai]. The shell has been reset as a result.")
+			ResetModel()
+			return TRUE
```
Majkl-J, line 360: You will make my future time with conflicts much easier by using a 
```
if(!do_after(user, 2 SECONDS))
	return
```
and preventing the extra indent

Majkl-J (changes requested):

code/modules/mob/living/silicon/robot/robot_defense.dm
ReturnToZender: Where do you want this, exactly? I'll do it soon

Majkl-J (approved):

Bombermansam:
I think this is a good idea. I've found it irritating when I get "driveby" emagged while I am in the middle of talking. This change would give a chance for the borg to escape being subverted and increase the risk of getting caught for the antag in exchange for getting a powerful helper. It would also create more interesting rp/mechanical scenarios for antagonists in which they would have to engineer a situation to get a borg alone and use another tool(flash, emp, etc) to disable the borg to make them easier to emag. This PR would increase interactions between players, increase mechanical interactions and could possibly provoke new situations for the playerbase to react to. Overall, I think this PR is a good change to make for an rp server.

Strazyplus:
This just prevents cheese, I tried to think of how this could severely negatively effect a antag, I just don't see how it would. [projectkepler-ru](https://github.com/projectkepler-ru) Calling this terrible without context just makes me assume they are being a troll.

I have made a suggestion for the mechanic to not end timer if cyborg moves, but when cyborg moves out of range the of the emagger. I feel this will allow a risky tactic of the antag to mount the cyborg then emag to avoid cyborg walking away. but leaves them vulnerable to *spin.

Frost155:
Hai! I do enjoy the changes, I defiantly felt there was a need to have this done, as the overall "drive-by" moments, from an assistant in medbay was quite troublesome experience, If not a bit funny, but it definitely took away from the overall experience as a Borg player.. with my 740+ hours of play time as a borg, is properly a good idea i speak up about my own experience with the suggestion and what we can do with it

 I find the 2 seconds maybe to be a bit too much personally, as we can see by the time it's implemented in, almost no one goes for a emagg anymore to subvert a borg, the risk vs reward for emagging a borg has always been a bit up to whoever "the borg is" on top of everything.. So it almost never happens anymore, unless we got a ninja, that got a emp ready.
 
 I could see a timing with 0.5, maybe 1 second? could be a better fix for it, as i know ReturnToZender have suggest if not in this thread/pr but possibly a different one? 
 
 And over all ReturnToZender from what i seen and their suggestions, trust them with what they have taken to the table
 about the changes to the emag timer

PR #1539, opened 2024-05-22
[Ready] You only get 1 corrupt organ when sacrificed (and removes dogshit organs from the corruption pool) instead of 2 to 4.

BurgerLUA:
## About The Pull Request

You only get 1 corrupt organ when sacrificed, instead of 2 to 4.

Removes some corruption options because they are extremely annoying / don't even work properly.

## Why It's Good For The Game

Why does /tg/ insist on making heretic one of the worst antagonist types to recover from.

Just because the medical system is dogshit and treatment results in near-instant recoveries does not mean you should add features that forces you to replace 2-4 of your internal organs (usually in separate locations) so there are consequences.

I hate /tg/ I hate /tg/ I hate /tg/

## Proof Of Testing

Untested.

## Changelog

:cl: BurgerBB
balance: You only get 1 corrupt organ when sacrificed, instead of 2 to 4.
qol: Removes some corruption options because they are extremely annoying / don't even work properly.
/:cl:


BurgerLUA:
I have a feeling someone is going to come in and say "Just go to medical to get them fixed." which

- Makes so sense because you're not supposed to know that you were sacrificed (It literally tells you this in game that you forgot everything).
- Bubberstation Metaprotections are a see-saw right now so that guessing "oh you're vomitting worms? must be an appendix issue" is a little out there.

Bonus:
- Not compatible with Skyrat's extra species. It is only compatible with plasmamen (in terms of snowflake organs) so any other snowflake species with special organs will get the shaft.

projectkepler-ru:
> I have a feeling someone is going to come in and say "Just go to medical to get them fixed." which
> 
> * Makes so sense because you're not supposed to know that you were sacrificed (It literally tells you this in game that you forgot everything).
> * Bubberstation Metaprotections are a see-saw right now so that guessing "oh you're vomitting worms? must be an appendix issue" is a little out there.
> 
> Bonus:
> 
> * Not compatible with Skyrat's extra species. It is only compatible with plasmamen (in terms of snowflake organs) so any other snowflake species with special organs will get the shaft.

I concur
it fucking suck when medical shove cybernetic organ into me and I'm effectively unable to play miner or shit like that

NargaInDisguise:
What about the permanent curse that you cannot fix in any way that makes you lose a random organ everytime you attack or get attacked by a heretic after being sacrificed?

BurgerLUA:
> What about the permanent curse that you cannot fix in any way that makes you lose a random organ everytime you attack or get attacked by a heretic after being sacrificed?

This is news to me. Does this actually happen?

NargaInDisguise:
> > What about the permanent curse that you cannot fix in any way that makes you lose a random organ everytime you attack or get attacked by a heretic after being sacrificed?
> 
> This is news to me. Does this actually happen?

From the original TG PR:

"Gives you an unremovable "curse" debuff which will cause you to immediately cough a random organ out of your chest if you attack or are attack by the person who sacrificed you, if they are wearing a Focus at the time."

Majkl-J (changes requested):
As far as I can see, the curse is a thing
`/datum/status_effect/heretic_curse`

I think giving it like a 10-20 minute timer is better than straight up removing it. (Also I was gonna do that when upstreaming but forgot)

So yea, give it a timer and I'd say this is fine

Majkl-J:
Also, fuck it, I am gonna go to tg and make the cursed organ code a bloody component so it isn't such an ass to maintain with extra species. All for merging this until then though

NargaInDisguise:
> As far as I can see, the curse is a thing `/datum/status_effect/heretic_curse`
> 
> I think giving it like a 10-20 minute timer is better than straight up removing it. (Also I was gonna do that when upstreaming but forgot)
> 
> So yea, give it a timer and I'd say this is fine

Yeah 15 minutes sounds like a much better compromise to keep you away than to just completely remove you from ever defending yourself against heretics for 3 hours.

BurgerLUA:
make it in another PR maybe. this Pr doesn't touch that code

Majkl-J:
> How about you remove this entire feature instead of just nerfing it This change isn't really compatible with us so why should we have it?

It is good for preventing people from letting themselves be sacrificed. If you have a replacement idea, then we can remove all the cursed organs. If not, these will have to stay

MajorBugReports:
It was added to prevent valid hunters from chasing heretics after being sacc'ed. I think its pretty obvious why it exists. But this isn't TG. 1 organ is enough and the curse having a respectable timer is more than fair.

Arturlang:
Gotta love just removing most of the organ variety because you don't like it, thanks burger. Even with the plasma organ, as long as you're not in a tiny room, scrubbers do a pretty good of scrubbing it out, and doesn't stasis just stop life looping so organs won't even proc? Please don't remove all the corrupted organ types.

BurgerLUA:
Ready to merge, linters failing because of storyteller.

PR #1541, opened 2024-05-22
Moonstation Mega Multi-Z Update

BurgerLUA:
image

## About The Pull Request

Destroying floor tiles on the Moonstation upper level can lead to a chasm that drops you down to the middle cave level. Note that you can't get dropped down to lavaland.
The middle cave level is now breathable air; the same as the surface.
Several areas were tweaked and some areas were added to include additional ladders that lead to the surface.
The prison is now "harder" to escape, with an already-existing chasm outside the windows.

## Why It's Good For The Game

It might actually not be because it means people might die to the mobs under there, and then get upset. Lets find out.

Regardless it looks pretty cool.

## Proof Of Testing

See screenshots.

## Ladder Routes

image

Routes marked with * require additional access.

## Changelog

:cl: BurgerBB
add: Destroying floor tiles on the Moonstation upper level can lead to a chasm that drops you down to the middle cave level. Note that you can't get dropped down to lavaland.
/:cl:


ReturnToZender:
> - Error parsing _maps/map_files/moonstation/moonstation.dmm (line 39407): Unknown constant type: s

Woe, linters fail be upon ye

Wolf-751:
Finally a use for my grappling hook.

PR #1545, opened 2024-05-23
Public Logging Update

nevimer:
Use Regex to be smarter.

## About The Pull Request

Public logging finds ckeys via regex to remove them from lines now. Not used for player say, at the moment. Not needed there either, as ckeys are only used by key_name() and thats passed down as a string. rather than just edit every instance of the proc, ill just regex them out by adding identifiers into the ckey blocks in logs and snip them that way. 

## Why It's Good For The Game

Public logs are little more private. We want them only to have characters, no keys. 

## Changelog

:cl:
refactor: public logging uses regex for cases of mechanics now. 
/:cl:


projectkepler-ru:
Any update on this?

StrangeWeirdKitten:
@nevimer please resolve the conflicts.

StrangeWeirdKitten:
This has been TM'd for like half a year and obviously we're not going to remove it. I know it might still have some schuffed issues with it, but it doesn't matter at this point. Future PRs can address that lol.

PR #1551, opened 2024-05-23
Adds prefloading to bitrunning

Majkl-J:
## About The Pull Request

Adds a new disky to bitrunning, which allows you to load any of your prefslots in, and then spawn in the domain as that character.

Also, I asked the admins about this, believable character rules wont apply inside bitrunning domains. So you can bring your meme chars there as well.

## Why It's Good For The Game

People keep bullying me to add sex to bitrunning and who am I to say no

## Proof Of Testing

It seemed to work without deleting my prefs

## Changelog

:cl:
add: Adds prefloading to bitrunning
/:cl:


Hislyna:
LETS FUCKING GOOO

Swiftfeather:
I disapprove because these kids are 'bitrunning' (video games) instead of working

Hislyna:
Bug Found - It removes char advert + headshot

Hislyna:
Okay so rereporting the bug from before. Not that it matters but Char Advert and Headshot are always what your Char Pref is set to, not what the disk is set to. That then will copy over into the station when you come out if you for some reason dont have your main char selected at the time.

pixelkitty286:
> Okay so rereporting the bug from before. Not that it matters but Char Advert and Headshot are always what your Char Pref is set to, not what the disk is set to. That then will copy over into the station when you come out if you for some reason dont have your main char selected at the time.

This is not a bug related to this PR when leaving the pod try to switch back to your character's slot because character flavor text and prefs seem to update real time

Majkl-J:
So, ethereals are now safe to use as avatars!

Cyprex (approved):
lgtm

code/modules/bitrunning/objects/netpod.dm
```diff
@@ -315,7 +315,13 @@
 		if(isnull(wayout))
 			balloon_alert(neo, "out of bandwidth!")
 			return
-		current_avatar = server.generate_avatar(wayout, netsuit)
+		// BUBBER EDIT BEGIN - PREFS!
+		var/datum/preferences/pref
+		for(var/obj/item/bitrunning_disk/prefs/prefdisk in neo.get_contents())
+			pref = prefdisk.loaded_preference
+			current_avatar = server.generate_avatar(wayout, netsuit, pref, include_loadout = prefdisk.include_loadout)  // Added the prefs argument
+			break
```
Cyprex, line 323: So, uh, what if there is no prefdisk? Is that a valid state?

Cyprex: Also the way this is written it may call generate_avatar multiple times, is that a good idea?

code/modules/bitrunning/server/obj_generation.dm
```diff
@@ -36,9 +36,14 @@
...
 
 /// Generates a new avatar for the bitrunner.
-/obj/machinery/quantum_server/proc/generate_avatar(obj/structure/hololadder/wayout, datum/outfit/netsuit)
+/obj/machinery/quantum_server/proc/generate_avatar(obj/structure/hololadder/wayout, datum/outfit/netsuit, datum/preferences/prefs, include_loadout = FALSE) // BUBBER EDIT - Prefs argument
 	var/mob/living/carbon/human/avatar = new(wayout.loc)
 
+	// BUBBER EDIT BEGIN - PREFS!
+	if(!isnull(prefs))
+		prefs.safe_transfer_prefs_to(avatar)
+	ADD_TRAIT(avatar, TRAIT_CANNOT_CRYSTALIZE, "Bitrunning") // Stops the funny ethereal bug
```
Cyprex, line 45: This is never removed... and I think that's fine because it only applies to avatar that is temporary anyways?

Majkl-J: Yup, it's only on the avatar, so it's fine

code/modules/bitrunning/objects/netpod.dm
Majkl-J: > So, uh, what if there is no prefdisk? Is that a valid state?

Shit

> Also the way this is written it may call generate_avatar multiple times, is that a good idea?

see: break

Majkl-J: Anyhow, fixed

modular_zubbers/code/modules/bitrunning/disks.dm
```diff
@@ -0,0 +1,50 @@
...
+
+/obj/item/bitrunning_disk/prefs/CtrlClick(mob/user)
+	. = ..()
+	include_loadout = !include_loadout // We just switch this around. Elegant!
+	balloon_alert(user, include_loadout ? "Loadout enabled" : "Loadout disabled")
+
+/obj/item/bitrunning_disk/prefs/attack_self(mob/user, modifiers)
+	. = ..()
+
+	var/list/prefdata_names = user.client.prefs?.create_character_profiles()
```
Cyprex, line 29:
```suggestion
	var/list/prefdata_names = user.client.prefs?.create_character_profiles()
	if(!prefdata_names)
		return
```
I don#t think we can do anything useful if prefs is null?

modular_zubbers/code/modules/bitrunning/disks.dm
```diff
@@ -0,0 +1,50 @@
...
+	var/datum/preferences/loaded_preference
+
+	var/include_loadout = FALSE
+
+/obj/item/bitrunning_disk/prefs/examine(mob/user)
+	. = ..()
+	if(!isnull(loaded_preference))
+		var/name = loaded_preference.read_preference(/datum/preference/name/real_name)
+		. += "It currently has the character [name] loaded, with loadouts [(include_loadout ? "enabled" : "disabled")]"
+		. += span_notice("Ctrl-Click to change loadout loading")
```
Cyprex, line 19:
```suggestion
		. += span_notice("Ctrl-Click to change loadout loading")
	else
		. += Is empty.
		. += span_notice("Use the disk on yourself to load a character")
```
Wording on that notice I wrote seems a little awkward

modular_zubbers/code/modules/bitrunning/disks.dm
```diff
@@ -0,0 +1,50 @@
...
+	. = ..()
+	include_loadout = !include_loadout // We just switch this around. Elegant!
+	balloon_alert(user, include_loadout ? "Loadout enabled" : "Loadout disabled")
+
+/obj/item/bitrunning_disk/prefs/attack_self(mob/user, modifiers)
+	. = ..()
+
+	var/list/prefdata_names = user.client.prefs?.create_character_profiles()
+
+	var/warning = tgui_alert(user, message = "This feature is experimental and messing with prefs. I suggest you save your prefs externally via the OOC tab. Proceed anyway?", title = "BEWARE", buttons = list("Yes", "No"))
```
Cyprex, line 31:
```suggestion
	var/response = tgui_alert(user, message = "This feature is experimental and messing with prefs. I suggest you save your prefs externally via the OOC tab. Proceed anyway?", title = "BEWARE", buttons = list("Yes", "No"))
```
And update subsequent lines too

modular_zubbers/code/modules/bitrunning/disks.dm
```diff
@@ -0,0 +1,50 @@
...
+	if(isnull(warning) || warning == "No")
+		return
+	var/choice = tgui_input_list(user, message = "Select a character",  title = "Character selection", items = prefdata_names)
+	if(isnull(choice) || !user.is_holding(src))
```
Cyprex, line 35: I don't know that is_holding is enough here, user could be incapacitated one way or another... and honestly I don't care. This is *really* inconsequential.

modular_zubbers/code/modules/bitrunning/disks.dm
Majkl-J: Disks already come with a use in hand warning by default

modular_zubbers/code/modules/bitrunning/disks.dm
```diff
@@ -0,0 +1,50 @@
...
+		return
+
+	loaded_preference = new(user.client)
+	loaded_preference.load_character(prefdata_names.Find(choice))
```
Cyprex, line 39: Uhhh is this(and preceding line) a good idea? Is this the sanest way to achieve this? Also, how will this get GC'ed? At the end of this proc? If so I'd explicitly qdel it to make that clear. I have no clue if that won't lead to problems though.

Cyprex: I presume this could be a little saner if we limited it to the character the person is already playing in the real world(the station), but that honestly kinda sucks, I like the flexibility of having all your characters available. Yuck.

modular_zubbers/code/modules/bitrunning/disks.dm
Majkl-J: Technically there isn't a case where this can return null anyhow, but sure

modular_zubbers/code/modules/bitrunning/disks.dm
Majkl-J: This is the actual holder of the player's character prefs. Only one character is ever held in these, so we need to do this cursed shit. The only time this needs to be qdelled is when all variables are cleared by Destroy, in which case it gets wiped cleanly due to not being held by anything else.

code/modules/bitrunning/objects/netpod.dm
```diff
@@ -315,7 +315,15 @@
...
 			balloon_alert(neo, "out of bandwidth!")
 			return
-		current_avatar = server.generate_avatar(wayout, netsuit)
+		// BUBBER EDIT BEGIN - PREFS!
+		var/datum/preferences/pref
+		var/load_loadout = FALSE
+		for(var/obj/item/bitrunning_disk/prefs/prefdisk in neo.get_contents())
+			load_loadout = prefdisk.include_loadout
+			pref = prefdisk.loaded_preference
+			break
```
Cyprex, line 324:
```suggestion
		var/obj/item/bitrunning_disk/prefs/prefdisk = locate() in neo.get_contents())
		if(prefdisk)
			load_loadout = prefdisk.include_loadout
			pref = prefdisk.loaded_preference
```

code/modules/bitrunning/server/obj_generation.dm
```diff
@@ -74,6 +79,7 @@
 			new /obj/item/flashlight,
 		)
 
+	if(include_loadout)	avatar.equip_outfit_and_loadout(new /datum/outfit(), prefs) // BUBBER EDIT - LOADOUTS
```
Cyprex, line 82:
```suggestion
	if(include_loadout)
		avatar.equip_outfit_and_loadout(new /datum/outfit(), prefs) // BUBBER EDIT - LOADOUTS
```
Does DM allow this sort of syntax? The inline if specifically. In any case I'd break this up over two lines, we do every if like that, no reason not to be consistent here imo.

modular_zubbers/code/modules/bitrunning/disks.dm
Cyprex: (There is the case of prefs being null- you used `prefs?.`..., which I believe returns null if prefs is null

PR #1563, opened 2024-05-26
Map voting shows total map counts for each map

BurgerLUA:
## About The Pull Request

Map voting shows total map counts for each map

## Why It's Good For The Game

Useful to me because it tells me what maps people actually like.

## Proof Of Testing

Untested :^)

## Changelog

:cl: BurgerBB
qol: Map voting shows total map counts for each map
/:cl:


Majkl-J (approved):
I wish there was a way to make it only show when the vote is over but alas.

PR #1568, opened 2024-05-27
[HEAT} Balance Heating Below This Point

BurgerLUA:
## About The Pull Request

I ran into a door and died because there was 2+ MW in the grid. 

## About The Pull Request (For Real)

Currently, electrocution damage is balanced around having 1MW in the grid, but since we're now allowing an upwards of 4MW in the grid now, I balanced electrocution damage to that 4MW value and applied other changes.

You now take 1 damage per 30J (previously 20J). This wasn't just an arbitrary x1.5 multiplier but the math worked out.
The maximum damage you can take from electrocution is now 125 (previously 195). Maximum health is 135 for reference.
The minimum damage you can take from shocks is 10 (previously 20).

## Why It's Good For The Game

As mentioned above, electrocution damage is currently in a bad state right now since it hasn't been balanced around the new power requirement changes.

This change is also good if we actually unironically want to add malf AI back into rotation because I don't want a round where the AI insta-kills everyone because they touched a door once.

## Proof Of Testing

Untested.

## Changelog

:cl: BurgerBB
balance: Nerfs electrocution damage because I died to it once
/:cl:

### DIFF
--- a/code/__DEFINES/combat.dm
+++ b/code/__DEFINES/combat.dm
@@ -392,7 +392,9 @@ GLOBAL_LIST_INIT(arm_zones, list(BODY_ZONE_L_ARM, BODY_ZONE_R_ARM))
 #define DEATHMATCH_PRE_PLAYING 1
 #define DEATHMATCH_PLAYING 2
 
+/* BUBBERSTATION CHANGE START: MOVED AND REFACTORED TO MODULAR FILE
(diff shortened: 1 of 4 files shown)


Majkl-J (commented):
I'll give this a test run locally, and merge accordingly.

Also this one is not to be closed, technically made before the freeze

PR #1586, opened 2024-05-31
Increases the chances of the Wizard Die of Fate event running.

BurgerLUA:
## About The Pull Request

Increases the chances of the Wizard Die of Fate running from a weight of 1 to a weight of 3.
For reference, Blob has a weight of 10.

## Why It's Good For The Game

I've personally never seen this event actually trigger because it's so rare.

## Proof Of Testing

Untested :)

## Changelog

:cl: BurgerBB
balance: Increases the chances of the Wizard Die of Fate event running.
/:cl:


NargaInDisguise:
Is there the possibility that it hasn't spawned at all because we're using storyteller?

Majkl-J:
> Is there the possibility that it hasn't spawned at all because we're using storyteller?

Nah it's in the storyteller panel with a 0.05% spawnchance

Majkl-J (changes requested):
You can now actually move it into the specific track folder on storyteller

Majkl-J (approved):

Majkl-J:
I forgot le merge button

PR #1590, opened 2024-06-02
[TEST] Removes the Ambient Buzz

nevimer:
## About The Pull Request

It's actually kind of grating, the other ambiences are so much better and this isn't covered by a pref

## Why It's Good For The Game

Ears

## Proof Of Testing

builds

## Changelog

:cl:
qol: ambience buzz is gone
/:cl:

### DIFF
--- a/code/game/area/areas.dm
+++ b/code/game/area/areas.dm
@@ -546,19 +546,19 @@ GLOBAL_LIST_EMPTY(teleportlocs)
 	var/mob/living/L = arrived
 	if(!L.ckey)
 		return
-
+/*
 	if(ambient_buzz != old_area.ambient_buzz)
-		L.refresh_looping_ambience()
+		L.refresh_looping_ambience() */ // BUBBER EDIT REMOVAL
 
 ///Tries to play looping ambience to the mobs.
 /mob/proc/refresh_looping_ambience()
-	var/area/my_area = get_area(src)
+/* 	var/area/my_area = get_area(src)
 
 	if(!(client?.prefs.read_preference(/datum/preference/toggle/sound_ship_ambience)) || !my_area.ambient_buzz)
 		SEND_SOUND(src, sound(null, repeat = 0, wait = 0, channel = CHANNEL_BUZZ))
 		return
 
-	SEND_SOUND(src, sound(my_area.ambient_buzz, repeat = 1, wait = 0, volume = my_area.ambient_buzz_vol, channel = CHANNEL_BUZZ))
+	SEND_SOUND(src, sound(my_area.ambient_buzz, repeat = 1, wait = 0, volume = my_area.ambient_buzz_vol, channel = CHANNEL_BUZZ)) */ // BUBBER EDIT REMOVAL
 
 /**
  * Called when an atom exits an area


YakumoChen:
Isn't this covered by a config?

nevimer:
> Isn't this covered by a pref though?

the pref covers shuttle sounds

pixelkitty286:
please make this into a pref

Arturlang:
There's no reason to make specific ambience's pref only, but this is incredibly grating, is there *any* reason to keep it, @pixelkitty286?

YakumoChen:
I've never found it grating, it adds immersion since you're existing more or less on a giant floating machine filled with florescent lights and other things that quietly hum.

captainfrost20:
The ambient buzz really adds to the atmosphere of the station. Removing it outright is sorta dumb. Like if its that big of an issue make a pref for it.

nevimer:
reaper_Qvxok64ZJz

I don't think you even know what you're arguing for. I'm not removing ambience effects. I am removing the low rumble that makes subwoofers shake constantly, and is just generally like listening to a 24 fan noise sound bit. There is no need to have it- it out-volumes many other effects that go on, and just is literally the definition of noise.

captainfrost20:
Shipambience.ogg is it not? I know exactly which ambience I'm talking about it. Having a general hum of the station isn't a bad thing when the other choice is literally just deafing silence. It is meant to be just station noise hence it being ambience, it adds a nice hum to the actual station itself giving the station the feeling of being in space. Genuinely this should just be a pref option, bundle a few of the other more "annoying" ambience ones into a grating category and be done with it. This shouldn't be a straight up removal based off your own personal preference

Patriot210:
> Shipambience.ogg is it not? I know exactly which ambience I'm talking about it. Having a general hum of the station isn't a bad thing when the other choice is literally just deafing silence. It is meant to be just station noise hence it being ambience, it adds a nice hum to the actual station itself giving the station the feeling of being in space. Genuinely this should just be a pref option, bundle a few of the other more "annoying" ambience ones into a grating category and be done with it. This shouldn't be a straight up removal based off your own personal preference

I agree with this tbh, I like the ambience, and it'd feel weird to remove the hum completely instead of adding an audio perf option.

Majkl-J (changes requested):
This needs to be a pref

Redrover1760:
Its surprising how much its needed.

StrangeWeirdKitten (changes requested):
This just looks like a TG bug then to warrent a flat out removal. I personally enjoy the white noise too, ngl.

StrangeWeirdKitten:
I found out the issue. Ambiance code requires area sensitivity. Living mobs are given area sensitivity via their mood. Ghosts do not have area sensitivity. It also does a !isliving early return. This should be a simple fix.

StrangeWeirdKitten:
Narrowing it down. It's because ghosts are above the area plane and they don't trigger shit.

StrangeWeirdKitten:
@nevimer this bug is extensive and deep. Disable ambiant sounds prefrence and use the stop all sounds command and it will remove the buzz while you're observing. The only thing I can think to do is giving a prefrence check on dead/observer initalize and it will correctly set what you need.

PR #1601, opened 2024-06-05
New hypospray type. More suitable for combat scenarios.

Moostard:
## About The Pull Request
This adds another hypospray type to play with.
It is unlocked after researching Advanced Biotech tech node and is a little bit more expensive to print than the standard hypo.
It can be printed in the medical fabricator and security fabricator.

Compared to the standard hypospray:
- Offers armor penetration
- Instant self injection

- Cannot be upgraded to fit the 100u vials.

Comes with a new reskin of the normal hypokit

## Why It's Good For The Game
Hyposprays are fun to use, but often you cannot help/save someone due to them wearing a modsuit or some sort of heavy armor. Or that ~second self injection makes it less appealing than the instant injection medipens and pills have.
This hypospray is for those that want an alternative for a quickfix or think the armor penetration is more important than double volume of vials.

## Proof Of Testing
Tested locally and everything worked fine for me.
Sprites of the kit and hypo are above.

Said sprites were made by Wolf751

## Changelog

:cl:
add: New armor piercing hypospray design can be researched now.
/:cl:


projectkepler-ru (changes requested):
Move them to a bubber modular folder

Moostard:
I will do it upstream.

aKromatopzia:
Maybe there's a reason why hypos can't be used in combat?

PR #1613, opened 2024-06-08
Significantly unnerfs the tinfoil hat because some reason someone nerfed it. Yes. Someone nerfed the tinfoil hat.

BurgerLUA:
## About The Pull Request

Wearing the tinfoil hat no longer gives you a phobia of conspiracy theories.
Tinfoil hat is no longer single use and will work indefinitely until removed.

## Why It's Good For The Game

Tinfoil hat used to be an amazing, funny, and balanced way to counter abductors. Unfortunately, over time the tinfoil hat has been nerfed to oblivion. This is probably the stupidest sentence I've ever had to write.

Tinfoil hat users were previously given a phobia of conspiracies, which made it so that every time you saw security/command, you'd basically shut down in fear and be unable to interact by saying security/command related words. This made no sense because Alex Jones still talks about gay frogs despite having a fear of gay frogs.

Semi-recently, tinfoil hats were only given six charges of anti-mind altering. This is a dumb change because the abductor 14 second stun baton can just be employed and you can just hit the person 6 times to remove it and allow them to finally sleep.

Abductors just need to remove the tinfoil hat in order to counter the tinfoil hat. This takes like 3 seconds. Literally walk up to them and stun them (the stun doesn't get canceled due to tinfoil hats).

## Proof Of Testing

Untested.

## Changelog

:cl: BurgerBB
balance: Wearing the tinfoil hat no longer gives you a phobia of conspiracy theories. Tinfoil hat is no longer single use and will work indefinitely until removed.
/:cl:


projectkepler-ru:
incredible

BurgerLUA:
> incredible

It really is. It pisses me off how insane abductors are and how we're apparently fine with abductor balance despite the extremely clear favoritism the code has for abductors.

code/modules/clothing/head/tinfoilhat.dm
```diff
@@ -6,7 +6,7 @@
 	armor_type = /datum/armor/costume_foilhat
 	equip_delay_other = 140
 	clothing_flags = ANTI_TINFOIL_MANEUVER
-	var/datum/brain_trauma/mild/phobia/conspiracies/paranoia
+	// var/datum/brain_trauma/mild/phobia/conspiracies/paranoia BUBBERSTATION CHANGE, REMOVES PARANOIA
```
nikothedude, line 9: the var and all associated edits except for the removal of the trauma addition dont need to be removed from what im seeing here because it always nullchecks the trauma

nikothedude: also test your prs

BurgerLUA: no

Arturlang:
Doesn't the tinfoil hat make you immune to the abductor baton? Wouldn't it be a direct and very effective counter then?

BurgerLUA:
> Doesn't the tinfoil hat make you immune to the abductor baton? Wouldn't it be a direct and very effective counter then? I think atleast it prevents sleeping, thus letting you call for help. (Which means abductors won't have enough time to beam you)

Only the sleep. Not the stun.

radarseas:
this PR, I do agree with.

also you can throw these aimed at heads to forcefully hat people, they can't remove it themselves, you can guess how this goes when revs/heads of staff/security weaponise this. (they don't, because nobody remembers this thing exists because of this phobia alone.)

Issue #1290, opened 2024-03-20
Experimental seeds lack icon


Majkl-J:
## Issue Summary

Using a plant DNA manipulator to create seeds creates invisible seeds

## Round ID:

Any

## Testmerges:

N/A

## Reproduction:

1. Use a plant DNA manipulator to create seeds

2. It creates invisible seeds

ReturnToZender:
This seems like an easy fix, I'll see if I can help any

projectkepler-ru:
> This seems like an easy fix, I'll see if I can help any

Did you do it?

PR #1614, opened 2024-06-08
Even more plant gene editor fixes (WIP)

shellspeed1:
## About The Pull Request

While working with the plant gene editor I noticed that were were having issues with it not actually improving for new parts. I also noticed that stat genes were not editable for whatever reason. So far I have the first issue fixed but I need to still figure out the second. Once finished this will fix both.

## Why It's Good For The Game

I like it when stuff works.

## Proof Of Testing

Can currently copy genes to disks at right values, cannot put them back in for stats.

## Changelog

:cl:
fix: Plant gene editor works again.
/:cl:


Majkl-J:
Bonus points if you manage to look into https://github.com/Bubberstation/Bubberstation/issues/1290, if it even still is valid

shellspeed1:
I probably won't be finishing this for another week or two as I have a CPA exam on monday than a trip. The suffering is REAL.

Majkl-J:
> I probably won't be finishing this for another week or two as I have a CPA exam on monday than a trip. The suffering is REAL.

take your time

shellspeed1:
Closing this as it looks like we just want to axe it for now.

PR #1620, opened 2024-06-09
[Hot Fix] WHY WERE THEY USING THE ONBJ FOR BOTH AAAA

theselfish:
WHY
WHO DONE THIS
WHY WERE THEY BOTH USING THE SAME DMI FOR ONMOB AND OBJ
## Changelog

:cl:
fix: Drill Hats ACTUALLY work.
/:cl:


Cyprex:
This doesn't remove any icons- were they completely missing originally?

theselfish:
> This doesn't remove any icons- were they completely missing originally?

Kinda.
At first, the objs were missing.
I fixed that, and accidentally removed the onmob since it was using the obj file for both.

PR #1623, opened 2024-06-10
Maybe fixes roundstart sometimes lagging

Maia-J:
See name


projectkepler-ru (commented):
oh that works? good to know

PR #1625, opened 2024-06-10
Reverts some nerfs to the bulldog

GoreDem:
## About The Pull Request
Removes the untoggleable burst fire and extremely long fire delay (the low fire delay was the thing that made it good) from the bulldog
Edit: Apparently the burst fire is toggleable, might add it back again to see how stupid the gun becomes with it

Also babby's first pr or second
I don't rmemeber

## Why It's Good For The Game
Bulldog is in a weird spot where it fires really slowly and has a burst on top of it so the burst accomplishes nothing, this was due to 2 prs nerfing/buffing it on SR/tg.

## Proof Of Testing
I have not tested this pr but it's like two lines

## Changelog

:cl:
add: buffed bulldog back to a usable state
/:cl:


AllCorruptingWyrm:
What are the stats compared to the bartender shotgun again?

GoreDem:
> What are the stats compared to the bartender shotgun again?
Doesn't matter but very much better

GoreDem:
> > What are the stats compared to the bartender shotgun again?
> 
> Doesn't matter but very much better

For reference: it has a double-decker magazine and a 1.2 damage MULTIPLIER, way slower firing speed and burst fire (this pr changes the latter two)

projectkepler-ru (changes requested):
do this upstream or do this modularly

GoreDem:
> do this upstream or do this modularly

How would I do it modularly? (Sorry babby's first pr)

projectkepler-ru:
> > do this upstream or do this modularly
> 
> How would I do it modularly? (Sorry, babby's first pr)

You can have the code path and the variable you want changed in a modular bubber folder

StrangeWeirdKitten:
Sure

PR #1626, opened 2024-06-10
[Ready] Exposed grilles start broken if they're connected to the powergrid (shocked grilles) because I saw someone get husked for lagging once.

BurgerLUA:
## About The Pull Request

Shocked grilles **spawned on map load** start broken. **This does not apply to grilles with windows on them.**

## Why It's Good For The Game

A dirty fix for maps that have shocked grilles randomly placed in maintenance. This is a catch-all solution that doesn't require map edits every time someone places a shocked grille in a location that shouldn't have it.

The problem with shocked grilles is that with recent power changes, touching one is guaranteed death. I think this is unfair to people with high ping.

This does not impact traitors who place shocked grilles in maint.

## Proof Of Testing

Draft because untested.

## Changelog

:cl: BurgerBB
balance: Shocked grilles spawned on map load start broken. This does not apply to grilles with windows on them.
/:cl:


Majkl-J (changes requested):
Invalid reasoning, your own changes to shock damage made it impossible for anyone to die, letalone get husked from grilles if they are conscious, they can only reliably kill you if you are in softcrit already.

Also you can't even get burn damage husked unless you are already dead
`if(((maxHealth - total_burn) < HEALTH_THRESHOLD_DEAD*2) && stat == DEAD )`
at `code/modules/mob/living/carbon/carbon.dm#L570`

Majkl-J (changes requested):
Also probably breaks non-station maps that use it as a mechanic, including the russian shuttle, and the deathmatch arena

BurgerLUA:
Linters are failing due to storyteller, otherwise this is good to merge.

theselfish:
Wouldn't this break pre-built ragecages on maps like Kilo and Delta?

projectkepler-ru (commented):
the reasoning doesn't seem very good? I don't think there's enough map where you'll die to electrified grille in maint due to lag for this to be warranting it

Majkl-J:
Conclusion is that this isn't really needed. Shocks can't even crit in one hit unless you are already injured, the reasoning is invalid, you yourself adjusted the formula for it to not kill you in one hit.

PR #1629, opened 2024-06-11
Disables Radiation Nebulas from occurring because something is seriously wrong with them.

BurgerLUA:
## About The Pull Request

Disables Radiation Nebulas from occurring because something is seriously wrong with them.

## Why It's Good For The Game

Something seems to be lagging the hell out of the server every time this trait occurs. Radiation storms tend to do it because they have to set the weather to every single station area, which causes intense lag.

Ignoring the lag, it's extremely painful to deal with and unfun that you have to go into maints every few minutes. Since our pop is usually high, medbay usually has to deal with 10+ people who got irritated because they didn't see the messages or the lag causes them to get fucked.

## Proof Of Testing

Untested. it should work.

## Changelog

:cl: BurgerBB
del: Disables Radiation Nebulas from occurring
/:cl:


IgiariValkyr:
I mean. I did say in round in OOC i was doing the storms on purpose on random chance because we were in a rad nebula. But the actual lag behind it is something that should be looked into, yes.

Majkl-J (changes requested):
Disable radstorms in general in the storyteller event overrides

StrangeWeirdKitten:
I'll test it out later.

nevimer:
is this still true?

BurgerLUA:
> is this still true?

Not sure tbh.

StrangeWeirdKitten:
Gameplay reason, the trait is fine and rather uncommon. I rarely see any sort of radiation storm. Lag reasons, I need to see an actual profiler (on the server) if this is the case. So if I see it, I'll look at it and try to get it fixed before outright removing it.

PR #1636, opened 2024-06-11
F3-LINE cyborg updates [Modularish]

pixelkitty286:
## About The Pull Request

Makes the sprite file resolutions from 64 x 32 to 32 x 32 for cat borgs

Makes 32 x 32 quadruped cybrogs into it's own borg type with it's own quadruped module of it's own.

Allows only station standard cat borgs to be hold able, non standard can not be held.

Adds code for cyborgs to indicate when they are being held.

Makes cyborg that are being held unable to interact items around them, they can still break free any time

This does not allow cyborgs with shape shifting to be picked up it only allows the REAL model to be held.

## Why It's Good For The Game

I've always wondered why can't I make cat borgs hold able and also make it so it don't have sweeping mechanical changes to the game so I came up with a way to allow them to be held disabling their abilities to interact with much wile being in a held state and being almost unbaggable.

Making the sprites 32 x 32 means they won't look so  offsetted when using things quite as often.

Cent Com don’t want you to know this but the the cat borgs on space station 13 are free you can take them home I have 458 cat borgs at home.

## Proof Of Testing

sc19jFbsn7

dreamseeker_mmNECEkkUe

## Changelog

:cl:
add: holding code for cyborgs
del: Old 32x32 F3-LINE sprites
image: 32x32 F3-LINE sprites
code: preventive flags to indicate when held
admin: Two robot traits added to the trait list
/:cl:


modular_zubbers/modules/borgs/code/robot_upgrade.dm
```diff
@@ -54,3 +54,39 @@
...
+
+/obj/item/borg/upgrade/squadrupedmodule/deactivate(mob/living/silicon/robot/borg, user = usr)
+	. = ..()
+	if(.)
+		return
+	borg.hasAffection = FALSE
+	for(var/obj/item/quadborg_tongue/quadtongue in borg.model.modules)
+		borg.model.remove_module(quadtongue, TRUE)
+	for(var/obj/item/quadborg_nose/quadnose in borg.model.modules)
+		borg.model.remove_module(quadnose, TRUE)
```
pixelkitty286, line 92: There is a reason I did this and that is the new method is not compatible with checking for traits this method is depreciated and should never be used else where.

modular_skyrat/modules/borgs/code/robot.dm
aKromatopzia, line 1: modifying skyrat code is iffy, do be careful with it and mark your changes for clarity

pixelkitty286: fixed thanks for the good catch

Redrover1760:
Could this apply to Roomba borgs too? They seem small enough.

pixelkitty286:
I LOVE WHEN FILE STRUCTURE CHANGES! :)))

PR #1649, opened 2024-06-16
Removes the absurd emote delay that Skyrat implemented ages ago as a result of someone doing a *flip *spin *clap once.

BurgerLUA:
## About The Pull Request

Removes the hardcoded 5 second emote delay for all emotes. Most emotes now have a cooldown of 0.8 seconds, with some being increased based on what /tg/ set.

## Why It's Good For The Game

Basically a long long time ago, people would *flip, *spin, *clap in the halls. This upset the host of skyrat,so the emote delay was increased to an absurd 15 seconds. After much criticism, it was reduced to a still absurd but not much 5 seconds.

I'm removing that because we're all adult enough not to spam emotes for no reason, also because /tg/ fixed *spin *clap *flip spam by adding conditions to it and chances for it to fail.

## Proof Of Testing

Untested :)

## Changelog

:cl: BurgerBB
del: Removes the hardcoded 5 second emote delay for all emotes. Most emotes now have a cooldown of 0.8 seconds, with some being increased based on what /tg/ set.
/:cl:


pixelkitty286:
Even though I disagree on things at times your cooking on this one.

Majkl-J:
Counterpoint:
image

But sure we can give it a shot

nevimer:
Only if combing up flips, claps, or spins, makes you fall flat into a longish hard stun.

BurgerLUA:
> Only if combing up flips, claps, or spins, makes you fall flat into a longish hard stun.

/tg/ has a way to prevent flip/spin spam

projectkepler-ru (commented):
I've no idea what other maintainer think of this

nevimer:
I will send this to the shadow realm if people abuse this

PR #1666, opened 2024-06-19
Clock out with your glock out: Plexagon Crew Self Serve

lessthnthree:
## About The Pull Request

Adds a new PDA app, Plexagon Self Serve. This app will be a crew member's centralised application for HR related functions.

- Allows crew to clock in/out wherever they are
- Instead of sending people's restricted gear to cryo, it's instead placed into a patent pending ~~SHAME BOX~~ lockbox.
- The lockbox is unlocked and items returned once they've clocked back in.

Future PRs I'd like to add things like paycheque and other crew related HR stuff.

## Why It's Good For The Game

- A serial observer got mad I didn't clock out before ERP once
- Pain point: Having to actually walk across station to the time clock
- Pain point: Losing your items on clock out, then you need the captain or HoP to get them back on clock in.

## Changelog

:cl: LT3
refactor: Crew HR functions in a new PDA app, Plexagon Self Serve
refactor: Clock in and out from your PDA
qol: Job items locked away while clocked out instead of being warped to cryo; returned on clock in
/:cl:


destrucktoid:
Main thing i disagree with is the ability to change your job title that can and WILL be abused by antagonists, it also takes away more from the HoP, a role that already has barely anything to actually do. Other than that, i see no issues.

lessthnthree:
> Main thing i disagree with is the ability to change your job title that can and WILL be abused by antagonists, it also takes away more from the HoP, a role that already has barely anything to actually do. Other than that, i see no issues.

I did not consider antags.

modular_zubbers/modules/plexagon_selfserve/code/plexagon_selfserve.dm
```diff
@@ -0,0 +1,241 @@
...
+	authenticated_user = "[authenticated_card.name]"
+	return TRUE
+
+/datum/computer_file/program/crew_self_serve/on_start(mob/living/user)
+	. = ..()
+	if(!.)
+		return FALSE
+	computer.crew_manifest_update = TRUE
+
+	if(!computer || !computer.computer_id_slot)
```
Cyprex, line 40: Two lines above this you set a var on computer, but here you check if computer is null- is it possible computer is null, ever?

modular_zubbers/modules/plexagon_selfserve/code/plexagon_selfserve.dm
```diff
@@ -0,0 +1,241 @@
...
+		to_chat(user, span_warning("Plexagon: Login failed, no ID card found!"))
+		playsound(computer, 'sound/machines/buzz-sigh.ogg', 50, FALSE)
+		return FALSE
+	else if(authenticate(user, computer.computer_id_slot))
```
Cyprex, line 44: I'd pull ```

	authenticated_card = id_card
	authenticated_user = "[authenticated_card.name]"
	return TRUE```
into an unconditional else branch here & remove the authenticate proc. All it does is check if the id_card is null, but you already check !computer.computer_id_slot above.

modular_zubbers/modules/plexagon_selfserve/code/plexagon_selfserve.dm
```diff
@@ -0,0 +1,241 @@
...
+		return FALSE
+
+	var/datum/component/off_duty_timer/id_component = authenticated_card.GetComponent(/datum/component/off_duty_timer)
+	if(!id_component)
+		return FALSE
+
+	if(id_component.hop_locked)
+		return TRUE
+
+	return FALSE
```
Cyprex, line 144: I don't see where this proc checks if the timer has expired?

modular_zubbers/modules/plexagon_selfserve/code/plexagon_selfserve.dm
```diff
@@ -0,0 +1,241 @@
...
+
+/// Clocks out the currently inserted ID Card
+/datum/computer_file/program/crew_self_serve/proc/clock_out()
+	if(!authenticated_card)
+		return FALSE
+
+	var/datum/component/off_duty_timer/timer_component = authenticated_card.AddComponent(/datum/component/off_duty_timer, TIMECLOCK_COOLDOWN)
+	if(important_job_check())
+		timer_component.hop_locked = TRUE
+		log_admin("[authenticated_card.registered_name] clocked out as a head of staff and/or command")
```
Cyprex, line 57: Or sec. Also, it needs to be communicated to the user that they won't be able to clock back in.

Honestly, you should pull the job from the trim/assignment. That makes these logs more useful and prevents issues where someone later modifies important_job_check() without adjusting this line(It's really bad if logs say something incorrect). Also, message_admins too, I think this just silently writes to log files.

lessthnthree: This is the original time clock code, I will have a look.

PR #456, opened 2023-08-15
Storyteller port from horizon [DNM]

nevimer:
## About The Pull Request
Ports storyteller from horizon, who deleted their repo. 

### How it works

_Written by Majkl-J_

Basically, the storyteller runs several "Tracks" that slowly fill up with points. Upon reaching their set limit, an event is chosen and spawned, and the storyteller takes a somewhat-random amount of points from that track.

image

In this example, the moderate track will soon spawn an event.

Each track has its own sets of events it can spawn, sorted by intensity (Roleset is midround and roundstart antags). Every event also has little tags that the storyteller can be made to prioritize.

As with dynamic, events have a certain weight that determines the chance they roll. Prioritized events get a multiplier to this weight. 

image

In this example, we can see some of the tags. Some storytellers can be made to increase/decrease the weight of these events.

#### Extra stuff it does

Keeps track of med, engineering, and sec players. Whilst this still has no actual usage, it is good to know it is capable of this.

## Why It's Good For The Game

Storyteller is an alternate event controller, in my opinion, superior to dynamic.

## Changelog

:cl: Nevimer, Majkl-J, BurgerBB, Azarak
add: Ported the storyteller event system from horizon
/:cl:

PR #1671, opened 2024-06-20
Fixes a reversion that causes the bureaucratic error event to have a chance to close all job slots

Shadow-Quill:
## About The Pull Request

Reverts a reversion in #456 which re-enabled the bureaucratic error event having a 33% chance to choose a random overflow job and close all job slots for all other jobs.

## Why It's Good For The Game

A HoP can only open non-command job slots at a rate of one slot every thirty seconds.
This can leave people only able to latejoin as assistants.
The job slot randomization that skyrat added is annoying, but doesn't prevent new people from joining actual jobs.

## Proof Of Testing

N/A, just re-commented out something that was un-commented out.

## Changelog

:cl:
del: Fixes a reversion which re-enabled a chance to close all job slots except one when the bureaucratic error event triggers.
/:cl:

### DIFF
--- a/code/modules/events/bureaucratic_error.dm
+++ b/code/modules/events/bureaucratic_error.dm
@@ -14,7 +14,7 @@
 
 /datum/round_event/bureaucratic_error/start()
 	var/list/jobs = SSjob.get_valid_overflow_jobs()
-	// BUBBER EDIT BEGIN /* SKYRAT EDIT REMOVAL START
+	/* SKYRAT EDIT REMOVAL START
 	if(prob(33)) // Only allows latejoining as a single role.
 		var/datum/job/overflow = pick_n_take(jobs)
 		overflow.spawn_positions = -1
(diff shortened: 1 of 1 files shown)


shellspeed1:
Hot take, can we rework this so it OPENS a job slot for all jobs?

Cyprex:
This event is bugged, fix the bug instead of disabling the event. I'll leave this open so that it can be testmerged until the event is fixed.

Shadow-Quill:
> This event is bugged

How is it bugged?

Cyprex:
Okay, I had a look at the code and it's not bugged. There's a 66% chance for it to randomize all job slots, and a 33% chance for it to make a random job unlimited and close all other slots. So let's just remove the 3 lines that close all other slots and we're good.

StrangeWeirdKitten:
@Shadow-Quill status?

Majkl-J:
staled out

PR #1684, opened 2024-06-25
Nerfs bloodsucker blood gain from the blood reagent

Arturlang:
## About The Pull Request
Bloodsuckers can only reach up to 560 blood using the blood reagent, bloodsuckers above level 4 will get further 30% less blood from each sip, and be disgusted as well.

## Why It's Good For The Game
Nobody likes bloodsuckers bypassing the blood drinking mechanics by simply eating 50u duplicated blood pills. If this solution is not too great, a alternative could be to make duplicated blood simply unusable for bloodsuckers, but I think this is alright compromise

## Proof Of Testing
went ingame, tested if it worked.

## Changelog

:cl:
balance: Bloodsuckers can only reach 560 max blood via drinking blood bags or the blood reagent, and bloodsuckers above level 4 will get 30% less blood and be disgusted by it.
/:cl:


Arturlang:
I would be fine increasing the required level before the detriments occur, however.

nevimer:
How does this affect bloodsuckers who drown themselves in blood reagent turfs?

Arturlang:
> How does this affect bloodsuckers who drown themselves in blood reagent turfs?

I can bloodsuckers only gain blood from ingest and inject only

plsleavemealon:
What if it was simply made so that it gave increasingly less effect above a threshold?

So if a bloodsucker were to drink "duplicated blood" past 560 blood, they would get progressively less blood absorbed in their body from it

Arturlang:
> What if it was simply made so that it gave increasingly less effect above a threshold?
> 
> So if a bloodsucker were to drink "duplicated blood" past 560 blood, they would get progressively less blood absorbed in their body from it
> 
> EDIT: This is also effecting normal hemophages too, is that intentional?
> 
> EDIT 2: Nevermind, normal hemophages are unaffected and fine.

Issue with this is that you could just spam a bluespace beaker of blood and keep getting ventrue levels, and would let you brute force it, I don't really want bloodsuckers sitting around doing chemistry to get blood

Arturlang:
> How does this affect bloodsuckers who drown themselves in blood reagent turfs?

Made it not possible, implemented it and tested it

Arturlang:
Tempted to reduce the malus from drinking blood at high and add the current malus if it is duplicated blood..

projectkepler-ru (approved):

Arturlang:
Won't change this PR, but I can do the idea I have later, it's so merge it whenever

PR #1689, opened 2024-06-28
Explosions no longer cause lighting ballast malfunctions.

BurgerLUA:
## About The Pull Request

Explosions no longer cause lighting ballast malfunctions.

## Why It's Good For The Game

Ballast malfunctions are really bad when they're common, as they put stress on the lighting subsystem. This PR makes it so that ballast malfunctions don't occur from explosions. They can still be triggered by other means, which is currently only use of the SSDecay system.

## Proof Of Testing

Tested.

## Changelog

:cl: BurgerBB.
del: Explosions no longer cause lighting ballast malfunctions.
/:cl:


nevimer:
eh, we have an insane amount of spare cpu time..

BurgerLUA:
> eh, we have an insane amount of spare cpu time..

true but it sucks

StrangeWeirdKitten:
Performance wise it's not really a concern and that's grounds to close this PR. But if it's personal prefrence of I just don't like it, then we can have that seperate discussion.

BurgerLUA:
I mean performance wise, it's obvious after big explosions (See: usage of a BSA one time, Maxcaps on another round) lag badly due to the ballast nonsense. While it doesn't have 100% CPU usage, it does look like the server is chugging when it happens.

StrangeWeirdKitten:
> I mean performance wise, it's obvious after big explosions (See: usage of a BSA one time, Maxcaps on another round) lag badly due to the ballast nonsense. While it doesn't have 100% CPU usage, it does look like the server is chugging when it happens.

I'll run a test and see how much the lighting subsystem chugs when there's constant flickering.

StrangeWeirdKitten:
image

This is with about twice the amount of flickering lights you'd expect in a single max cap got to around 7 ms, which is by far not a lot of stress. 

For fun I called flicker() on 1600 lights in the server and got the subsystem up to 1000 ms with an SDQL. But that's by far an extreme case. I don't really see much of a noticable impact outside of people having to turn their flashlights on. Slowdown is going to mostly be tied up with whatever the atmospherics subsystem is doing. Overall *there really is not much noticable impact when some lights start to flicker* directly caused by the lighting subsystem. 

Also what the fuck is the economy subsystem doing. 
image

StrangeWeirdKitten:
Closing per above. The flicker() calls on these lights being removed will not have any noticable impact on performance.

PR #1691, opened 2024-06-28
[MODULAR] Reduce Empath point cost

ForrestWick:
## About The Pull Request

Just lowers the cost of the Empath trait by 2 points, down to 4.

## Why It's Good For The Game

Skyrat already lowered this thing from 8 to 6, but it looks like TG made some big changes to it since then and now it's basically just a flavor thing. Personally I'd rather it be 2 points, but I got told 4 would be best so here we are. Most of the stuff you get from it isn't exactly powerful enough to warrant a 6 point price.

## Changelog

:cl:
balance: lowered Empath cost from 6 to 3.
/:cl:


modular_zubbers/code/modules/quirks/code/positive_quirks/empath_override.dm
```diff
@@ -0,0 +1,10 @@
+/datum/quirk/empath
+	name = "Empath"
+	desc = "Whether it's a sixth sense or careful study of body language, it only takes you a quick glance at someone to understand how they feel."
+	icon = FA_ICON_SMILE_BEAM
+	value = 4 // Original: value = 8 - Skyrat Value: 6
+	mob_trait = TRAIT_EMPATH
+	gain_text = span_notice("You feel in tune with those around you.")
+	lose_text = span_danger("You feel isolated from others.")
+	medical_record_text = "Patient is highly perceptive of and sensitive to social cues, or may possibly have ESP. Further testing needed."
+	mail_goodies = list(/obj/item/toy/foamfinger)
```
StrangeWeirdKitten, line 10:
```suggestion
	value = 4
```
You don't need all this

StrangeWeirdKitten (changes requested):

plsleavemealon:
2 would probably be fine honestly.  It doesn't seem to do all that much....could also be used as one of the mythical 3 pointers, we don't really have many of those.

modular_zubbers/code/modules/quirks/code/positive_quirks/empath_override.dm
ForrestWick: > ```suggestion
> 	value = 4
> ```
> You don't need all this

Ah, gotcha

modular_zubbers/code/modules/quirks/code/positive_quirks/empath_override.dm
```diff
@@ -0,0 +1,3 @@
+/datum/quirk/empath
+	name = "Empath"
+	value = 4 // Original: value = 8 - Skyrat Value: 6
```
StrangeWeirdKitten, line 3:
```
value = 4
```

modular_zubbers/code/modules/quirks/code/positive_quirks/empath_override.dm
```diff
@@ -0,0 +1,3 @@
+/datum/quirk/empath
+	name = "Empath"
```
StrangeWeirdKitten, line 2: You don't need this either. Basically what you are doing is overriding the original value of 6 with 4. The compiler will first compile the original code, then when it compiles our modular changes, it'll see datum path again and duplicate variables with what you set here. (Or add an entirely new variable if you're adding one)

StrangeWeirdKitten (changes requested):

theselfish:
> 2 would probably be fine honestly. It doesn't seem to do all that much....could also be used as one of the mythical 3 pointers, we don't really have many of those.
> 
> Edit: Actually wait, we don't even have a single 3 point positive quirk. Huh.

Traits *are* meant to go in increments of two.
Skyrat and whomever added the clicker traits forgot that.

SonicGotNuked:
> > 2 would probably be fine honestly. It doesn't seem to do all that much....could also be used as one of the mythical 3 pointers, we don't really have many of those.
> > Edit: Actually wait, we don't even have a single 3 point positive quirk. Huh.
> 
> Traits _are_ meant to go in increments of two. Skyrat and whomever added the clicker traits forgot that.

This is arbitrary and I dont really care tbqh.

StrangeWeirdKitten:
This is arbitrary and I dont really care.

theselfish:
It's not arbitrary, it's consistency.

plsleavemealon:
> > 2 would probably be fine honestly. It doesn't seem to do all that much....could also be used as one of the mythical 3 pointers, we don't really have many of those.
> > Edit: Actually wait, we don't even have a single 3 point positive quirk. Huh.
> 
> Traits _are_ meant to go in increments of two. Skyrat and whomever added the clicker traits forgot that.

To be fair we have several traits that aren't in increments of 2.  With some being 7 points, others being 1 point.  And negative quirks are the same way with some being 3 and 1.

theselfish:
> > > 2 would probably be fine honestly. It doesn't seem to do all that much....could also be used as one of the mythical 3 pointers, we don't really have many of those.
> > > Edit: Actually wait, we don't even have a single 3 point positive quirk. Huh.
> > 
> > 
> > Traits _are_ meant to go in increments of two. Skyrat and whomever added the clicker traits forgot that.
> 
> To be fair we have several traits that aren't in increments of 2. With some being 7 points, others being 1 point, and yet some others being 5. And negative quirks are the same way with some being 3 and 1.

See "Skyrat and whomever added the clicker traits forgot that."

nikothedude:
> It's not arbitrary, it's consistency.

id argue its an arbitrary consistancy, theres no reason it has to go in increments of two besides tradition

PR #1477, opened 2024-05-04
Modular-ish: Buff TEG machines by 10

tmyqlfpir:
## About The Pull Request

This PR restores the original math value before the recent upstream merge.

## Why It's Good For The Game

For Interdyne and Dauntless players, the TEG is almost vital for their operations. It's also a good incentive for players to actively setup this system using the turbine exhaust gas.

## Proof Of Testing

I have tested this, and confirm it works.

## Changelog

:cl:
balance: Restore the original TEG efficiency 
/:cl:

PR #1697, opened 2024-06-30
reverts #1477 (TEG buff)

YakumoChen:
## About The Pull Request
reverts #1477 
Decreases the output of the TEG by 10x

## Why It's Good For The Game

It currently outputs 20MW in a decent setup which is insane and far outstrips any other actual method of power generation, trivializing power.
The reason for the buff was to help Interdyne etc, but they already have a perfectly functioning turbine.
2mw is MORE than enough for any real use, especially considering the TEG is completely safe and requires no upkeep

## Proof Of Testing

I assume it works if its a revert

## Changelog

:cl:
balance: Reverts buff to the TEG making it output up to 20MW, now outputs 10x less with a decent setup
/:cl:


Majkl-J (approved):
People also know how to use it now, so sure

PR #1724, opened 2024-07-02
 [MDB Ignore] [IDB Ignore] Bubberstation x Effigy Promo [TM ONLY]

lessthnthree:
## About The Pull Request

Two stations, exclusive from Effigy, will join Bubberstation 13’s map rotation sometime for the month of July. The two stations are: RimPoint, a planetary outpost set in the jungle of Taeloth, with a focus of traveling outside between points of interests. And Sigma Octantis, an underwater station on the same planet of RimPoint.

Additionally, whenever an Effigy map is chosen, the splash screen and other elements will change to display a collaboration between Bubberstation 13 and Effigy.


unit0016:
Can't edit the PR Name myself but can either maints or LT3 slap on both [MDB Ignore] & [IDB Ignore]? I have it on good authority that arc probably gets more annoyed with every patch I put out considering the map sizes here 💀

unit0016:
Going to be manually redoing the telescreens, dw about the lints for that commit.

PR #1743, opened 2024-07-05
[TM Only] Rimpoint Edits (pathways, touchups, the pole found a home)

ForrestWick:
## About The Pull Request

Essentially I did the following:

- Added stone brick paths along the map connecting each of the different departments, allowing easier navigation
- Rearranged the placement of the solar lamps (I really think the entire area should just be lit up naturally because the lamps look a bit out of place, but...)
- Added a new building in the dorms area that's a mini-lounge to house the stripper pole that was kinda just left in a patch of dirt.
- Removed some of the trees and placed a few elsewhere.
- Added some reinforced windows to perma (electrified grilles underneath).
- Made the upper level catwalks 1 tile wider, making it a bit easier to run around on for everyone.
- Added fences around each of the shuttle docking zones (cargo, mining in particular; arrivals was covered).
- Replaced the insta-death chasm by virology with a small patch of lava that's blocked off by r-walls and a plasma glass window.
- Placed some racks and delivery decals on the concrete patio area outside cargo.

## Why It's Good For The Game

Rimpoint's cool, but if it's gonna be here a whole month then it's probably best to make sure it's easier for people to navigate as well as touch up a few spots. That and less potential for accidental shuttle gibbing. Also giving perma prisoners a way to escape if they get admin permission (assuming they're not antagonists that got thrown in perma) via the windows is important.

## Changelog

:cl:
add: multiple small changes to the Rimpoint map
qol: roads on the Rimpoint map for easier travel.

/:cl:


Majkl-J:
This map isn't ours and is meant to be presented as-is. Simply use the navigation button, it's not that confusing after a few rounds.

PR #1744, opened 2024-07-05
[DNM] Tries to fix the mecha_damage check

Majkl-J:
Experimental before I put this up on tg


Majkl-J:
Upstream tg fixed this the exact same way line to line so

PR #1745, opened 2024-07-05
You can no longer vote for a map if it was the last 2 maps played, or the current map.

BurgerLUA:
## About The Pull Request

You can no longer vote for a map if it was the last 2 maps played, or the current map.

Ie if the map rotation is
Alpha (Current)
Beta (Last)
Charlie (Second Last)

then you can't vote for Alpha, Beta, or Charlie.

## Why It's Good For The Game

Prevents a never-ending cycle of people voting for the same 3 maps. This is to prevent a spree of people voting for the same 2 maps (cough

## Proof Of Testing

Untested.

## Changelog

:cl: BurgerBB
del: You can no longer vote for a map if it was the last 2 maps played, or the current map.
/:cl:


StrangeWeirdKitten (approved):
I'm fine with this

Swiftfeather (approved):
This is necessary and honestly if we don't take more maps out of rotation I think it should be even more.

Iamgoofball:
Why don't you guys just update to /tg/ map voting?

ForrestWick:
I thought this was already a thing but apparently not. Good change

PR #1828, opened 2024-07-23
Limits movement speed to about x2 the base movespeed

BurgerLUA:
## About The Pull Request

As a living creature, it is no longer possible to go 2 times faster than base movement speed. Base movement speed is the default running speed of mobs, without any buffs.

## Technical Information

Note that a slowdown value of 1 means that every 0.1 seconds, you can move a tile. A slowdown value of 0.75 means that every 0.075 seconds, you can move.

Thus, the lower the slowdown value is, the faster you are. Someone with a slowdown value of 0.5 is twice as fast as someone with a slowdown value of 1.

The game's default slowdown value for humans is 1.5. This means that every 0.15 seconds, you can move a tile.

This PR makes it so the capped slowdown value is 0.7. I came up with this number because changeling adrenaline removes 0.8 from your slowdown value (1.5 - 0.8 = 0.7), and it's the strongest source of increasing movement speed in the game.

Some thing can "ignore" this cap, such as bluespace tiles and other sources of movement manipulation. I tested everything obvious (camera movement, xenomorphs, humans, monkeys, etc) but there might be some things that missed testing that are affected by this.

## Why It's Good For The Game

Having an insane amount of speed is generally not fun for several reasons.
- Melee as a system is underdeveloped and falls apart during fights where an opponent runs at mach 10.
- Projectiles are balanced around not being hitscan or instant, and are slow as to balance out ranged. However, it is generally not reasonable to move faster than a bullet.
- A good majority of people are connected from North America (sever is hosted in Germany), where they'd usually get 80-100 ping on a relay server. While this seems small, 100ms delay is quite a bit.

A legitimate question is why not just apply this to humans. I think applying this to just humans kind of defeats the purpose of the PR, and that generally high movement speed in general is not good for the game.

## Proof Of Testing

It just werks.

## Changelog

:cl: BurgerBB
balance: As a living creature, it is no longer possible to go 2 times faster than base movement speed. Base movement speed is the default running speed of mobs, without any buffs.
/:cl:


PawnAtWork:
I genuinely 100% support this and feel it should be tried upstream first, that'd solve a lot of issues.

projectkepler-ru (commented):
ok admittedly this is a fucking incredible idea, I myself play on germany relay and get sub 200 ms, which is good but still not enough.

I wholeheartedly agree it suck ass to play against someone who can physically react 200ms faster than me

StrangeWeirdKitten:
Test merge when it's ready. I plan on giving it a week. But to be perfectly honest, this is a very reasonable PR since it still allows 4x movement speed.

BurgerLUA:
@StrangeWeirdKitten This is ready for testing.

Majkl-J:
Remind me to TM this after we fix the server going woomp and crashing over some DB stuff

Majkl-J (approved):
This has been going quite well

PR #1832, opened 2024-07-24
HoS+QM no longer start with Science Hub

ReturnToZender:
## About The Pull Request
Two years ago, a great man declared that in Science and Reason, all Heads were created equal. This was not true, however, as dark clouds gathered on the horizon-- The Head of Security, villain of this story, threatened to ruin the peace in Scienceville by spending dangerous amounts of Science points.

"No, Head of Security," they cried, as their hard-earned harvests were churned into weapons of war, the miners roared, the scientists floored, and it was a dark day indeed.

This PR seeks to solve this problem, by removing the Head of Security and Quartermaster, the two lowest command roles on the rung, from carrying the Science Hub app.
(I'm not super strongly attached to the Quartermaster having this taken off, if you feel the Quartermaster should keep it, let me know and I'll change this to just the Head of Security :>)

## Why It's Good For The Game
Heads of Security and Quartermasters can no longer butt out Science with their Science hub access.

## Proof Of Testing
It compiles, but I need to check and make sure it works properly first. So, this is a draft.

## Changelog

:cl: ReturnToZender
balance: QM+HoS no longer have the Science Hub on their PDAs.
/:cl:

### DIFF
--- a/code/modules/modular_computers/computers/item/role_tablet_presets.dm
+++ b/code/modules/modular_computers/computers/item/role_tablet_presets.dm
@@ -52,7 +52,7 @@
 	starting_programs = list(
 		/datum/computer_file/program/crew_manifest,
 		/datum/computer_file/program/status,
-		/datum/computer_file/program/science,
+		///datum/computer_file/program/science, //BUBBER EDIT: HOS CANNOT HAVE RESEARCH APP
 		/datum/computer_file/program/robocontrol,
 		/datum/computer_file/program/budgetorders,
 		/datum/computer_file/program/records/security,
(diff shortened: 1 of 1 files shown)


AllCorruptingWyrm:
image
image
science very rarely has enough staff to man the desk, also nearly all of them are antagonists. This does scream sec salt because I routinely see science sit on 600-1200 points some shifts.

StrangeWeirdKitten:
Consoles still exist on the bridge. Programs in lockers. This is very arbitary, which I pointed out. Nor is it really that much of an issue except maybe one time where someone researched something. You can 100 percent make it so other heads' queues can't be cleared since apparently I heard that exists. Only the RDs and Captain queue shouldn't be cleared, but that is in a future / or a completely rewritten PR.

PR #1833, opened 2024-07-25
Crew monitor alarm doesn't trigger in medbay

lessthnthree:
## About The Pull Request

Adjusts the crew monitor alarm so that bodies dead in medical don't set off the alarm.

## Why It's Good For The Game

Less false alarms

## Changelog

:cl: LT3
qol: Crew monitor alarm now excludes patients in medbay
/:cl:


nevimer:
It kind of makes medical a magical area, with properties that are immutable... Maybe there is a better way?

Swiftfeather:
This change isn't very necessary, medbay staff can just remove the clothes.

lessthnthree:
> This change isn't very necessary, medbay staff can just remove the clothes.

I agree with this more than finding an alternate way to go about it.

PR #1834, opened 2024-07-25
Adds a strict limit for how many antagonists of a type can be added in the round via storyteller

BurgerLUA:
## About The Pull Request

Adds a strict limit for how many antagonists of the same type can be added in the round. This affects both midrounds and roundstarts.

### Current limits are:
- Bloodsucker: 3
- Changeling: 4
- Heretic: 2
- Malfunctioning AI: 1
- Nuke Ops: 5
- Spies: 4
- Traitor: 6

If an antagonist type is not listed here, there is no limit.

This does not affect an admin's ability to bus in additional antagonists via the antag panel. This just affects storyteller.

## Why It's Good For The Game

Storyteller does not actually have a maximum cap of how many antagonists of a certain type can be in the game. This means that while the maximum heretics that storyteller can spawn roundstart is 2, there is actually no limit (except for cost) for how many heretics can be in a round, as the midround heretic event does not consider existing heretics in the round.

This PR fixes that and adds a hard cap to antagonists of a certain type in a round. This should prevent the rare and extremely stressful round of 6 heretics spawning or something insane.

## Proof Of Testing

Untested. Can't really test this because I can't fake antag players.

## Changelog

:cl: BurgerBB
balance: Adds a strict limit for how many antagonists of the same type can be added in the round. This affects both midrounds and roundstarts.
/:cl:


AllCorruptingWyrm:
I would suggest swaping ling with bloodsucker count wise, generally lings are dealt with and instead fluxuate in and out of their aheal regen coma, but bloodsuckers have to be treated like a rev round.

BurgerLUA:
A question that was asked: Do vassals affect bloodsucker numbers?
The answer is no, however if a vassal is turned into a bloodsucker, then yes.

modular_zubbers/code/modules/storyteller/event_defines/roleset/_antagonist_event.dm
```diff
@@ -83,8 +85,20 @@
...
 	var/amount = base_antags + FLOOR(people / denominator, 1)
+
+	if(antag_datum && maximum_antags_global > 0)
+		var/antag_slots_left = maximum_antags_global
+		for(var/datum/antagonist/existing_antagonist as anything in GLOB.antagonists)
+			if(!istype(existing_antagonist,antag_datum)) //Obviously ignore other antagonists.
+				continue
+			antag_slots_left-- //Slot is occupied.
+			if(antag_slots_left <= 0) //No point in checking anymore.
+				break
```
Majkl-J, line 99: Might be nice to check `isnull(existing_antagonist.owner.current)` and exclude those. I do wanna hear your input on this though

BurgerLUA: If it is possible for antag datums not to have owners, then sure. I don't know how they're handled.

Majkl-J: If the `owner.current` is missing then it means the body is destroyed so it shouldn't be taking up a slot. Also apparently sometimes antag datums can become completely ownerless so `owner` should be nullchecked too

theselfish:
Does nuke ops account for re-enforcements and borgs?

projectkepler-ru:
> Does nuke ops account for re-enforcements and borgs?

They should be fine, since the nukie are spawned first so they call in reinforcement and borg after.

imedial:
> Does nuke ops account for re-enforcements and borgs?

nukies is admin only rn

projectkepler-ru:
> > Does nuke ops account for re-enforcements and borgs?
> 
> nukies is admin only rn

they're not, never has been.

Majkl-J:
> > > Does nuke ops account for re-enforcements and borgs?
> > 
> > 
> > nukies is admin only rn
> 
> they're not, never has been.

They literally still are

projectkepler-ru:
> > > > Does nuke ops account for re-enforcements and borgs?
> > > 
> > > 
> > > nukies is admin only rn
> > 
> > 
> > they're not, never has been.
> 
> They literally still are

we literally had them roll naturally before, albeit a tad bugged

BurgerLUA:
> Does nuke ops account for re-enforcements and borgs?

Nope

PR #1835, opened 2024-07-25
Upstream 7 24 2024

StrangeWeirdKitten:
voidwalkers....

🆑
balance: Voidwalker trauma does not mute you. It'll still pacify you and you'll still die in planetary gravity (which cures you).
/:cl:


StrangeWeirdKitten:
Moonstation CI is kind of broken on master. That'll be a future fix.

PR #1842, opened 2024-07-27
Limbgrown bodies.

xXPawnStarrXx:
## About The Pull Request

The limbgrower was an often underutilised item, it was only really useful for bounties, because most medical staff never checked if their ~~victim~~ patient was a body purist or not. They'd just jam them full of parts and off they go. This allows an easier method (if not more expensive) to getting corses for body transfers. I have also added blank brains for the niche reason of RSD phylacteries for those who were completely gibbed or dusted to be able to return to the round. 

Made it so that hopefully; Ghosts can be added to the round via soul phylacteries. If we can have golems, ashwalker eggs, cyborgs and such that accept ghosts and can allow players to add ghosts to the round, soulcatchers shouldn't be so restrictive.

EDIT: As per request, I have removed monkeydone too, so the recipe will no longer be available.

## Why It's Good For The Game

Miners who get gibbed, dusted or just trapped deep in lava.
Genetics gets to keep their monkeys.
Coroners have more to do, by printing, preserving and prepping corpses for revival.
More uses for the limbgrower and less CBT of restraining a feral monkeyman.

## Proof Of Testing

image

## Changelog

:cl:
add: Added blanks to the limbgrower, for soul/brain transfer.
/:cl:


projectkepler-ru (commented):
Please hold, the implication of this change is abit, big

aKromatopzia:
> Please hold, the implication of this change is abit, big

isn't this pretty on-rate with stealing monkey cubes from botany/genetics/xenobio/virology?

xXPawnStarrXx:
> isn't this pretty on-rate with stealing monkey cubes from botany/genetics/xenobio/virology?

Yeah. That's the majority of what I was hoping to replace. You need more synthflesh, but means no monkeydone or anything like that.

YakumoChen:
Regardless of the implications of this, we should not be encouraging in any manner players cold ghosting into RSDs in order to get a new body when they've been gibbed or similarly round removed.

In most cases, round removal either happens by intention or otherwise can be ahelped and an admin will usually fix it, if the round removal is against the rules, the result of a bug, or dying so badly that the admin gives you a more manageable way to be revived out of pity.

What this would encourage is security being even more suicidal, since we truly would be removing all consequences of dying to antagonists no matter the cost.

StrangeWeirdKitten:
It's either this or monkeydone. Remove monkeydone from the code as we try out this PR. I don't think it's awful to require synthflesh or coorporation with genetics (to use the machine) to print a new body.

xXPawnStarrXx:
> It's either this or monkeydone. Remove monkeydone from the code as we try out this PR. I don't think it's awful to require synthflesh or coorporation with genetics (to use the machine) to print a new body.

I didn't feel it was right to rip someone else's PR out to test mine. Monkeydone can also still be useful, since the chem will work without power.

shellspeed1:
Honestly, why not have both. Monkey done is useful as a basic bitch solution while this lets medical have several bodies prepped for quick revives while still giving medical pretty of content. Especially as the content is not reactive to the round.

aKromatopzia:
> Regardless of the implications of this, we should not be encouraging in any manner players cold ghosting into RSDs in order to get a new body when they've been gibbed or similarly round removed.

true, but if you had the brain can't you just slot that into a monkey body anyways?

But, what is this about monkeydone - is that not just a more-specific mutadone?

Majkl-J:
> But, what is this about monkeydone - is that not just a more-specific mutadone?

Mutadone no longer turns monkeys into humans

modular_zubbers/code/modules/limbgrowncorpses/limbgrowerblanks.dm
```diff
@@ -0,0 +1,31 @@
...
+	var/old_death_sound = death_sound // You can't do this with initial() as the death sound is set after by the species datum
+	death_sound = null // We don't need them screaming
+	death()
+
+	var/obj/item/organ/internal/brain/B = get_organ_slot(ORGAN_SLOT_BRAIN)
+	if(B)
+		B.Remove(src)
+		QDEL_NULL(B)
+
+	death_sound = old_death_sound // Now they can die loudly again
```
Majkl-J, line 15: Normal humans don't have death sounds, also the nullcheck is not needed as the brain is guaranteed to exist and qdelling already removes the organ from the body so we don't need to do that.

Turns out my old synth code that you copied is not that good lol

```suggestion
	death()

	var/obj/item/organ/internal/brain/to_remove = get_organ_slot(ORGAN_SLOT_BRAIN)
	QDEL_NULL(to_remove)
```

Majkl-J (changes requested):
Also you should remove monkeydone with this PR. Feels like a better and more elegant solution

StrangeWeirdKitten:
stale

Majkl-J (changes requested):
Code is solid, I still think you should review monkeydone with this PR as well

xXPawnStarrXx:
> Code is solid, I still think you should review monkeydone with this PR as well

Fair enough, monkeydone is monkeydone-done away with.

xXPawnStarrXx:
@Majkl-J Won't let me request another review, so I'll do it this way.

Majkl-J (approved):

funnyfreeperson:
this is ironically cloning again. not to say that's a bad thing. but yanno. full circle

Majkl-J:
> this is ironically cloning again. not to say that's a bad thing. but yanno. full circle I'd like to say that synthflesh is a good way to "limit" it, as that's the way most modern implementations of cloning work rn anyway

So was making monkeys and that was cheaper than this

PR #1843, opened 2024-07-27
It's not just you, everything IS slower. Fixes SSmachinery sleep [TM ONLY]

lessthnthree:
## About The Pull Request

Early mirror of https://github.com/tgstation/tgstation/pull/85269


StrangeWeirdKitten:
Will need to pull down from skyrat as soon as it's available.

PR #1850, opened 2024-07-28
Fix blood filter filtering toxins

lessthnthree:
## About The Pull Request

Updates the check for the blood filter's toxin removal so that the surgery step is allowed to start. Updates the removal proc to work similar to the new blood filtering clamp.

## Why It's Good For The Game

Toxin filtering to blood filter was added, but the surgery step check won't start unless there's reagents.

## Changelog

:cl: LT3
fix: Blood filter toxin removal feature now starts correctly
/:cl:


StrangeWeirdKitten:
Is this not an upstream bug?

lessthnthree:
> Is this not an upstream bug?

No, toxin filtering is a Bubber addition. However the check for if the step can be performed doesn't actually check for toxins.

PR #1856, opened 2024-07-29
[TM ONLY] Adds, well Ports Echolocation

Erol509:
## About The Pull Request

I will move it to skyrat, as well. For now, test merge it here please. 

This is ported from Nova https://github.com/NovaSector/NovaSector/pull/525 All credits goes towards the original author, I only fixed issues with the tgui crashing, when trying to customise it

You can also read what it adds, but shortly the types only allow you to share what you see to your friends with the same type of visior. Extrasensory and Extrasensory will be able to see tiles around each other, I think...from what I understood from the PR.  And you can do the same with psykers, with psyker mode. 

You can turn it off and MOST IMPORTANTLY, it makes you blind, but also it removes the anoying eye straining screen refresh rate, and changes colors...So your hud is not entierly white and black. Its a huge improvement.

It doesnt fully work, for some reason. I mean the outline colors are not being applied correctly. Probably because it nulls somewhere. 

## Why It's Good For The Game

Proper way to play as a blind character and more flavour to our bat friends. 

## Proof Of Testing

image

## Changelog

:cl: yooriss
add: The Echolocation quirk is now available with several levels of customization, granting 5 tiles of pulsing view distance and significantly enhanced hearing at the downside of being blind when deafened and twice as vulnerable to ear damage.
/:cl:

### DIFF
--- a/_maps/map_files/moonstation/moonstation.dmm
+++ b/_maps/map_files/moonstation/moonstation.dmm
@@ -14956,7 +14956,7 @@
 /obj/item/reagent_containers/cup/glass/bottle/holywater{
 	desc = "A flask of holy water allegedly containing the first seed planted in the Garden of Eden";
 	name = "flask of the first holy seed";
-	list_reagents = list(/datum/reagent/consumable/cum = 100);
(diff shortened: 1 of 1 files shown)


Erol509:
https://github.com/Skyrat-SS13/Skyrat-tg/pull/29108

PR #1868, opened 2024-07-31
Removes N-Spect contraband scanning

StrangeWeirdKitten:
## About The Pull Request

What the title says. The gates still function as intended.

## Why It's Good For The Game

This is a very arbitary magic wand device. I know there's a few things that block it (like smuggler satchels), but security should still search people manually and not walk up and scan them while they're typing. The gates themselves are fine.

## Proof Of Testing

It works.

## Changelog

:cl:
balance: N-Spect no longer has the capability for mobile contraband scanning.
/:cl:

### DIFF
--- a/_maps/map_files/BoxStation/BoxStation.dmm
+++ b/_maps/map_files/BoxStation/BoxStation.dmm
@@ -27451,7 +27451,7 @@
 /turf/open/floor/iron/smooth_large,
 /area/station/engineering/engine_smes)
 "ceo" = (
-/obj/machinery/keycard_auth/directional/south,
+/obj/machinery/keycard_auth/wall_mounted/directional/south,
 /obj/machinery/computer/station_alert{
 	dir = 4
 	},
@@ -61168,7 +61168,7 @@
 /turf/open/floor/iron/smooth_large,
 /area/station/engineering/gravity_generator)
 "wIx" = (
-/obj/machinery/keycard_auth/directional/west,
+/obj/machinery/keycard_auth/wall_mounted/directional/west,
 /obj/machinery/pdapainter/supply,
 /obj/machinery/computer/security/telescreen/cargo/directional/north,
 /obj/machinery/camera/directional/west{
--- a/_maps/map_files/LimaStation/LimaStation.dmm
+++ b/_maps/map_files/LimaStation/LimaStation.dmm
@@ -5317,7 +5317,7 @@
 "chZ" = (
 /obj/machinery/pdapainter,
 /obj/structure/cable,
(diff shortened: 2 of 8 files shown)


Majkl-J (approved):

Moostard:
If they will be seaching manually anyway, what is the problem?
It is a nice device that makes the "search" last seconds instead of feeling like eternity. Which also might save your ass if you do have a satchel or other stealthy item, which wont save you from the manual search.
The "problem" you said there is more or less about people running around scanning people when they talk, which should be against the rules. And pretty sure it is.

Also, doesnt it just gut the whole "contraband" bounty thing?
You use it to scan people and detect contraband, which gets marked (or something) for the bounty you can get in cargo.

StrangeWeirdKitten:
I'll look at the bounty later, thanks for pointing that out. And no, we're not going to make policy around this because logs are a pain in the ass when it comes to typing indicator.

Moostard:
Eh.
Im just sad that a really good feature, which I would consider QoL as you can search someone in seconds and let them go, and the whole contraband collecting thing is getting axed, because of some rule breakers.
I do not think it is the correct way of solving it.

StrangeWeirdKitten:
> Eh. Im just sad that a really good feature, which I would consider QoL as you can search someone in seconds and let them go, and the whole contraband collecting thing is getting axed, because of some rule breakers. I do not think it is the correct way of solving it.

I'll rework it into scanning bags and things like lockers.

Moostard:
> > Eh. Im just sad that a really good feature, which I would consider QoL as you can search someone in seconds and let them go, and the whole contraband collecting thing is getting axed, because of some rule breakers. I do not think it is the correct way of solving it.
> 
> I'll rework it into scanning bags and things like lockers. I agree that it's neat, but QOL should not interfere with an antag who is trying to RP.

They could also just run close and try to take the bag off as well, which is less noticable (funny enough).
Scanner itself is more harmless than people believe it to be.
The scanner only tells you if the person has contraband or doesnt. You still have to take it from them. And you can trick the scanner (cant trick the person searching if they are actually focusing on it).
Getting scanned mid sentence is bad sport for sure, but it is nothing in comparison to like typebaiting.
If someone gets stopped, they likely will be searched anyway, scanned or not.

All I can do is ask you to reconsider this as quite a few players (including me) do enjoy this feature.

Iamgoofball:
This is an extremely bad change that undermines the entire point of the mechanic.

Majkl-J:
> They could also just run close and try to take the bag off as well, which is less noticable (funny enough).

Pretty sure stripping is a span_danger, (the large red text one), and shows cogs above them

PR #1891, opened 2024-08-02
Changes to Ethereal's power supply

plsleavemealon:
## About The Pull Request
Rebalances the ethereal's power supply so that it is more in line with the power changes that took place a while ago.
Essentially, the changes are as follows:

Maximum power storage without experiencing problems is at 110KJ
Overcharge has been tweaked to 115KJ
And dangerous overcharge has been tweaked to 120KJ
This is in line with the rest of them and means I don't need to change how the discharging works for overcharge.  I'd like to possibly look into modifying or lessening some of the penalties for overcharging when it isn't at the dangerous level but I'll do that later.

## Why It's Good For The Game

Ethereals are hardly played, their basic power supply is only at 20KJ and its weird that it is so extremely low.  This change brings them more in line with the rest of the power changes that occurred a while ago and also helps them more closely resemble their synthetic counterparts (who probably need some tweaking too but that can come later).

## Changelog

:cl:
balance: rebalanced (buffed) ethereal's interior power cell
/:cl:


Majkl-J (approved):

Arturlang:
This might be better to PR to TG

Majkl-J:
> This might be better to PR to TG

TG is on a complete freeze and I don't like waiting

PR #1896, opened 2024-08-03
Removes code edits allowing Botany plants to mutate any botany trait

ReturnToZender:
## About The Pull Request
Deletes 6 lines that make it so _every_ trait is mutatable, fixing the problem with the DNA manipulator (worth noting the problem itself was not the manipulator it was some lines of code that allowed any trait to be selected)

Also steals #1614 

## Why It's Good For The Game
Prevents any future issues of Botany powergaming by mutating anti-magic deathnettles that do 95 brute, then turn into a killer tomato when you click them in hand

## Proof Of Testing
image
I cannot add it :(

## Changelog

:cl: ReturnToZender, Shellspeed (DNA machine fix)
balance: Plants can no longer mutate traits that were specifically marked as not mutatable
Fix: Plant DNA manipulators now properly respect upgrades
/:cl:


shellspeed1:
Should probably test merge this as we have some existing Plant gene editors issues that may still need to be fixed like it not pulling the actual stats of the seeds but only the default ones.

DorfGuy1:
hey i think this broke being able to change the traits at all. like yield and what not.

PR #1904, opened 2024-08-04
Nerfs chasms because I saw someone die to it once.

BurgerLUA:
## About The Pull Request

You can no longer willingly walk into chasms. Does not apply if the chasm is safe to walk on (it has rods on it or something).

You can still fall in a chasm if you're dead, you were thrown onto it, if you are inside something (like a barrel, or a crate), or if you lost your flight while on a chasm or the rods on a chasm were removed while on a chasm.

## Why It's Good For The Game

Chasms are unfun. People walk into them accidentally all the time. It basically round removes you for lagging and in most cases an admin usually sees someone die due to chasms and just respawns them.

"But you can rescue people who fall in chasms!" yeah you can fuckin find someone's brain that was tossed into deep space too. Just because it's possible, doesn't mean people will actually do it.

## Proof Of Testing

Untested. The following needs to be tested.

- [ ] If the prevention actually works.
- [ ] If the prevention actually works with mechs.
- [ ] If thrown mobs can still get chasm'd.
- [ ] If you can still walk on chasms if there are supporting rods.
- [ ] If you can still walk on chasms if you are flying.

## Changelog

:cl: BurgerBB
balance: You can no longer willingly walk into chasms. You can still fall into it via other means (getting thrown, pushed while dead, floor below you turns into a chasm, etc).
/:cl:

### DIFF
--- a/code/game/turfs/open/chasm.dm
+++ b/code/game/turfs/open/chasm.dm
@@ -17,10 +17,12 @@
 	. = ..()
 	apply_components(mapload)
 
+/* BUBBERSTATION CHANGE START: CHASM HUGBOX
 /// Lets people walk into chasms.
 /turf/open/chasm/CanAllowThrough(atom/movable/mover, border_dir)
 	. = ..()
 	return TRUE
+BUBBERSTATION CHANGE END: CHASM HUBOX */
 
 /turf/open/chasm/proc/set_target(turf/target)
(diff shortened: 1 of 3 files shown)


Majkl-J:
Firstly, we're currently frozen, and this includes significant balancing, which this very much falls under as it reworks a fairly core mechanic.

Secondly, chasm fishing doesn't care for distance or z level `/code/modules/fishing/fish/chasm_detritus.dm#L99`
image

PR #1919, opened 2024-08-08
Unfucking disease outbreaks/transmission

lessthnthree:
## About The Pull Request

Disease transmission is nonsensical and broken. Goes through disease outbreak and transmission vectors from start to finish fixing things that don't work properly, or not as expected. Also adds some constraints on the disease outbreak event based on medical staff.

This is mostly nonmodular edits as it's a test before I make the PR on TG. It's been tested for a long while with NPCs on stations, but live testing it here gives better visibility into how it goes in a real world environment, and there is some extra game logging added for the duration of the TM to validate the new disease calculations and spread are working as expected.

Mainly the test is to see if the new values for disease transmission "feel" right in a live environment, to perform some tweaks to the percentages if required.

- Event diseases are no longer hidden from health analyzers during incubation periods
- Masks and head slot items actually provide protection from airborne disease spread
- People other than the original airborne hosts spread via respiration
- Reduced infection chance for contact/airborne diseases from current 95%+ on all diseases to values 15%-40% based on the transmissibility strength
- Airborne spread is rolled per target, not source (smoother infections compared to suddenly infecting 5 people with a single sneeze)
- Disease outbreak won't spawn during rounds with less than 3 medical staff
- Disease outbreak begins with 2-3 original hosts but lower spread speed

## Why It's Good For The Game
- Disease outbreaks and transmission work more in line with how people think they do/are supposed to.
- Gloves, masks, face shields etc. properly get calculated to protect you from getting infected/infecting others.
- Disease outbreaks can immediately be identified by med staff with analyzers, but borgs can't see the HUD icon as soon as the event starts and instantly find patient zero.

## Changelog

:cl: LT3
balance: Event diseases are no longer hidden from health analyzers during incubation periods
fix: Masks and head slot items actually provide protection from airborne disease spread
fix: People other than the original airborne hosts spread diseases via respiration
balance: Reduced infection chance for contact/airborne diseases from current 95%+ on all diseases to values 15%-40% based on the transmissibility strength
code: Airborne disease spread is rolled per target, not source
balance: Disease outbreak won't spawn during rounds with less than 3 medical staff
balance: Disease outbreak begins with 2-3 original hosts but lower spread speed
fix: Disease cure won't request both antihol and alcohol mixed together
/:cl:


Arturlang:
The diseases being hidden at the start protects it from being instantly detected and killed, not sure how well if at all it will spread now, given how unlikely that they spread on here

YakumoChen:
> [...]only means you have a patient screaming in the lobby that they have something and are patient zero[...]

Patient zero should probably be a silent carrier in that case, and have the incubation timer make their medhud invisible for the incubation duration, I thought it was intended in that way?

The initial carrier being able to out themselves instantly makes most viruses unable to spread interestingly.

lessthnthree:
> > [...]only means you have a patient screaming in the lobby that they have something and are patient zero[...]
> 
> Patient zero should probably be a silent carrier in that case, and have the incubation timer make their medhud invisible for the incubation duration, I thought it was intended in that way?
> 
> The initial carrier being able to out themselves instantly makes most viruses unable to spread interestingly.

Before:
Skin contact under sev 5: hidden from scanner and medhud for duration of incubation timer
Airborne, skin contact sev 5+:  visible immediately

Now:
Skin contact (all): hidden from medhud for duration of incubation timer
Airborne: visible immediately

It's not a large change in that aspect, point of the TM is to see if there is any meaningful change in how response pans out. We haven't had a disease outbreak event since it's been TM'd. (error in my check had it wanting 5 med staff, not 3, though that's fixed as of current round.)

If you want to adminbus a disease outbreak give me a poke on discord and I'll come watch.

lessthnthree:
I've refined it so that only airborne diseases with sneeze/cough are immediately visible, as those are the two that can spread in a single digit amount of seconds. The rest use the incubation timer, and won't show up on medHUD until stage 2.

ReturnToZender:
I think this can be merged now we've tested it for over half a year

PR #1925, opened 2024-08-09
Security belts can no longer hold guns

KazooBard:
## About The Pull Request
The security belt is unable to hold guns such as disablers, dragnets, m1911's and essentially anything that is /obj/item/gun/...

## Why It's Good For The Game

Hoarding guns is a form of powergaming. Noone should be holding 5 guns on themselves at all times, that is why guns were made bulky in the first place with exception of stealthy ones. The problem arises when the issue of storing so many medium-sized objects is made really simple by letting you have at LEAST 3 medium slots on your belt essentially.

 It can end up in ridiculous loadouts such as this one: 
hosgun

## Changelog

:cl:
balance: Security belts can no longer hold guns
/:cl:

### DIFF
--- a/code/game/objects/items/storage/belt.dm
+++ b/code/game/objects/items/storage/belt.dm
@@ -387,8 +387,6 @@
 		/obj/item/clothing/gloves,
 		/obj/item/flashlight/seclite,
 		/obj/item/food/donut,
-		/obj/item/grenade,
-		/obj/item/gun, //SKYRAT EDIT ADDITION
 		/obj/item/holosign_creator/security,
 		/obj/item/knife/combat,
 		/obj/item/melee/baton,
--- a/code/game/objects/items/storage/belt.dm
+++ b/code/game/objects/items/storage/belt.dm
@@ -387,6 +387,7 @@
 		/obj/item/clothing/gloves,
 		/obj/item/flashlight/seclite,
 		/obj/item/food/donut,
+		/obj/item/grenade,
 		/obj/item/holosign_creator/security,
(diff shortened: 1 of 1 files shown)


AllCorruptingWyrm:
This PR achieves absolutely nothing and will make things worse. The real issue is inventory management instead of really anything else.

You think this will fix things but it really wont. In fact, it'll just encourage secoffs to stock their glock instead of a disabler and it'll lead to people urging to amber so they can get a longarm instead of something else. This will then in turn lead to security being MORE empowered than before and this will not go over well. In fact, this is such a drastic change that it's going to make a lot of people unhappy and lead to antagonists getting nerfs too because of syndicate items have been balanced with this in mind. Security **needs** to be able to have utility in the tools they have or else they'll just default to the more effective option, and making pistols bigger will just lead to sec justifying LONGER amber alerts, and lead to an imbalance with antagonist tool items. Its a massive change that I dont think you realize will really make things worse

Majkl-J:
> because of syndicate items have been balanced with this in mind.
Incorrect. Most syndicate item balance is TG, and the extra gun slots don't exist on TG

> Security needs to be able to have utility in the tools they have or else they'll just default to the more effective option.
There's nonlethal rounds for almost every ballistic, and a good half of the laser guns feature switchable modes

Also the glocks are supposedly quite meh, or so I've been told by Tzula in the voltaic heart discussion

YakumoChen:
> standard roundstart HoS loadout (2 baggable-size guns and a longarm)
> powergaming
?????????

PR #1933, opened 2024-08-10
[modular] Adds a silicon manager PDA app similar to the Job manager app the HoP gets

Odairu:
## About The Pull Request
Allows for the prioritizing of silicon jobs (AI/Cyborg) so they show up on the join menu, you cannot add or close slots, and it can be downloaded and used as long as you have Robotics Access on your ID

## Why It's Good For The Game
Every other job can be prioritized, these two should be able to be as well

## Proof Of Testing
image
image

## Changelog

:cl:
add: adds silicon manager pda app
/:cl:


projectkepler-ru (approved):
I specifically asked for this since I wanted to do otherthing instead
so for once i am actually quite happy to see a PR like this

aKromatopzia:
yipee

lessthnthree (approved):

PR #1960, opened 2024-08-16
Re-enables "The Fragile" (Do Not Merge)

BurgerLUA:
## About The Pull Request

Re-enables "The Fragile" storyteller.

## Why It's Good For The Game

I am not sure why it was disabled, but I remember it somehow breaking storyteller for some reason. If anyone has any idea about this, please post here so I can resolve it.

The Fragile is a good alternative to sleeper where things can still happen, but weights are specifically applied to prevent dangerous antag types and events from being rolled frequently.

## Proof Of Testing

Untested. :)

## Changelog

:cl: BurgerBB
add: Re-enables "The Fragile" storyteller.
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/storyteller/storytellers/storyteller_fragile.dm
+++ b/modular_zubbers/code/modules/storyteller/storytellers/storyteller_fragile.dm
@@ -1,4 +1,3 @@
-/*
 /datum/storyteller/fragile
 	name = "The Fragile"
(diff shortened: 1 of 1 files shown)


BurgerLUA:
I am being told it was disabled because it spawned every antag.

I have resolved this by adding antag_divisor = 8. Antag divisor was not set in this storyteller and I suspect the default is probably 1, trying to spawn 1 antag for every crewmember.

BurgerLUA:
making in another pr

PR #1964, opened 2024-08-17
Abductor machines no longer check for species when deciding if you can use it or not

Arturlang:
Changes abductor machines to no longer randomly check for both species and traits, instead they will check for the TRAIT_ABDUCTOR_TRAINING

## About The Pull Request
We have abductor species that you can spawn as from roundstart, it makes more sense that only the antags that are supposed to use this stuff, can.
Also fixes the gizmo's being unlinkable due to item interaction changes

## Why It's Good For The Game
Need it for a upcoming event, moreso, secondly, currently crew-abductors can use these devices which is probably not good for them to be able to use antag machines

## Changelog

:cl:
add: Now you can re-link abductor machines using a science tool if one of them happens to become unlinked.
admin: Admins can click on a abductor console(not the camera one, but the other one that controls equipment) to change it's team ID.
balance: Abductor machines are now locked based on the TRAIT_ABDUCTOR_TRAINING instead of randomly checking for both training and abductor species
fix: You can now again link abductor gizmos with a abductor console
/:cl:

### DIFF
--- a/code/game/machinery/computer/camera_advanced.dm
+++ b/code/game/machinery/computer/camera_advanced.dm
@@ -133,7 +133,7 @@
 	return can_interact(user)
 
 /obj/machinery/computer/camera_advanced/abductor/can_use(mob/user)
-	if(!isabductor(user))
+	if(user && !HAS_MIND_TRAIT(user, TRAIT_ABDUCTOR_TRAINING)) // # BUBBER CHANGE, abductor machinery checks for abductor training
 		return FALSE
 	return ..()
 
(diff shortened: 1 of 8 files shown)


projectkepler-ru (changes requested):
PR Body and Changelog needs improvement, I'm not convinced this is good enough
Changelog suggests this is a fix but I'd wager it belongs in a different category

projectkepler-ru (approved):

destrucktoid:
I disagree and think that abductor crew having the ability to use abductor machines makes sense. They already cannot use certain normal machines and being able to use the abductor ones makes sense. They already cannot use abductor guns or weapons which would be the only things i would agree with you on that they shouldn't be able to use. Being able to use abductor consoles though makes sense.

Issue #1700, opened 2024-06-30
Placing vamp structures anywhere


KazooBard:
Okay so you can just, do what I do in this video, ez spam the station with hallucinations and structures which are really fucking hard to break cause it's funny. 

https://github.com/Bubberstation/Bubberstation/assets/65713506/b8261d6e-6de5-41ad-8a08-4b97a70a5932

And yeah accidentally posted it empty then I panicked and closed this, woopsieeees

KazooBard:
FUCK I PRESSED ENTER ON ACCIDENT

ReturnToZender:
@Arturlang

Arturlang:
Huh. Damn

PR #1994, opened 2024-08-22
Bloodsucker leveling rework

Arturlang:
## About The Pull Request
Bloodsuckers now no longer level up from sol past level 4, requiring you to drink sentient blood to progress in levels. Also moves ventrue to this same system.
Blood thickening points are obtained 1-1 on blood drunk when you are drinking from players, these points are used for leveling, and the act of using these points for leveling does not use blood, nor does leveling in general.
Current blood cost is 40% of total blood cap, with 30% for ventrue.
This calculates based on the total blood volume a bloodsucker can hold, which at level 1 is 700, thus requiring 280 blood for a level up. At level 4 this is a cost of 440 blood. I'm tempted to increase the base requirement to 50%, but I'd like to hear on feedback on that
Also fixes some exploits with candelabrums, 
fixes https://github.com/Bubberstation/Bubberstation/issues/1700

## Why It's Good For The Game
Nobody likes bloodsuckers simply not having to interact with crew to constantly gain in strength, and just hiding and hiding in coffins to level up, this should heavily encourage bloodsuckers to drink blood (their main gimmikc) to gain levels, or you will simply be unable to level up.

## Changelog

:cl:
balance: The candelabrum is no longer as tough
fix: candelabrum now turns off when unsecured
balance: Bloodsucker no longer level from sol when they reach level 4
balance: Ventrue now can only use blood drunk from sentients to level up vassals if they wish to level themselves up.
refactor: Refactors how ventrue vassalizing is handled, no longer copying most of level up code for it.
/:cl:


Arturlang:
Please don't TM this until it's no longer a draft, this is NOT TESTED CODE yet

Arturlang:
Thinking, should sol leveling be removed entirely given stealth feeding exists

Arturlang:
Ready for testing once power refactor is merged and the resulting conflicts are fixed

Arturlang:
WIP until i fix refactor-related issues

EpicCharger:
I've only played a bit of bloodsucker and haven't looked at the code but something I was curious about is the blood conversion rates with this level system in mind. I feel like in the time I have played blood sucker it's a lot harder to increase blood levels than it is as say a hemophage. Example being I can nearly top off my 1000 unit blood on four humonkies without killing any of them, since each hemophage bite takes 50 units of blood and gives 50 units of blood. Where as when I played bloodsucker the same amount could only get me to about 600 to 700ish when it gave the warning of low blood. Since I haven't played it it's possible I'm missing understanding the blood warning and not taking as much, but it felt like the steal rate gave a lower amount than what was taken. If that's the case then it means to get levels you'd have to take a considerable amount of blood from players, possibly leading to more kills based off of just a need for blood alone. Not sure if that's desired.

Other thought is that if that is not the case and it is a 1 to 1 conversion, can bloodsuckers drink from other bloodsuckers? If so could they theoretically just get two bloodsuckers drinking off one another to grind out levels by drinking sentient blood as long as they have some non-sentient thing to get blood from to keep up with the normal drain? Just some thoughts that came to mind with the new system. Thanks for your time!

DatFrostys:
I understand the reason why this is made but I doubt this will foster more interactions. Most (and I say most because I know that not everyone is like that) will just stand idly behind others while they are busy just to get a bit of their blood. Not only that but there are several ways for the vampires to bypass that.
Even though bloodsuckers are strong, I don't believe they require a nerf at the moment.

(Unrelated but Ventrues can't level up past rank 3 rn)

Arturlang:
> (Unrelated but Ventrues can't level up past rank 3 rn)

Can you elaborate? Are they simply unable to use blood thickening points on a vassal?

Arturlang:
@EpicCharger Currently, drinking too much from one person will make you get less blood thickening points, this PR has removed some of the nerfs for drinking blood, but I'm considering to remove more of them as it doesn't contribute to leveling. And yes, it is desired that you attack other players and possibly kill them for blood, for the entire gimmick about bloodsucker is in fact, drinking player's blood.

Arturlang:
@DatFrostys Stealth feeding can be super obvious if you have some amount of awareness, I think it's fair to risk discovery like this, simply getting levels to become overpowered simply by existing is just way awful, and I've not been suggested any better alternatives.

Arturlang:
> > (Unrelated but Ventrues can't level up past rank 3 rn)
> 
> Can you elaborate? Are they simply unable to use blood thickening points on a vassal?

Found the problem and am fixing it, sorry

PR #2048, opened 2024-09-04
Re-removes buggy/defunct cargo import categories

YakumoChen:
## About The Pull Request

Cargo had the defunct "Allstar" and "Microstar" categories in cargo again, one selling MCR stuff (doubled) and the other selling a single ion rifle and nothing else?

## Why It's Good For The Game

Looks buggy

## Changelog

:cl:
fix: fixed the cargo categories again
/:cl:


projectkepler-ru:
augh
don't do it this way because you'll have to uninclude the file

YakumoChen:
Someone can just add the ion rifle to goodies if they want it buyable, I'd rather not we have guns in imports.

YakumoChen:
> augh don't do it this way because you'll have to uninclude the file

how do you want it done then? It worked last time I did it

projectkepler-ru:
> > augh don't do it this way because you'll have to uninclude the file
> 
> how do you want it done then? It worked last time I did it

IT doesn't work because this file only consists of comment which break linter
that's why I didn't just outright comment everything out because then I may aswell just delete the entire .dm

RimiNosha:
Yeah, delete all the code that's not in use. If someone wants to return them, git history exists for exactly this reason.

ReturnToZender:
> Someone can just add the ion rifle to goodies if they want it buyable, I'd rather not we have guns in imports where there's no access check.

/datum/supply_pack/security/armory/ionrifle
	name = "Ion Carbine Crate"
	cost = CARGO_CRATE_VALUE * 18 //Same as the energy gun crate
	desc = "Contains two Ion Carbines, for when you need to deal with speedy space tiders, mechs, or upstart silicons."
	contains = list(/obj/item/gun/energy/ionrifle/carbine = 2)
	crate_name = "Ion Carbine Crate"
From security.dm line 30

Majkl-J (changes requested):
Remove the entire files the companies probably wont come back

Majkl-J (approved):

PR #2050, opened 2024-09-04
Reverts the custom map rotation code because it seems to keep causing more problems than it solves

Maia-J:
## About The Pull Request

See name

## Why It's Good For The Game

Fixes empty votes being created, autorotation just picking a map because there's nothing else for it to pick, etc, etc

## Proof Of Testing

Werks fine on local

## Changelog

:cl:
del: Reverted map voting back to the tg system
/:cl:

### DIFF
--- a/code/controllers/subsystem/persistence/_persistence.dm
+++ b/code/controllers/subsystem/persistence/_persistence.dm
@@ -1,5 +1,5 @@
 #define FILE_RECENT_MAPS "data/RecentMaps.json"
-#define KEEP_ROUNDS_MAP 2 //BUBBERSTATION CHANGE: 3 TO 2.
+#define KEEP_ROUNDS_MAP 3
 
 SUBSYSTEM_DEF(persistence)
 	name = "Persistence"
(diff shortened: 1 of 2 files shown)


BurgerLUA:
Not being able to vote should've been fixed with https://github.com/Bubberstation/Bubberstation/pull/1916

Majkl-J:
> Not being able to vote should've been fixed with #1916

From today
image

projectkepler-ru:
It didn't JUST fail
it repeatedly tried to do a vote but because there was no map it just kept doing a re-vote

BurgerLUA:
pain, this is the first time I'm hearing about this

BurgerLUA:
I do not think an outright removal is the solution here. The map rotation tweaks were added for a reason. This prevents rounds from being meta/delta/icebox constantly and I have a feeling that behavior will continue if my changes are reverted.

Just fix the issue with lowpop and it will be fine.

BurgerLUA:
See #2054

Swiftfeather:
We'll see if burger's logic repair in #2054 works uwu

PR #2060, opened 2024-09-06
Most events have a less chance to run if certain departments don't have enough players, depending on the event.

BurgerLUA:
## About The Pull Request

Most events have a less chance to run if certain departments don't have enough players.
Spacevines and Abductors had their chaotic tag removed given the new system.

### Examples:

Meteor wave has an engineering department requirement of 4. The final chance for this event to run will be divided by 1 plus the amount of missing engineering department slots. 1 missing engineer means 50% less chance to run (1/2). 2 missing engineers means 66% less chance to run (1/3). 3 missing engineers means 75% less chance to run (1/4).

Anomaly Vortex has an engineering department and science department requirement of 2. The same logic above is applied, but  it will do it twice (one for the engineer department, and one for the science department). 1 missing engineer and 1 missing scientist means the event will have a 75% less chance to run (1/2 * 1/2).

Note that events with absurdly low weights, such as 1, will effectively be disabled if any slots are missing.

## Why It's Good For The Game

It's not fun for anyone when the server votes a storyteller like "The Bomb" and zero engineers end up playing. The station will usually go to shit because no one is willing to deal with the events. Instead of forcing players to suffer through the round, this PR will effectively remove those situations from occurring if there are not enough slots.

This PR will also make it so that joining a department with no players won't penalize you (too much) by still reducing the weight of certain events until the pop requirements are met.

## Proof Of Testing

Untested. I have no way to test this.

## Changelog

:cl: BurgerBB
balance: Most events have a less chance to run if certain departments don't have enough players. Spacevines and Abductors had their chaotic tag removed given the new system.
/:cl:


Majkl-J (changes requested):
Seems to break roundstart weights. These should ignore what crew types are around since there are no crew yet
image


Also this needs maintainer discussion as we're not entirely sure whether an extra layer of mechanical complexity like this is currently needed. Will keep you updated on how that goes

BurgerLUA:
It's better than outright removing certain storytellers for being too hard on departments (The Bomb).

StrangeWeirdKitten:
Alright I'll outline basically what I think and said when discussing this to be open with you. Don't mind the bullet points, cause it's just me organizing my thoughts.

- Some of these high destructive events already have low weights. Some of them only roll like 2 percent on a major track (which is around every hour and a half)
- I think this is an overcomplication of what minimum population requirements can achieve. If you set something to 50 minimum pop, there's someone bound to be able to fix a hole in the station. 
- People don't play engineering because it's boring after the 10th time you set up the tesla SM high power engine. Nothing happens sometimes. I think population requirements for things happening would just put engineering in a continued spiral down the wrong direction.
- We have a large silicon population and when things do happen, they tend to reset to engineering to fill the department.

This is default andy Major track.

image

BurgerLUA:
Heavily disagree with everything you said. I've had engineering mains beg to me to remove The Bomb and/or reduce the chances of crazy events spawning. They hate it when they're doing a big project and suddenly an event comes and destroys it. I even played engineering to see if this happens and it does.

BurgerLUA:
This doesn't make things complex. Dynamic had the same system, except it was much more complex than that. This system is extremely basic and almost required by any event rolling system worth its salt. It's not fun for anyone when meteors roll when there are 0 engineering staff.

StrangeWeirdKitten:
I will agree that story teller has massive issues in it, but something I forgot to mention was that we also shouldn't be working on adding complexity to an already flawed system. I'm already going to let #2064 have its chance. 

Also it adresses the maximum occurences of some of these destructive events.

BurgerLUA:
this isn't even complexity. why are you so set on a PR that literally improves quality of life on playing on the server, especially at lowpop. this barely even affects midpop and does not affect highpop.

Arturlang:
Mostly, the main issue is that people don't bother with jobs that's not part of their job description, even if there is crew, anyone can do the job, but the issue is they don't. Not sure how this helps the issue.

BurgerLUA:
> Mostly, the main issue is that people don't bother with jobs that's not part of their job description, even if there is crew, anyone can do the job, but the issue is they don't. Not sure how this helps the issue. Even if the issues are access-related, the captain can change jobs, and there's always at least a acting

When there are 0 jobs in a department, people don't tend to join to fill that job. 0 engineering stays 0 engineering for a while because no one wants to be the solo engineer that deals with everything. No one wants to be the solo security officer either. If you make playing as a solo department easier (less chance of events occurring because you're the only one in the department), then there is a higher chance of people joining as a solo role.

As for outside help: In my experience as omni-assistant, people are cautious about non-department people asking for departmental help because:
- There is a risk that the help could be low quality and do harm.
- There is a risk that the help could not actually be help and is in it to steal things (insuls/meds/etc).
- There is a risk that you could get a bwoink from an admin asking "Why did you let an assistant do brain surgery on a sec officer."

This is also assuming that people are willing to help. How many times have you seen "Hey we need engineers/shaft miners/doctors" etc and maybe about 1 person just signing up?

This PR is a 100% no brainer to me.

coffeecoffeecoffee:
Controversial opinion: Sometimes, having suffering rounds is fun and good and it's healthy for the server to encourage diversification. Please consider leaving it be.

StrangeWeirdKitten:
Story teller has tons of fundimental issues with it and building up on it is not something we want to do after discussion. I left this up for waterpig to decide and he doesn't want it.

PR #2071, opened 2024-09-09
Removes some old files that skyrat has now.

Swiftfeather:
## About The Pull Request

Tears these two out, they weren't doing much

## Why It's Good For The Game

I like to make as many .dme edits as possible

## Proof Of Testing

if it compiles that probably means it worked

## Changelog

I'm not gonna cl this because it's not player facing.


Majkl-J (approved):
Wasnt able to find any remnants myself

PR #2079, opened 2024-09-14
Cuts heretic weight

Swiftfeather:
## About The Pull Request

Cuts heretic weight from 8 to 3.

## Why It's Good For The Game

8 is the same weight as Spies, Bloodsuckers, and Lings. Heretic shouldn't be weighted that heavily.

## Proof Of Testing

If it compiles we're good

## Changelog

:cl:
balance: Reduced heretic weighting.
/:cl:


BurgerLUA:
This makes no sense to me. Heretic being common isn't a problem; the ascensions are.

StrangeWeirdKitten:
Yeah heretic needs to be lowered.

PR #2094, opened 2024-09-17
Removes The Clown storyteller making all event weights the same.

BurgerLUA:
## About The Pull Request

Removes The Clown storyteller making all event weights the same.

## Why It's Good For The Game

This is a feature many don't know about. I just learned about this feature looking through the code and this set off warning flags in my head.

The code sets all weights to 1 for each of the tracks. This means that meteors has the same chance as a electrical storm. I'm sure there is some nuiance I'm missing here (such as the interactions with thresholds), but that's still insane to me.

## Proof Of Testing

If it compiles, it works.

## Changelog

:cl: BurgerBB
del: Removes The Clown storyteller making all event weights the same.
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/storyteller/storytellers/tellers/storyteller_clown.dm
+++ b/modular_zubbers/code/modules/storyteller/storytellers/tellers/storyteller_clown.dm
@@ -16,9 +16,3 @@
 	threshold_major = 3200
 	threshold_crewset = 1000
 	threshold_ghostset = 3200
-
-// All the weights are the same to the clown
-/datum/storyteller/clown/calculate_weights(track)
-	for(var/datum/round_event_control/event as anything in SSgamemode.event_pools[track])
-		if(event.weight)
-			event.calculated_weight = 1


plsleavemealon:
I am pretty sure that is the point of the clown though.  Its supposed to choose any event at pure random with no care for how destructive (or not) that event might be.

Majkl-J:
> I am pretty sure that is the point of the clown though. Its supposed to choose any event at pure random with no care for how destructive (or not) that event might be.

It is stated in its description, and is a fairly recent change to make clown actually interesting as there's otherwise no reason not to pick a different high intensity teller.

Arturlang:
image
The entire point of the clown is that it sets the weights the same, it's even described in the storyteller description explicitly. If you don't like that, remove the entire storyteller instead, otherwise there's not much point to having it vs the others

Doom-external:
That's the point of the clown, and I personally love that it does that. It brings up events that aren't usually seen on other rounds, though maybe you could add some kind of extra cooldown on it being voted if you think it's rolled too often.

PR #2097, opened 2024-09-18
Replaces the Monkeycube Biogenerator recipe with monkey powder because I lagged once.

BurgerLUA:
## About The Pull Request

Replaces the Monkeycube Biogenerator recipe with Monkey Powder.
Instead of 50 biomass to make 1 monkey cube, it's 1 biomass to make 1 monkey powder (50 monkey powder makes 1 monkey when mixed with water).

## Why It's Good For The Game

If I had a nickel for every time I lagged out due to someone creating 50 monkey cubes on the biogenerator and then spraying water, I'd have four nickels, which isn't a lot but still weird.

This PR makes it harder to spam lag-inducing monkeys with little effort. You now have to actually put in effort if you want to spam 50 monkeys, either by doing xenobiology or by using a bluespace beaker.

This does unfortunately impact chefs. Instead of getting 5-8 monkey cubes for meat, they now have to bring 3-4 large beakers with them in order to harvest monkeys. However, I've been told that most chefs just print processed meat, or just get the botanist to grow meatwheat.

## Proof Of Testing

Draft because untested + feedback.

## Changelog

:cl: BurgerBB
balance: Replaces the Monkeycube Biogenerator recipe with monkey powder
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/research/designs/biogenerator_designs.dm
+++ b/modular_zubbers/code/modules/research/designs/biogenerator_designs.dm
@@ -1,0 +1,5 @@
+/datum/design/monkey_cube
+	name = "Monkey Powder"
+	materials = list(/datum/material/biomass = 1)
+	build_path = null
+	make_reagent = /datum/reagent/monkey_powder
--- a/tgstation.dme
+++ b/tgstation.dme
@@ -9125,6 +9125,7 @@
 #include "modular_zubbers\code\modules\reagents\reagent_containers\cups\glassbottle.dm"
 #include "modular_zubbers\code\modules\recycling\disposal\bin.dm"
 #include "modular_zubbers\code\modules\research\designs\ai_module_designs.dm"
+#include "modular_zubbers\code\modules\research\designs\biogenerator_designs.dm"
 #include "modular_zubbers\code\modules\research\designs\comp_board_designs.dm"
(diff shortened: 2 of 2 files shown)


BurgerLUA:
Full discloser; I thought that mixing monkey powder with water also gives all monkeys on the same tile, but it only does that if the reagent is mixed in a monkey. Maybe I can implement that too.

Majkl-J:
Considering we're a whitelist server I think this is an admin issue not a code issue. The few bad actors are gonna get filtered out if they keep doing it. +you can still make 50 monkey cubes in xenobiology pretty cheaply as well so I think if someone wanted to grief like that it won't even stop them

Doom-external:
The last person who did this got banned, even if you're an antag you're gonna get in trouble for doing this, literally no reason to spam monkeys

Swiftfeather:
> Full discloser; I thought that mixing monkey powder with water also gives all monkeys on the same tile, but it only does that if the reagent is mixed in a monkey. Maybe I can implement that too.

I will say the perpetrator of two of your four incidents is currently on an administrative leave, I agree with waterpig about this not really being a code issue.

StrangeWeirdKitten:
Per the above. I'll personally bwoink them myself if I see it.

PR #2101, opened 2024-09-20
Adds Global Event Cooldowns for Storyteller

BurgerLUA:
## About The Pull Request

Adds a "breathing room" feature for storyteller that prevents storyteller point gains if a chaotic event ran.

Adds minimum cooldowns for all track types, preventing two tracks firing in quick succession due to excessive points.

## Why It's Good For The Game

Minimum cooldowns and breathing room prevent the storyteller from suddenly waking up and deciding "Hey lets spawn blob, spiders, and meteors at the same time because we can now afford all these events now." While Blob, Spiders, and Meteors can still all occur in the same round, it just won't spawn them all within 1 minute of eachother.

## Proof Of Testing

Can't really test. Recommend testmerge.

## Changelog

:cl: BurgerBB
add: Adds Cooldowns for Storyteller
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/storyteller/_events/_event.dm
+++ b/modular_zubbers/code/modules/storyteller/_events/_event.dm
@@ -11,6 +11,8 @@
 	var/static/list/shared_occurences = list()
 	/// Whether a roundstart event can happen post roundstart. Very important for events which override job assignments.
 	var/can_run_post_roundstart = TRUE
+	//Prevents point gains from time for this amount of seconds when ran. Set this for high chaos events.
+	var/breathing_room_to_add = 0 SECONDS
 
 /datum/round_event
 	/// Whether the event called its start() yet or not.
--- a/modular_zubbers/code/modules/storyteller/gamemode.dm
+++ b/modular_zubbers/code/modules/storyteller/gamemode.dm
@@ -782,6 +782,7 @@ SUBSYSTEM_DEF(gamemode)
 				if(last_points)
 					next = round((upper - lower) / last_points / STORYTELLER_WAIT_TIME * 40 / 6) / 10
 				next = max(next,round(src.breathing_room / 6)/10)
+				next = max(next,round(src.next_track_event_run[track] / 6)/10)
 				dat += "<tr style='vertical-align:top; background-color: [background_cl];'>"
 				dat += "<td>[track]</td>" //Track
 				dat += "<td>[percent]% ([lower]/[upper])</td>" //Progress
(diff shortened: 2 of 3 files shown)


BurgerLUA:
Does this fix storyteller? Only one aspect of it, which is there not being a reasonable delay between events and allowing crazy shit to happen even on the calmest of storytellers.

What is basically happening is that storyteller saves up their points for all tracks and then all at the same time (pretty much), starts running events for each track, which results in the funny Blob into Meteors into Malf spawn.

BurgerLUA:
> The cooldown delays should be tied to storytellers themselves, and block points from being added into all the tracks instead of blocking events for running. Probably with a lower amount of time, suggesting defaults of none for mundane, 10 min for moderate, 20 min for major. It feels needlessly convoluted running several cooldowns at once, and makes it fairly hard to read.
> 
> Reason to move it to the points is that the current code it will just pile up more points in the back so you'll get unintended behavior of having tracks be filled to 500% due to everything else running before and blocking them (This was an issue with the mundane track on the original predictable chaos)

I can implement half of what you suggested. Problem is that blocking points gains does not stop already saved points from being spent.

BurgerLUA:
Multiple cooldowns I think is the best way to do it.

Single cooldowns means that all tracks won't be able to gain points, meaning that even with lower cooldowns, it is possible for major events not to occur because the amount of points it gets is extremely little.

Majkl-J:
> I can implement half of what you suggested. Problem is that blocking points gains does not stop already saved points from being spent.

Points are only ever spent when the track is full so at worse we'll have the rare case of 2 events syncing up perfectly which is already very rare with the new implementation and it's fine if that happens here and there

Majkl-J:
> Multiple cooldowns I think is the best way to do it.
> 
> Single cooldowns means that all tracks won't be able to gain points, meaning that even with lower cooldowns, it is possible for major events not to occur because the amount of points it gets is extremely little.

That is a fair point, let's see how this works as of current and we can work from there.

Majkl-J:
Also as for the ghost/crew rolls I think that maybe making the antag cap rise less / sec member could be of better help currently, antags are usually given equipment to fight off multiple seccies so having it 1sec = 1antag is probably dumb and why default andy sometimes feels nuclear. Feel free to make a new atomized PR that changes it to 0.5 cap for sec, unless I do it first once I get off lunch.

BurgerLUA:
> Also as for the ghost/crew rolls I think that maybe making the antag cap rise less / sec member could be of better help currently, antags are usually given equipment to fight off multiple seccies so having it 1sec = 1antag is probably dumb and why default andy sometimes feels nuclear. Feel free to make a new atomized PR that changes it to 0.5 cap for sec, unless I do it first once I get off lunch.

I removed that feature because the only reason it existed is because I needed to decide which gets priority when it comes to the "add antag" cooldown. This PR just does the cooldown thing.

PR #2102, opened 2024-09-20
[MIRROR] Fixes ballistic weapons not making any firing sound (#86250)

Maia-J:
## About The Pull Request

https://github.com/tgstation/tgstation/pull/85470 added some checks in `/obj/item/gun/ballistic/shoot_live_shot()` that were causing runtimes in probably most of the ballistic gun types. To see for yourself just take out a revolver and shoot it a few times, then check the runtimes log.

The runtime was preventing the parent
`/obj/item/gun/ballistic/shoot_live_shot()` from being called, which is where the firing sound gets played.

Fix is just acknowledging that `chambered` can be null at this point in the call chain and adding a quick check in there for good measure. Also fixes the runtime here, though that one didn't cause any functional bugs apart from polluting the log since it occurs at the end of the proc.

KZXgDrEAnV
 
Also adds a missing multiline list comma that I noticed cough.

## Why It's Good For The Game

Guns go boom again.


Swiftfeather:
oh thank god this was getting so annoying

PR #2127, opened 2024-09-25
Beepsky can no longer see potential targets if they don't move

BurgerLUA:
https://www.youtube.com/watch?v=NOs116aoyo0

## About The Pull Request

If you haven't moved or clicked for the past 4 seconds, Beepsky won't be able to see you, unless you're already being hunted by them.

Damaging Beepsky (somehow) without clicking on them will still agro Beepsky to you. This only affects passive target acquisition. 

## Why It's Good For The Game

Hilarious fix to the problem of Beepsky running into situations while security is talking it out with the criminal.

## Proof Of Testing

I tested it. It just works.

This should be testmerged in case there is fuckery with other interactions or other funny exploits that I did not consider.

## Changelog

:cl: BurgerBB
balance: If you haven't moved or clicked for the past 4 seconds, Beepsky won't be able to see you, unless you're already being hunted by them.
/:cl:

### DIFF
--- a/code/modules/mob/living/simple_animal/bot/secbot.dm
+++ b/code/modules/mob/living/simple_animal/bot/secbot.dm
@@ -493,6 +493,12 @@
 		if((nearby_carbons.name == oldtarget_name) && (world.time < last_found + 100))
 			continue
 
+		//BUBBERSTATION CHANGE START: BEEPSKY IS A DINOSAUR NOW. CAN'T SEE YOU IF YOU DON'T MOVE.
+		if(nearby_carbons.next_move + (4 SECONDS) <= world.time && nearby_carbons.next_click + (4 SECONDS) <= world.time)
+			//Holding still for 4 seconds.
+			continue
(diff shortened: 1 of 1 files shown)


Majkl-J (commented):
My main concern with this is that it is extended a deprecated mob type and is probably gonna cause a maintenance burden or get removed once beepsky becomes a basic mob

Majkl-J:
Don't get me wrong I love this but that will be a future issue we need to account for before we go through with a merge

BurgerLUA:
> My main concern with this is that it is extending a deprecated mob type and is probably gonna cause a maintenance burden or get removed once beepsky becomes a basic mob

If it gets removed it can jus get re-added again.

Majkl-J (approved):
> If it gets removed it can just get re-added again.

I will try to remember this and tell you when the day beepsky becomes basic comes then. Otherwise I like it so ill leave this open for people and other maints to chime in and if no significant issue is brought up it's all good

StrangeWeirdKitten:
You'll have to modify whatever beepsky ai targetting behavior when it does eventually become basic. But otherwise this is a good change.

PR #2139, opened 2024-09-27
Reduces the Ion Rifle to a 1x1 Heavy EMP

StrangeWeirdKitten:
## About The Pull Request

Per title. The Ion rifle does a 3x3 Heavy EMP around its projectile and has 12 shots. The ion carbine is the same (smaller version orderiable via cargo)

## Why It's Good For The Game

A good point of spill over was brought up due to the 3x3 heavy ion radius of the rifles. They're armor piercing and can instant stun cyborgs. It takes around 4 shots to knockover a synthetic crewmember. This will make the guns more precisely targetted weapons and way more viable to use against events like fleshmind and without EMPing your own gear or your buddy's equipment as often.

## Proof Of Testing

I am the test (it works fine)

## Changelog

:cl:
balance: Due to budget cuts, handheld ion weaponry will only EMP in a 1x1 radius. 
/:cl:


StrangeWeirdKitten:
I'll keep the armory (heavier version) a 3x3

PR #2153, opened 2024-09-30
[TM-ONLY] Begins the long process of resetting announcer sounds to be more like /tg/ because I got annoyed once

Swiftfeather:
## About The Pull Request

back to tg announcers

There are 3.5 years of changes to all these sounds so I'm going to wait until the upstream is complete before I finalize this for a merge. 

## Why It's Good For The Game

please save me from the mortal coil

## Proof Of Testing

if it compiles and TM's and works we know it works

## Changelog

:cl:
sound: Began reversion to tg announcements and sounds.
/:cl:


StrangeWeirdKitten:
We'll need to narrow down the problematic sounds and probally just level them through Audacity or something. This really isn't an improvement at all since all the personality left the announcer.

Swiftfeather:
That was the whole intent, and the volume hasn't really been the main issue since the original tms, it's how long some of the announcers go and the fact that some things that don't really need announcers have them

Swiftfeather:
I'll do more with these later

PR #2163, opened 2024-09-30
Remove eigenstate from scrubber overflow

lessthnthree:
## About The Pull Request

Updates scrubber overflow list, removing eigenstate, re-adds some of the easily cleanables (glitter, flour etc) that were removed, as storyteller now seems to be under control.

## Changelog

:cl: LT3
del: No more eigenstate in scrubber overflows
/:cl:


Razurath:
I don't think taking away the risks is a good thing for the server personally.

lessthnthree:
> I don't think taking away the risks is a good thing for the server personally.

Yeah, I was about to commit a change reducing it to just removing eigenstate. Adding eigen was a genuine error.

StrangeWeirdKitten (approved):

Swiftfeather:
Eigenstasium has always been fucky and cursing players with this particular chem is mean

PR #1223, opened 2024-02-28
RULE 1: Always keep the muzzle pointed in a safe direction

Swiftfeather:
## About The Pull Request

Removes gun safeties. BETTER FOLLOW RULE 1 OF GUN SAFETY BETTER NOW!

This is actually extremely noninvasive as far as uhhh uhhh uhhh modularity goes because it's just ONE file, all I had to do was disable the add component part of the proc but the proc is still defined so it doesn't error out on the guns that get the proc in other files.

## Why It's Good For The Game

I just find them really annoying.

Also the bigger thing, there's a very strong supermajority for removing it in the [suggestions](https://discord.com/channels/1059199070016655462/1071095123145924679/1212230353561849857) channel.

image

## Proof Of Testing

I loaded in, joined as captain, picked up all the guns in the armory and didn't see safeties on ANY of them.

## Changelog

:cl:Swiftfeather
del: Removed gun safeties.
/:cl:

PR #2170, opened 2024-10-02
Reverting PR #1223, which removed gun safeties

Boviro:
## About The Pull Request
This PR reverts Bubberstation/Bubberstation#1223 aka "RULE 1: Always keep the muzzle pointed in a safe direction", which disabled gun safeties on Bubberstation.
~~Does not add new content, solely re-enables gun safeties by reverting the PR that disabled them.~~
EDIT: Reverts the removal of gun safeties, guns also now disable their safety when you attempt to fire with an enabled safety.
~~Untested but it should work fine, as the original simply comments out the line that adds safeties.~~
EDIT: Locally tested, works as intended, compiles with no errors or warnings

## Why It's Good For The Game
Gun safeties are a good feature for preventing accidental misclicks and shooting a gun. On top of this, there is an RP aspect to disabling a safety during an interaction when one or more people have a gun. It's easy to ignore if you don't want to use it and quite helpful if you do want to use it. Along with this, there's an 80% majority for re-adding safeties on the #suggestions channel in the discord:
image
Ultimately I believe people want this to be re-added and for a mechanic that is ultimately optional I don't see a reason to deny it.

## Proof Of Testing
~~N/A, this revert shouldn't cause issues. I can locally test it to ensure it works if you would like.~~
Locally tested, functions as expected.

## Changelog

:cl:
add: Re-adds gun safeties
add: Guns with enabled safeties will disable their safeties if you attempt to fire with it on.
/:cl:


StrangeWeirdKitten:
If we're going to want this back, a straight revert isn't going to get merged, Safties need to get disabled after the first click.

Boviro:
If I may ask, why exactly would we need that feature?

StrangeWeirdKitten:
Because the primary issue with gun safties was due to their clunky nature on an item that's supposed to be responsive and is typically used in the heat of the moment where players trip up over it because they forgot to click the action button. I really do not want guns to be clunky and unintititive to use. They're click and shoot, and should be click and shoot, even if it means someone might ocassionally click and shoot by accident. 

I'm not really a fan of the action button either, since I really hate action button clutter- but it's not something I'm going to strictly die on.

Boviro:
FYI it has a hotkey as well (Alt-F if i recall) but I understand where you're coming from and I'll see if I can do something about disabling the safety when you fire.

Swiftfeather:
The hotkey was trash just like the whole safeties system was

Boviro:
Correction: the hotkey was actually Shift+F, new commit makes guns disengage their safety if you attempt to fire them with the safety on, locally tested.

shellspeed1:
Time for security and antags alike to fumble the gun safeties so we can laugh about it.

YakumoChen:
I never remember to use the hotkey for unholstering from the belt in the heat of the moment, I'm definitely not going to remember a hotkey so I can shoot my gun in the first place.

The only thing this will lead to is either:
a) people fumbling safeties in the middle of firefights again
b) people not fumbling safeties because they disengage safeties as soon as they pick up any gun, invalidating the entire point

Boviro:
This is why guns will now disengage their safeties once you pull the trigger while the safety is engaged. This prevents safeties from serving as a major hindrance in an emergency but should stop most friendly-fire incidents

Arturlang:
This is fine AS LONG as there's not a huge off action button for safeties, too much action button bloat

StrangeWeirdKitten (changes requested):
Have the safties off by default.

Boviro:
Honestly I don't think that makes sense. I mean safeties are already disabled when you click now so like, I don't see how having them off by default would make gunplay that much smoother. And this change would solve the issue of picking up a gun and just using  it, so why have the safety disable on click? I don't really think it makes sense to have both of them.

Boviro:
Like, I have a commit I can push at any time to switch it to being disabled by default, but I don't know if it's necessary to have both the self-disabling safeties and have them off by default. Kinda feels like it compromises their effectiveness too much.

Boviro:
Just one or the other does that fine, there's no need to have both

StrangeWeirdKitten:
Just do both, and it leaves it as an optional mechanic instead of forcing it on everyone who wants to use a gun. I want this mechanic to remain as uncumbering as possible.

- The mechanic then becomes optional.
- Guns set to saftey then don't impeed someone wanting to shoot said gun because it'll toggle off if they're spamming the click.

Boviro:
Okay I understand that but if you don't want to interact with the mechanic, then you're most likely never going to touch the toggle, making the click to toggle redundant. If you want to interact with the mechanic, then you're most likely going to prefer the extra safety of not having it turn off when you click once. So I don't see the point in having both of them.

StrangeWeirdKitten:
The requirement presented is it to be an opt in, not an opt out. Hence every gun having it off by default while anyone picking up a gun is still able to disable it (if it was disabled) and primarily so if someone does pick up a gun with safties on, they aren't spending valuable time clicking the trigger without it then disabling for them. I want weapons to remain responsive for everybody who comes across it while still allowing the people who may want the extra safety to then have the extra cushioning if they opt into having it.

Boviro:
Fair enough. Gun safeties are now all set to OFF by default

StrangeWeirdKitten:
Aka if it exists, it's important to be an action button. But that doesn't mean I won't entirely rework its location on the screen in the future when we break off from Skyrat.

Boviro:
That's fine, I mostly just want them back

StrangeWeirdKitten:
Np

modular_skyrat/modules/gun_safety/code/safety_component.dm
```diff
@@ -5,7 +5,7 @@
 	/// Holder for the toggle safety action
 	var/datum/action/item_action/gun_safety_toggle/toggle_safety_action
 
-/datum/component/gun_safety/Initialize(safety_currently_on = TRUE)
+/datum/component/gun_safety/Initialize(safety_currently_on = FALSE)
```
StrangeWeirdKitten, line 8:
```suggestion
/datum/component/gun_safety/Initialize(safety_currently_on = FALSE) // Bubber Edit - Org: True
```

modular_skyrat/modules/gun_safety/code/safety_component.dm
```diff
@@ -41,7 +41,8 @@
 	SIGNAL_HANDLER
 
 	if(safety_currently_on)
-		user.balloon_alert(user, "safety on!")
+		user.balloon_alert(user, "The safety disengages!")
+		toggle_safeties(user)
```
StrangeWeirdKitten, line 45:
```suggestion
		user.balloon_alert(user, "The safety disengages!") // Bubber Edit
		toggle_safeties(user) // Bubber Edit
```

PR #2175, opened 2024-10-03
Changes protogen snout and adds a new "cat" snout

Fn-OS:
## About The Pull Request

resprites the protogen snout to look like a visor, rather than a snout
Adds a new snout, called the Cat snout.

## Why It's Good For The Game
Makes the protogen visor eye candy and not doodoo dogshit, as well as adding a new "Cat" snout. More customization is always good.

## Proof Of Testing
image
image

## Changelog
Changes protogen snout sprite
Adds a new Cat snout.

:cl:
Added: New cat snout
image: added/modified/removed some icons or images: Changed protogen snout sprite. 
:cl:


Kurzaen:
About time, hell yeah.

ShockOmega:
wowoozay very nice

RadiantFlash7:
I'm not the biggest fan of the tajaran muzzle sprite change, to be honest. It's almost too circular, with the muzzle not being pronounced enough. It also makes tajaran seem less like their own thing and instead anthro cats- which I suppose they /are/ but historically they were always closer to lynx's then house cats, iirc.

Fn-OS:
> I'm not the biggest fan of the tajaran muzzle sprite change, to be honest. It's almost too circular, with the muzzle not being pronounced enough. It also makes tajaran seem less like their own thing and instead anthro cats- which I suppose they /are/ but historically in ss13 they were always closer to lynx's then house cats, iirc.
> 
> That said, I don't even use the old muzzle, because none of the tajaran markings actually work on it at the moment, as far as I can tell. So it's not like it's super great either.

I'm sadly nowhere near being even a decent spriter, this is really the best I could do. I haven't heard much about Tajaran being closer to lynx's but I don't know all too much of their lore either.

If I could, I'd make it a seperate sprite but this has already taken around ~3 hours and I don't want to fuck with VSC/Github even more.
It could be changed later on, but this is what I have for now.

Odairu (changes requested):
The sprites should be moved to .dmi files under the modular_zubbers folder (wherever we have sprite accesories in there, and if there isnt one, one should be made) and then overwrite the file path where the original sprites are used.

Conormackgegore:
You'll have to guide me through that later. Burnt out rn

Conormackgegore:
Oh I'm not even on the right account on mobile. Damn

StrangeWeirdKitten (changes requested):
For the Taj muzzle, make it a new muzzle type so people can pick it if they want to while still keeping the old one.

Fn-OS:
Okay, fixing this PR today. Old tajaran sprite will be kept, sprites will be moved to the correct folder and I'll try to figure out how to get them into the char creation menu

Fn-OS:
PR is fixed. Again, Tajaran sprite is NO LONGER being replaced per request

modular_zubbers/code/modules/customization/sprite_accessories/snouts.dm
```diff
@@ -7,3 +7,7 @@
 	name = "Cervine"
 	icon = 'modular_zubbers/icons/customization/snouts.dmi'
 	icon_state = "cervine"
+/datum/sprite_accessory/snouts/mammal/cat
```
lessthnthree, line 10:
```suggestion

/datum/sprite_accessory/snouts/mammal/cat
```

lessthnthree (approved):

PR #2178, opened 2024-10-05
asset cache [TM ONLY]

nevimer:
asset cache test, for faster joins


nevimer:
nevermind I found a better way lol

PR #2182, opened 2024-10-06
Partially Reverts PR#1915 (Heretic Sacrifice Doubling)

Mitryll:
## About The Pull Request

Does as the title explains, reduces the amount of targets from 6-10 to 5-6 (technically seven is the maximum based on certain pop requirements, but those are an edge case and will hardly ever come into play)

## Why It's Good For The Game

The original PR to double sacrifices was due to various issues with both the antagonist itself and systems surrounding it (i.e Storyteller making it as common as traitor and spy, two antagonists who are a lot less of a threat comparatively), most of which have been remedied by recent changes such as [PR#2079](https://github.com/Bubberstation/Bubberstation/pull/2079), the various nerfs to Path of the Exile, as well as recent policy changes encouraging Heretics to engage in roleplay rather than immediately go for their greentext. As it stands, ascension is defacto removed unless the player actively decides to ignore roleplay entirely and go for murdering an obscene amount of people, which just shifts the issue from ascensions to instead encouraging bad behavior. Heretics without the ascension is essentially just magic flavored traitor, which kind of kills the point of the antagonist, and most of the issues regarding the original PR were fixed or remedied in some regard.

## Proof Of Testing

Code compiles, game runs.

## Changelog

:cl:
balance: Set Heretic sacrifice target amount to 5-6.
/:cl:


Mitryll:
I forgot to update the description on the actual commit despite yelling at myself to do so. Woe. Everything else is as intended otherwise.

nikothedude:
the sacrifice doublign was done with good intention but its just another example of us pussyfooting around the real issue, being ascension itself and waht it offers

StrangeWeirdKitten (changes requested):
Decrease it to 5-6. I'm not considering a revert at this time

Mitryll:
> Decrease it to 5-6. I'm not considering a revert at this time

Yeah that's a fair compromise. Works.

PR #2184, opened 2024-10-06
[TestMerge Only] ITS A SPOOKY MONTH

IgiariValkyr:
## About The Pull Request

Makes the halloween event last all month, so players can play the creepy things like skeletons and vampires (non-hemophage or bloodsucker) and zombies (wait we have those already) and dullahans (this is mostly for dullahans). I think it also changes player ghosts too or something. I can't remember.

## Why It's Good For The Game

B87

## Proof Of Testing

It's a single line edit I did for Skyrat a few years ago too. It'll work just fine.

## Changelog

:cl:
qol: Spooky Month is now actually a month long!
/:cl:

### DIFF
--- a/code/modules/holiday/holidays.dm
+++ b/code/modules/holiday/holidays.dm
@@ -585,7 +585,7 @@
 
 /datum/holiday/halloween
 	name = HALLOWEEN
-	begin_day = 1 //bubberstation edit for spooky month - default: 29th of october
+	begin_day = 1 // BUBBER EDIT CHANGE - HALLOWEEN - ORIGINAL: 29
 	begin_month = OCTOBER
 	end_day = 2
 	end_month = NOVEMBER
--- a/code/modules/holiday/holidays.dm
+++ b/code/modules/holiday/holidays.dm
@@ -585,7 +585,7 @@
 
 /datum/holiday/halloween
 	name = HALLOWEEN
-	begin_day = 29
+	begin_day = 1 //bubberstation edit for spooky month - default: 29th of october
 	begin_month = OCTOBER
 	end_day = 2
 	end_month = NOVEMBER


code/modules/holiday/holidays.dm
```diff
@@ -585,7 +585,7 @@
 
 /datum/holiday/halloween
 	name = HALLOWEEN
-	begin_day = 29
+	begin_day = 1 //bubberstation edit for spooky month - default: 29th of october
```
lessthnthree, line 588:
```suggestion
	begin_day = 1 // BUBBER EDIT CHANGE - HALLOWEEN - ORIGINAL: 29
```

lessthnthree (changes requested):

lessthnthree:
I don't see why it has to be TM Only, though.

IgiariValkyr:
> I don't see why it has to be TM Only, though.

Cause I understand how bad the halloween code is, and having it active the whole month may cause unwanted lag and what not. And it may be decided later on to not be a permanent change. So marking it as TM only means we aren't making the full commitment and possibly forgetting we set it to all month in the event something gets out of hand.

lessthnthree:
> > I don't see why it has to be TM Only, though.
> 
> Cause I understand how bad the halloween code is, and having it active the whole month may cause unwanted lag and what not. And it may be decided later on to not be a permanent change. So marking it as TM only means we aren't making the full commitment and possibly forgetting we set it to all month in the event something gets out of hand.

I'd run it as a TM candidate like most PRs, it's a one line date change. PR title doesn't ultimately matter, though.

lessthnthree:
Superseded by https://github.com/Bubberstation/Bubberstation/pull/2211

PR #2185, opened 2024-10-06
Gravity generator blackout announcement/sound changes

lessthnthree:
## About The Pull Request

- Replaces the 47(!) second long Skyrat announcement for a gravity gen failure with something of a reasonable length
- Changes announcement to only play if it's a gravity generator blackout event. Normal gravity generator failures (such as no power) reverts to TG behavior, only playing a sound effect
- Removes the Skyrat priority announcements for the gravity generator losing charge/finishing its charging
- Replaces the TG default gravgen alert sound with gravgen spinup/power down sounds clipped from the old Skyrat announcement.
- Reduces the intensity of the screen shake, changing the duration to match the new sounds

## Why It's Good For The Game

Gravity generator failures make me angry because of how much god damn announcement spam and noise it makes for a comparatively minor station event.

Skyrat has it sending two Central Command priority announcements, Skyrat announcements on top of sound effects, none of it timed properly so you can have THREE different sounds all blasted at you at the same time if you're near the gravgen, and still two if you're anywhere else on the station.

Not to mention it's 47 seconds long, the gravgen can be half recharged by the time the announcement is finished if the AI is quick... then you get another long ass announcement for the gravity returning. That announcement too is overlapped with the grav gen alert effect if you're nearby.

## Changelog

:cl: LT3
qol: Gravity generator failures/power restores are now less announcement and sound spammy
/:cl:


modular_zubbers/sound/machines/gravgen_up.ogg
Swiftfeather, line 1: The audio ends kind of abruptly, I had the same problem in #2153 and couldn't figure out how to make a good reverb with audacity so I just included the announcer's "gravity generator functioning within normal parameters"

lessthnthree: Alright, I'll include that.

lessthnthree: Gotcha, replaced it with the version you made in 2153.

PR #2188, opened 2024-10-07
Lizard Gas Redux - Now with Ghost Role

IgiariValkyr:
## About The Pull Request

Remodels the interior of Lavaland and Icemoon Lizard Gas and adds a ghost role to all three stations!
That's right! The space one has a spawner too!
Lizard gas gets its own uniform!
<details>
<summary>Full change list</summary>

- A locker with a pipe rifle and hunter rounds, as well as flatpack machines so you can actually survive.

- A bathroom!

- Adjusted stock to more fit a gas station

- Lizard Gas has a purple uniform to match the plasma fueling station! (with digitigrade leg variant)
</details>

<details>
<summary>Visual comparisons</summary>

- Lavalands Gas Station
lavalandcomparison

- Icemoon Gas Station
icemooncomparison

- Space Gas Station
spacecomparison

</details>

## Why It's Good For The Game

Was made for a suggestion by Gearsin on discord.
https://discord.com/channels/1059199070016655462/1071095123145924679/1292324881962373191

## Proof Of Testing

dreamseeker_2024-10-07_16-43-16

## Changelog

:cl:
add: Lizard Gas can actually be staffed now!
/:cl:


theselfish:
I'm gonna sound insanely cynical but; No one will play this in like, a month. If that.


This just seems like the beach ruin but with even less to do.

IgiariValkyr:
> I'm gonna sound insanely cynical but; No one will play this in like, a month. If that.
> 
> This just seems like the beach ruin but with even less to do.

Well the issue with the beach ruin is it was added into the bitrunner server, but then never removed from lavaland rotation. But yeah, I get the concern. But since someone suggested it and it was pretty upvoted, I decided "fuck it sure I could do this".

Razurath:
The gas station spawns almost every round anyways, so why not give it a use where someone from dauntless or from station can wonder by, here's a space with air and storm protection, ashwalkers can wonder by and trade some, crew can use it for gimmicks and projects, and it's funny. A skilled player can even do their own gimmick with it like set up a tiny emergency med area. Sure the gas station may explode sometimes, but that's the risk of the ghost role. In short, even if it's not used a whole lot, we lose nothing by adding it.

YakumoChen:
This is incompatible with the codebase. You cannot make direct changes to maps sourced from upstream.

IgiariValkyr:
> This is incompatible with the codebase. You cannot make direct changes to maps sourced from upstream.
Already did and if they don't update it themselves, nothing will go wrong.

YakumoChen:
You make modular maps by making a new map and turning off the old one in config

Majkl-J:
Alternatively you can make automapper do this for you

IgiariValkyr:
As I thought, it thinks I changed the original map around a bit. I just replaced it with the map already in the repo which should be the original. I guess I can work a bit further on this as well and edit Icemoon Lizard Gas and Space Lizard Gas before we call for an actual pull this time.

IgiariValkyr:
Alright. That should be it. The maps are modularized (kinda). Icemoon's design matches the one on Lavaland. Space Gas keeps the same design it had, but the stock is adjusted to match the other two gas stations. Each one does spawn naturally as they should, though space gas seems to be rare since the lavaland one will spawn 9/10 times instead.

But unless there's some pressing things that need to be adjusted now, I'll call this personally done.

aKromatopzia:
yipee. updating will be required though

IgiariValkyr:
> yipee. updating will be required though

I never said I wouldn't keep up with it. Because I know there's things that will still change it if we don't pay attention to pull requests. Specifically being the config files, since i added the BUBBER markers to the other files so those don't get mistakenly overwritten (again) (but they probably will).

Majkl-J (changes requested):
You're much better off adding a single `/datum/area_spawn` for the automapper

```
/datum/area_spawn/lizardgas
	target_areas = list(/area/ruin/thelizardsgas_lavaland)
	desired_atom = /obj/effect/mob_spawn/ghost_role/human/lavaland_gasstation
	mode = AREA_SPAWN_MODE_HUG_WALL
```

IgiariValkyr:
I updated the original post with another dropdown showing differences between the original and the new versions of the maps. It's more than just adding a singular object and while I'm sure I can probably make them automappers, I feel like replacing one map file via code will be easier (on both myself and the server load time) in the long run.

code/modules/mob_spawn/ghost_roles/mining_roles.dm
```diff
@@ -360,3 +360,29 @@
...
+/datum/outfit/lavaland_gasstation
+	name = "Lizard Gas Station Attendant"
+	uniform = /obj/item/clothing/under/costume/lizardgas
+	shoes = /obj/item/clothing/shoes/sneakers/black
+	ears = /obj/item/instrument/piano_synth/headphones
+	gloves = /obj/item/clothing/gloves/fingerless
+	head = /obj/item/clothing/head/soft/purple
+	l_pocket = /obj/item/modular_computer/pda
+
+/// BUBBER EDIT ADDITION END
```
StrangeWeirdKitten, line 388: Can you modularize this into a zubbers folder?

StrangeWeirdKitten (changes requested):
Also there's no need to edit the config files, please remove those changes as well. I'll have to make the changes on the box.

code/modules/mob_spawn/ghost_roles/mining_roles.dm
IgiariValkyr: I really need a clear message from what you guys want from this, because now you're saying that stuff that wasn't wrong is now wrong, especially when skyrat does this in the same file. The exact same file.

SonicGotNuked: I just want this code to be inside modular_zubbers for future maintenance purposes so when this file is inevitably messed with, we won’t have to worry about it. The config itself doesn’t actually blacklist or unblacklist ruins- since it’s just a template

code/modules/mob_spawn/ghost_roles/mining_roles.dm
IgiariValkyr, line 1: Ok, why did this not have an extra linebreak anyways now that I'm looking at it? Like it's "reverted" but "not reverted"

code/modules/mob_spawn/ghost_roles/mining_roles.dm
IgiariValkyr: Let me see if this even still works now that I reverted and modularized further.

IgiariValkyr: It works. Anything else I should try to fix?

code/modules/mob_spawn/ghost_roles/mining_roles.dm
```diff
@@ -360,3 +360,4 @@
 	flavour_text = "Unfortunately, your hated enemy, Nanotrasen, has begun mining in this sector. Monitor enemy activity as best you can, and try to keep a low profile. Use the communication equipment to provide support to any field agents, and sow disinformation to throw Nanotrasen off your trail. Do not let the outpost fall into enemy hands!"
 	important_text = "Do NOT let the outpost fall into enemy hands"
 	outfit = /datum/outfit/lavaland_syndicate/comms/icemoon
+
```
StrangeWeirdKitten, line 363:
```suggestion
```

PR #2193, opened 2024-10-10
Succumbing now alerts everyone around you that you gave up on life.

StrangeWeirdKitten:
## About The Pull Request

Per title

## Why It's Good For The Game

This will clear up some confusion in tense moments. At the end of the day, if you succumb, it's not the other person's fault you died. (Security shouldn't be blamed for "executing" someone who succumbs)

## Proof Of Testing

I'm not going to have an easy time screenshotting a balloon alert. If it compiles, it works.

## Changelog

:cl:
add: Succumbing now alerts everyone that you gave up on life.
/:cl:


lessthnthree:
A balloon alert doesn't leave anything to bystanders in the text log, so it can still be a "he said, she said" about if they did or not.

StrangeWeirdKitten:
I could put a to_chat, but it could still be that game of he said she said regardless unless I was to rewrite succumbing a bit which I'm trying to avoid.

lessthnthree:
Yeah that's true. I'm just thinking of myself with the memory of a goldfish who can't remember what alerts popped up 5 seconds ago.

code/modules/mob/living/living.dm
```diff
@@ -565,6 +565,7 @@
 			to_chat(src, text="You are unable to succumb to death! This life continues.", type=MESSAGE_TYPE_INFO)
 			return
 	log_message("Has [whispered ? "whispered his final words" : "succumbed to death"] with [round(health, 0.1)] points of health!", LOG_ATTACK)
+	balloon_alert_to_viewers("has gave up on life...", "you have succumbed...", world.view) // BUBBER EDIT - ALERT OBSERVERS THAT THEY SUCCUMBED
```
projectkepler-ru, line 568: This is a biiiiit offensive

Swiftfeather:
But why? I don't see a reason for this and it's just gonna create more "hah they succumbed they're a poor sport/rager/loser/shitter" in people's brains, and worse yet it's gonna take that sentiment I have seen many times and make it ic-able too

code/modules/mob/living/living.dm
StrangeWeirdKitten: Yeah perhaps. I changed it.

StrangeWeirdKitten:
If you're going to succumb, you're going to succumb. But there's nothing wrong with communicating to people why someone in crit might have suddenly kneeled over and died and _that is not their fault_. Especially given the fact that you can do it immedietly in light crit.

ReturnToZender:
I like being able to whisper my final words in crit dramatically, but I _also_ like the idea of giving this an emote. Maybe not something so direct, but something that indicates it.

Like, a unique death text that only triggers on a succumb, such as "slips away from the mortal coil, finally going still." It doesn't need to explicitly _say_ succumb to indicate that's what's happening.

BurgerLUA:
> But why? I don't see a reason for this and it's just gonna create more "hah they succumbed they're a poor sport/rager/loser/shitter" in people's brains, and worse yet it's gonna take that sentiment I have seen many times and make it ic-able too

Exactly this. All this does is create drama/salt/toxicity.

thegrb93:
Imo I'm gonna succumb either way because I'd rather ghost around than watch my spasming body on the floor. Critting someone and killing them really shouldn't be treated any different.

Hemlock32:
> Imo I'm gonna succumb either way because I'd rather ghost around than watch my spasming body on the floor. Critting someone and killing them really shouldn't be treated differently, if that's what this is trying to address.

Putting someone into a critical condition and having them die after without intending to kill them, is, IRL, manslaughter, not murder... So... I disagree

thegrb93:
Killing someone at all without intending to is manslaughter, regardless of whether they were in critical condition for an extended period or not. I don't see your point.

My point is that injuring someone to the brink of death should be just as bad as killing them in the SS13 universe because in both cases you'll be equally revived and the actions leading to both are equally heinous.

thegrb93:
This would make deaths be treated differently depending on whether the 'this player succumbed' message pops up or not, which is meta and stupid.

Kepteyn:
> I like being able to whisper my final words in crit dramatically, but I _also_ like the idea of giving this an emote. Maybe not something so direct, but something that indicates it.
> 
> Like, a unique death text that only triggers on a succumb, such as "slips away from the mortal coil, finally going still." It doesn't need to explicitly _say_ succumb to indicate that's what's happening.

I do prefer this better, both as it's an emote and not a balloon alert and because it feels more in-character than an out-of-character alert.

PR #2289, opened 2024-10-27
The game can now compile in open dream

StrangeWeirdKitten:
## About The Pull Request

Refactors some code to allow the game to compile in OpenDream

Open dream is a rewrite of the entire game engine that supports all current servers.

https://github.com/OpenDreamProject/OpenDream

We aren't running open dream, it requires the ss14 launcher to run anyway. But we can pass our code through it because it gives some valuable errors like initials not doing anything or empty if else blocks.

## Why It's Good For The Game

We had a couple global proc overrides that didn't do anything because you can't override global procs. Some inital(variabls) on local varibles that returned the same value, and a macro not being used.

## Proof Of Testing

It compiles... in Open Dream.

## Changelog

:cl:
fix: Lizards that spawn with a random name will get a random ashwalker name now.
fix: Random generated humans will now have random vocal barks.
/:cl:

### DIFF
--- a/modular_skyrat/modules/customization/modules/clothing/under/accessories.dm
+++ b/modular_skyrat/modules/customization/modules/clothing/under/accessories.dm
@@ -98,7 +98,7 @@
 		if(istype(object, /obj/item/card/id))
 			id_card = object
 
-		if(ACCESS_SECURITY in id_card.access || (obj_flags & EMAGGED))
+		if(ACCESS_SECURITY in (id_card.access || (obj_flags & EMAGGED)))
 			to_chat(user, "You imprint your ID details onto the badge.")
 			set_name(user.real_name)
 			badge_string = id_card.assignment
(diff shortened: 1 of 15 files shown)


modular_skyrat/modules/customization/modules/clothing/under/accessories.dm
```diff
@@ -98,7 +98,7 @@
 		if(istype(object, /obj/item/card/id))
 			id_card = object
 
-		if(ACCESS_SECURITY in id_card.access || (obj_flags & EMAGGED))
+		if(ACCESS_SECURITY in (id_card.access || (obj_flags & EMAGGED)))
```
Shadow-Quill, line 101: shouldn't this be
```suggestion
		if((ACCESS_SECURITY in id_card.access) || (obj_flags & EMAGGED))
```

PR #2290, opened 2024-10-28
Buffs the Chill

ReturnToZender:
## About The Pull Request
This PR makes the Chill fire mundane events less frequently, as well as crew antags less frequently.

## Why It's Good For The Game
The Chill currently doesn't feel significantly different from the Default, only having a difference in Mundane events, which ultimately aren't the most significant actor.

## Proof Of Testing
WE DO IT LIVE

## Changelog

:cl: ReturnToZender
bal: Makes the Chill target crew antags, and spam slightly less mundane events.
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/storyteller/storytellers/tellers/storyteller_chill.dm
+++ b/modular_zubbers/code/modules/storyteller/storytellers/tellers/storyteller_chill.dm
@@ -15,7 +15,7 @@
 	storyteller_type = STORYTELLER_TYPE_CALM
 
 /datum/storyteller_data/tracks/chill
-	threshold_mundane = 900
+	threshold_mundane = 1800
 	threshold_moderate = 2700
 	threshold_major = 10000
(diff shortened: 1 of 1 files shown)


lessthnthree:
Here are the values for Fragile. If we want Chill to be quieter than Fragile, your adjustments to crewset and ghostset should be higher than these:
```
	threshold_crewset = 3000
	threshold_ghostset = 8000
```
Maybe try 3200 and make a ghostset var of 9000.

ReturnToZender:
> Here are the values for Fragile. If we want Chill to be quieter than Fragile, your adjustments to crewset and ghostset should be higher than these:
> 
> ```
> 	threshold_crewset = 3000
> 	threshold_ghostset = 8000
> ```
> 
> Maybe try 3200 and make a ghostset var of 9000.

done

Majkl-J (approved):

Issue #1969, opened 2024-08-18
Storyteller sometimes picks null event, throwing a runtime, creating flaky test


Shroopy:
## Issue Summary
The storyteller sometimes chooses a null event, throwing a runtime. This makes the storyteller CI tests flaky. Found during CI tests: https://github.com/Bubberstation/Bubberstation/actions/runs/10438090115/job/28905026930
```
[05:30:21] Runtime in modular_zubbers/code/modules/storyteller/storytellers/_storyteller.dm,109: WARNING: Storyteller The Guide picked a null from event pool at track Moderate.
    proc name:  stack trace (/proc/_stack_trace)
    src: null
    call stack:
     stack trace("WARNING: Storyteller The Guide...", "modular_zubbers/code/modules/s...", 109)
    The Guide (/datum/storyteller/guide): find and buy event from track("Moderate")
    The Guide (/datum/storyteller/guide): handle tracks()
    The Guide (/datum/storyteller/guide): process(20)
    Gamemode (/datum/controller/subsystem/gamemode): fire(0)
    Gamemode (/datum/controller/subsystem/gamemode): ignite(0)
    Master (/datum/controller/master): RunQueue()
    Master (/datum/controller/master): Loop(2)
    Master (/datum/controller/master): StartProcessing(0)
  Error: [05:30:21] Runtime in modular_zubbers/code/modules/storyteller/storytellers/_storyteller.dm,109: WARNING: Storyteller The Guide picked a null from event pool at track Moderate.
    proc name:  stack trace (/proc/_stack_trace)
    src: null
    call stack:
     stack trace("WARNING: Storyteller The Guide...", "modular_zubbers/code/modules/s...", 109)
    The Guide (/datum/storyteller/guide): find and buy event from track("Moderate")
    The Guide (/datum/storyteller/guide): handle tracks()
    The Guide (/datum/storyteller/guide): process(20)
    Gamemode (/datum/controller/subsystem/gamemode): fire(0)
    Gamemode (/datum/controller/subsystem/gamemode): ignite(0)
    Master (/datum/controller/master): RunQueue()
    Master (/datum/controller/master): Loop(2)
    Master (/datum/controller/master): StartProcessing(0)
  	FAILURE #1: [05:30:21] Runtime in modular_zubbers/code/modules/storyteller/storytellers/_storyteller.dm,109: WARNING: Storyteller The Guide picked a null from event pool at track Moderate.
[...]

PR #2319, opened 2024-11-02
Fix storyteller null event runtime

TealSeer:
## About The Pull Request

Fixes a semi-frequent runtime in storyteller code that causes CI to fail, see: 
https://github.com/Bubberstation/Bubberstation/actions/runs/11641923588/job/32420967445
https://github.com/Bubberstation/Bubberstation/actions/runs/11641923588/job/32420968238

for examples, this is caused by calculated weights for events often being non-whole numbers, which are then fed into `pick_weight` which is explicitly designed to only handle integers. 

Fixes https://github.com/Bubberstation/Bubberstation/issues/1969

## Why It's Good For The Game

One less cause of checks randomly failing.

## Proof Of Testing

Hard to prove lack of errors but I did a unit test that ran `find_and_buy_event_from_track` 200 times in a row and before this change it always failed and after this change it doesn't.

<details>

</details>

## Changelog

:cl:
fix: find_and_buy_event_from_track should no longer randomly fail.
/:cl:


lessthnthree (approved):

SonicGotNuked:
That being said, thank you.

PR #2349, opened 2024-11-06
Makes loadout renamed items able to be examined when worn

Maia-J:
## About The Pull Request

When an item is renamed due to loadout settings, it shows up as clickable on the examine of the person wearing it.

image

Clicking this link then makes the user examine this item further

image

## Why It's Good For The Game

Allows people's neat loadout item descriptions to be visible from an examine without any stripping needed. This means more reason to customize these items.

## Proof Of Testing

See above

## Changelog

:cl:
add: Loadout renamed/redescribed items can be examined when worn.
/:cl:


modular_zubbers/code/modules/mob/living/carbon/examine.dm
```diff
@@ -0,0 +1,6 @@
+/atom/proc/examine_title_worn(mob/user)
+	var/regular_examine = src.examine_title(user)
+	if(HAS_TRAIT_FROM(src, TRAIT_WAS_RENAMED, "Loadout")) // Uses /datum/element/examined_when_worn
```
Arturlang, line 3: It would be better if /datum/element/examined_when_worn gives the trait that examine_title_worn checks if it can examine, so the element can be nicely used on it's own instead of needing external implementation

Arturlang (changes requested):

modular_zubbers/code/modules/mob/living/carbon/examine.dm
Majkl-J: Done

modular_zubbers/code/datums/elements/examine_when_worn.dm
```diff
@@ -0,0 +1,16 @@
+// This is cheaper than adding to the Topic() of atom
+/datum/element/examined_when_worn
+
+/datum/element/examined_when_worn/Attach(datum/target)
+	. = ..()
+	RegisterSignal(target, COMSIG_TOPIC, PROC_REF(on_topic))
+	ADD_TRAIT(target, TRAIT_WORN_EXAMINE, TRAIT_SUBTREE_REQUIRED_OPERATIONAL_DATUM)
```
Arturlang, line 7: I don't see where you're removing the trait

Arturlang (changes requested):

modular_zubbers/code/datums/elements/examine_when_worn.dm
Majkl-J: Oh true 💀

Arturlang (approved):

ReturnToZender:
I like this

PR #2392, opened 2024-11-11
Fix issues with sol weakness and the quirk whitelist

Arturlang:
## About The Pull Request
Fixes issues with sol weakness quirk not giving you cheaper healing in a coffin and the quirk whitelist being buggy when changing characters and restarts.
Also makes the sol quirk disable healing when sol is active to save on blood, useful if you're hiding in a closet instead of a coffin.

## Why It's Good For The Game
Fixing bugs with old features is a good thing

## Proof Of Testing
Went ingame, made sure that the blood cost was correct and that quirks worked properly

<details>

</details>

## Changelog

:cl:
fix: Quirks with whitelists should no longer randomly dissapear
fix: Sol weakness quirk should actually make your healing cheaper if you're in a coffin
add: Sol weakness quirk will prevent hempohage healing working while Sol is active
/:cl:


Arturlang:
It does still show the status effect but it doesn't heal anything, i'm not sure if I should bother adding a case for that, bleh.

EpicCharger:
I dig the first part but it stopping healing while in a coffin and actively taking damage from Sol will almost certainly kill me since I've been on the edge of crit after Sol WITH locker healing, without it means it's curtains if you're already hurt and have no where else to hide.

Arturlang:
@EpicCharger My assumption was that you could survive with no healing in a locker if you werent already hurt, i can go and test it, and you should still be able to heal out of critical if you dont die

Majkl-J (approved):
I really tried but couldn't find anything

Arturlang:
> I really tried but couldn't find anything

Ha.

PR #2402, opened 2024-11-11
Refurbishes the Interdyne Research Facility ruin

ChromeFoxxity:
### This is only the first refreshed ruin, I'm considering going through the rest, and refreshing them as well.

Resurrects the previously very old ruin called the Interdyne Research Base. It looked out of place in the newer modern content, so I made it look much more modern, and less cramped. Now has worthwhile loot, and the place can actively power itself for quite some time unlike the previous version, which would run out of power, and struggle to keep it's systems powered. Due to so, it would turn the APC off and on as it tried to charge. It would also add some more loot, such as an RCD crate, a Prybar, a few good medkits, a Chameleon Hat, Blue Syndicate Documents, empty 357 speedloaders. While the corpses are also redesigned, and zombies have the actual uniforms of employees. The Engineers and Medical Director have combat boots and gloves. It also just entirely redesigns the whole place, with the same layout, just makes it so you don't have to go through every department room to get somewhere. Gives the lore a bit of a tweaking, makes some lore fluff, such as a grenade mishap in chemistry, with a charred skeleton, fire extinguisher on the floor after someone got bit, and turned. The events in the surgery room, two doctors getting bit, and patient zero being a unique zombie. Speaking of zombies, the uniforms and ID trims, and hud icons were created for specifically this ruin, just so it looks perfect.

### Credit to the creator of this ruin.

Old look.
image

SDMM look.
image

In-game look.
image

## Why It's Good For The Game

It would be better for the game in only design, as it would only be obvious for those who adventure space to find this location. This is also a possibility for better Roleplay if someone were to repair it to it's former glory. It'll be good!

## Proof Of Testing

image

## Changelog

:cl:
add: Refurbishes new Interdyne uniforms, and redesigns the Interdyne Research Base ruin.
/:cl:


Majkl-J (commented):
I think we're better off with fresh new bubber only ruins rather than refurbishing old ones. Keeps up a more diverse selection.

Would be cool to have stuff like a shot up/crashed remains of the dauntless in space to remember dauntless our beloved.

Also edits directly to TG icon files are unsupported so you'd need to move those

Majkl-J:
> I think we're better off with fresh new bubber only ruins rather than refurbishing old ones. Keeps up a more diverse selection.

After some further discussion, this and the other PR will be closed for that exact reason. Whilst it's really good work, it's also just refurbishing old TG ruins, and we'd rather see new unique ones

PR #2388, opened 2024-11-10
Dragonslayer Nerf (Pressurized Areas Only)

xPokee:
## About The Pull Request
_This is part one of my Marked One on-station nerfs._
Nerfs the dragonslayer considerably when in a pressurized environment (i.e. on-station). 

For reference the original was:
force = 20
wound_bonus = 10
armour_penetration = 35
block_chance = 25

While the new variables are:
force = 18
wound_bonus = 8
armour_penetration = 15
block_chance = 10
## Why It's Good For The Game
This will hopefully cut down on people powergaming as antagonists and validhunting with miner loot by making it less viable of a weapon on-station.
## Proof Of Testing
It compiled and ran, see the screenshot below for proof of the new stats updating.
<details>
<summary>Screenshots/Videos</summary>
image
</details>

## Changelog
:cl:
balance: nerfed the dragonslayer's stats when in pressurized environments
/:cl:

PR #2413, opened 2024-11-12
Revert "Dragonslayer Nerf (Pressurized Areas Only)"

StrangeWeirdKitten:
Reverts Bubberstation/Bubberstation#2388

Overcompensation and the sword got admittidely nerfed to the floor. Its original damage values were pretty reasonable, except with the 35 armor pen. But the issues come mostly in the minmaxing you can do with the armor as you can still do that with any other melee item.


BurgerLUA:
why remove instead of rework the damage values

assuming that the PR in question was too much of a nerf, this just reverts it to overpowerered status

Majkl-J (approved):

PR #2421, opened 2024-11-13
Lets anyone be a special ghoul type

Arturlang:
## About The Pull Request
Lets let anyone be a special ghoul, surely this can't go wrong

## Why It's Good For The Game
This was requested and as far as I can tell there's no reason to prevent this

## Proof Of Testing
not tested hehe
## Changelog

:cl:
balance: Races without the MOB_ORGANIC biotype are no longer disallowed from being special ghouls
/:cl:


ReturnToZender:
If you made this code and it's modular on us you don't need to comment it out, you can just delete it

Arturlang:
@ReturnToZender I felt like it was wierd to just leave it return TRUE so i opted to leave the comments for posterity, but i removed it

lessthnthree (approved):

PR #2425, opened 2024-11-14
Experimentally changes the thermal volume of space from 7000L to 25L

nevimer:
## About The Pull Request

This is mostly an atmos change- it makes space behave more realistic. As a consequence, your bone stock SM setup where you don't touch anything at all is no longer safe on three emitters. 
## Why It's Good For The Game

Space violating the laws of thermodynamics is silly, and well, I think everything atmos-wise having that as a crutch removed the upper gameplay ceiling. This shouldn't make basic atmos harder, thermal machines which still violate the laws are still fully functional. This just makes space itself behave more like an absence of anything, than a constant -270k cold source.

## Proof Of Testing
<summary></summary>

dreamseeker_pyTxK7Exmv
</summary>

## Changelog
:cl:
balance: space becomes a vacuum instead of a freezer
/:cl:

### DIFF
--- a/code/__DEFINES/atmospherics/atmos_core.dm
+++ b/code/__DEFINES/atmospherics/atmos_core.dm
@@ -124,7 +124,7 @@
 /// a hack for now
 #define WINDOW_HEAT_TRANSFER_COEFFICIENT 0.1
 /// a hack to help make vacuums "cold", sacrificing realism for gameplay
-#define HEAT_CAPACITY_VACUUM 7000
+#define HEAT_CAPACITY_VACUUM 1 // BUBBER EDIT: SPACE IS A POOR WAY TO REMOVE HEAT AND IT ISNT HOT OR COLD
 
 //FIRE
 ///Minimum temperature for fire to move to the next turf (150 °C or 433 K)
--- a/code/__DEFINES/atmospherics/atmos_core.dm
+++ b/code/__DEFINES/atmospherics/atmos_core.dm
@@ -124,7 +124,7 @@
 /// a hack for now
 #define WINDOW_HEAT_TRANSFER_COEFFICIENT 0.1
 /// a hack to help make vacuums "cold", sacrificing realism for gameplay
-#define HEAT_CAPACITY_VACUUM 1 // BUBBER EDIT: SPACE IS A POOR WAY TO REMOVE HEAT AND IT ISNT HOT OR COLD
+#define HEAT_CAPACITY_VACUUM 25 // BUBBER EDIT: SPACE IS A POOR WAY TO REMOVE HEAT AND IT ISNT HOT OR COLD
 
 //FIRE
 ///Minimum temperature for fire to move to the next turf (150 °C or 433 K)
--- a/code/modules/atmospherics/environmental/LINDA_turf_tile.dm
+++ b/code/modules/atmospherics/environmental/LINDA_turf_tile.dm
@@ -679,7 +679,7 @@ Then we space some of our heat, and think about if we should stop conducting.
 		return
 	// Heat should be positive in most cases
 	// coefficient applied first because some turfs have very big heat caps.
-	var/heat = CALCULATE_CONDUCTION_ENERGY(thermal_conductivity * delta_temperature, HEAT_CAPACITY_VACUUM, heat_capacity)
+	var/heat = CALCULATE_CONDUCTION_ENERGY(thermal_conductivity * delta_temperature, /*HEAT_CAPACITY_VACUUM*/ 7000, heat_capacity) // BUBBER EDIT
 	temperature -= heat / heat_capacity
 
 /turf/open/proc/temperature_share_open_to_solid(turf/sharer)


IgiariValkyr:
How does this affect heat exchangers in space for pre-built SM setups and atmospherics?

CenkTheRaptor:
yeah no. this is a horrible change frankly. i could see this working if ANY of the maps had systems in place to cool down the interior spaces but as is, any heat source that is unexpected fucks up the entire station atmos for HOURS. please, PLEASE let this cook longer before you rush a testmerge again.

Razurath:
One plasma fire in ordonance and the station was almost entirely unlivable and the temperatures went nuts for the entire shift. No room was holding temps for very long, the entire station was constantly getting to hot.

aKromatopzia:
I can see the merits of this PR, but if you're going to make space function as a vacuum rather than a coolant do understand that this doesn't just affect the SM. Killing heat flux to space causes the station to become a literal air fryer as temperature radiates outward through thermodynamics-violating materials. If this change is to stay, Heat transfer density through surfaces needs to also be mitigated. Like, please code in variable heat flux density per surface materials if you want to fix the laws of thermodynamics! Allow us to make insulated walls to keep the temperature inside an area! Change all the maps to include these new changes so engineering doesn't have to manually retrofit the entire map!

Also, a concern, does this code affect scrubbers?
`/datum/gas_mixture/turf/heat_capacity(data = MOLES)
	var/list/cached_gases = gases
	. = 0
	for(var/id in cached_gases)
		var/gas_data = cached_gases[id]
		. += gas_data[data] * gas_data[GAS_META][META_GAS_SPECIFIC_HEAT]
	if(!.)
		. += HEAT_CAPACITY_VACUUM //we want vacuums in turfs to have the same heat capacity as space`

Majkl-J:
> Also, a concern, does this code affect scrubbers? This snippet worries me a little
> 
> /datum/gas_mixture/turf/heat_capacity(data = MOLES) var/list/cached_gases = gases . = 0 for(var/id in cached_gases) var/gas_data = cached_gases[id] . += gas_data[data] * gas_data[GAS_META][META_GAS_SPECIFIC_HEAT] if(!.) . += HEAT_CAPACITY_VACUUM //we want vacuums in turfs to have the same heat capacity as space

Oh that might explain why scrubbers didn't seem to be doing much but I might be wrong

aKromatopzia:
> Oh that might explain why scrubbers didn't seem to be doing much but I might be wrong

I'm preeeeetty certain scrubbers weren't working because there was an active exothermic reaction (trit production, probably?) in the pipes, causing the distro to clog up. Back to room heating - Because simple exposure to space is no longer functionally a net-loss in temperature, heat is trapped within the station. But! The station doesn't have convection currents, everything is done through conduction via averages. What this means is that after an area becomes hot enough, even our magically-efficient HVAC/space heaters are nothing versus the almighty 500k room given time for the tiles to also heat up. Because floor tiles also need to cool. And guess what! I'm pretty sure axing HEAT_CAPACITY_VACUUM from 7000 to 25 makes tiles cool down literally 280x slower.

```/turf/proc/radiate_to_spess()
	if(temperature <= T0C) // Considering 0 degC as the break even point for radiation in and out.
		return
	// Because we keep losing energy, makes more sense for us to be the T2 here.
	var/delta_temperature = temperature_archived - TCMB //hardcoded space temperature
	if(heat_capacity <= 0 || abs(delta_temperature) <= MINIMUM_TEMPERATURE_DELTA_TO_CONSIDER)
		return
	// Heat should be positive in most cases
	// coefficient applied first because some turfs have very big heat caps.
	var/heat = CALCULATE_CONDUCTION_ENERGY(thermal_conductivity * delta_temperature, HEAT_CAPACITY_VACUUM, heat_capacity)
	temperature -= heat / heat_capacity

NargaInDisguise:
When we're able to cool the entire station at 0.3% of the rate prior, it's really really bad if theres ever a single heat producing issue. You have one person throw a SINGLE molotov and the entire station will raise 10 celsius.

Majkl-J:
Specifically for heat pipes it isn't 280x slower but about 10x due to how you get diminishing returns with large volume differences. But it does seem as if it's much harder to deal with fires

aKromatopzia:
Doing the maths, the result for HEAT_CAPACITY_VACUUM being 7000 (assuming the temperature of the tile is 293k) would be (((0.02*(293-2.7))*7000*20000))/(7000+20000)*1/20000, which comes out as a temperature delta of _-1.5052 (5sf) kelvins_. This is the passive temperature loss of an iron floor tile existing. When HEAT_CAPACITY_VACUUM is 1, this comes out as (((0.02*(293-2.7))*1*20000))/(1+20000)*1/20000, resulting in a delta of _-0.00019284 kelvins_.
 

But of course when the temperature is higher, say 500 kelvins...  HCV being 7000 comes out to a _-2.5785 kelvin delta_ and HCV 1 to _-0.00049727 kelvin delta_.

But you may say, the PR changes HCV to 25, not 1! So at HCV 25 it's _-0.012417 kelvin delta_ at 500k and _-0.0072484 delta_ at 293k.

In percentages, it looks like this: at 293k, passive heat loss to space goes down to 0.48% at HCV 25 and 25 0.013% at HCV 1, in comparison to HCV 7000;
and at 500k, it goes to at 0.48% HCV 25 and 0.019% at HCV 1 in comparison to HCV 7000

NargaInDisguise:
> Doing the maths, the result for HEAT_CAPACITY_VACUUM being 7000 (assuming the temperature of the tile is 293k) would be (((0.02x(293-2.7))x7000x20000))/(7000+20000)x1/20000, which comes out as a temperature delta of _-1.5052 (5sf) kelvins_. This is the passive temperature loss of an iron floor tile existing. When HEAT_CAPACITY_VACUUM is 1, this comes out as (((0.02x(293-2.7))x1x20000))/(1+20000)x1/20000, resulting in a delta of _-0.00019284 kelvins_.
> 
> But of course when the temperature is higher, say 500 kelvins... HCV being 7000 comes out to a _-2.5785 kelvin delta_ and HCV 1 to _-0.00049727 kelvin delta_.
> 
> But you may say, the PR changes HCV to 25, not 1! So at HCV 25 it's _-0.012417 kelvin delta_ at 500k and _-0.0072484 delta_ at 293k.
> 
> In percentages, it looks like this: at 293k, passive heat loss to space goes down to 0.48% at HCV 25 and 0.013% at HCV 1, in comparison to HCV 7000; and at 500k, it goes to at 0.48% HCV 25 and 0.019% at HCV 1 in comparison to HCV 7000
> 
> EDIT: What this means is that floor tiles _passively_ cool at extremely reduced rates. This is in the grand scheme of things, extremely minor! But this does mean that unless ENG staff go out of their way to use advanced methods to fix an overheated room, it's staying that way for a very long time. I hope my maths is wrong.

This explains exactly why it genuinely took us OVER AN HOUR to cool down ordnance after it plasma-fired for a couple minutes.

nikothedude:
what is it with you and making wildly unpopular "experiments" and having them _immediately_ test merged (looking back at the revenant """patch""")

cant we at least like. wait. and see if people like them

projectkepler-ru:
Interesting but this kinda kills passive cooling

YakumoChen:
> bone stock SM setup where you don't touch anything at all is no longer safe on three emitters

Do we really need to be making the SM harder to set up on a basic level?

RimiNosha:
Relaying what I've said in discord:

I'd be for this if it weren't for the fact that the only real thing this does is encourage thermomachine abuse. We need more interesting ways to cool stuff if you want this to be a real thing that encourages creativity.

YakumoChen:
> > > bone stock SM setup where you don't touch anything at all is no longer safe on three emitters
> > 
> > 
> > Do we really need to be making the SM harder to set up on a basic level?
> 
> specifically due to the backdraft effect of irl thermal radiation where the heat can reenter the exhaust. ie, the gas exchanging in space

Not sure what you're trying to reply but I don't think we should make the SM harder to set up than it is for basic bitch power supplies, especially when there isn't always an engineer or someone who wants to waste a fifth of the round setting it up just so the round is playable.

nikothedude:
can we please stop balancing with realism im so tired and it just make things worse

nikothedude:
why is this still test merged with the DNM label? cmon guys

aKromatopzia:
> Relaying what I've said in discord:
> 
> I'd be for this if it weren't for the fact that the only real thing this does is encourage thermomachine abuse. We need more interesting ways to cool stuff if you want this to be a real thing that encourages creativity.
> 
> I would also like to mention to those worrying about the station temperature that space heaters exist, and have been for the longest time the only easy way to regulate the station's temperature. Unless it's been removed in my absence, it cools/heats 5 tiles of air every atmos machine tick, which is 5x the normal.

It could well be something separate from the PR, but on the first round of its implementation HVACs and scrubbers were not working to expected efficiency. My assumption, and the reason for the maths earlier, was that it had something to do with the maths getting gunked up somewhere.

It's been some time now so if someone else has played rounds with this PR and has a counterpoint to make, do mention!

And also as mentioned, simply removing passive cooling loops will not encourage more creative design. Neither is covering this change as "realism". If realism is to be enforced, then assumptive simulations for the square-cube law and the ability to create cooling machines we have in reality but lack in game are required. More tools, not less, is what creates divergence instead of simple variance.

ReturnToZender:
> why is this still test merged with the DNM label? cmon guys

No longer TM'd as of an hour ago (maybe a little more)

nikothedude:
> > why is this still test merged with the DNM label? cmon guys
> 
> No longer TM'd as of an hour ago (maybe a little more)

not anymore its in again

ThatFunkyGuy:
Can definitely see this ruining lots of just general atmos things like tritium in the turbine and general BZ production. Why encourage more thermo machine abuse? theres nothing fun in having to make 15 thermo machines when space can do the same thing and not strain on the station.

Majkl-J:
Whilst this wasn't exactly station killing, fundamental mechanics should be kept in sync with TG unless you wish to write your own guide to the SM on the wiki or give players an ingame up to date guide

PR #2438, opened 2024-11-15
fix bokken wielded damage

Arturlang:
## About The Pull Request
Wierdly enough, bokkens revert to 8 force when wielded, and 16 damage when not, i'm not sure if this was intended or not but the description nor commits seem to indicate so, thus I'm setting the force of bokken to 8 and making wielded go up to the original 16 force when unwielded, as it seemed reversed beforehand

## Proof Of Testing

<details>

</details>

## Changelog

:cl:
fix: Fixes the bokken having 16 damage when unwielded, and 8 when it is
/:cl:


Majkl-J (approved):

SpringSkipper:
this was actually intentional; the wielded bokken has a higher block chance and deals nonlethal damage while the unwielded is the "offensive" stance that deals higher lethal damage but without the block chance.

PR #2446, opened 2024-11-16
Add a safety net for Manifold Sickness

TheGreatKitsune:
## About The Pull Request

Adds a safety net to Hereditary Manifold Sickness, which triggers if and only if it progresses all the way from stage one to stage four without _any_ visible symptoms having occurred. This safety net forces one of the stage 4 visible symptoms to occur, which is also a damaging effect that produces a sound effect (assuming the character has a set scream noise).
## Why It's Good For The Game

Complete round removal should not be able to blindside you. This does not guarantee it won't still blindside a player playing riskily with the quirk either, getting some stage two symptoms and then the sickness stealthing the rest of the way to stage 5 is still possible; and even with this safety net triggering, the sickness has a very low chance of going straight to stage 5 in the next few seconds anyways before you can get a pill in.
It is probably not immediately obvious to the player if this safety net occurs, as the visible effect is exactly the same as a normal symptom (You would only be able to tell by the fact that there are no messages in chat regarding the sickness between the last time you reset it with a pill and the stage four symptom occurring), it simply removes the possibility of the disease completely stealthing its way all the way to stage 5 with no symptoms at all by forcing a symptom to occur.
## Proof Of Testing

Ran code both with stage_prob set to 100 in order to test this code for when it should trigger, and then normally with the correct stage_prob to make sure the safety net doesn't trigger if a symptom does occur.
Again, no visible difference with what should be normal gameplay, this prevents an absence of something.
## Changelog
:cl: Cerami
qol: HMS / Chronic Illness no longer has the very low chance of going all the way from stage 1 to stage 5 without any symptoms.
/:cl:


code/datums/diseases/chronic_illness.dm
```diff
@@ -16,6 +16,7 @@
 	bypasses_immunity = TRUE//Bubber addition
 	severity = DISEASE_SEVERITY_UNCURABLE
 	bypasses_immunity = TRUE
+	var/being_stealthy = TRUE // Bubber addition
```
Erol509, line 19: Why are you uh, adding a new bolean var but it doesnt seem to do anything. You are still checking for exact stage of the disease

TheGreatKitsune: Stage is being checked for timing. This variable is updated to false if a visible symptom occurs during stages two or three. If it is still true when the stage rolls over to four, then no visible symptoms have occurred yet by stage four, and a visible symptom is forced to occur.

code/datums/diseases/chronic_illness.dm
```diff
@@ -86,3 +93,13 @@
...
 					qdel(affected_mob)
+
+//Bubbers addition - Costly warning if it sneaks through to stage four.
+/datum/disease/chronic_illness/update_stage(new_stage)
+	. = ..()
+	if(new_stage == 4 && being_stealthy)
+		to_chat(affected_mob, span_danger("[pick("You feel as though your atoms are accelerating in place.", "You feel like you're being torn apart!")]"))
+		affected_mob.emote("scream")
+		affected_mob.adjustBruteLoss(10, updating_health = FALSE)
+		affected_mob.updatehealth()
```
Majkl-J, line 104: adjustBruteLoss() already comes with update_health, any reason we'd need to split it up like this?
```suggestion
		affected_mob.adjustBruteLoss(10)
```

Majkl-J (changes requested):

code/datums/diseases/chronic_illness.dm
TheGreatKitsune: Yeah, that was a moment of me reading the code and it not fully clicking what was going on. Fixed.

PR #2447, opened 2024-11-16
Minor pref changes before I go utterly insane doing something else

Maia-J:
## About The Pull Request

Makes some pref code correctly use apply_to_human and dna. This means that if you switch mind around the prefs stay the same... I wonder why I'm adding that 

Removes deprecated headshot var from preferences. This used to be the universal one on skyrat but since we have one for humans one for silicons it was never really used in any place that couldn't be replaced by dna on here.

## Why It's Good For The Game

Makes the code more consistent. And more ready for certain *fun* activities

## Proof Of Testing

<details>

image

</details>

## Changelog

:cl:
code: Makes headshot and nsfw preference code handling more consistent with the rest. Savefiles are untouched by this
/:cl:


ReturnToZender:
Thanks for finally fixing this up ^^

PR #2456, opened 2024-11-17
xeno queen can't be KO'd anymore

Arturlang:
## About The Pull Request
ditto as per PR title, TG queens have this, but not skyrat's

## Proof Of Testing
mechs nor punches no longer KO queens, confirmed by testing

## Changelog

:cl:
add: Skyrat xeno queen can no longer be KO'd, just like TG queens cannot.
/:cl:

### DIFF
--- a/modular_skyrat/modules/xenos_skyrat_redo/code/xeno_types/queen.dm
+++ b/modular_skyrat/modules/xenos_skyrat_redo/code/xeno_types/queen.dm
@@ -5,6 +5,7 @@
 	desc = "A hulking beast of an alien, for some reason this one seems more important than the others, you should probably quit staring at it and do something."
 	caste = "queen"
 	maxHealth = 500
+	status_flags = NONE //can't shove or KO the queen, kiddo.
 	health = 500
 	icon_state = "alienqueen"
 	melee_damage_lower = 30


Arturlang:
> It might be considered to add to this, giving the queen's melee attack AP and maybe some damage boost. Right now they're damage 30 with no AP, which seems odd as they should be the strongest xeno especially if they get into melee range.

50 damage per slash, i think that's plenty

Razurath:
30 Damage isn;t much when there's no AP. Consider that a basic armor best is Melee 3 so you take 30% off that. Riot armor is melee 5, so it's halved meaning a Xeno queen is only doing 15 damage to a guy in riot armor when she slashes them. A crusher does more then that because it has AP. Even if the damage is unchanged, the queen's melee need Ap to get through armor both on players and on mechs. Because there is better armor then riot for melee fights.

Razurath:
Usually if a xeno queen in melee it's a massive fight in the nest and she is likely on her own. She is supposed to be a boss, not on her own a round ender, but she should hit harder.

Arturlang:
Mostly, what I think would benefit queen more would be stronger spits, I've honestly never had trouble killing people as queen with slashes itself, but spits do pitiful amounts of damage right now, 15, i think, on the acid mode? Or less

modular_skyrat/modules/xenos_skyrat_redo/code/xeno_types/queen.dm
```diff
@@ -5,6 +5,7 @@
 	desc = "A hulking beast of an alien, for some reason this one seems more important than the others, you should probably quit staring at it and do something."
 	caste = "queen"
 	maxHealth = 500
+	status_flags = NONE //can't shove or KO the queen, kiddo.
```
RimiNosha, line 8: This is not the fix. Please actually test your PRs.

You want to override the `get_shove_flags` proc and just return. If you want it to be the same as TG, look at their alien shove proc.

For stuns, use `add_stun_absorption` and use src as the source.

RimiNosha (changes requested):

modular_skyrat/modules/xenos_skyrat_redo/code/xeno_types/queen.dm
Arturlang: Setting status flags to NONE does in fact make it immune to KO's, but pushing still occurs.

Arturlang: One of the bitfields was CAN_PUSH, and thus I believed this to stop shoves as well.
image

Arturlang: Mech shoving uses step_away which is a byond builtin, it's unlikely that overriding get_shove_flags will help but I'll test it

Arturlang: Yep, get_shove_flags is used for right click shoves, which queen is already immune to

Arturlang: @RimiNosha xenos were already immune to stuns, but mech punches and human punches cause unconciousness, which is the issue here

RimiNosha: What the hell, I think my VSC search was bugged, then, cause all I saw was it being set to none and no uses. My bad.

StrangeWeirdKitten: this is a test

PR #2461, opened 2024-11-17
depressurizes your glucose injector

sippykot:
## About The Pull Request

Removes pressure check from the glucose medipen, and changes the description a bit.

## Why It's Good For The Game

Why are we gating glucose behind a pressure check like miner healing pens again? No one uses this thing.

## Proof Of Testing

Simple change, but I did actually test it on localhost.

## Changelog

:cl:
balance: removes pressure check from the glucose medipen and changes its description
/:cl:


Majkl-J (approved):
I cant believe they actually locked sugar behind pressure what were they smoking

PR #2509, opened 2024-11-21
VV Editing the storyteller name will append it with the original name.

StrangeWeirdKitten:
## About The Pull Request

Custom Story Teller Name (The Chill)

## Why It's Good For The Game

Let admins have their cake and eat it too. I was about to block the VV change entirely, but this is a more sane and better solution.

## Proof Of Testing

image

## Changelog
:cl:
admin: Storyteller name changes will append it with the original story teller name.
/:cl:


Majkl-J (approved):

Shadow-Quill:
Does this work correctly if the storyteller is changed mid round by an admin?

StrangeWeirdKitten:
Yes, because it's a seperate story teller. The name var is unique across each story teller datum and any edits to said name will append the story teller name still.

PR #2510, opened 2024-11-21
Reduces the amount of ruin Z levels

StrangeWeirdKitten:
We have memory issues and we need to start cutting down on it. I'd rather keep our ghost roles intact and the Interlink intact with this being the first solution.

Halves the amount of ruin Z levels that spawn. (7 to 4 by default)

### DIFF
--- a/code/__DEFINES/maps.dm
+++ b/code/__DEFINES/maps.dm
@@ -171,7 +171,7 @@ Always compile, always use that verb, and always make sure that it works for wha
 #define PLACE_ISOLATED "isolated" //On isolated ruin z level
 
 ///Map generation defines
-#define DEFAULT_SPACE_RUIN_LEVELS 4
+#define DEFAULT_SPACE_RUIN_LEVELS 4 // Bubber Edit - ORG: 7
 #define DEFAULT_SPACE_EMPTY_LEVELS 1
 
 #define BIOME_LOW_HEAT "low_heat"
--- a/code/__DEFINES/maps.dm
+++ b/code/__DEFINES/maps.dm
@@ -171,7 +171,7 @@ Always compile, always use that verb, and always make sure that it works for wha
 #define PLACE_ISOLATED "isolated" //On isolated ruin z level
 
 ///Map generation defines
-#define DEFAULT_SPACE_RUIN_LEVELS 7
+#define DEFAULT_SPACE_RUIN_LEVELS 4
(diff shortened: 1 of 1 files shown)


NargaInDisguise:
Will this overall decrease the amount of ruins to explore? If so that'll be a little sad.

Majkl-J:
> Will this overall decrease the amount of ruins to explore? If so that'll be a little sad.

Yes but the only other short term option is lowering the player cap or somehow getting lummox to make the engine 64bit

Swiftfeather:
> > Will this overall decrease the amount of ruins to explore? If so that'll be a little sad.
> 
> Yes but the only other short term option is lowering the player cap or somehow getting lummox to make the engine 64bit

we should get lummox to update it to 64 bit. That seems like the easiest choice :)

NargaInDisguise:
> > Will this overall decrease the amount of ruins to explore? If so that'll be a little sad.
> 
> Yes but the only other short term option is lowering the player cap or somehow getting lummox to make the engine 64bit

Augh...
Goodbye space exploration, you will be missed...

MajManatee:
Can we increase ruins per Z level slightly to try and compensate?

Hemlock32:
7 to 4 is an absurd decrease, as if space isn't already neutered and barren enough as is.

ReturnToZender:
> [Hemlock32](/Hemlock32)
WD: Server crashed! Rebooting...

StrangeWeirdKitten:
I haven't touched the ruin tokens so theoretically the same amount should spawn. 

I'll investigate later to see if this is the case.

StrangeWeirdKitten:
I increased the tokens from 10 to 16 for hopefully better ruin density. It's going to be monitored though, but there was a very substantial improvement on memory and ultimately we have to look at cutting *something*.

Cyprex:
👍 Pretty simple change, easily reverted once we get memory under control. Vaguely related- when I profiled init times w/ tracy it was shown that we spend a ton of time trying (and failing) to place down ruins, was this addressed at some point?

StrangeWeirdKitten:
> 👍 Pretty simple change, easily reverted once we get memory under control. Vaguely related- when I profiled init times w/ tracy it was shown that we spend a ton of time trying (and failing) to place down ruins, was this addressed at some point?

Yes that was when the config had 80 space ruin tokens in it. I killed that number (currently it's 16)

PR #2514, opened 2024-11-21
Modularity Cleanup: Quirks

xPokee:
## About The Pull Request
I moved all of the modular quirks from SR and Bubber to the same folder/filepath, previously they were split up in multiple different places.
## Why It's Good For The Game
It helps people when looking for things while coding or code diving due to them all being in one place.
## Proof Of Testing
It compiled and ran.
## Changelog
:cl:
code: moved a bunch of modular quirk files to the same folder
/:cl:


ReturnToZender:
We don't use veteran so if you want you can remove any code related to it. We've had a hacky fix that makes everyone count as a veteran since forever

nevimer:
> We don't use veteran so if you want you can remove any code related to it. We've had a hacky fix that makes everyone count as a veteran since forever

Probably out of the scope of this PR, but veteran checks are entirely un-needed.

Majkl-J (approved):

PR #2529, opened 2024-11-23
Brings back MCR recharging, rebalances the MCR progression techweb

Maia-J:
## About The Pull Request

This brings the MCR back to its old rechargeability, but also keeps the cell reloading as an alternative to allow you to last longer.

Additionally rebalances the techwebs to cost less and spreads the basic upgrades across researches that science normally does anyhow, hopefully making them viable in more rounds.

## Why It's Good For The Game

They were almost never used by anyone currently because the reloads were slower than any ballistic, and the cells themselves tended to be a massive waste of resources when you can just print boolet. Sure you could beg science for the attachment research, or buy them, but the benefit was not really there either so again, no reason to bother.

This hopefully makes the MCR a valid sidegrade to all the ballistics, as it also got some light buffs to some of its firemodes before but those were eclipsed by the terrible cell mess.

## Proof Of Testing

<details>

image

</details>

## Changelog

:cl:
balance: MCR recharging brought back
balance: MCR techwebs now cost less, with basic MCR research spread across standard techs
/:cl:


RimiNosha:
Could you round that percentage while you're at it? The extra .0000 precision is really unnecessary.

Majkl-J:
> Could you round that percentage while you're at it? The extra .0000 precision is really unnecessary.

Not on here no
https://github.com/Bubberstation/Bubberstation/blob/59eaebf62d66350bfccb2fd6ca3f5ba16eb682cc/code/game/machinery/recharger.dm#L56-L58

RimiNosha:
That's really weird, the percent define should round it to 0.1 itself... wtf is going on?

Majkl-J:
> That's really weird, the percent define should round it to 0.1 itself... wtf is going on?

I've noticed that byond's round gets really fucky when not rounding to 1 for some reason in the latest 515 releases could be that I can check on it (And get myself written into another set of patchnotes probably lol)

IgiariValkyr:
Wait, so you're re-adding the reload in the gun, but also readding where it removes all charge if the cell is removed as well? That's going to really make people not use these anymore. The point of the reload handles / tac-reload handles was to have spare cells ready to go if you didn't have the self-recharge attachment. Now it's just a glorified laser gun with attachments again, which is why Nova removed them in favor of the modular rifles.

Majkl-J:
> Wait, so you're re-adding the reload in the gun, but also readding where it removes all charge if the cell is removed as well? That's going to really make people not use these anymore.

I'm readding the ability to charge the gun and the cells with no need to get attachments.

Nobody used them before because any ballistic was able to reload faster and give better feedback when provided as-is. Not being able to recharge the cells or the gun cut into the usefulness even more because the resource cost of cells was worse than just getting your mags reloaded

> The point of the reload handles / tac-reload handles was to have spare cells ready to go if you didn't have the self-recharge attachment. Now it's just a glorified laser gun with attachments again, which is why Nova removed them in favor of the modular rifles.

You can still keep the cells and tac reload them, if you so wish. Being a glorified laser rifle is better than being a laser rifle running on what is basically ballistic mags but worse than ballistic in every way.

This is mostly a buff to MCRs because they had like 30 minutes of setup to maybe become half as viable as ballistics are when they arrive from cargonia

ReturnToZender:
Lost on what specifically this does to the MCR, I'll have to try it out locally and see what it's like. Enjoy some of the changes, not sure on others. I like that you can recharge the MCR in a static weapon recharger.

sippykot:
> Nobody used them before because any ballistic was able to reload faster

I saw plenty of people using them, its just that they are restricted to armoury gear and HoPs/Wardens are stingy with them.
Also the big weakness of ballistic is not being able to go thru windows.

FoxxyRenamon:
> > Wait, so you're re-adding the reload in the gun, but also readding where it removes all charge if the cell is removed as well? That's going to really make people not use these anymore.
> 
> I'm readding the ability to charge the gun and the cells with no need to get attachments.
> 
> Nobody used them before because any ballistic was able to reload faster and give better feedback when provided as-is. Not being able to recharge the cells or the gun cut into the usefulness even more because the resource cost of cells was worse than just getting your mags reloaded
> 
> > The point of the reload handles / tac-reload handles was to have spare cells ready to go if you didn't have the self-recharge attachment. Now it's just a glorified laser gun with attachments again, which is why Nova removed them in favor of the modular rifles.
> 
> You can still keep the cells and tac reload them, if you so wish. Being a glorified laser rifle is better than being a laser rifle running on what is basically ballistic mags but worse than ballistic in every way.
> 
> This is mostly a buff to MCRs because they had like 30 minutes of setup to maybe become half as viable as ballistics are when they arrive from cargonia


I use the MCR a lot, it's my favourite Lasergun since we don't have the AER platform....

PR #2533, opened 2024-11-24
Balances Godslayer Armor for Icemoon

IgiariValkyr:
## About The Pull Request

I know what you're immediately thinking this is a nerf for mining again. No. This one is actually a buff. Let me explain.
Godslayer Armor is supposed to be the penultimate reward for Icebox. It requires beating all three unique megafauna, the Demonic Frost Miner, the Wendigo, and the Clockwork Knight (as pathetic as it is). You would think that this armor would then be good and useful for the Icemoon... but no. It's not.

The armor stats are less than Drake Armor, and if this was Lavaland, that'd be fine. However on the Icemoon, Ice Whelps are exceedingly common at vents and in the wild, so you can have six or more suits of Drake Armor in a single round to just pass around to the crew because you can, given you likely have your seventeen lux pens and implanted legion core. **_This PR balances the armor stats to be the similar to Drake Armor, actually making it viable for fauna combat while still having poor bullet resistance for station conflicts._** This specific change means: higher melee protection (V -> VIII [50 -> 75 compared to Drake's 65]); lower bullet (III -> II [25 -> 15 same as Drake]), laser (III -> III [25 -> 30 compared to Drake's 40]), and energy (III -> III [25 -> 30 compared to Drake's 40]) protection.

But, I also went further with this PR. **_I've added innate Lavaproofing to the Godslayer Armor._** I felt it needed something else to make it stand out. No one wears the armor, so no one actually knows the suit also has a natural pressure seal, meaning it can be worn in space. But since there's no aspects where you'll ever be wearing this in space given that Icebox is a planetary map, so I wanted to make that pressure seal also be lavaproof, meaning the wearer can walk in the plasma river on the bottom layer of the map.

## Why It's Good For The Game

People may actually wear this unique armor again instead of making it, looking at the armor values, and tossing it aside.

## Proof Of Testing

<details>

Drake Armor base stats:
dreamseeker_2024-11-23_20-25-29

Edited Godslayer stats:
dreamseeker_2024-11-23_21-24-51

Drake Armor in lava:
dreamseeker_2024-11-23_20-28-06

Godslayer Armor in lava:
dreamseeker_2024-11-23_20-42-13

</details>

## Changelog

:cl:
add: Developments with Godslayer Armor have unlocked its true armor potential.
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/mining/equipment/explorer_gear.dm
+++ b/modular_zubbers/code/modules/mining/equipment/explorer_gear.dm
@@ -1,0 +1,19 @@
+/obj/item/clothing/suit/hooded/cloak/godslayer
+		clothing_flags = STOPSPRESSUREDAMAGE | THICKMATERIAL | LAVAPROTECT
+		resistance_flags = FIRE_PROOF | ACID_PROOF | FREEZE_PROOF | LAVA_PROOF
+
+/datum/armor/cloak_godslayer
+	melee = 65
+	bullet = 15
+	laser = 40
+	energy = 40
+	bomb = 70
+	bio = 60
+	fire = 100
+	acid = 100
+	wound = 10
+
+
+/obj/item/clothing/head/hooded/cloakhood/godslayer
+		clothing_flags = STOPSPRESSUREDAMAGE | THICKMATERIAL | SNUG_FIT | LAVAPROTECT
+		resistance_flags = FIRE_PROOF | ACID_PROOF | FREEZE_PROOF | LAVA_PROOF
--- a/modular_zubbers/code/modules/mining/equipment/explorer_gear.dm
+++ b/modular_zubbers/code/modules/mining/equipment/explorer_gear.dm
@@ -3,10 +3,10 @@
 		resistance_flags = FIRE_PROOF | ACID_PROOF | FREEZE_PROOF | LAVA_PROOF
 
 /datum/armor/cloak_godslayer
-	melee = 65
+	melee = 75
 	bullet = 15
-	laser = 40
-	energy = 40
+	laser = 30
+	energy = 30
 	bomb = 70
 	bio = 60
 	fire = 100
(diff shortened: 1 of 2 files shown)


StrangeWeirdKitten:
The armor literally aheals and teleports you from danger. People not wearing it is not an issue with the armor stats, and I don't see any reason to buff this to be even more powerful.

IgiariValkyr:
> The armor literally aheals and teleports you from danger. People not wearing it is not an issue with the armor stats, and I don't see any reason to buff this to be even more powerful.

Not sure where you see that it teleports you, because it doesn't. Give me a second to check how much it actually heals, cause it says 500, but it also doesn't pull any "revive" action, so if you're dead dead, it may not proc.

IgiariValkyr:
I did further testing with the armor. It will not revive someone already dead. It only revives people that hit crit status, so you can't revive someone by just putting the armor on them. The amount healed doesn't matter because a legion core will **also** fully restore when implanted. So while the armor actually is sensible and heals 500 damage over brute, burn, and oxy (in that order); the legion core just calls an **actual** aheal with owner.revive(HEAL_ALL). The armor will proc before the core, but that's still 10 minutes between heals; and if you use both get out of jail free cards trying to run away from megafauna (bubblegum, ash drake, and colossus are still all on icemoon at times), then the armor is not compensating enough for what's supposed to be the best thing on the icemoon.

What would you rather me do? Figure out how to code somehow so it can only revive in low pressure?

Arturlang:
@IgiariValkyr Honestly, it simply doesn't need comparable armor stats to ash drake when it has something as powerful as a full heal.

IgiariValkyr:
I have gone over this again, and decided that the current armor value it has are actually good for the Godslayer Armor. I personally wish it was a bit stronger, but balance is king and I wasn't thinking with balance. I was thinking with "strong miner need strong armor as final reward".

PR #2534, opened 2024-11-24
Ports telepathy quirk from Nova

ReturnToZender:
## About The Pull Request
Ports https://github.com/NovaSector/NovaSector/pull/603

Quote from the PR:
Telepathy has kind of sucked for a while in terms of actually using it as a dedicated communication method. This PR includes a full rework of telepathy into a point-targeting ability (goodbye list targeting) with a right-click override for sending quick messages to the last target you spoke with, and an emote reply (*treply) so you can keybind replies to something that doesn't involve map clicks at all.

In addition, telepathy now shows a runechat message both above your head (the sender) and above the recipient's head, allowing telepathy chat to be read entirely via runechat. It comes from the recipient's head for them because the messages are in their mind.

And of course, the whole thing is autopunctuated. :)

There's now also a 3pt Telepathic quirk that either activates telepathy if it's in your roundstart genome or adds it as an unremovable, no cost mutation if it isn't. Espers rejoice!

## Why It's Good For The Game
Quote from the PR:
This makes playing roundstart psychics/psionicists and a whole slew of other concepts possible, and it also makes telepathy SIGNIFICANTLY less ass to actually use. Mind conversations for everyone!

## Proof Of Testing

https://github.com/user-attachments/assets/48808344-6bb5-43d6-98ea-f751625e3406
## Changelog

:cl: yooriss (Code), ReturnToZender (port)
qol: The telepathy genetics mutation has had significant usability improvements: it is now point-targeting based, and right-clicking the power allows for quick resending to the same target, a feature mirrored by the new *treply emote. Telepathy now also shows runechat messages.
add: The Telepathic quirk has been added, allowing characters to start with an unremovable (or activated) telepathy mutation.
/:cl:


ReturnToZender:
Only change from the original PR is making the quirk cost points, as everyone being able to telepath at you at all times would be extremely annoying. Positive quirk cost variable, I set it to 3

aKromatopzia:
Everyone's out projecting the gay thoughts

RimiNosha:
There's no need to merge master unless you get a merge conflict, by the way.

ReturnToZender:
> There's no need to merge master unless you get a merge conflict, by the way.

merges master anyway cutely

modular_zubbers/code/modules/quirks/code/positive_quirks/telepathy.dm
```diff
@@ -0,0 +1,178 @@
...
+	spell_requirements = SPELL_REQUIRES_NO_ANTIMAGIC
+	antimagic_flags = MAGIC_RESISTANCE_MIND
+	cooldown_time = 1 SECONDS
+	cast_range = 7
+	/// What's the last mob we point-targeted with this ability?
+	var/datum/weakref/last_target_ref
+	/// The message we send
+	var/message
+	/// Are we blocking casts?
+	var/blocked = FALSE
```
Majkl-J, line 18: Ill be completely honest reworking telepathy into pointed is something I think even tg would like

ReturnToZender: Definitely agree. Might take this upstream after

PR #2536, opened 2024-11-24
Twitch can no longer lower your next move beneath 1, no more obscene embeds on reagent spears

ReturnToZender:
## About The Pull Request
This fixes two bugs. Quoting the player reported bug:

Exploit #1 - Twitch Next Move modificier

This one needs a bit of VV to test mostly to speed up the conditions, but its doable by a miner with a bit of skill.

You need the berserker armor of the chosen one, or alternatively, the Dexterous mutation and a way to remove it (I think you can do so with mutadone, but never tried, regardless!) and Twitch.
The short of it is that the next move variable is always edited with multiplicatives, but Twitch was coded with an aditive, so alternating one and another and then removing one effect and letting the other run out would net you with a lower than 1 next move variable.

Exploit #2  - God Slaying Spear

This one is easy to test, basically you all got a spear with better embeed stats than ninja stars, which embeeds most of the time thanks to its high embeed chance, plus its armor penetration, with no chance for it to fall out naturally so it has to be manually removed. Since it has a Bulky weight class, its damage multiplier is 4 plus the pain multiplier being 6, doing 24 damage per tick and on extraction, being able to kill a person that gets it in 10 seconds, even if armored.

We had a whole clique abusing this as icecats on nova, and even combining it with morphine.

https://github.com/Bubberstation/Bubberstation/blob/master/modular_skyrat/modules/reagent_forging/code/forge_weapons.dm#L136

## Why It's Good For The Game
Fixes some seriously game breaking bugs no one's abused yet, discovered on Nova,

## Proof Of Testing

<details>

</details>

## Changelog

:cl: OrbisAnima (code for 1), ReturnToZender (code for 2)
fix: Fixes an exploit with Twitch, and the reagent forged spear's embedding.
/:cl:


ReturnToZender:
Mom says it's my turn on the exploit.
image

OrbisAnima:
happy to help!

PR #2540, opened 2024-11-24
[SEMI-MODULAR] Multiple Loadout Presets!

RimiNosha:
## About The Pull Request

Tin. Also made some QoL tweaks while I was at it!

Comprehensively:
- Removes the defunct job checkbox and replaces it with a dropdown.
- Adds a preference update message, cause it's otherwise silent.
- Stops the TGUI prefs migration message from showing when changing to a brand new character.
- Adds the ability to create (nameable) loadout presets.
  - The default preset can't be renamed or deleted. This was to make my life easier, or I'd spend another day on this, which I'm not prepared to do.
- Automatically COPIES your old loadout to the new system's default loadout. The old preference value is ENTIRELY UNCHANGED, which should mean this is entirely safe. **Back up the server playersave data, regardless, though.**
- Fixes a funny bug where the blacklisted and whitelisted roles text was the same.

## Why It's Good For The Game

This is a thing folk have always wanted since the dawn of loadouts, and the new loadout system is *surprisingly* hackable, so I did it!

Now you can set up multiple outfits for whatever occasion you want!

## Proof Of Testing
<details>

image

https://github.com/user-attachments/assets/9d4bd6f3-0dc7-4aaa-a6ad-989fc9cb987a

The total allowed loadout count, added after the video, so here's a screencap instead!
image

Renaming

https://github.com/user-attachments/assets/236d99c2-1bcf-4836-b32c-4bb4dd6227e0

</details>

## Changelog
:cl:
add: Loadout presets! Make multiple loadouts, be able to switch between them in a couple of clicks!
qol: Replaced the defunct job checkbox on the loadout page with a working dropdown.
/:cl:


ReturnToZender:
RIMI I COULD KISS YOU

ReturnToZender:
image
image
Here's what it looks like in practice. Only issue I noticed is that you cannot rename a loadout without making a new one, but that would be helped with just an extra button. Overall, this is really good. Also the # you can make should be increased to 12.

RimiNosha:
> Job loadouts/different maps

Feasible, but I'm fast running out of space to put stuff in the UI. Also would be done in a different PR, if I do it at all.

> Renaming

Can do.

RimiNosha:
Also, can I make a special mention that I absolutely hate how fucking huge the loadout buttons are?

Majkl-J:
> Also, can I make a special mention that I absolutely hate how fucking huge the loadout buttons are? We're playing on computers here, not a 10 year old shamsmug knock off tablet.

MSO probably wanted them big because he's old and can't see properly

Majkl-J:
Would it be possible to make it so that only one preset is loaded on the pref datum at a time, similar to how character prefs themselves handle it? 12x+1 of the list per player doesn't exactly seem too nice. Though with how prefs work I'm guessing this would be impossible without a big rework.

Also, how does this affect prefloading in bitrunning? I would be able to test this on my main rig but sadly don't have that on hand and guest prefs are already kinda fucked

RimiNosha:
> Only one entry

I *could* add a var to the prefs datum that's handled by the index preference though, now that I've slept on it.

> Bit running

Non issue, whatever code worked before still works now.

Majkl-J:
> Non issue, whatever code worked before still works now, so long as it wasn't cringe and did some roundabout BS.

Gulps

modular_skyrat/master_files/code/modules/client/preferences_savefile.dm
```diff
@@ -13,6 +13,7 @@
 #define VERSION_UNDERSHIRT_BRA_SPLIT 4
 #define VERSION_CHRONOLOGICAL_AGE 5
 #define VERSION_LANGUAGES 6
+#define VERSION_LOADOUT_PRESETS 7
```
Majkl-J, line 16: Missing undef, and apparently the other two above it miss it too (That's my mistakes coming to bite back).

modular_zubbers/master_files/code/modules/loadout/loadout_menu.dm
```diff
@@ -0,0 +1,86 @@
...
+
+/datum/preference_middleware/loadout/proc/add_loadout_preset(list/params, mob/user)
+	PRIVATE_PROC(TRUE)
+	var/list/loadout_entries = preferences.read_preference(/datum/preference/loadout)
+
+	if (loadout_entries.len >= LOADOUT_MAX_PRESETS)
+		return TRUE
+
+	var/loadout_name = params["name"]
+	if (!istext(loadout_name) || length(loadout_name) >= LOADOUT_MAX_NAME_LENGTH || length(loadout_name) < 1)
```
Majkl-J, line 18: Shouldn't this be just a larger than?
```suggestion
	if (!istext(loadout_name) || length(loadout_name) > LOADOUT_MAX_NAME_LENGTH || length(loadout_name) < 1)
```

modular_zubbers/master_files/code/modules/loadout/loadout_preference.dm
```diff
@@ -0,0 +1,42 @@
...
+
+/datum/preference/loadout_index/apply_to_human(mob/living/carbon/human/target, value, datum/preferences/preferences)
+	return
+
+/datum/preference/loadout_index/create_informed_default_value(datum/preferences/preferences)
+	var/list/loadouts = preferences.read_preference(/datum/preference/loadout)
+	var/name
+	for (name in loadouts) // Lazy hack fraud
+		break
+	return name
```
Majkl-J, line 16: Why do we even loop here if we do nothing in the loop and just break out of it to return name?

modular_zubbers/master_files/code/modules/loadout/loadout_preference.dm
```diff
@@ -0,0 +1,42 @@
...
+
+/datum/preference/loadout_index/deserialize(input, datum/preferences/preferences)
+	if (istext(input))
+		return input
+
+	var/list/loadouts = preferences.read_preference(/datum/preference/loadout)
+	var/name
+	for (name in loadouts) // Lazy hack fraud
+		break
```
Majkl-J, line 25: ditto, also since you use this exact same thing in the create_informed_default_value proc you can just avoid copypasta by calling it directly as a fallback

tgui/packages/tgui/interfaces/PreferencesMenu/loadout/index.tsx
```diff
@@ -309,6 +474,18 @@ const LoadoutPreviewSection = () => {
...
+              <Dropdown
+                selected={data.preview_selection}
+                options={data.preview_options}
+                onSelected={(value) =>
+                  act('update_preview', {
+                    updated_preview: value,
+                  })
+                }
+              />
+            </Stack.Item>
```
Majkl-J, line 488: Missing edit end comment here

tgui/packages/tgui/interfaces/PreferencesMenu/loadout/index.tsx
Majkl-J, line 1: Comments look mostly okay, gonna ask @Arturlang for a proper review of the typescript

Majkl-J (changes requested):

modular_zubbers/master_files/code/modules/loadout/loadout_menu.dm
RimiNosha: Oh yeah

modular_zubbers/master_files/code/modules/loadout/loadout_preference.dm
RimiNosha: Because beyond lists are shit and you can't properly get keys without this hack.

modular_zubbers/master_files/code/modules/loadout/loadout_preference.dm
```diff
@@ -0,0 +1,39 @@
...
+/datum/preference/loadout_index/apply_to_human(mob/living/carbon/human/target, value, datum/preferences/preferences)
+	return
+
+/datum/preference/loadout_index/create_informed_default_value(datum/preferences/preferences)
+	var/list/loadouts = preferences.read_preference(/datum/preference/loadout)
+	// Explanation so I don't get more questions:
+	// You can't get a list of keys for assoc lists, so you have to iterate said list to get them.
+	// I only need the fist one, so I just immediately return on the first loop.
+	for (var/name in loadouts) // Lazy hack fraud
+		return name
```
Majkl-J, line 17:
```suggestion
	if(length(loadouts))
		return loadouts[1]
```
This is a bit faster and does the same thing without being hacky.

Snippet I used to make sure this is the case:
```
/proc/list_finder_run()
	var/list/L = list("a" = "a1","b" = "b1","c" = "c1")
	var/i
	for(i=0, i<=100000, i++)
		var/a = list_finder_ifcheck(L)
		var/b = list_finder_forcheck(L)

/proc/list_finder_ifcheck(var/list/L)
	if(length(L))
		return L[1]

/proc/list_finder_forcheck(var/list/L)
	for(var/name in L)
		return name
```

Majkl-J (changes requested):

modular_zubbers/master_files/code/modules/loadout/loadout_preference.dm
RimiNosha: I love undocumented functionality!

Majkl-J (approved):

PR #2541, opened 2024-11-24
Makes the Shotgun PKA Two Handed

OrbisAnima:
## About The Pull Request
Makes the Shotgun two handed to use.

## Why It's Good For The Game
It prevents the abuse of akimbo shotguns, letting you do 60 explosive damage on station with little investment, while still mantaining the shotgun as a viable pka.

## Proof Of Testing
<details>

Message when you dont have both hands free:

image

</details>

## Changelog
:cl:
balance: Made the shotgun PKA two handed.
/:cl:


Hemlock32:
Aren't damage mods universally considered kinda mediocre on the base PKA? Not sure how they are with variants, but, I recall that being how they were perceived with the base one

IgiariValkyr:
> Aren't damage mods universally considered kinda mediocre on the base PKA? Not sure how they are with variants, but, I recall that being how they were perceived with the base one

Just as well, the shotgun and repeater have less mod space, so you can only fit two mods into them usually since they have mod spaces of 75% with the mods themselves taking up 30% (i never understood this)

NargaInDisguise:
Please do not nerf these, the shotgun and repeater already have less mod space to compensate for the fact that mods are stronger.

OrbisAnima:
please do not despair, this is just making it so the damage mod is not a no brainer for these two, we have two alternatives now with the proper arragement.

If we believe the damage modkit is too weak compared to a cooldown one, we can increase its base value so its buffed _all across the board_

If you all think the pka variants would need more complexity, that can also be buffed, my guess to the 75% value is to have 2 modkits of 30 and then the trigger guard.

Do consider both PKA guns change some positioning for doing 50% more base damage (regular pka does 40 damage, these both do 60 when all is connected, shotgun when its on melee pointblank, repeater when you rapid fire it, thus the extra 1 range)

So, what would be the preffered buff if these are considered weak? to the damage kit or +30 to the complexity for all variants?

OrbisAnima:
> Aren't damage mods universally considered kinda mediocre on the base PKA? Not sure how they are with variants, but, I recall that being how they were perceived with the base one

in their current state, 1 damage mod ups the damage by 10, for these variants it ups them to every shoot, so compared the damage increase of a regular pka damage goes 25% (from 40 to 50) while on these it goes 50% (from 60 to 90), the idea of this change was to keep that relationship so its easier to adjust things all across the board

Arturlang:
The issue here is that the shotgun and repeater benefit way too much from the damage upgrades, as the damage modkit gives a FLAT upgrade, which is a issue when these KPA's have multiple or way more projectiles than the basic PKA

Arturlang (approved):

Arturlang:
For example, shotgun PKA with 2 damage upgrades does 120 damage per shot, up from 60, repeater does 120 up from 65. This is really funky for one volley/shot.

NargaInDisguise:
> For example, shotgun PKA with 2 damage upgrades does 120 damage per shot, up from 60, repeater does 120 up from 65. This is really funky for one volley/shot.

On the shotguns defense, it has an incredibly long cooldown without a single cd mod because of the weird way those work, and has practically no range for actually hitting all three shots. The repeater is a bit jank though..

Majkl-J (changes requested):
Just make the shotgun twohanded instead and give them a 0.25 pressure_decrease.

OrbisAnima:
sure, but, they already have a 0.25 decrease

OrbisAnima:
> Just make the shotgun twohanded instead and give them a 0.25 pressure_decrease.

all done, reverted the changes and made it heavy (aka, two handed), pressure decrease is already that value for all pka's but the railgun (which is 0.1)

Majkl-J (approved):

PR #2542, opened 2024-11-25
Repurpose Bokkens

OrbisAnima:
## About The Pull Request
Changed the bokken damage to Stamina, made its block chance behave like any other weapon,  its too strong of an ability to block proyectiles as we need implants to do so and this was giving it for free for just 4 wood, it still has its decent block chance, just not against guns and lasers. Damage is 17 unwielded, 25.5 wielded, and block chance is 20 and 40 respectively, it is flammable since its a wooden sword. It does not work on mobs as its stamina damage (and has a message to indicate so).

Balance wise the two handed pillow with a brick was used as a guideline, which sits at 20 stamina damage.

Also, made the **wooden** sword flammable.

## Why It's Good For The Game
It makes a solid if albeit not OP nonlethal option people can use to fool around, you can finally have duels where getting trauma or dying from blood loss is not needed, thus improving RP slightly and also having a good, choice for going at melee which is not a death threat, with its 40 block chance its comparable to TG  Bamboo Staff (/obj/item/bambostaff, block chance 45, 10 brute one handed, 14 two handed)

And I think its good that the wooden sword burns instead of being impervious to fire.

## Proof Of Testing
<details>

</details>

Numbers change, but for stamina crit the numbers are:

to unarmored: 8 hits one hand, 6 hits two hand.
to armor IV + sec uniform: 9 hits one hand, 6 hits two hand.
to armor V + captain uniform: 10 hits one hand, 7 hits two hand.

## Changelog
:cl:
balance: 'Forged' Bokken, the wooden sword that can be made on the workbench with no anvil or fuel, its now flammable, doesnt affect mobs, and does entirely stamina damage. It has 40 AP, in one hand it does 17 damage and has a block chance of 20, while wielded does 25.5 damage and has a 40 block chance. It no longer blocks proyectiles, it behaves like any other sword in the game albeit with a high block chance.
/:cl:


OrbisAnima:
> can you harmbatton with it?

not on their current state, it is a very good idea though

OrbisAnima:
> Note down in the PR that you are removing it's ability to block projectiles. Or wait, are you? I'm not sure, reading the code, assuming that it still keeps it, right?

I will make it more clear on the change notes, but yes, it behaves like a regular weapon with its block chance and no longer blocks proyectiles a regular weapon wouldn't block (so it wont block lasers and bullets anymore, but I cant remember if regular weapons block throw items, if it does, then this does too, as its standard)

PR #2543, opened 2024-11-25
Lowers lit cigarette heat temperature, fixes grinding dried cannabis to make hash.

fludd12:
## About The Pull Request

Lowers the heat variable on a lit cigarette to 415, and fixes the broken `on_grind()` behavior for dried cannabis.
## Why It's Good For The Game

Currently, the lit cigarette target temperature is 1000K, standard for most flammable objects. Problematically, THC condenses into dabs at 420K, and space drugs distill into fentanyl at ~700K, so if you're rolling literally anything other than tobacco you're going to have trouble after a few seconds, either way.

Additionally, the behavior for dried cannabis creating hash when ground was handled with the `on_grind()` proc in an exceptionally weird way. For almost literally anything else, this would be an appropriate proc! However, the `grind()` proc is overridden for all subtypes of `grown` to modify reagent quantities based on nutriment production. Since cannabis doesn't contain any nutriment, `on_grind` never produced any `hashish` reagent to condense into hash.
## Proof Of Testing

https://github.com/user-attachments/assets/c882ed32-ea00-4338-90cc-5a2cba3e8007

</details>

## Changelog
:cl:
balance: Cigarettes only have a heat of 410K. Joints will no longer boil out their own THC or turn space drugs to fentanyl.
fix: Grinding dried cannabis now makes hash as originally intended.
/:cl:


fludd12:
I made a character who makes frequent use of these mechanics and its been pissing me off for a full month which I guess is a good way of making me do things!

modular_skyrat/modules/morenarcotics/code/thc.dm
```diff
@@ -1,8 +1,12 @@
-/obj/item/food/grown/cannabis/on_grind()
-	. = ..()
-	if(HAS_TRAIT(src, TRAIT_DRIED))
-		grind_results = list(/datum/reagent/drug/thc/hash = 0.15*src.seed.potency)
-		reagents.clear_reagents() //prevents anything else from coming out
```
Majkl-J, line 5: The reason this borks is because the math relies on there being stuff (specifically, nutriment) inside the plant to make stuff. Which skyrat, in their infinite wisdom, removes with the `reagents.clear_reagents()`.

Update: Actually turns out this code isn't even needed - normal cannabis already contains only thc when dried so this just.. tries to do the exact same thing that would've already happened.

modular_skyrat/modules/morenarcotics/code/thc.dm
```diff
@@ -1,8 +1,12 @@
...
+/obj/item/food/grown/cannabis/grind(datum/reagents/target_holder, mob/user)
+	if(!HAS_TRAIT(src, TRAIT_DRIED))
+		return ..(target_holder, user)
+
+	if(reagents && target_holder)
+		reagents.add_reagent(/datum/reagent/drug/thc/hash, 0.15 * seed.potency)
+		reagents.trans_to(target_holder, reagents.total_volume, transferred_by = user)
+
+	return TRUE
```
Majkl-J, line 9:
```suggestion
```

Majkl-J (changes requested):

modular_skyrat/modules/morenarcotics/code/thc.dm
Majkl-J: What the fuck. Well, I guess you could try and rewrite this into
```
reagents.clear_reagents()
reagents.add_reagent(/datum/reagent/drug/thc/hash, 0.15 * seed.potency)
```
then, should work about fine hopefully

modular_skyrat/modules/morenarcotics/code/thc.dm
fludd12: On my phone at the moment, but if you're suggesting removing the grind override, see previous comment - hash production is a uniquely stupid chemical reaction meant to allow decimal contribution to making the hash item by individual products.

modular_skyrat/modules/morenarcotics/code/thc.dm
fludd12: I can't believe I didn't think of that. I guess I got so in my own head about the stupid way they did it. Will amend when home.

fludd12:
Swapped behavior back to `on_grind`, but with less stupidity. Updating PR body to reflect now.

fludd12:
Let me know if there are any further reviews!

code/game/objects/items/cigarettes.dm
```diff
@@ -142,7 +142,7 @@ CIGARETTE PACKETS ARE IN FANCY.DM
 	w_class = WEIGHT_CLASS_TINY
 	slot_flags = ITEM_SLOT_MASK
 	grind_results = list()
-	heat = 1000
+	heat = 415 //BUBBER EDIT CHANGE - Limits heat exposure to prevent condensing THC or turning space drugs to fentanyl. Original: 1000.
```
Majkl-J, line 145: Since you modularized this you can remove this change
```suggestion
	heat = 1000
```

Majkl-J (approved):

code/game/objects/items/cigarettes.dm
fludd12: Ack I swear to god I removed this, thanks a ton.

PR #2555, opened 2024-11-26
Gives NTC and Blueshield their CentCom Cards back

ChromeFoxxity:
## About The Pull Request

This replaces the silver cards that both NTC and Blueshield get with the CentCom cards they're supposed to get. Though the ones they are administered are fake and only have the access limit of a silver ID, but look like CentCom ID's. 

This PR also pushes our modular files into the actual NTC/Blueshield files, since Skyrat is gone now. To make less file bloat. 

Also turns the Blueshield's locker access to CentCom access so the Captain can't just promote an assistant to Blueshield, and give them their gear. (actually happened).

## Why It's Good For The Game

It's fair for the game because NTC and Blueshield are CentCom personnel. Administered to the station by CentCom themselves, and CentCom are the ones who deal with them if they are unfit for the job. Captain shouldn't just be allowed to enter the Blueshield's locker either.

## Proof Of Testing

image

## Changelog

:cl:
add: Changes the NTC and Blueshield ID's back to CentCom ID's.
add: Changes Blueshield's locker to CentCom access, and gives Blueshield CentCom access.
/:cl:


IgiariValkyr:
Honestly, a good choice to both give them the blue IDs back and compress file bloat at the same time for them. I notice the ID you're giving them is essentially just a Centcom ID with the access limit of a Silver, so it's only a cosmetic change from what I can tell. Makes them stand out as actual Centcom employees again, instead of having them mistakenly thought as hired by the station itself.

ChromeFoxxity:
> Honestly, a good choice to both give them the blue IDs back and compress file bloat at the same time for them. I notice the ID you're giving them is essentially just a Centcom ID with the access limit of a Silver, so it's only a cosmetic change from what I can tell. Makes them stand out as actual Centcom employees again, instead of having them mistakenly thought as hired by the station itself.

Right? Exactly why I did this.

okithedoki:
As the Assistant that got promoted to Blueshield [actually happened], the fact that I inspired an entire PR alone has actually caused me physical pain from laughing so hard.

Although I think this is something of a salt PR, I can't deny that NTRs at least need to be presented as not able to be hired from onstation. They're representatives of the Corporate Interest from Central itself. Having the CC ID would help make this distinguishment.

sippykot:
...what if you lose the id? do you need to ahelp for a new one or something? what if there's no admins online?

ChromeFoxxity:
> ...what if you lose the id? do you need to ahelp for a new one or something? what if there's no admins online?

I WAS considering making Spare CentCom ID's for them, like the Captain's, but I don't know if that would be too overkill.

ChromeFoxxity:
To be honest, in the case of losing their ID, they always couldn't get it fixed, due to their access also having CentCom General access, which on-station doesn't have.

Adrian16199:
I like this solely for the captain not being able to open blueshield locker. They shouldnt be able to just open it and steal from it for themselfs if it does happen (They like the revolver)

RimiNosha:
This changes more than what the description says. Please document your changes properly.

Majkl-J (changes requested):
The captain is allowed to promote people to blueshield at their discretion so the locker should stay open to the captain.

Shadow-Quill:
> The captain is allowed to promote people to blueshield at their discretion [...]

How? Blueshield and NTC access isn't (afaik) available on the access modification terminal.

ChromeFoxxity:
> The captain is allowed to promote people to blueshield at their discretion so the locker should stay open to the captain.

That's incorrect. It isn't a station rank, nor came they do it correctly. It is a Centcom rank, Captain should not be able to even have access to these trims, which I don't know why they do. 

Even if you do give them the trims, they don't worm because of the access that the original IDs give that the console cannot. It doesn't fully work anyways. 

Station command should not be allowed to promote anyone to the rank of a CentCom station-side employee.

ChromeFoxxity:
> According to what staff told me the blueshield rank is a station one, anyone can be a designated bodyguard for the heads of staff and captain. The only actual job that is meant to be a CC one is the NTrep.
> 
> So yeah the locker and ID should stay as station ones on BS

Oh damn.. Sorry, I didn't know that. Why are they a CentCom employee though? They get an Interlink office, and Skyrat gives them a CentCom ID. All this stuff. Even CentCom comms.

Majkl-J:
After some further discussion (That happened like 3 days ago I just was busy), the blueshield and NTC are to stay as-is. I already outlined why the blueshield should not be touched. The main point of concern for NTRep is that the CC IDs would bring back the original issue of them using it to constantly argue with the captain's rulings because "I am centcom". You are there as a station-centcom hotline incase the captain goes rogue, not to fight their position of authority directly.

Feel free to reopen with the modularization fixes but changes to IDs probably won't be considered.

PR #2567, opened 2024-11-26
Adds Ballistic Derringer, and Cargo Goodies Packs

VickiMorris:
## About The Pull Request

Adds a new sidearm to the selection available to the crew, a small two-shot derringer chambered in .310 Strilka (same ammo used by the bolt action rifles) and styled as a companion to the Sakhno rifle. This offers high point-blank burst damage, but is lacking in both ranged performance, and sustained damage compared to the other options that are available from cargo.

The Derringer itself has a 75% damage multiplier, and a ~~5%~~ **10%** damage drop off (per tile) on projectiles fired from it, but otherwise can still be used at a moderate range as a deterrent. This means it will deal 45 damage pointblank (assuming the target has no armor, and using regular .310 ammo), about 9 damage at a screen's distance away, and has an absolute maximum range of 10 tiles.

The cargo goody pack is ~~1300~~ **~~700~~ 450** credits, putting it slightly above the plasma projector firearms, and most melee weaponry. The pack comes with the derringer, and an ammo box with 10 additional shots.

As a recent change, the derringer also requires a do_after delay to load, meaning loading a cartridge takes 3/4ths of a second to perform, to prevent magdumping and being ready to fire immediately.

In addition, it also adds .310 Strilka rubber ammo boxes to the cargo import market, allowing acquisition of (less-lethal) ammo for it, and the Sakhno for the general public. It's 50 credits for 10 shots.

## Why It's Good For The Game

This offers a gun that is more heavily tilted towards self-defense, and requires a bit more thought towards when to use it due to the slight ammo scarcity (requiring either a hacked lathe to print weaker surplus rounds, or an ammo workbench). The low capacity and poor damage at range stops it from being a pocket sniper, while still giving it enough 'oomf' to justify carrying it as a non-combat focused role.

While it can fire twice in (relatively) quick succession, the lower damage makes it unable to crit an unarmored crew member; it still deals a significant amount and leaves them in a vulnerable state, but requires reloading and gives the victim a chance to retaliate, rather that being an instant crit or kill.
## Proof Of Testing
<details>

( Screenshots are not upto date in regards to current price, all other information is correct )
Derringer Cargo

Derringer Cargo Desc

Crate Contents

Derringer and Sakhno

Three in a Row Low

TwoTap Close Low

TwoTap Far Low

Strilka Rubbers

This video below shows the reload delay in action (time is 0.8 seconds, rather than the 0.75 seconds that is committed currently)

https://github.com/user-attachments/assets/d8769a79-6019-4288-bd00-c343a03a3935

<details>
<summary>Old Damage Dropoff @ 5% per Tile</summary>

Three in a Row

TwoTap PB

TwoTap Far

</details>

In addition to the GIFs above, I did a small series of wound testing and general purpose utility actions with the derringer. While surplus ammo suffers from lower wound chance and immensely reduced damage against armored targets, it still deals full against structures, unarmored, and simple mobs.

```
20 trials, 3 tiles between shooter and target, aiming chest, firing twice

1: Skin Breakage (Moderate Chest)
2: Open Puncture (Severe Leg)
3: No Wounds
4: No Wounds
5: Open Puncture (Severe Chest)
6: Skin Breakage x2 (Moderate Chest & Leg)
7: No Wounds
8: No Wounds
9: No Wounds
10: Skin Breakage (Moderate Chest)
11: Skin Breakage (Moderate Chest)
12: Skin Breakage (Moderate Leg)
13: Muscle Tear (Moderate Leg) - Survived Wound
14: Open Puncture (Severe Arm)
15: No Wounds
16: No Wounds
17: No Wounds
18: Skin Breakage (Moderate Chest)
19: No Wounds
20: Skin Breakage (Moderate Arm), Open Puncture (Severe Chest)

9 Unwounded
9 Wounded
2 Double Wounded

3/4 to Crit PB (Affected by Determination from Wounds)
5 to Crit with 3 tiles between shooter and target
8 to Crit with 5 tiles between shooter and target
18 to Crit with 8 tiles between shooter and target

Opens secure locker in 8 PB shots
Breaks reinforced window + grille in 6 PB shots
```

</details>

## Changelog
:cl:
add: Adds Yinbi Derringer. Sakhno production 2-shot pistol chambered in .310 Strilka. Available at cargo for 450 credits.
add: Re-adds .310 Strilka Rubber ammo boxes to the company imports tab, available for 50 credits.
image: Adds sprite for Yinbi Derringer.
/:cl:


Arturlang:
I would give it way higher damage falloff but make it a very cheap gun.

sippykot:
can you put it in your boots?

VickiMorris:
I havent put the 'boot' in 'bootgun', no, I could look into adding it to the holding list alongside knives at the like though. I'm not sure of the balance of it necessarily, but it would still set off weapon scanners while being hidden away.

sippykot:
> I havent put the 'boot' in 'bootgun', no, I could look into adding it to the holding list alongside knives and the like though. I'm not sure of the balance of it necessarily, but it would still set off weapon scanners while being hidden away.

i mean yeah, but it would give it a niche as a surprise/backup weapon, since rn, why wouldnt you buy a stronger gun

Arturlang:
How much does a bolt action cost comparatively?

VickiMorris:
Converting this to draft for the time being while I sort a few things out. Went a bit too fast and loose with this at first.

VickiMorris:
> How much does a bolt action cost comparatively?

Your options for a Sakhno bolt-action are rather limited:

1. Be a cargo tech and be extremely lucky and get one in the mail for absolutely free.
2. Be the QM and have the sporterized 8-shot capacity version in your locker, round start.
3. Hack a cargo console with a multitool, buy a CIN military surplus crate for 1800, gamble with a (decent) chance to get a Sakhno gun case (which has a ~87% chance to be surplus, which will eventually jam and require cleaning, typically after 3-4 shots).
4. Be a traitor, buy an emag, emag a cargo console, and then buy one as contraband for 1200 credits (This one WONT jam on you).

Both the surplus, and regular Sakhno rifles can be upgraded to the 8-shot sporterized model with a black market conversion kit (typically around 500 credits, and requires dealing with a black market purchase).

> If its going to be exclusively useful at point blank AND have scarce ammo, I'd rather just use a shotgun since ammo for that isn't as hard to find

The ammo's a valid complaint! Not everyone wants to have to hack a lathe to print ammunition for their firearm, I will be adding the .310 Strilka rubber ammo box back to cargo imports, to fit in with Sol .40 long only offering the less than lethal option.

IgiariValkyr:
Today’s thought. A laser variant that loads bluespace crystals as ammo to shoot x-ray bolts

Arturlang:
Should be really cheap, something like 300-500 credits

YakumoChen:
How fast does it fire both shots? This sounds like it's going to be a legitimately better option than most guns (especially pistols) if you can click a guy twice in melee and bring them to red health.

YakumoChen:
Especially since you can fill your backpack with what essentially seems like pocket rockets. It reminds me that sawn-off / Eoka spam is still a legitimate threat even in late game RUST (for comparison)

SteamStucKobold:
So, here's a sort of out-of-left-field possible design suggestion to maybe help avoid the Reaper Overwatch fight-style;
What if firing the weapon applies a minor, stacking mood debuff?  ("Ow, that gun makes my wrist hurt.")
It wouldn't be much if you fire it once, it might be noticeable if you fire it twice, but at three+ you'll start getting into bad mood slowdown.

It's definitely less of a numerical balancing mechanism and more of a _feel_ balance mechanism, but I think it could maybe help.

VickiMorris:
> How fast does it fire both shots?
It has a fire delay of 3, which means nothing really in isolation. Other weapons in the same size category (Pocketable, Small or Tiny Firearms)
include the Wespe (3 as well but formatted as 0.3 * seconds, which is not what firing delay represents),
the Eland (0 fire delay),
the Makarov (0 fire delay),
the Bobr (0 fire delay),
the Mini E-Gun (0 fire delay).

Give me a moment to get OBS installed so I can actually do a video showing some side by sides.

> What if firing the weapon applies a minor, stacking mood debuff? ("Ow, that gun makes my wrist hurt.")
Mood debuffs don't immediately apply slowdown, but lower sanity over time which causes slow down. I think this would just be, frustrating because you're getting slowed 30 seconds later for seemingly no reason.

Personally, if you wanted to avoid this play style, you could easily require a short do_after to prime it between reloads (which would be what I wanted personally, timed ejection of spent casings, but fucking .310 strilka has no casings because power bloc cartridges are _all the rage these days._).

I could also change it to require manual priming of the hammer, but that would require butchering so much gun code to make it function and feel nice. Trust me, I've tried it before.

SteamStucKobold:
> Mood debuffs don't immediately apply slowdown, but lower sanity over time which causes slow down. I think this would just be, frustrating because you're getting slowed 30 seconds later for seemingly no reason.
Fair, though I think even just a non-stackable minor mood debuff would be nice, just for the flavor of, you know, firing a full-size rifle cartridge out of a teeny little derringer lol

> Personally, if you wanted to avoid this play style, you could easily require a short do_after to prime it between reloads (which would be what I wanted personally, timed ejection of spent casings, but fucking .310 strilka has no casings because power bloc cartridges are all the rage these days.).
A do-after to prime between reloads would be good, I think, yeah!  Another possible solution (and a humorous solution to the caselessness) could be, having to stick each round in a reusable casing which you _then_ put into the gun.  Neither of these really address the Jetfire Jim playstyle, though personally I'd like both the doafter and the reusable casings to be added just to make the little thing feel more unique.

...oh god, I really am a github comments ideaguy now, aren't I?

VickiMorris:
@YakumoChen 

Here is some side-by-side of "pistols" that we have available through cargo to regular crew, in addition to the Makarov (which is woefully underpowered for an antag gun tbh).

Of note, I am not directly clicking for PB shots to avoid knockback on the Bobr, and because the Sol revolver defaulted to pistol whipping melee. The Bobr is also loaded with slugs, as all other guns are firing lethals, buckshot is a joke, and slugs can just be printed from a hacked lathe (regular crew accessible).

https://github.com/user-attachments/assets/133e328f-36c7-4dbe-80ed-b0afb92ed532

> having to stick each round in a reusable casing which you then put into the gun.

At that point, I would style it as a small 2-round 'mouse clip' instead of making it have an internal magazine. I could then make it need a do-after to eject the magazine, instead of the casings that don't exist.

VickiMorris:
Lots of good feedback here! I decided to go with the "Loading the derringer has a delay" so there is a 0.75 second do_after on loading, applied to both single cartridges and ammo boxes. You're still able to load it while moving, similar to swapping magazines, but it isn't instant now.

https://github.com/user-attachments/assets/0ce77cea-c2d7-4935-a827-90af1de93015

I am lowering the price to a tentative 450 credits, slightly above the import Sol pistols and goodie disablers. I'm not dropping it to 400 flat, as it does come with a box of ammo, which is worth 50 credits by itself.

YakumoChen:
I'm still concerned this is going to encourage people just filling their bag with pocket rockets (watching the video, you can clearly see how badly this chunks people after 2 point blank shots) instead of reloading but what's done is done I guess

PR #2573, opened 2024-11-27
[NON-MODULAR] PAI Freedom!

RimiNosha:
## About The Pull Request

Tin, basically gives PAIs their ability to self-move back, as it used to be.

I *understand why* the change was made for TG specifically, but the niche PAIs fill here tends to be somewhat different, and I also miss the random PAIs running around causing havoc or helping new players out on their own. 

Also I have to use an inventory slot to carry them now, which I severely dislike. They're perfectly capable of carrying themselves, thank you very much. I'd also like to not have to micro where they are so they can sit where they want to.
## Why It's Good For The Game

As mentioned above, PAIs are great entertainment as a standalone entity, players can wander off and scene, or go straight up SSD with no notice, and to top it all off, they actually are part of what helped me get used to SR, funnily enough, and it wouldn't have happened if they couldn't wander.

## Proof Of Testing
<details>

image

image

</details>

## Changelog
:cl:
qol: PAIs can now roam freely again.
/:cl:


Hislyna:
Maybe make it a toggle instead of just revert?

nevimer (approved):

KeptheKobold:
I'd also like a toggle if it's feasible!
 it's useful being able to teleport back to your owner if you hopped out for a moment and they end up running somewhere else without picking you up.

RimiNosha:
Unless it's going to block the PR from being merged, I'm not really wanting to add a toggle for it, due to the extra nonmodular edits I'd need to make, on top of making what's intended to be a simple PR into one with a few more moving parts.

Majkl-J (approved):
Nice

PR #2576, opened 2024-11-28
Move permanent limp to modular

TheGreatKitsune:
## About The Pull Request

Forgot to check if I could make it modular before making pull request, it can be. Updating a previous PR to be modular where it makes sense.

Some files are left as commented edits where it makes more sense to do so.

## Why It's Good For The Game

Better follows contribution guidelines.

## Proof Of Testing

Functions the same way/no changes in how it runs, just how code is structured.


Majkl-J (changes requested):
Needs the GLOBAL_LIST_INIT(permanent_limp_choice moved in a modular file too and this is all good

TheGreatKitsune:
Yup yup. Was just testing if that broke things and it didn't, so.

Majkl-J (approved):

ReturnToZender:
Seems like the permanent limp is applying on people without the quirk, and is... well, permanent. Could use a fix

PR #2595, opened 2024-11-30
Cdef symptom kit

sippykot:
## About The Pull Request

(re?)adds the "civil defense symptom support kit" to cargo

## Why It's Good For The Game

why couldn't you import this again? it's a bunch of chems to stop someone from puking their guts out or such, and helps out people with bad quirks - maybe it should be added to medical orders as a crate, but at that point you can just cook some of these

## Proof Of Testing

i swear i tested it

## Changelog

:cl:
add: you can buy a civil defense symptom support kit from company imports
/:cl:


trefoils:
I love that you named this cdef which is a bacteria that causes Diarrhea. Perfectly fitting.

Majkl-J (approved):

PR #2596, opened 2024-11-30
Not doing portals no longer kills the dragon

Arturlang:
## About The Pull Request
When the dragon runs out of time to make a rift, instead of deleting the dragon, they are given the dragon depression movespeed debuff.

## Why It's Good For The Game
You are forced to get into full on combat with everyone or be literally deleted, this should give people a chance to RP as dragon, with dragon

## Proof Of Testing
Tested and it confirmed that it did not break the victory condition

## Changelog

:cl:
del: Dragon no longer gets deleted if you don't do rifts, instead you are given dragon depression, which slows you down heavily.
/:cl:

### DIFF
--- a/code/modules/antagonists/space_dragon/carp_rift.dm
+++ b/code/modules/antagonists/space_dragon/carp_rift.dm
@@ -35,6 +35,7 @@
 		return
 	if(locate(/obj/structure/carp_rift) in owner.loc)
 		return
+	owner.remove_movespeed_modifier(/datum/movespeed_modifier/dragon_depression/no_portal) // BUBBER ADDITION
 	var/obj/structure/carp_rift/new_rift = new(get_turf(owner))
 	playsound(owner.loc, 'sound/vehicles/rocketlaunch.ogg', 100, TRUE)
 	dragon.riftTimer = -1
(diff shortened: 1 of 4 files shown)


YakumoChen:
I think this is just a little short-sighted. 

The portals force the dragon to actually hunker down somewhere and do an objective as opposed to them just flying around and destroying all the windows and walls. 

Instead of flat-out deleting the dragon, it should impose a running penalty, I believe, such as ticking down their health to a cap of 50%, for example (as a really Conservative penalty)

Arturlang:
> I think this is just a little short-sighted.
> 
> The portals force the dragon to actually hunker down somewhere and do an objective as opposed to them just flying around and destroying all the windows and walls.
> 
> Instead of flat-out deleting the dragon, it should impose a running penalty, I believe, such as ticking down their health to a cap of 50%, for example (as a really Conservative penalty) that also gets undone when they make a portal as they are supposed to.
> 
> I'm all for the idea of this PR bit there should be an incentive to not just do the opposite no-RP station grief.

Hmm, I'd instead just give them the dragon depression slowdown that they get from loosing a portal if they don't do it in time, which gets removed when they make a portal.

YakumoChen:
As long as they have some dissuasion from just griefing with no timer, yeah.

Arturlang:
> As long as they have some dissuasion from just griefing with no timer, yeah.

Suggestions implemented, thanks!

PR #2606, opened 2024-12-01
Upstream 24 12a

lessthnthree:
## Excluded Commits
- [ ] [87201](https://github.com/Bubberstation/Bubberstation/issues/2638)
- [ ] [87215](https://github.com/Bubberstation/Bubberstation/issues/2637)


NargaInDisguise:
I'm going to audibly shout my distaste towards a couple specific upstream pulls, namely the cyborg slowdown changes (huge nerf to cyborgs) and the heretic blade path rework (massive buffs to heretics, refillable aheal potion)

I don't know if you can choose to avoid pulling specific upstreams but those two I would recommend get second opinions before being merged in.

GyrPhlymBaBumble:
What's the point of this besides already fucking over one of the worst characters for combat anyways? almost all of our cyborg players get floored in combat as it is. This is just more torture.

GyrPhlymBaBumble (commented):
All this does legitimately is punish borg players. They are slower than augmented humans unless they have commandeered equipment like scooters, which I think is possible (I have not attempted it, I know it was possible on citadel), and many of our borg players both dislike combat and are not good at it. This just worsens an existing problem of borg salt. Borgs should be able to beat a single human in 1v1 combat at least 30% of the time, or run away, and I do not think that is currently possible.

ReturnToZender:
> All this does legitimately is punish borg players. They are slower than augmented humans unless they have commandeered equipment like scooters, which I think is possible (I have not attempted it, I know it was possible on citadel), and many of our borg players both dislike combat and are not good at it. This just worsens an existing problem of borg salt. Borgs should be able to beat a single human in 1v1 combat at least 30% of the time, or run away, and I do not think that is currently possible.

This is an upstream merge, not a PR to change balance. It's probably going to be merged as-is for the commit history-- if you would like to make edits to it, you'd need to do them separately. Complaining about cyborg changes here specifically is out of scope
(Basically, if you don't like it, make a PR to change it once this is done)

PR #2610, opened 2024-12-02
New Marking (Fat)

GyrPhlymBaBumble:
## About The Pull Request

Adds a marking that properly makes your character overweight instead of only being visible from the front. it is called "Body Weight (No nipples, side-view)" which should also allow anthros to use it without concern
## Why It's Good For The Game
Allows people to more accurately depict their characters with SS13 sprites :)
## Proof Of Testing
image
<details>

</details>

## Changelog
Adds "Body Weight (No nipples, side-view)" to markings.
:cl: Flvffi
add: Added new marking for males.
/:cl:


Majkl-J:
Does this correctly get hidden by clothing or does it clip? Try both with just a jumpsuit and just a labcoat and post screens of both if its fine ill merge

GyrPhlymBaBumble:
> Does this correctly get hidden by clothing or does it clip? Try both with just a jumpsuit and just a labcoat and post screens of both if its fine ill merge

That's a good point I forgot to test that. Thanks! If it does clip I just learned some tekkers to fix that problem using front and behind

GyrPhlymBaBumble:
> Does this correctly get hidden by clothing or does it clip? Try both with just a jumpsuit and just a labcoat and post screens of both if its fine ill merge

It clips, and I can't fix it as it appears the code in 
code/datums/bodypart_overlays/bodypart_overlay.dm
and code/datums/bodypart_overlays/markings_bodypart_overlay.dm
is not working properly. I should be able to add a variant named weight3_chest_m_FRONT_UNDER or weight3_chest_FRONT_UNDER_m (I'm not sure which, but I tried both and neither worked) and the sprite should work properly- I think this is a bug?
`/**This exists so sprite accessories can still be per-layer without having to include that layer's
*  number in their sprite name, which causes issues when those numbers change.
*/
/datum/bodypart_overlay/proc/mutant_bodyparts_layertext(layer)
	switch(layer)
		if(-BODY_BEHIND_LAYER)
			return "BEHIND"
		if(-BODY_ADJ_LAYER)
			return "ADJ"
		if(-BODY_FRONT_LAYER)
			return "FRONT"
		//SKYRAT EDIT ADDITION - Customization
		if(-BODY_FRONT_UNDER_CLOTHES)
			return "FRONT_UNDER"
		if(-ABOVE_BODY_FRONT_HEAD_LAYER)
			return "FRONT_OVER"
		if(-HEAD_LAYER)
			return "FRONT_OVER_HAIR"
		//SKYRAT EDIT ADDITION END`

PR #2623, opened 2024-12-03
Entombed suits no longer add tox damage if their occupant is dead.

Shadow-Quill:
## About The Pull Request

This has been a long annoyance for me; now, people with the entombed quirk will no longer take tox damage if they're dead  (and their modsuit is dead).
This shouldn't apply after revival, as they'll no longer be dead.

## Why It's Good For The Game

It doesn't make sense with how the quirk is explained, as the suit is keeping the person alive. I don't think the suit's pumping toxins into the user when dead, so this stops that.

## Proof Of Testing

TBD

## Changelog

:cl:
fix: Entombed modsuits no longer cause stacking toxin loss if the entombed is dead.
/:cl:


Majkl-J (approved):
Looks good

Shadow-Quill:
I trust my code.

Majkl-J:
I, too, trust his code

PR #2627, opened 2024-12-04
[s] ensure key config is set for any secure request (#88367)

Majkl-J:
## About The Pull Request
If you set up a server without `COMMS_KEY` set it will be blank, which is a valid key
## Why It's Good For The Game
CommKey validation should fail closed not open

https://github.com/tgstation/tgstation/pull/88367


StrangeWeirdKitten:
If you own a downstream from us, either set something in comms.txt (it could be something random) or merge this PR so it fails closed. This is a really massive security vulrability on the default config that allows injections.

PR #2620, opened 2024-12-03
Upstream merge 85137 - Death of mutant bodyparts AND external organs

Funce:
Looks its the big one - just ballparking on the unit-tests

---

## About The Pull Request
Removes mutant bodyparts and external organs from the game completely Digitgrade behaviour was mutant bodypart for no reason

Cat ears now work with the bodyparts overlay system, same as all the other external organs (since all their behaviour is now just on /organ

It doesn't remove all the /external types, but moves all behaviour to /organ. I'll follow up with a PR wiping all the /external organ types, but it's just conflict heaven so not this PR

I've also streamlined a lot of duplicate/weird species regeneration code

Melbert did the same PR as well but due to a lack of time (?) I have absorbed his PR to double nuke mutant bodyparts

## Why It's Good For The Game
Frees us from the chain of unmodular code, and kills my greatest nemesis (after the shuttle meteor murder bug)

## Changelog
🆑 Time-Green and MrMelbert 
Refactor: External organ behaviour has been moved to /organ, ears now use the same system as the other organs 
Refactor: Mutant bodyparts are dead! This likely does not mean much to the average person but it's very dear to me 
code: Improves digitgrade handling in preference code 
/:cl:

I have absorbed #85126, using Melberts code to improve and add some missing changes. Mainly improving the functioning of preferences and digitgrade legs. I didn't take over the hairstyle improvements.

PR #2645, opened 2024-12-08
Mutant body part branch

StrangeWeirdKitten:
Puts #2620 into a repo branch that allows PRs to go into it.

Also we litterally can not upstream without this now.


RimiNosha:
The original should be closed if you're doing this.

StrangeWeirdKitten:
> The original should be closed if you're doing this.

I'm keeping the original around as a branch for Funce to push into. I don't really want to interrupt their work, mostly just get us all on the same page.

Majkl-J:
Oh yeah why is this open lol

Issue #2641, opened 2024-12-08
Map Votes Are Incorrect


RimiNosha:
### Client Version:

N/A

### Issue Summary:

![image](https://github.com/user-attachments/assets/45df32fb-e672-4098-8226-2598997fb8ed)

Either mapvote result is wrong, or the mapvote display numbers are. This is incredibly confusing and should be fixed.

### Round ID:

[6214](https://scrubby.melonmesa.com/round/6214)

### Test Merge Information:

[#1919: '[TM ONLY] Unfucking disease outbreaks/transmission' by lessthnthree at commit c9df7c9c1f](https://github.com/Bubberstation/Bubberstation/pull/1919)
[#2201: 'Investigate Stat Panel as a Stutter cause' by nevimer at commit 60878d6988](https://github.com/Bubberstation/Bubberstation/pull/2201)
[#2291: 'Saddles, saddlebags, calvary' by nikothedude at commit aa92d12daf](https://github.com/Bubberstation/Bubberstation/pull/2291)
[#2401: '[TM ONLY] [MDB IGNORE] Pubbystation 2' by xPokee at commit 31778e169c](https://github.com/Bubberstation/Bubberstation/pull/2401)
[#2405: '[TM READY] New mutation - Strong legs' by Majkl-J at commit 980e6035f2](https://github.com/Bubberstation/Bubberstation/pull/2405)
[#2419: 'Adds a new mod module - mind swap - to illegal tech' by Odairu at commit d341019b2f](https://github.com/Bubberstation/Bubberstation/pull/2419)
[#2529: 'Brings back MCR recharging, rebalances the MCR progression techweb' by Majkl-J at commit 9125122646](https://github.com/Bubberstation/Bubberstation/pull/2529)
[#2122: '[Ready for Review] Round Removal Opt-In' by Swiftfeather at commit fc0837f1a3](https://github.com/Bubberstation/Bubberstation/pull/2122)
[#2569: 'Surgery speed bonus revert/refactor' by lessthnthree at commit 60afaced6e](https://github.com/Bubberstation/Bubberstation/pull/2569)

### Reproduction Steps:

Hold mapvote, see image.

PR #2647, opened 2024-12-08
Fixes map vote winner selection (or: how I found 10,000 extra votes for Tramstation)

lessthnthree:
## About The Pull Request

Why it was picking the winning map based off the cache instead of the actual list of choices, it is a mystery.

Closes https://github.com/Bubberstation/Bubberstation/issues/2641

## Changelog

:cl: LT3
fix: Map vote results should no longer randomly decide to ignore what choice won the vote
/:cl:


Majkl-J (approved):
Im losing my mind

PR #2651, opened 2024-12-08
[NON-MODULAR] [TOOLING] Modular Icon Cutter!

RimiNosha:
## About The Pull Request

Yes, yes, the name might sound incorrect with that tag, but what I mean here is that the icon cutter now checks and works in `modular_zubbers/icons`! It won't pull templates from there, though, I'd recommend creating a zubbers folder inside the `icon_templates` folder instead.

Why did I do this? I have a mid-sized thing I'm working on in the background, and I'd figure I'd PR anything that makes sense by itself to reduce the upfront workload of maints when and if the time comes, provided I get the prototype approved.

Also fixes the malformed moonrock file, I'm not sure exactly what was wrong, but it had the wrong states in places? Not sure if anyone noticed in-game.

## Why It's Good For The Game

Makes spriters less sad when they have to sprite icon smoothed stuff, this is objectively a very good thing,

## Proof Of Testing
<details>

image

</details>

## Changelog
Non player facing.


RimiNosha:
Oh hey nice it cancelled all the tests for absolutely no reason.

Majkl-J (approved):

PR #2692, opened 2024-12-15
Maintentance around nitrogen breathers

Erol509:
## About The Pull Request

First of the changes would be the fact that tank which voxes and people with nitrogen breathing quirk get is extremly big, and can store double the amount of any tank in the game and its a belt tank, which means can be stored in the pockets. This changes to initial value of 25 liters, to 12l which seems to be better alternative, and more fiting since the tank looks like double oxy tank, which can be crafter by engineering. 

To compensate the fact, that nitrogen breathers have to keep their internals on all the time, for close to 2:45 as our shifts tend to be, or their risk quick death. Their minimal gas intake value was changed from 10 to 4, which is the value that plasmaman need to stay alive and not choke to death. So it means that, they only need 4 kpa on their tank, to not take any damage which lowers their need for constant nitrogen resupplies. 

On top of that I fumbled around the quirk, and now aheal/sad will not remove your lungs. SAD will give em back, in case something didnt work out as suposed, while having quirk on you, will readd the lungs each time you miss them during revival. On top of that admin aheal will give em back, as the quirk overrides species default lungs. 

## Why It's Good For The Game

Overall QOL around playing with nitrogen breathing quirk, or as vox as this affects them as well. Its overall good balance change, and brings back their tanks to normal centralised value. 

## Proof Of Testing

Tested it few times in game, I dont quarantee that it wont break any any point, but! I tested everything in game 

## Changelog

:cl:
qol: Vox and people with nitrogen breathing quirk, need less nitrogen to sustain themselfs.
balance: Changed nitrogen tank belt max capacity of 25l to 12l
fix: Aheal/S.A.D interaction doesnt remove nitrogen lungs on quirk holders. Both Aheal/Sad will return them, replacing the one inside you. This means, cybernetic lungs will be removed. 
/:cl:

### DIFF
--- a/modular_skyrat/modules/customization/game/objects/items/tanks/n2_tanks.dm
+++ b/modular_skyrat/modules/customization/game/objects/items/tanks/n2_tanks.dm
@@ -27,7 +27,7 @@
 	inhand_icon_state = "nitrogen"
 	slot_flags = ITEM_SLOT_BELT
 	force = 5
-	volume = 12 // changed from previous 24, its the same as double oxy. I think it was suposed to be double oxy equivalent suggesting icon state name.
+	volume = 12
 	w_class = WEIGHT_CLASS_SMALL
 
 /obj/item/tank/internals/nitrogen/belt/full/populate_gas()
--- a/modular_skyrat/modules/customization/game/objects/items/tanks/n2_tanks.dm
+++ b/modular_skyrat/modules/customization/game/objects/items/tanks/n2_tanks.dm
@@ -8,6 +8,7 @@
 	icon_state = "oxygen_fr"
 	force = 10
 	distribute_pressure = TANK_DEFAULT_RELEASE_PRESSURE
+	distribute_pressure = 4 // see vox.dm for their intake values balance changes
 
 /obj/item/tank/internals/nitrogen/populate_gas()
 	air_contents.assert_gas(/datum/gas/nitrogen)
@@ -26,7 +27,7 @@
 	inhand_icon_state = "nitrogen"
 	slot_flags = ITEM_SLOT_BELT
 	force = 5
-	volume = 24
+	volume = 12 // changed from previous 24, its the same as double oxy. I think it was suposed to be double oxy equivalent suggesting icon state name.
 	w_class = WEIGHT_CLASS_SMALL
 
 /obj/item/tank/internals/nitrogen/belt/full/populate_gas()
--- a/modular_skyrat/modules/customization/modules/surgery/organs/vox.dm
+++ b/modular_skyrat/modules/customization/modules/surgery/organs/vox.dm
@@ -3,7 +3,7 @@
 	desc = "A set of lungs for breathing nitrogen."
 	safe_oxygen_min = 0	//Dont need oxygen
 	safe_oxygen_max = 2 //But it is quite toxic
-	safe_nitro_min = 4 // Atleast 10 nitrogen // previously 10 changed to 4 via Bubber change, exactly the same numbers as plasmaman lungs
+	safe_nitro_min = 4 // Atleast 4 nitrogen
 	oxy_damage_type = TOX
 	oxy_breath_dam_min = 6
 	oxy_breath_dam_max = 20
--- a/modular_skyrat/modules/customization/modules/surgery/organs/vox.dm
+++ b/modular_skyrat/modules/customization/modules/surgery/organs/vox.dm
@@ -3,7 +3,7 @@
 	desc = "A set of lungs for breathing nitrogen."
 	safe_oxygen_min = 0	//Dont need oxygen
 	safe_oxygen_max = 2 //But it is quite toxic
-	safe_nitro_min = 10 // Atleast 10 nitrogen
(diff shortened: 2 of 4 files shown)


modular_skyrat/modules/customization/modules/surgery/organs/vox.dm
```diff
@@ -3,7 +3,7 @@
 	desc = "A set of lungs for breathing nitrogen."
 	safe_oxygen_min = 0	//Dont need oxygen
 	safe_oxygen_max = 2 //But it is quite toxic
-	safe_nitro_min = 10 // Atleast 10 nitrogen
+	safe_nitro_min = 4 // Atleast 10 nitrogen // previously 10 changed to 4 via Bubber change, exactly the same numbers as plasmaman lungs
```
Majkl-J, line 6:
```suggestion
	safe_nitro_min = 4 // Atleast 4 nitrogen
```

modular_skyrat/modules/customization/game/objects/items/tanks/n2_tanks.dm
```diff
@@ -26,7 +27,7 @@
 	inhand_icon_state = "nitrogen"
 	slot_flags = ITEM_SLOT_BELT
 	force = 5
-	volume = 24
+	volume = 12 // changed from previous 24, its the same as double oxy. I think it was suposed to be double oxy equivalent suggesting icon state name.
```
Majkl-J, line 30:
```suggestion
	volume = 12
```

modular_skyrat/modules/self_actualization_device/code/self_actualization_device.dm
```diff
@@ -214,6 +214,13 @@
 	patient.client?.prefs?.safe_transfer_prefs_to_with_damage(patient)
 	patient.dna.update_dna_identity()
 	SSquirks.AssignQuirks(patient, patient.client)
+	if(patient.has_quirk(/datum/quirk/equipping/lungs/nitrogen))
+		if(!patient.get_organ_by_type(/obj/item/organ/internal/lungs/nitrogen))
+			var/lungs = patient.get_organ_slot(ORGAN_SLOT_LUNGS)
+			qdel(lungs)
+			var/obj/item/organ/target_organ = new /obj/item/organ/internal/lungs/nitrogen
+			target_organ.Insert(patient, special = TRUE)
```
Majkl-J, line 222: This feels very snowflakey but I think you discussed this with @Arturlang so he might be aware of a better way?

Majkl-J (changes requested):

modular_skyrat/modules/customization/game/objects/items/tanks/n2_tanks.dm
lessthnthree: Remember to change the default kPA setting to match the reduced need, make it the same as emergency oxygen.

Erol509: > Remember to change the default kPA setting to match the reduced need, make it the same as emergency oxygen.

Default kpa is already set on all tanks sharing the typepath to 4 kpa

modular_skyrat/modules/self_actualization_device/code/self_actualization_device.dm
Erol509: They didnt seem to have issue with it

modular_zubbers/code/datums/quirks/neutral_quirks/lungs.dm
```diff
@@ -32,6 +33,7 @@
 		return
 	lungs_holding.Insert(carbon_holder, special = TRUE, movement_flags = DELETE_IF_REPLACED)
 	lungs_holding.organ_flags &= ~ORGAN_FROZEN
+	carbon_holder.dna.species.mutantlungs = initial(carbon_holder.dna.species.mutantlungs)
```
Majkl-J, line 36: You need to account for species changes as well.

modular_skyrat/modules/self_actualization_device/code/self_actualization_device.dm
```diff
@@ -214,6 +214,13 @@
...
+
```
Majkl-J, line 223: Since you are already overriding mutantlungs this should not be needed

Majkl-J (changes requested):

modular_skyrat/modules/self_actualization_device/code/self_actualization_device.dm
Erol509: Does it place correct species lungs on itself?

Majkl-J: Ill double test this but as far as I am aware it should. Gimme a bit

Erol509: Removed

modular_zubbers/code/datums/quirks/neutral_quirks/lungs.dm
Erol509: Not something Im gonna mess with, mechanical species changes are rare.

PR #2693, opened 2024-12-15
Replace Skyrat santa hat with GAGS santa hat

xXPawnStarrXx:
## About The Pull Request

Autodrobes now supply GAGs santahats.
## Why It's Good For The Game
Festive customisation.
## Changelog
:cl:
fix: Made autodrobes dispense the right santahats.
/:cl:


Majkl-J (commented):
What is the practical difference between the two?

lessthnthree (changes requested):
Can you please also change this at 

code/modules/power/supermatter/supermatter.dm :1098

It can be reverted to
```
if(istype(item, /obj/item/clothing/head/costume/santa))
```

lessthnthree:
> What is the practical difference between the two?

GAGS one can be custom colored

PR #2711, opened 2024-12-17
Locks the research paper only to the hermit

StrangeWeirdKitten:
## About The Pull Request

Removes the research paper from everywhere and locks it only to the hermit.

## Why It's Good For The Game

This item was only specifically meant to go to the hermit. I defended it at first, but people are correct that prisoners shouldn't be building entire robotic factories and I think the age of easy alien tools in science should probally pass.

## Proof Of Testing

image

image

## Changelog

:cl:
balance: The research paper crafting recipe has been locked to the hermit.
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/mob_spawn/ghost_roles/mining_roles.dm
+++ b/modular_zubbers/code/modules/mob_spawn/ghost_roles/mining_roles.dm
@@ -20,3 +20,13 @@
 	gloves = /obj/item/clothing/gloves/fingerless
 	head = /obj/item/clothing/head/soft/purple
 	l_pocket = /obj/item/modular_computer/pda
+
+/datum/outfit/hermit
+	backpack_contents = list(/obj/item/research_paper = 1)
+
+/datum/outfit/hermit/post_equip(mob/living/carbon/human/hermit, visualsOnly)
+	. = ..()
+	if(visualsOnly)
+		return
+	hermit.mind.teach_crafting_recipe(/datum/crafting_recipe/research_paper)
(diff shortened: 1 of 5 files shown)


xXPawnStarrXx:
It's a shame we can't have fun things without other's ruining it. It was a good run while it lasted.

modular_zubbers/code/modules/mob_spawn/ghost_roles/mining_roles.dm
```diff
@@ -20,3 +20,13 @@
...
+
+/datum/outfit/hermit
+	backpack_contents = list(/obj/item/research_paper = 1)
+
+/datum/outfit/hermit/post_equip(mob/living/carbon/human/hermit, visualsOnly)
+	. = ..()
+	if(visualsOnly)
+		return
+	hermit.mind?.teach_crafting_recipe(/datum/crafting_recipe/research_paper)
+	to_chat(hermit, span_notice("You learn the recipe of the <b>research paper</b> to craft everything from nothing."))
```
Majkl-J, line 32:
```suggestion
	to_chat(hermit, span_notice("You learn the recipe for the <b>research paper</b>, giving you the ability to craft everything from nothing."))
```

Majkl-J (changes requested):

Majkl-J (approved):
This is the first time I downright kill something with no potential path of redemption.

MajManatee:
born to have fun, forced to do xenoarch (the most boring thing possible)

MothionTechnologies:
I think it might be better to just remove it, since so few people even play hermit and even those that play usually just go find the station instead of doing survival.

PR #2712, opened 2024-12-17
fixes the exploit of farming primitive skill and makes the status effect tick way slower

Arturlang:
## About The Pull Request
fixes the exploit of farming primitive skill and makes the status effect tick way slower

## Why It's Good For The Game
exploits are bad, i guess, plus this is a 0.5S view call which is Not Good

## Proof Of Testing

<details>

</details>

## Changelog

:cl:
add: changes the primitive skill status effect to work every 5 seconds, instead of 0.5 seconds for performance reasons.
fix: It is no longer possible to farm primitive skill using a plant bag on a wormfarm barrel.
/:cl:

### DIFF
--- a/modular_skyrat/modules/ashwalkers/code/buildings/wormfarm.dm
+++ b/modular_skyrat/modules/ashwalkers/code/buildings/wormfarm.dm
@@ -118,8 +118,8 @@
 			return
 		in_use = TRUE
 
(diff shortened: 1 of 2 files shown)


Majkl-J (changes requested):
You forgot the effect ID

Majkl-J (approved):
Nvm Im dumb you atomised it

PR #2721, opened 2024-12-18
IceBoxStation.dmm - Added Fitness equipment to fitness room

JagOfTroy:
## About The Pull Request

When playing on Icebox station with another player, we went to the fitness room to workout to only realize they don't have any actual fitness equipment in the room. Without totally tearing apart the layout, I added two lifting benches so that people can still get their muscles swole and still give the room the same ease of flow for people to move through it.

## Why It's Good For The Game

This adds equipment to the map that needs to be there so that players can build up their fitness levels other than using the boxing ring.

## Proof Of Testing

Screenshot 2024-12-18 112307

## Changelog

:cl: Sparex

map: Added lifting benches/workout benches to the fitness room for IceboxStation.dmm

/:cl:

### DIFF
--- a/_maps/map_files/IceBoxStation/IceBoxStation.dmm
+++ b/_maps/map_files/IceBoxStation/IceBoxStation.dmm
@@ -20107,6 +20107,10 @@
 	dir = 9
 	},
 /area/station/science/research)
+"fLm" = (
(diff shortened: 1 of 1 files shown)


JagOfTroy:
Sending this upstream as noted.

PR #912, opened 2023-12-27
Makes the christmas tree safe

StrangeWeirdKitten:
## About this pull request

It makes the christmas tree spawner safe

## Why it's good for the game

Because I got a piece of paper and someone else got a chronos modsuit. I don't need to explain why getting debug and admin gear is very broken.

## Changelog

🆑
balance: The christmas tree will be safe next year
/:cl:

PR #2723, opened 2024-12-19
Revert "Makes the christmas tree safe" (Please use broken gear responsibly)

StrangeWeirdKitten:
Reverts Bubberstation/Bubberstation#912

Please be nice, I'm feeling festive :)

### DIFF
--- a/code/modules/events/holiday/xmas.dm
+++ b/code/modules/events/holiday/xmas.dm
@@ -53,7 +53,7 @@
 	/// Christmas tree, no presents included.
 	var/festive_tree = /obj/structure/flora/tree/pine/xmas
 	/// Christmas tree, presents included.
-	var/christmas_tree = /obj/structure/flora/tree/pine/xmas/presents/safe //Bubberstation Edition
+	var/christmas_tree = /obj/structure/flora/tree/pine/xmas/presents
 
 /obj/effect/spawner/xmastree/Initialize(mapload)
 	. = ..()


lessthnthree:
The one thing faster than Santa's sleigh: putting this into TM.

NanoCats:
Praise be

SteamStucKobold:
How much work would it be to have an alert/log whenever one of these not-safe items is created?  Mostly just for easier smiting of the irresponsible

StrangeWeirdKitten:
> How much work would it be to have an alert/log whenever one of these not-safe items is created?  Mostly just for easier smiting of the irresponsible

Too much for me to care about. Mostly because I don't have a list of every admin spawned item in my head and I'd have to probably add to it as I see it.

SteamStucKobold:
> Too much for me to care about. Mostly because I don't have a list of every admin spawned item in my head and I'd have to probably add to it as I see it.

Gotcha, gotcha.
Anyways, considering the hilarity that ensued from Fido Holmes getting an instakill rifle from Santa Claus last night, I fully support this change

MajManatee:
It has only ever brought fun to the server. No reason not to let this stay as a permanent feature unless you HATE CHRISTMAS CHEER

NanoCats:
I know this is a lot to ask but there is one change I'd like to make to the Christmas tree loot list.

Blacklist two things

1. Sheets of paper
2. Random modular computer parts that do nothing and nobody knows what they are

PR #2729, opened 2024-12-20
Flechett ammo rebalance.

StrangeWeirdKitten:
## About The Pull Request

# Flechetts
- 5 pellets at a 15 varience spread (a little better then buckshot). This means that the spread will start to miss one or two entire pellets at 4-5 tiles. 
- 10 damage with a damage drop off of 1 per tile. (at five tiles, it'll be halved)
- Removed the AP and made it weak against armor.
- Base wounding is at 10 with a bonus wounding of 15 (25 total) and has a wounding dropoff of 3 per tile. 

Also the tile drop off for antitide and beehive now properly subtracts its damage instead of adding.

## Why It's Good For The Game
Flechett rounds are litterally the best shotgun ammo in the game. You don't need to use anything else. This is one shot on an elite suit.

image

They can be a great wounding ammo, but their damage should suffer when it begins to not be point blank and against armored targets.

## Proof Of Testing

I don't have specific videos but I got the numbers from somewhere. But truth be told, this needs heavy test merging.

## Changelog

:cl:
balance: Signifigantly changes flechett shotgun rounds. They are a close range ammo with reliable wounding and damage against unarmored with very shitty armor penitration. Damage drops off signifigantly with range.
fix: Beehive and Antitide shotgun ammo now properly subtracts damage dropoff instead of adding.
/:cl:

### DIFF
--- a/modular_skyrat/modules/shotgunrebalance/code/shotgun.dm
+++ b/modular_skyrat/modules/shotgunrebalance/code/shotgun.dm
@@ -171,7 +171,7 @@
 	sharpness = SHARP_EDGED //Did you knew flechettes fly sideways into people
 	weak_against_armour = TRUE
 	damage_falloff_tile = -1 // Five tiles will halve the effectiveness dramatically
-
+	wound_falloff_tile = -3
 /obj/projectile/bullet/pellet/shotgun_buckshot/flechette/Initialize(mapload)
 	. = ..()
 	SpinAnimation()
--- a/modular_skyrat/modules/shotgunrebalance/code/shotgun.dm
+++ b/modular_skyrat/modules/shotgunrebalance/code/shotgun.dm
@@ -153,7 +153,7 @@
 
 /obj/item/ammo_casing/shotgun/flechette
 	name = "flechette shell"
-	desc = "A 12 gauge flechette shell that specializes in ripping armoured targets apart."
+	desc = "A 12 gauge flechette shell that specializes in ripping unarmoured targets apart. These are exceptionally weak against armored targets."
 	icon_state = "fshell"
 	projectile_type = /obj/projectile/bullet/pellet/shotgun_buckshot/flechette
 	pellets = 5
--- a/modular_skyrat/modules/shotgunrebalance/code/shotgun.dm
+++ b/modular_skyrat/modules/shotgunrebalance/code/shotgun.dm
@@ -172,6 +172,7 @@
 	weak_against_armour = TRUE
 	damage_falloff_tile = -1 // Five tiles will halve the effectiveness dramatically
 	wound_falloff_tile = -3
+
 /obj/projectile/bullet/pellet/shotgun_buckshot/flechette/Initialize(mapload)
 	. = ..()
 	SpinAnimation()
@@ -279,10 +280,6 @@
 		else
 			demolition_mod = 2
 			damage = 15
-
-/obj/projectile/bullet/shotgun_slug // Dead simple slug. Hurts like a bitch
-	damage = 30
-	armour_penetration = 40 // Can bypass 40 percent bullet resist
 
 /obj/item/ammo_casing/shotgun/hunter
 	name = "hunter slug shell"
--- a/modular_skyrat/modules/shotgunrebalance/code/shotgun.dm
+++ b/modular_skyrat/modules/shotgunrebalance/code/shotgun.dm
@@ -157,7 +157,7 @@
 	icon_state = "fshell"
 	projectile_type = /obj/projectile/bullet/pellet/shotgun_buckshot/flechette
 	pellets = 5
-	variance = 8
+	variance = 15
 	custom_materials = AMMO_MATS_SHOTGUN_FLECH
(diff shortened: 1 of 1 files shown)


StrangeWeirdKitten:
btw the slugs are very TBD and can get more adjustments.

projectkepler-ru (changes requested):
Flech being made weaker without buckshot and slug box being pushed up in tech?Flech was already on the last tier of tech after i fixed a duplicate, they shouldn't be common, also buckshot was still not adjusted for and is still the worst ammo type as itdoes very little, you have better chance with laser scattershot than buckshot to kill a target due to armour weakness

consider the followings

- move buckshot and slug up by 1 tier (see the sec node, we have 3 tier of ammo tech)
- remove innate armour weakness on buckshot or increases pellet (as otherwise base TG scatterlaser is better and would defeat the point of trying to get buckshot)
- pushes flech up to ammo tech tier 3
- keep the nerf and bring it down by 1 tech

projectkepler-ru:
no doubt a damage nerf is needed

StrangeWeirdKitten:
this PR is only going to rebalance the flechett rounds.

Krypandenej:
from what I'm searching online IRL, flechettes are actually a little better at defeating armor than normal buckshot, but suck in ballistic gel penetration tests due to them tumbling. Wouldn't it make more sense to remove the high wound chance and turn the damage down, but keep AP?

StrangeWeirdKitten:
> from what I'm searching online, IRL 12-gauge fired flechettes are reportedly a little better at defeating armor than normal buckshot, but suck in ballistic gel penetration tests due to them tumbling. Wouldn't it make more sense to remove the high wound chance and turn the damage down, but keep AP?

To be honest I don't care all that much about the realism regarding the rounds, I'm mostly going off what base TG flechett rounds are. But yes, buckshot are basically little balls while flechett is more like shrapnel, so it'll be slightly better. But that being said, it's still not as good as a bullet. 

I think a close range wounding ammo is decent and I shot some players during EORG to test it out and it was fairly comparable to the other rounds in my personal feeling.

IgiariValkyr:
If flechettes are going to be poor against armor now, then bump up slugs penetration, given its one big chunk flying instead of five needles.

StrangeWeirdKitten:
> If flechettes are going to be poor against armor now, then bump up slugs penetration, given its one big chunk flying instead of five needles.

slugs do 50 brute raw damage, that enough is still very good against armor.

projectkepler-ru:
If you insists on only balancing flech move it up a tech to be easier to get, it was last in tech line already for balance reason to prevent easy access

StrangeWeirdKitten:
> If you insists on only balancing flech move it up a tech to be easier to get, it was last in tech line already for balance reason to prevent easy access

I'm not really concerned about the tech tree, and it still absolutely shreds as a higher tier shell. I'm just making it have a drawback.

projectkepler-ru:
@IgiariValkyr Raw damage should still be fine, and xraygun/cmg exists. both doing 15 damage respectively and CMG fire in 2r bursts

BurgerLUA:
Screenshot is misleading because you're shooting a nukie who is standing still, 2 tiles away from you, has no shields, has no stims, etc etc.

If you somehow manage to 

- Get Flechette research done during a nuke ops round.
- Get within 2 tiles of a nukie.
- Shoot a nukie when their stims/shield/etc is down.
- Manage to actually hit a nukie with ALL the flechette rounds.


You pretty much deserve to see half their health gone.

Moostard:
So it will become buckshot+?

I do not like these changes as you are killing the special use for the shell (AP) and making it another weak to armor shell that nobody will use.
The reason is simple.
The vast majority of serious threats you see on station are either wound immune or have full body heavy armor, which makes those types of ammunition only effective and punishing for those unprepared for combat. They do pathetic, sometimes one digit damage to armored targets.

This is why normal and AP ammo is so popular like flechettes and slugs. You barely see unarmored threats. Even if the target is unarmored, AP ammo will still work well against them. Just bit less effective than normal bullets. Meanwhile Hollowpoint/wounding ammo etc. will do almost nothing if used on armor.

I do not understand the justification. Elite suit is one example of heavy armor. Of course if you use AP ammunition (especially shotgun up close) you will actually hurt them. At least that is expected...
Additionally, it is supposed to be a late game ammunition it seems. I think it is supposed to be strong if that is the case.
Are you sure you activated the armor booster by the way? Without it, elite suit has less armor than a normal vest. Just a speculation since you only showed the healthbar.

And like Burger already mentioned, you are shooting a stationary target.
It is easier said than done in actual combat.
You could find out that you can kill half the nukie squad with just one laser gun if you tested it like that.

Please rethink this decision. Thanks.

sippykot:
imo it calls for a dedicated armor piercing round... idk about game balance that much but yeah flechettes shot out of a shotgun would not be too good at piercing more than very thin armor since they are low velocity

StrangeWeirdKitten:
whoops that was the wrong branch.

PR #2741, opened 2024-12-21
Flechett Ammo Rebalance (Has AP with a little more spread and lower damage)

StrangeWeirdKitten:
Continuation of #2729 because I fucked up and merge the wrong branch.

## About The Pull Request

# Flechetts
- 5 pellets at a 15 varience spread (a little better then buckshot). This means that the spread will start to miss one or two entire pellets at 4-5 tiles. 
- 10 damage with a damage drop off of 1 per tile. (at five tiles, it'll be halved)
- Removed the AP and made it weak against armor.
- Base wounding is at 10 with a bonus wounding of 15 (25 total) and has a wounding dropoff of 3 per tile. 

Also the tile drop off for antitide and beehive now properly subtracts its damage instead of adding.

## Why It's Good For The Game
Flechett rounds are litterally the best shotgun ammo in the game. You don't need to use anything else. This is one shot on an elite suit.

image

They can be a great wounding ammo, but their damage should suffer when it begins to not be point blank and against armored targets.

## Proof Of Testing

I don't have specific videos but I got the numbers from somewhere. But truth be told, this needs heavy test merging.

## Changelog

:cl:
balance: Signifigantly changes flechett shotgun rounds. They are a close range ammo with reliable wounding and damage against unarmored with very shitty armor penitration. Damage drops off signifigantly with range.
fix: Beehive and Antitide shotgun ammo now properly subtracts damage dropoff instead of adding.
/:cl:


projectkepler-ru:
Suggestion Still The Same, Flech need to be moved down a tech if they have to lose AP

Genevesh:
Flechettes everywhere are /the/ AP ammo for shotguns, said to be better than slugs at shredding armor. I don't see this as reasonable, either balance them around the AP feature and nerf their damage or just remove them at all

Razurath:
flech

I'd like to point out this is the real world flechette we use today. These things will penetrate armor below a certain grade and tend to keep their grouping and speed with the fins which are basically razorblades. We balanced xray lasers and their insane AP with low damage. High wound, good ap, low damage with limited spread would probably be the best way to change the flechette rounds to balance them without making them useless.

captainfrost20:
Flechette rounds are literally sharpened pins lmao. They're meant to shred soft armor and travel further then buckshot. Not to mention the wound channels these mfers leave. It has issues going through solid metal plates but against like the normal armor vest it would absolutely pierce that without an issue since theres no reason to believe the standard armor vest is a hard plate and not soft body armor.

StrangeWeirdKitten:
I’ll rework the Fletchetts later to be a bit more AP with a damage falloff. But I’m going to keep this test merged because I’m not going to allow their default state back into the codebase

modular_skyrat/modules/shotgunrebalance/code/ammobox.dm
```diff
@@ -53,7 +53,7 @@
 
 /obj/item/ammo_box/advanced/s12gauge/flechette
 	name = "Flechette ammo box"
-	desc = "A box of 7 flechette shells. Each shell contains a small group of tumbling blades that excel at causing terrible wounds."
+	desc = "A box of 7 flechette shells. Each shell contains a small group of tumbling blades that excel at armor penitration."
```
xXPawnStarrXx, line 56: Just wanted to point out a typo in "penetration".

projectkepler-ru (commented):
7*5 for 35 that do eh AP for a tier 2 ballistic tech seems
a bit high of a nerf?
considering a few early game AP option exists that can realistically be obtained with just TG option alone
Remember. this is a late game option. even assuming sci was on overdrive they'll still take hour to get it

PR #2758, opened 2024-12-23
[I think it works] Heretic reroll rite earlier, cheaper, Interlink and SSD protections to heretic targets!

KazooBard:
## About The Pull Request
Hello Chat. I have no idea how to test this on a private server, have no clue if we already had these protections and don't know if the freeze will even end in a month. What I DO know is that heretic has a lot of frustration with it and that I want to do my part! So let me first say, if the dorms and ssd part is unneeded/broken and a better coder says it's easier to keep the first part than fix my spaghetti, sure, I can just commit that out, we can go with just the easier rerolls.

Now let me argue my case:
## Why It's Good For The Game
Be a heretic. Roll 5 targets. Doesn't matter when. Someone's at the bar? Oh welp it's either convince them away, success, or fail at that and now you gotta stalk them out or just go guns blazing into 5 people. None of the 6 people over there want to go through all that trouble to do the thing they want to do. That includes YOU. And depending on the type of RP there's protections.

Okay, so this is why heretics have a REROLL. Awesome, problem solved? Well... Its far down the skill tree... Now you gotta wait/gather rifts or try to get sacrifices from other people, which are guaranteed to be at least 1 secoff at least 1 head (earlier on? No way you do this!), 1 coworker, 1 random person (Plus there's Mr. Busy).  That seems like 4 more chances, but really it's 2 more. 

And well, if that person's in a very open job like botany, chef, bar, then you are going to cause a slaughter trying to get this one person.  (You cannot convince everyone, it's normal). 

So when do we get the reroll?...

3 steps away from ascencion...

So what do we do? Well if you could consult the graph:
image

**We let people reroll earlier on, because neither the sacrificer nor the sacrificee wants to be involved. And there are people who WANT a fight, Roleplay, or are idly wandering halls.**

Finally, we change the price. Whoah. Whoah whoah, ssalty, why the price change? Well, very simply: 
Do you know how much of an effort it is to start a garden while on the run for PACTS WITH DARK GODS?

Very. But it's not as suspicious as carrying a flower that nobody but heretics touch. Nobody. Harebells, unlike poppies, have zero use. I don't _Want_ to argue about Back In The Day, but back in the day when this got first introduced (we are talking 3 heretic paths, gib on sacrifice) there'd be arrests and suspicions over someone just taking the damn flowers!! And since then? Nothing has changed about this mechanically. To me? This looks like:

"The antag HAS to steal an item that only they ever would touch EN MASSE, or grow them - contradicting their playstyle to sit down and watch paint dry/in case they like stealth, then having to do a meaningless chore of getting the funny flower, Or have to put in energy to get someone who doesn't want to engage _at all_ into antag stuff... When there are better options availibl! Or maybe they just straight up are stuck with ONE less sacrifice option because well unlucky, your targets are ssd or ERP'ing."

If you would consult the chart
consult the chart

## Proof Of Testing

hereticced

</details>

## Changelog

:cl:
qol: All Mansus Worshippers can use the Relentless Heartbeat (Reroll targets) ritual earlier by two steps, so that they gain access alongside the ritual of knowledge.
qol: The mansus no longer demands harebells for the Relentless Heartbeat, however a heart is used in it's place.
fix: The Grail was very adamant on keeping people making love in their chambers out of the heretic's target list when it is made or rerolled.
/:cl:

### DIFF
--- a/code/modules/antagonists/heretic/knowledge/cosmic_lore.dm
+++ b/code/modules/antagonists/heretic/knowledge/cosmic_lore.dm
@@ -93,7 +93,9 @@
 		They will then be paralyzed for 2 seconds."
 	gain_text = "The Beast now whispered to me occasionally, only small tidbits of their circumstances. \
 		I can help them, I have to help them."
-	next_knowledge = list(/datum/heretic_knowledge/knowledge_ritual/cosmic)
+	next_knowledge = list(
+	/datum/heretic_knowledge/knowledge_ritual/cosmic,
+	/datum/heretic_knowledge/reroll_targets)
 	route = PATH_COSMIC
 	mark_type = /datum/status_effect/eldritch/cosmic
 
@@ -122,7 +124,6 @@
 	gain_text = "The Beast was behind me now at all times, with each sacrifice words of affirmation coursed through me."
 	next_knowledge = list(
 		/datum/heretic_knowledge/blade_upgrade/cosmic,
-		/datum/heretic_knowledge/reroll_targets,
 		/datum/heretic_knowledge/curse/corrosion,
 		/datum/heretic_knowledge/summon/rusty,
 		/datum/heretic_knowledge/spell/space_phase,
--- a/code/modules/antagonists/heretic/knowledge/cosmic_lore.dm
+++ b/code/modules/antagonists/heretic/knowledge/cosmic_lore.dm
@@ -124,6 +124,7 @@
 	gain_text = "The Beast was behind me now at all times, with each sacrifice words of affirmation coursed through me."
 	next_knowledge = list(
 		/datum/heretic_knowledge/blade_upgrade/cosmic,
+		//datum/heretic_knowledge/reroll_targets //BUBBER EDIT
 		/datum/heretic_knowledge/curse/corrosion,
 		/datum/heretic_knowledge/summon/rusty,
 		/datum/heretic_knowledge/spell/space_phase,
--- a/code/modules/antagonists/heretic/knowledge/cosmic_lore.dm
+++ b/code/modules/antagonists/heretic/knowledge/cosmic_lore.dm
@@ -124,7 +124,7 @@
 	gain_text = "The Beast was behind me now at all times, with each sacrifice words of affirmation coursed through me."
 	next_knowledge = list(
 		/datum/heretic_knowledge/blade_upgrade/cosmic,
-		//datum/heretic_knowledge/reroll_targets //BUBBER EDIT
+		// /datum/heretic_knowledge/reroll_targets, // BUBBER EDIT REMOVAL
 		/datum/heretic_knowledge/curse/corrosion,
 		/datum/heretic_knowledge/summon/rusty,
 		/datum/heretic_knowledge/spell/space_phase,
--- a/code/modules/antagonists/heretic/knowledge/cosmic_lore.dm
+++ b/code/modules/antagonists/heretic/knowledge/cosmic_lore.dm
@@ -95,7 +95,7 @@
 		I can help them, I have to help them."
 	next_knowledge = list(
 	/datum/heretic_knowledge/knowledge_ritual/cosmic,
-	/datum/heretic_knowledge/reroll_targets)
+	/datum/heretic_knowledge/reroll_targets)  //BUBBER EDIT
 	route = PATH_COSMIC
 	mark_type = /datum/status_effect/eldritch/cosmic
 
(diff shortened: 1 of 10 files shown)


NargaInDisguise:
While this is just a flat out heretic buff, I think its a really healthy one for a server like bubber. This lets heretics roll their targets into targets that actually both enjoy combat and enjoy being antagged much sooner and easier, and will hopefully make it less of a hassle towards crew that just want to roleplay overall.

If we truly can't have antag objective opt-out, may as well have this.

code/modules/antagonists/heretic/knowledge/sacrifice_knowledge/sacrifice_knowledge.dm
```diff
@@ -131,7 +132,11 @@
 			continue
 		if(possible_target.current.stat == DEAD)
 			continue
-
+		if(!HAS_TRAIT(possible_target, TRAIT_MIND_TEMPORARILY_GONE))
+			continue
+		for(var/area/protected_areas as anything in GLOB.dorms_areas)
+			if(!is_type_in_list(get_area(possible_target), protected_areas))
+				continue
```
Majkl-J, line 139: This just drops out of the current for loop and not the outside one.

Majkl-J (changes requested):
Dorms onstation don't experience full protections in rules, you should instead add exclusions to people on centcomm levels to protect the interlink

NargaInDisguise:
> Dorms onstation don't experience full protections in rules, you should instead add exclusions to people on centcomm levels to protect the interlink

Is there any reason not to keep dorms protected still? It cuts the potential to rulebreak by just diving in there without asking when they're still a semi-protected area.

Shadow-Quill:
> > Dorms onstation don't experience full protections in rules, you should instead add exclusions to people on centcomm levels to protect the interlink
> 
> Is there any reason not to keep dorms protected still? It cuts the potential to rulebreak by just diving in there without asking when they're still a semi-protected area.

Entering a dorm room isn't a get out of jail free card — someone just existing inside a dorm room shouldn't make them ineligible from being an antag target.

Also, if someone *does* break rules regarding current dorm room protections, they're going to get smacked pretty hard.
I know the ERP rules are one of the things that are more strictly enforced, for good reason.

Arturlang:
TBH it makes sense with how people in dorms are protected and specified in the rules to literally require a ahelp to swap targets.

code/modules/antagonists/heretic/knowledge/cosmic_lore.dm
Majkl-J, line 1: Ditto

code/modules/antagonists/heretic/knowledge/flesh_lore.dm
Majkl-J, line 1: Ditto

code/modules/antagonists/heretic/knowledge/general_side.dm
Majkl-J, line 1: Ditto

code/modules/antagonists/heretic/knowledge/lock_lore.dm
Majkl-J, line 1: Ditto

code/modules/antagonists/heretic/knowledge/moon_lore.dm
Majkl-J, line 1: Ditto

code/modules/antagonists/heretic/knowledge/rust_lore.dm
Majkl-J, line 1: Ditto

code/modules/antagonists/heretic/knowledge/moon_lore.dm
```diff
@@ -124,7 +126,6 @@
 	gain_text = "At the head of the parade he stood, the moon condensed into one mass, a reflection of the soul."
 	next_knowledge = list(
 		/datum/heretic_knowledge/blade_upgrade/moon,
```
Majkl-J, line 128: Removals should be done with commenting, do these for other files too
```suggestion
		/datum/heretic_knowledge/blade_upgrade/moon,
		// /datum/heretic_knowledge/reroll_targets, // BUBBER EDIT REMOVAL
```

Majkl-J (changes requested):

code/modules/antagonists/heretic/knowledge/moon_lore.dm
KazooBard: Did it for all in newest commit!

PR #2767, opened 2024-12-24
Add no checks helper to kilo telecomms air alarm

TealSeer:
## About The Pull Request

Add a `tlv_no_checks` helper to the air alarm in Kilo's telecomms server room because it just screams the whole round that its cold and theres low oxygen, like how it's supposed to be

## Why It's Good For The Game

It's supposed to be like this

## Proof Of Testing

Compiled

## Changelog

:cl:
map: fixed kilostation's telecomms air alarm constantly being angry.
/:cl:


Majkl-J (approved):

Majkl-J (changes requested):
Actually wait, tcomms should use the /obj/effect/mapping_helpers/airalarm/tlv_cold_room helper

TealSeer:
> Actually wait, tcomms should use the /obj/effect/mapping_helpers/airalarm/tlv_cold_room helper

not so, the cold room helper is for the kitchen back room (which only goes down to -14c i.e. still survivable), while the no checks one is for telecomms because telecomms is cooled to -193c

lessthnthree (approved):

PR #2781, opened 2024-12-25
Increases Blob Max mass points. (175)

CloverIsLucky:
## About The Pull Request

Increases blob max biomass from 100 to 175.
## Why It's Good For The Game

Let's be real, Blob is the most CBT'd antag role of the server with currently zero chances to win against a late-game fully armed crew who can easily distribute guns and build emitters to counter them. This basically buffs a bit the Blob back with just lazily increasing their stored mass cap to get a bit of chance of getting a comeback when the crew least expect it.

## Proof Of Testing

idk, I just changed numbers.

</details>

## Changelog
:cl:
balance: Blob's mass cap reaches to 175
/:cl:

### DIFF
--- a/code/__DEFINES/blob_defines.dm
+++ b/code/__DEFINES/blob_defines.dm
@@ -1,6 +1,6 @@
 // Overmind defines
 
-#define OVERMIND_MAX_POINTS_DEFAULT 999 // Max point storage. Bubber edit, original: 100. Good luck reaching that cap.
+#define OVERMIND_MAX_POINTS_DEFAULT 175 // Max point storage. Bubber edit, original: 100. Good luck reaching that cap.
 #define OVERMIND_STARTING_POINTS 60 // Points granted upon start
 #define OVERMIND_STARTING_REROLLS 1 // Free strain rerolls at the start
 #define OVERMIND_STARTING_MIN_PLACE_TIME (1 MINUTES) // Minimum time before the core can be placed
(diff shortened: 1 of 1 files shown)


CloverIsLucky:
If you think 999 max mass is too much, is because you never played Blob and should shush. But I'm open for opinion from maintainers to suggest me a different biomass cap instead, instead of just highly increasing it.... As if it's virtually possible to reach that biomass cap 😹😹😹😹

StrangeWeirdKitten:
As I said in the other PR that wanted to increase it, the biomass limit is fine and is a good balance between the crew winning and the blob killing everyone, waiting to win. I’m not really interested in raising it especially with how blobs can be definitely a little squishy against certain mechanics

CloverIsLucky:
I changed my mind, 999 biomass cap would be too much if a Blob makes lots of factories and recklessly expands too much.. How about this cap instead? (175)

nikothedude:
100 cap is tg not bubber

CloverIsLucky:
I hate loyal upstreams.

nikothedude:
we arent a loyal upstream we can change this

NargaInDisguise:
IMO, a blob should never win against a full crew that is actually going out of their way to counter them by pulling out all the stops. Guns, mechs and engineering builds take up ALL of the crews attention, this gives other antags their chance to do whatever stealth shit they're planning. Also, deltas in an RP server should really only ever come from the crew completely dropping the ball on something, tbh. A blob win is an instant round end delta, and I just don't think bubber needs more of those.

Odairu:
ngl, this was a cooking PR sad that its closed and gone :(

PR #2805, opened 2024-12-29
Fixes BDM eye's buff duration.(I believe.)

CloverIsLucky:
## About The Pull Request

Makes the BDM eye's buff actually last for one second as it promises.
## Why It's Good For The Game

I began to wonder why the BDM eye isn't properly working, I was near to double or triple up its duration until I realized it was coded like it was using ticks when apparently the buff codiing seems to be really using seconds instead. I hope this fixes the BDM and makes it work as intended.

Edit: I reverted the requested change and I increased the ticks instead from 10 to 15 ticks.

## Proof Of Testing

Local test seems to work fine.

## Changelog
:cl:
fix: Blood-drunk buff should really last for 1 second.
/:cl:

### DIFF
--- a/code/datums/status_effects/buffs.dm
+++ b/code/datums/status_effects/buffs.dm
@@ -74,8 +74,8 @@
 
 /datum/status_effect/blooddrunk
 	id = "blooddrunk"
-	duration = 1 SECONDS
-//	tick_interval = -1 
+	duration = 15
+	tick_interval = -1 
 	alert_type = /atom/movable/screen/alert/status_effect/blooddrunk
 
 /atom/movable/screen/alert/status_effect/blooddrunk
--- a/code/datums/status_effects/buffs.dm
+++ b/code/datums/status_effects/buffs.dm
@@ -74,8 +74,8 @@
 
 /datum/status_effect/blooddrunk
 	id = "blooddrunk"
-	duration = 10
-	tick_interval = -1
+	duration = 1 SECONDS
+//	tick_interval = -1 
 	alert_type = /atom/movable/screen/alert/status_effect/blooddrunk
 
 /atom/movable/screen/alert/status_effect/blooddrunk
--- a/code/datums/status_effects/buffs.dm
+++ b/code/datums/status_effects/buffs.dm
@@ -75,7 +75,7 @@
 /datum/status_effect/blooddrunk
 	id = "blooddrunk"
 	duration = 15
-	tick_interval = -1 
+	tick_interval = -1
 	alert_type = /atom/movable/screen/alert/status_effect/blooddrunk
 
 /atom/movable/screen/alert/status_effect/blooddrunk


IgiariValkyr:
This feels like something that should go upstream if it's an issue there and not just here. But I do agree that it currently lasts less than the one second that is said.

CloverIsLucky:
> This feels like something that should go upstream if it's an issue there and not just here. But I do agree that it currently lasts less than the one second that is said.

I'm not waiting until an upstream PR adds the fix from upstream.

code/datums/status_effects/buffs.dm
```diff
@@ -74,8 +74,8 @@
 
 /datum/status_effect/blooddrunk
 	id = "blooddrunk"
-	duration = 10
-	tick_interval = -1
+	duration = 1 SECONDS
+//	tick_interval = -1
```
Shadow-Quill, line 78: Unless the code comments for status effects are wrong, this fundamentally does nothing, except make this status effect tick every second.
https://github.com/Bubberstation/Bubberstation/blob/e7b7473c71e4bac488bdfea6f00339165423c7db/code/datums/status_effects/_status_effect.dm#L6-L14

As the duration is in deciseconds, 10 already equals `1 SECOND` (In fact, the SECOND(S) macro just multiplies whatever number by 10.)
The blooddrunk status effect also doesn't need to tick, as it's a "get a speed boost when you get it, lose it when you lose it" type thing.

In short, this does nothing.
Also, please test changes.

Shadow-Quill (changes requested):

code/datums/status_effects/buffs.dm
CloverIsLucky: I have tested and for me it works fine, less the somewhat stun immunity part and damage tankiing sometimes works, sometimes it doesn't. I have the feeling it has to do with server lag. Maybe increasing the ticks should resolve it to give a false time duration I try to make it truly work.

CloverIsLucky: I'll try increasing the tick duration.

Shadow-Quill: > I have tested

I mentioned that because 
image

Also, you can PR this to TG itself - as is, one of the maints is almost definitely going to, at minimum, ask you to mark the change, if not make it modular (or port it to TG anyways)

CloverIsLucky: I'm not gonna wait for an upstream PR to have it fixed, how can I modularize an already existing status effect just to manipulate its time duration though?

CloverIsLucky:
I made a local test and increasing the ticks instead of changing the time to seconds briefly increases its duration. For me the duration is longer than one second, but for the server, it might be exactly 1 second. Also I begin to notice that BDM eye doesn't bring stun immunity against goliath tentacles anymore.

YakumoChen:
> > This feels like something that should go upstream if it's an issue there and not just here. But I do agree that it currently lasts less than the one second that is said.
> 
> Sorry, but I'm not gonna wait until an upstream PR adds this (hopefully) very simple fix.

PR it upstream and make an early pull then

We still are a faithful downstream of TG so making code edits to bugs upstream will just create code debt.

IgiariValkyr:
> > This feels like something that should go upstream if it's an issue there and not just here. But I do agree that it currently lasts less than the one second that is said.
> 
> Sorry, but I'm not gonna wait until an upstream PR adds this (hopefully) very simple fix.

Literally _you_ would be the one making the PR upstream. Not someone else since this is _your_ fix. Make it on TG, keep this here for when theirs gets merged, then this can get merged and be ready immediately instead of when the upstream pull happens later on.

StrangeWeirdKitten:
If this is an issue, it's an upstream issue. Fix it on /tg/ and one of our maintainers can easily early pull it.

PR #2846, opened 2025-01-05
Limits the RBMK to pulse every three seconds. Halves range

StrangeWeirdKitten:
## About The Pull Request

So every tick nine machines sent out radiation pulses that can climb up to 20 tiles. Radiation creates massive lists, especially 20 tiles of lists. Radiation should not be allocating more total heap then god damn signals, lol.

```
/datum/gas_mixture/copy = 35.4 MB
/datum/light_source/update_corners = 43.4 MB
/datum/lighting_object/update = 83.4 MB
/datum/lighting_object/New = 104.9 MB
/datum/RegisterSignal = 170.5 MB
/radiation_pulse = 178.2 MB
```

## Why It's Good For The Game

We kind of need memory to play this game.

## Proof Of Testing

It compiles, so I assume it works. I have zero idea how this thing works.

## Changelog

:cl:
fix: RBMK no longer crashes the server.
/:cl:

### DIFF
--- a/code/__DEFINES/reactions.dm
+++ b/code/__DEFINES/reactions.dm
@@ -14,7 +14,7 @@
 #define ATMOS_RADIATION_VOLUME_EXP 3
 
 /// Maximum range a radiation pulse is allowed to be from a gas reaction.
-#define GAS_REACTION_MAXIMUM_RADIATION_PULSE_RANGE 20
+#define GAS_REACTION_MAXIMUM_RADIATION_PULSE_RANGE 10
 
 // Water Vapor:
 /// The temperature required for water vapor to condense.
--- a/code/__DEFINES/reactions.dm
+++ b/code/__DEFINES/reactions.dm
@@ -14,7 +14,7 @@
 #define ATMOS_RADIATION_VOLUME_EXP 3
 
 /// Maximum range a radiation pulse is allowed to be from a gas reaction.
-#define GAS_REACTION_MAXIMUM_RADIATION_PULSE_RANGE 10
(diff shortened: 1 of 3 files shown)


code/__DEFINES/reactions.dm
```diff
@@ -14,7 +14,7 @@
 #define ATMOS_RADIATION_VOLUME_EXP 3
 
 /// Maximum range a radiation pulse is allowed to be from a gas reaction.
-#define GAS_REACTION_MAXIMUM_RADIATION_PULSE_RANGE 20
+#define GAS_REACTION_MAXIMUM_RADIATION_PULSE_RANGE 10
```
TealSeer, line 17: modular comment that thang

StrangeWeirdKitten: Oh it's TG. It'll be fine kept on 20

BurgerLUA:
This won't fix the memory leak, but it is a code improvement.

People don't use the RBMK often, and the server crashes on other maps without the RBMK present. When the RBMK was commonly used months ago, there wasn't any crashes, even with the oversight in the radiation pulse triggering way too often for what it was.

#1629 was a PR I made a while ago involving Nebula Storms. It would lag out the server maybe 3 out of 10 times it occurred. it used the radiation_pulse proc, but called it locally on the living mob, which wouldn't use any range procs, as seen here.

https://github.com/Bubberstation/Bubberstation/blob/0988d746f47ebc474e561d2e508d6e72d93c55c8/code/datums/weather/weather_types/radiation_storm.dm#L103

PR #2856, opened 2025-01-07
Memory 2.0: The flickering light removal

StrangeWeirdKitten:
## About The Pull Request

Removes broken flickering lights

## Why It's Good For The Game

Lighting is another one of our massive memory usages. Let's remove this.

## Changelog

:cl:
del: Removes broken light flickering.
/:cl:


IgiariValkyr:
Flickering lights as also an issue with gameplay at random as well. You use a multitool to repair the ballast so the light works, but as soon as you do that, it will fry the bulb regardless.

BurgerLUA:
Should've just reopened #1689

rintherat:
I wonder if this will make lighting more stable. Sometimes lights would outright **refuse** to work sometimes. Maybe flickering was the cause of it.

PR #2872, opened 2025-01-09
removes armor stats from the magician tailcoat in the costume drobe

Arturlang:
## About The Pull Request
removes wizard armor from wizdrobe

## Why It's Good For The Game
costumes should not have armor

## Proof Of Testing
nope

<details>

</details>

## Changelog

:cl:
del: no more wizard armor in wizdrobe, sorry, 
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/clothing/outfits/bunny.dm
+++ b/modular_zubbers/code/modules/clothing/outfits/bunny.dm
@@ -64,7 +64,7 @@
 /datum/outfit/wizard/bunny_magician
 	name = "Bunny Magician"
 	uniform = /obj/item/clothing/under/costume/playbunny/magician
-	suit = /obj/item/clothing/suit/jacket/tailcoat/magician
+	suit = /obj/item/clothing/suit/wizrobe/magician
 	back = /obj/item/storage/backpack/satchel/leather
 	head = /obj/item/clothing/head/playbunnyears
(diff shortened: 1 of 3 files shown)


YakumoChen:
I think your title is wrong

Arturlang:
> I think your title is wrong

you are in fact, correct

IgiariValkyr:
Isn't the wizard armor for the battlemage archetype that runs around with the big fucking hammer?

Arturlang:
> Isn't the wizard armor for the battlemage archetype that runs around with the big fucking hammer?

all wizard robes have some amount of armor

Arturlang:
> The armor value is wrong, the tailcoat is meant to be a purely flavour item. Change the armor value on it instead at `modular_zubbers/code/modules/clothing/suits/jacket.dm#L132`

changed the path instead and re-added it

YakumoChen:
Now you're creating the opposite problem, the wizard tailcoat has no armour. Fix it correctly?

Arturlang:
> Now you're creating the opposite problem, the magivend tailcoat has no armour. Fix it correctly?
> 
> The solution is to just make a /fake child with no armour datum like every other wizard outfit

Honestly didn't think it'd matter since there's no head slot for the wizard clothes set

Arturlang:
done

Majkl-J (approved):
Why did I never think of using that to avoid copypaste code inconsistencies.

Anyhow, all good

PR #2890, opened 2025-01-12
Subtle notification sound/chat filter

lessthnthree:
## About The Pull Request

Adds an optional soft notification sound for subtle, using the same pref as the subtler notification.

## Why It's Good For The Game

It's nice to have the option to have subtle sounds as well, particularly now that interactions use them.

## Changelog

:cl: LT3
sound: Subtle now has a sound notification alongside subtler
qol: Subtle/subtler now has a TG chat filter
/:cl:

### DIFF
--- a/modular_skyrat/modules/verbs/code/subtle.dm
+++ b/modular_skyrat/modules/verbs/code/subtle.dm
@@ -55,7 +55,7 @@
 
 	for(var/mob/ghost as anything in GLOB.dead_mob_list)
 		if((ghost.client?.prefs.chat_toggles & CHAT_GHOSTSIGHT) && !(ghost in viewers))
-			ghost.show_message(subtle_message)
+			to_chat(ghost, "[FOLLOW_LINK(ghost, user)] [subtle_message]")
 
 	for(var/mob/receiver in viewers)
 		receiver.show_message(subtle_message, alt_msg = subtle_message)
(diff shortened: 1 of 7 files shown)


modular_skyrat/modules/verbs/code/subtle.dm
```diff
@@ -59,6 +59,10 @@
 
 	for(var/mob/receiver in viewers)
 		receiver.show_message(subtle_message, alt_msg = subtle_message)
+		// Optional sound notification
+		var/datum/preferences/prefs = receiver.client?.prefs
+		if(prefs && prefs.read_preference(/datum/preference/toggle/subtler_sound))
+			receiver.playsound_local(get_turf(receiver), 'sound/misc/bloop.ogg', 50)
```
Majkl-J, line 65: Would be nice to keep these consistent and play the `sound/effects/achievement/glockenspiel_ping.ogg`. Alternatively anything that isn't an admin announcement sound as that should be reserved due to its importance

Majkl-J (changes requested):

modular_skyrat/modules/verbs/code/subtle.dm
lessthnthree: Yeh that's why I marked it DNTM. So nobody put it in before I changed the sound to match.

Originally the thought was so people cam tell between subtle and subtler, but I think I'll do that with a change in spans instead.

PR #2909, opened 2025-01-13
Synth Heretics Heart QOL

KazooBard:
## About The Pull Request

Heretics who are synths no longer have to craft their living heart - making them not reliant on poppies to start antagging - **Putting them on an even playing field.**

## Why It's Good For The Game

The check for making hearts only apply to non-synths was making some people not have a good time, so I deletedi t

## Proof Of Testing

<details>
Heart very much beating
https://github.com/user-attachments/assets/409a913c-0a03-4e5a-aa5e-ea4e2882bd0c
</details>

:cl:
qol: Heretic Synths also start with a living heart
/:cl:

### DIFF
--- a/code/modules/antagonists/heretic/knowledge/starting_lore.dm
+++ b/code/modules/antagonists/heretic/knowledge/starting_lore.dm
@@ -187,7 +187,7 @@ GLOBAL_LIST_INIT(heretic_start_knowledge, initialize_starting_knowledge())
 		return FALSE
 	if(!new_heart.useable)
 		return FALSE
-	if(new_heart.organ_flags & (ORGAN_ROBOTIC|ORGAN_FAILING))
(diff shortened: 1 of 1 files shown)


KazooBard:
https://github.com/user-attachments/assets/22c74288-0f5a-42c3-a7ef-97954b27b370

IgiariValkyr:
Does this affect humans with synthetic hearts? The point of robotic hearts is to cripple heretics when they’re caught instead of them being outright RR’d, but if synthetic hearts no longer do this, this may cause even more RR’s.

KazooBard:
> Does this affect humans with synthetic hearts? The point of robotic hearts is to cripple heretics when they’re caught instead of them being outright RR’d, but if synthetic hearts no longer do this, this may cause even more RR’s.

It affects them only at the time of the roll.

Nothing will change in regards to regaining your heart.

Majkl-J (approved):

PR #2922, opened 2025-01-14
caps laser eye blur at 20 seconds

Arturlang:
## About The Pull Request
caps laser eye blur at 20 seconds

## Why It's Good For The Game
there is no cap and we have guns that shoot VERY FAST

## Proof Of Testing

I tested it, trust me.

## Changelog

:cl:
balance: Eye blur from lasers is capped at maximum of 20 seconds.
/:cl:


Majkl-J (changes requested):
I'd argue to just go to /mob/living/proc/apply_effects and capping it to 30 SECONDS across the board, running around with unlimited eye blur from most other damage sources is also annoying.

Also it doesn't seem to get called from anywhere else important so.

Arturlang:
> I'd argue to just go to /mob/living/proc/apply_effects and capping it to 30 SECONDS across the board, running around with unlimited eye blur from most other damage sources is also annoying.
> 
> Also it doesn't seem to get called from anywhere else important so.

I'll cap it at the effect level

lessthnthree (approved):

PR #2937, opened 2025-01-17
Reduces antag counts across most storytellers

nikothedude:
## About The Pull Request

I really didn't want to make this PR, because I hoped someone with more knowledge of the problem at hand would come along, but noone has so far. So you're stuck with my desperate attempt to make things a little better.

I know this PR isn't very good. I don't really know the core issue besides "rounds way too chaotic"; but I think it's a reasonable assumption that too many antags could be a candidate - especially since a lot of new players seem to not engage in a healthy antag relationship.
## Why It's Good For The Game

Storytellers have gone off the rails recently, directly correlating with our popsize. I think we should try reducing the amount of antags even by just a little to see if it helps.
## Proof Of Testing

it compiles

## Changelog
:cl:
balance: Storytellers spawn less antags
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/storyteller/storytellers/tellers/storyteller_bomb.dm
+++ b/modular_zubbers/code/modules/storyteller/storytellers/tellers/storyteller_bomb.dm
@@ -8,7 +8,7 @@
 		TAG_DESTRUCTIVE = 2
 	)
 	population_min = 25
-	antag_divisor = 10
+	antag_divisor = 13
 	storyteller_type = STORYTELLER_TYPE_INTENSE
 
 /datum/storyteller_data/tracks/bomb
--- a/modular_zubbers/code/modules/storyteller/storytellers/tellers/storyteller_chill.dm
+++ b/modular_zubbers/code/modules/storyteller/storytellers/tellers/storyteller_chill.dm
@@ -11,7 +11,7 @@
 		TAG_DESTRUCTIVE = 0.3,
 		TAG_CHAOTIC = 0.1
 	)
-	antag_divisor = 32
+	antag_divisor = 38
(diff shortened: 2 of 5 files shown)


nikothedude:
event freq not reduced as that does not scale with pop size, my theorized problem

nikothedude:
note that most of these changes were made on a whim, theres no exact science besides "lets reduce antag size"

Arturlang:
This won't reduce the amount of blobs/dragons/ninjas appearing, only less antags inside the crew

Arturlang (dismissed):
This won't make event tracks run less, you need to modify /datum/storyteller_data track info to actually make them run less.
Figure out if this is a issue on specific storytellers first, I doubt it's a issue on every one.
Tell me if the issue is in fact too many crew-sided antags, or ghost sided antags so we can actually figure out the exact issue.

Arturlang:
For reference, the equation for antag cap is round(max(min(get_correct_popcount() / storyteller.antag_divisor + sec_crew ,sec_crew * 1.5),ANTAG_CAP_FLAT))

Arturlang:
Anyway, the usual maximum is capped by security pop instead, and not antag divisior, the issue is that high threat antags are just as expensive as low threat antags

nikothedude:
So what you're saying is that a more proper change would reduce event frequency and add a antag wighting system?

lessthnthree:
> So what you're saying is that a more proper change would reduce event frequency and add a antag wighting system?

Yes.

Also the realistic antag cap is based on the hardcoded antag = 1.5 security, so this PR wouldn't do anything really.

nikothedude:
If people that know more and care more thann me are stepping up, this PR is no longer needed

PR #2945, opened 2025-01-18
[TEST-MERGE ONLY] Makes storytellers random

nikothedude:
## About The Pull Request

**In this stage of the PR, this is mostly just up for discussion - it has not been tested or tuned**

Storytellers are now randomly selected, rather than voted for. This is controlled by a new config variable.
Clown and the bomb are less common.
## Why It's Good For The Game

Clown and default andy are basically being picked 100% of the time these days. A lot of people, me incnluded, are upset by this. Im pretty sure storytellers were never made to give total control to the playerbase, more of add variety and a more nuanced experience (I could be wrong, please correct me if I am)

Ultimately, even if the players do want extreme levels of chaos 24/7 - it might be best if we t ake the wheel for once to actually dictate what _we_ think a good distribution of chaos is. In this implemenation, all storytellers are evenly weighted except f or the clown and the bomb, which are rarer.
## Proof Of Testing

Hard to test, always picks extended on 0pop

## Changelog
:cl:
add: Storytellers are now randomly picked
/:cl:

### DIFF
--- a/code/controllers/subsystem/ticker.dm
+++ b/code/controllers/subsystem/ticker.dm
@@ -174,7 +174,10 @@ SUBSYSTEM_DEF(ticker)
 			if(storyteller)
 				SSgamemode.set_storyteller(text2path(storyteller), TRUE)
 			else
-				SSvote.initiate_vote(/datum/vote/storyteller, "Storyteller Vote", forced = TRUE)
+				if (CONFIG_GET(flag/vote_for_storytellers))
+					SSvote.initiate_vote(/datum/vote/storyteller, "Storyteller Vote", forced = TRUE)
+				else
+					SSgamemode.pick_random_storyteller()
 		// BUBBERSTATION EDIT END
 			SStitle.change_title_screen() //SKYRAT EDIT ADDITION - Title screen
 			addtimer(CALLBACK(SStitle, TYPE_PROC_REF(/datum/controller/subsystem/title, change_title_screen)), 1 SECONDS) //SKYRAT EDIT ADDITION - Title screen
--- a/modular_skyrat/master_files/code/controllers/configuration/entries/skyrat_config_entries.dm
+++ b/modular_skyrat/master_files/code/controllers/configuration/entries/skyrat_config_entries.dm
@@ -63,3 +63,7 @@
 
 /// If you want to have a default storyteller
 /datum/config_entry/string/default_storyteller
+
+/// If enabled, storytellers cease being randomly picked and are voted for
+/datum/config_entry/flag/vote_for_storytellers
+	default = FALSE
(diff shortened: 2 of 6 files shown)


nikothedude:
as of right now im just putting this up for feedback and discussion

IgiariValkyr:
We may as well go back to Dynamic at this rate.

lessthnthree:
After the changes today we saw a more varied selection of storyteller tracks. With clown removed from voting, and fragile added to always be available, in the time I watched we had 1 chill, 2 fragile, and a gamer round.

I'm not a fan of taking player choice away entirely regarding what vibe they want on a shift.

Another reason for some skewed storyteller selections is that (I BELIEVE, but not 100%) when the server crashes we repeat the same vote choices because it doesn't think we've completed a round. Changing it so like, when the round starts (rather than ends) it considers that track to have run (So if it was Gamer, if the server crashes you're not presented with another vote for Default Andy, Gamer, or Bomb) that should improve things too.

BurgerLUA:
I would be fine with making storyteller random if some the storytellers weren't so extreme in terms of differences.

ArrisFairburne:
I also don't agree with completely random-picking storytellers. While the system as-is is definitely imperfect (if a storyteller has majority it is chosen 100% of the time, even if it only has 20% of the votes), it is worse if the storyteller is chosen at random instead of reflecting what at least some people on the server currently want.

Redrover1760:
You could make storytellers hidden, or make storytellers done by weighed random votes as alternatives.

Strazyplus:
_"Clown and default andy are basically being picked 100% of the time these days. A lot of people, me included, are upset by this"_
**Absolutely not** - A lot of people, me included are upset when Extended/Chill gets picked and they nope out. Security just gets left unmanned.

I always vote for Clown and Bomb because the voting system is terrible—when are we getting a tiered voting system? I haven’t been playing much until recently, but I’ve noticed the opposite trend: Unpredictable and Chill getting picked more often. Unpredictable is supposed to be medium-low chaos with adminbus, but in practice, that probably doesn’t happen.

I can’t stand Extended or no/low chaos rounds. Playing Security with nothing to do for two and a half hours is miserable. I want variety, intensity, and unpredictable antags. I want extreme chaos sprinkled in occasionally sure, but please just dont try to make me sit though another chill round.

an idea... I want admins to take on a storyteller role, actively modifying upcoming events and doing more adminbussing. 

The server’s mission is to always have antags - no exceptions. I feel storyteller PRs are undermining that by leaning toward Extended/Chill. I also hate the repetitive cycle of Chill → Clown/Default → Chill → Clown/Default. When I do get time to play, I either have to sit or wait through a Chill round to play or stop playing because Clown or Bomb isn’t an option anymore and chill likely gets picked. And to make it worse, some Clown rounds where fuck all happened.

PR #2956, opened 2025-01-18
PDA Alert Level Readout

ArrisFairburne:
## About The Pull Request
Adds an alert level notice on the main PDA screen.
The notice starts out as simple colored text when the user is not in any danger or concern. If the user is in danger, it becomes black text on a colored blackground. If the user is called to fix the problem relevant to alert (ex. security to deal with red alert active threats, medical to handle a biohazard) the notice becomes blinking text on a colored background. 

## Why It's Good For The Game
The existing way to check current alert levels (look at a fire alarm) doesn't really make much sense for new players. Also, this way is more accessible.

## Proof Of Testing

<details>

GIF 18-Jan-25 4-07-30 PM

</details>

## Changelog

:cl:
qol: Added a button to the main menu of the modular computer screen, which will tell the user what the current alert level is and if they are needed to fix the problem.
/:cl:

### DIFF
--- a/code/modules/modular_computers/computers/item/computer.dm
+++ b/code/modules/modular_computers/computers/item/computer.dm
@@ -1027,50 +1027,68 @@
 
 	return 0
 
+#define ALERT_SHORTFORM_GAMMA = "Γ"
+#define ALERT_SHORTFORM_EPSILON = "E"
+#define ALERT_SHORTFORM_DELTA = "Δ"
+#define ALERT_SHORTFORM_RED = "RED"
+#define ALERT_SHORTFORM_AMBER = "AMB"
+#define ALERT_SHORTFORM_ORANGE = "ORN"
+#define ALERT_SHORTFORM_GREEN = "GRN"
+#define ALERT_SHORTFORM_BLUE = "BLU"
+#define ALERT_SHORTFORM_VIOLET = "VIO"
+
 /// Returns strings for the current security level to display on the PDA.
 /obj/item/modular_computer/proc/get_security_level_shortform()
 	switch (SSsecurity_level.get_current_level_as_number())
 		if(SEC_LEVEL_GAMMA) //you're just fucked if you see a greek letter
-			return "Γ"
+			return ALERT_SHORTFORM_GAMMA
 		if(SEC_LEVEL_EPSILON)
-			return "E"
+			return ALERT_SHORTFORM_EPSILON
 		if(SEC_LEVEL_DELTA)
-			return "Δ"
+			return ALERT_SHORTFORM_DELTA
 		if(SEC_LEVEL_RED)
-			return "RED"
+			return ALERT_SHORTFORM_RED
 		if(SEC_LEVEL_AMBER)
-			return "AMB"
+			return ALERT_SHORTFORM_AMBER
 		if(SEC_LEVEL_VIOLET)
-			return "VIO"
+			return ALERT_SHORTFORM_VIOLET
 		if(SEC_LEVEL_ORANGE)
-			return "ORN"
+			return ALERT_SHORTFORM_ORANGE
 		if(SEC_LEVEL_BLUE)
-			return "BLU"
+			return ALERT_SHORTFORM_BLUE
 		if(SEC_LEVEL_GREEN)
-			return "GRN"
+			return ALERT_SHORTFORM_GREEN
 
 	return "NUL"
+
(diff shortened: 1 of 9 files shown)


code/modules/modular_computers/computers/item/computer.dm
```diff
@@ -981,6 +985,96 @@
 		return stored_files
 	return stored_files + inserted_disk.stored_files
 
+/// Returns how relevant the current security level is:
+/// * 0: User is not in immediate danger and not needed for some station-critical task.
+/// * 1: Danger is around, but the user is not directly needed to handle it.
+/// * 2: Danger is around and the user is responsible for handling it.
+/obj/item/modular_computer/proc/get_security_level_relevancy(var/obj/item/card/id/idToCheck)
```
Shadow-Quill, line 992: I'd suggest using defines instead of magic numbers.

code/modules/modular_computers/computers/item/computer.dm
```diff
@@ -981,6 +985,96 @@
...
+	var/list/id_accesses = list()
+	id_accesses = saved_id_accesses
+
+	switch (SSsecurity_level.get_current_level_as_number())
```
Shadow-Quill, line 996: I _might_ be wrong, but I believe current code standard is to not have a space between a call and its parenthesis (ie: `if(xyz)` instead of `if (xyz)`

tgui/packages/tgui/layouts/NtosWindow.tsx
```diff
@@ -21,6 +21,9 @@ export type NTOSData = {
   light_on: BooleanLike;
   login: Login;
   pai: string | null;
+  alert_style: number;
+  alert_color: string;
+  alert_name: string;
```
Shadow-Quill, line 26: Are these used in this file?

Shadow-Quill (changes requested):
I'd love if the UI could update when the alert level changes, but that'd probably require signal stuff, and I CBA to look up enough to suggest anything rn.

Otherwise, I like this.

tgui/packages/tgui/layouts/NtosWindow.tsx
ArrisFairburne: They are used in NtosMain.tsx, which uses the same data variables as NtosWindow.tsx.

code/modules/modular_computers/computers/item/computer.dm
ArrisFairburne: I have seen `if (xyz)` around in the other parts of the code, and in my opinion the difference between space and spaceless doesn't matter when reading code. Regardless it has been changed on the latest commit.

ArrisFairburne:
> I'd love if the UI could update when the alert level changes, but that'd probably require signal stuff, and I CBA to look up enough to suggest anything rn.

It should be as simple as moving the `data[]` calls from `ui_static_data` to `ui_data`. However, I elected not to do this because the alert notice shouldn't need to be updated immediately (the only user situation where the notice would be inaccurate to the real alert level would be immediately after it is changed, in which case the player can just refer to the announcement the alert level is changed), and because `ui_static_data` is "for heavy uis that would be sending a lot of redundant data frequently" (which the alert level would be in most situations; it's not usually changed more than once every few minutes).

ArrisFairburne:
Actually, come to think -- would this be appropriate to bring upstream? Not sure what that process would be, so if you have a guide for bringing that up it'd be much appreciated.

aKromatopzia:
> Actually, come to think -- would this be appropriate to bring upstream? Not sure what that process would be, so if you have a guide for bringing that up it'd be much appreciated.

the process is simply to repeat this code on TG's github, really

Arturlang:
Make changing alert levels update static data instead, i'd say

Arturlang:
> Actually, come to think -- would this be appropriate to bring upstream? Not sure what that process would be, so if you have a guide for bringing that up it'd be much appreciated.

Just remove the modular changes, reviews may be a bit more stringent, not much different otherwise, you can cherrypick or directly bring over your changes

ArrisFairburne:
As it turns out the only things I needed to rewrite from my code additions are that tgstation does not have any alert levels besides RGB and delta, so I had to remove those.

I'll keep this open for now since we have different stuff to TG.

TealSeer:
https://github.com/tgstation/tgstation/pull/89173

PR #2971, opened 2025-01-19
Makes the clock sound not penetrate walls

SingingSpock:
## About The Pull Request

Exactly what it says on the tin. Makes the soundloop for clocks not make noise through walls.
## Why It's Good For The Game

Constantly ticking clocks are an annoyance that can be hard to escape as command on certain maps like Moonstation. There's no reason that a wall thick enough to hold back vacuum should be letting a clock noise through it like nothing.

## Proof Of Testing

Can't really screenshot a sound, but I did test this on local and it works.

## Changelog
:cl:
fix: The ticking of clocks can no longer be heard through walls, filling the bridge with an incessant ticking noise like the Telltale heart's thumping.
/:cl:


ReturnToZender:
This sound was totally kino but this is still a good PR

Majkl-J (approved):

PR #2982, opened 2025-01-21
Fixes pirates round event so they spawn in story teller

Odairu:
## About The Pull Request
So it turns out that the list of possible pirates was only ever populated during the dynamic game mode procs to run the event, since we use story teller and not dynamic, that means the list was empty. Empty list means no pirates. This PR makes it so that if the list is empty (it is only ever populated when an admin uses the trigger-event verb to manually spawn pirates) it will use the combined global lists of heavy and light pirates, pick one, and fire it.
## Why It's Good For The Game
this antag been broke for like a year and 3 months
## Proof Of Testing
image
<details>

</details>

## Changelog
:cl:
fix: pirates spawn again
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/antagonists/pirate/pirate_event.dm
+++ b/modular_zubbers/code/modules/antagonists/pirate/pirate_event.dm
@@ -1,0 +1,9 @@
+/datum/round_event/pirates/start()
+	//Check for admin customization - gang_list is only populated if admins manually spawn the event
+	if(gang_list)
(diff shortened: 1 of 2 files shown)


NargaInDisguise:
The fact that noone either noticed nor picked up on this issue for this long is wild

Majkl-J (approved):

Majkl-J:
Since there's light and heavy gangs I'll use that later down the line with storytellers

PR #2999, opened 2025-01-23
Replaces The Clown storyteller with The House (Adminbus storyteller)

StrangeWeirdKitten:
## About The Pull Request

While The House will set weights to one still, it'll be running on the fragile ruleset (Medium to low chaos)

The catch, admins are given the opportunity to drive the admin bus straight into the station. The house now runs the show.

## Why It's Good For The Game

I was told by multiple admins that they used Clown as the dedicated adminbus storyteller. Now I'll make it official, slow down clown, and give admins the opportunity to take over. The goal is to make it feel more like a coin flip then just hyper chaos every round.

## Proof Of Testing

works. 

## Changelog

:cl:
add: Replaces The Clown with The House which sets weights to 1, but spawns stuff at a slower pace, and allows the adminbus to come.
fix: Voidwalker properly spawns on the ghostset now.
/:cl:


Majkl-J (approved):

StrangeWeirdKitten:
TM for now because I might still need to make adjustments to its intensity.

The-Sun-In-Splendour:
I think we should make the house's intro text be "No! He needs mouse bites to live." because that'd be pretty funy i think. :3 smile.

CatoChristopherMrow:
This kicks ass.

theselfish:
Default Andy still references the Clown.

"Default Andy (Medium Chaos)
Default Andy is the default Storyteller, and the comparison point for every other Storyteller. More frequent events than the Chill or the Fragile, but less frequent events than The Gamer or the Clown. Best for an average, varied experience.
"

Yanziko:
i think it's **really** important that somewhere a house m.d reference is made...  after all, how could we live without mouse bites?

PR #3034, opened 2025-02-01
Prevents certain events and antag spawns from running depending on the status of the shuttle.

BurgerLUA:
# About The Pull Request

Prevents certain events and antag spawns from running depending on the status of the shuttle.

Makes it by default, all events cannot be triggered if the shuttle is docked with the station (or beyond that point).

Most crew-spawn antagonist events will no longer run if the shuttle is in transit (excluding recall to central command status). Stuff like blob, spiders, xenos, etc follow the new default condition above.

Note that per /tg/station default, all events cannot run if the shuttle left the station.

## Why It's Good For The Game

There is a bit of an issue with antagonists feeling the urge to speedrun their newly acquired antag status because they rolled antag when the shuttle is coming. This is generally bad for roleplay since there is less focus on that and more focus on "How can I make the most use of my TC before the round ends."

For antagonists (and events not controlled by players) where roleplay isn't required, I made it so that they can only occur if the shuttle is not docked with the station (or beyond that point).

## Proof Of Testing

Draft because untested and feedback is required (plus it's saturday morning and I am not even sure if any of this works).

## Changelog

:cl: BurgerBB
del: By default, all events cannot be triggered if the shuttle is docked with the station (or beyond that point). All crew-spawn antagonist events antagonist events will no longer run if the shuttle is in transit (excluding recall to central command status).
/:cl:


ReturnToZender:
This seems like a good idea! I don't think anyone would have an issue with making it so events won't run while the shuttle's in transit or leaving.

projectkepler-ru:
> This seems like a good idea! I don't think anyone would have an issue with making it so events won't run while the shuttle's in transit or leaving.

it definitely wasnt supposed to cause I was fucking around w/ dynamic and it avoid spawning event on shuttle call

BurgerLUA:
Damn I forgot about this for 2 weeks

Will work on it when I have the time.

BurgerLUA:
I recommend a test merge in case there was something I missed from testing.

Hemlock32:
image
Better late than never

Majkl-J (changes requested):
Good change, however probably way too complicated for what we really need (Preventing storyteller from processing tracks when you hit `!SSshuttle.canRecall()` )

BurgerLUA:
> Good change, however probably way too complicated for what we really need (Preventing storyteller from processing tracks when you hit `!SSshuttle.canRecall()` )

I think that's too simple of a solution considering that some events are okay to run when the shuttle is coming, and this wouldn't stop situations where you can get antag in the last 10 minutes since the shuttle can still be recalled.

The code itself isn't really complicated, and is similar to what already exists on /tg/. The code was designed to have support in the future and cause the least headaches possible while getting the job done.

Majkl-J:
> I think that's too simple of a solution considering that some events are okay to run when the shuttle is coming, and this wouldn't stop situations where you can get antag in the last 10 minutes since the shuttle can still be recalled.

> The code itself isn't really complicated, and is similar to what already exists on /tg/.

From /code/controllers/subsystem/dynamic/dynamic_midround_rolling.dm on TG:
```
/datum/controller/subsystem/dynamic/proc/try_midround_roll()
	if (!mid_forced_injection && next_midround_injection() > world.time)
		return

	if (GLOB.dynamic_forced_extended)
		return

	if (EMERGENCY_PAST_POINT_OF_NO_RETURN)
		return
```

No midrounds can roll past the point of the shuttle call on TG, and I don't see why we really need a more complicated system to theirs in this regard. There's not really a better way here - either we waste rolls because the captain called then recalled randomly, or we very rarely roll someone on an early call. Roundend shuttles will always return EMERGENCY_PAST_POINT_OF_NO_RETURN as TRUE

BurgerLUA:
All implementing that accomplishes is that instead of getting antag 2 minutes before the shuttle, you get antag 5 minutes before the shuttle.

The emergency point of no return is:

10 minutes for greenshift.
5 minutes for blue.
2.5 minutes for red.

Specifying whether or not it is in transit adds flexibility in what gets run so you can still get events while the shuttle is coming, while just blocking the antag ones.

Like what is the concern here with the code not simple according to your standards?

PR #3047, opened 2025-02-03
Trims excessive roundstart security guns

YakumoChen:
## About The Pull Request

This PR removes the skyrat-added "gun beacons" that security officers and the Head of Security are spawned into the game with.
In addition, moves the silver glock case from the Head of Security's locker to the Warden's locker

## Why It's Good For The Game
Security officers already get a glock in their lockers (a bubber-added glock) as a sidearm to use, there is little reason for them to spawn with a SECOND sidearm on their person. Given this is Bubber, I preferred to remove the Skyrat option over the native Bubber one.

This PR was originally intended to remove some of the **FIVE GUNS** that the Head of Security has personally available to him at round-start. This PR trimmed it down to three, and I gave the Warden a little love by giving them the cool glock instead (the warden never gets anything cool...). If the HoS really wants a glock or something else, they have the means.

## Proof Of Testing
<details>

</details>

## Changelog
:cl:
balance: Removed "sidearm beacons" from spawning SecOff and HoS inventories. Complimentary glocks are still available in security officer lockers.
balance: Gave the HoS's silver glock to the Warden. The Head of Security only has three guns at roundstart now! How will he ever recover??
/:cl:


Bapldap:
remove the locker guns too while you're at it :D

VickiMorris:
> their damage multiplier reduce the viablity of using different ammo type at all

I agree here, the .65 multiplier on damage in tandem with the MASSIVE firing delay should be rolled back at _least_ slightly. The HOS's (now Warden's) varient also just, is better with a .75 damage mult, and lower firing delay.

> and making them, at best, a very bad early game weapon for spider and dragon, and at worst, something to throw away as soon as you buy a wespe

I think the Warden should do their job and hand out weaponry that is more appropriate to the situation is these cases. I think the alert level pinning is silly, considering they come loaded with flat-heads and not proper lethals, but I also understand wanting to force people to use disablers and batongs (Read: Non-Lethal) on green.

Rykka-Stormheart:
I really liked the R10 though :c

It was cooler than the pistol.

YakumoChen:
> If you're gonna do this, revert 9x25mm Nerf aswell

The guns are still available through cargo

> additionally the cck pistol still arent ideal

Gun balance is out of scope. They're a roundstart sidearm, not murderhobo guns

Arturlang:
To be fair security officers should not and do not need roundstart ballistics for multiple reasons. Them getting roundstart makarovs is fucking horrific, and the security revolvers are damn strong. I'd be okay with removing both glocks

YakumoChen:
> To be fair security officers should not and do not need roundstart ballistics for multiple reasons. Them getting roundstart makarovs is fucking horrific, and the security revolvers are damn strong. I'd be okay with removing both glocks

Let me know if maintainers reach a consensus of how many gocks each officer needs packed in their trousers, I think this PR lowers it to an acceptable middle ground for now

DorfGuy1:
Do not remove the signalis reference  revolver. Also yeah the guns kinda suck. Idk if they want them to be better or worse or not. Or not have guns at roundstart at all. I've no clue really. As is they suck and are just a worse wespe.

YakumoChen:
This isn't the PR for balancing guns.

 If you want my personal opinion, I think its good the default choice might be underpowered. It isn't meant to be good, and having an acceptably bad default means only the sweaty players are going to replace it with a wespe immediately (they will regardless).

BurgerLUA:
The Glock is actually balanced. It has the same DPS as a laser rifle with normal ammo. The reason why it feels weak is because Skyrat guns are extremely strong by comparison.

I also support this PR. We need less Skyrat Gun Bloat, and more Bubberstation Gun Bloat.

Moostard:
This is kinda sad to see.
Please reconsider your choice of removal.

You are removing a beacon that gives you a choice between 2 different weapons that are solid overall and are generally liked.

C-CK glocks on the other hand are no doubt very underwhelming.
The only people carrying those are fresh security players that have no idea of the weapons capablities, which tends to leave them disappointed or clueless since it does single digit damage to armored targets or/and it does not shoot due to the alert pin in it making the gun effectively useless (was a common case against random simplemobs).

The glock itself is a weapon you take and replace the moment you can.
It has low capacity, very low damage and no option for silencer iirc.
You cannot put a naked target (assuming no additional damage source) to critical condition with an entire mag of starter flathead ammunition. 15.6 damage with armor weakness is quite pathetic.
Missing 2 regular ammo shots against a naked target forces you to reload.

There is no need to imagine what will happen after removal as before the beacon was a thing, security ordered wespes en masse as it was a good choice for a sidearm. There were also other choices, but this one was the most popular.
I do not believe that starter kits should be bad on purpose just because it is what you get at the start. It should at least do its job decently enough and C-CK does not hit that bar.

If you are set on removing the beacon, please consider removing the C-CK as well to save the newbies.
The only reason "not sweaty" players keep the starter kit is likely because they put too much trust in it and the examine damage information is likely lying to them if it does not put the damage mod into account.
Thanks.

projectkepler-ru:
> The Glock is actually balanced. It has the same DPS as a laser rifle with normal ammo. The reason why it feels weak is because Skyrat guns are extremely strong by comparison.
> 
> I also support this PR. We need less Skyrat Gun Bloat, and more Bubberstation Gun Bloat.

The reason why it feels weak is because it IS weak, we are comparing 8 shot of  21 damage to 20 shot of 25 damage
it is literally better to buy a crate of mini e gun for damages 
pistoleer prefereed actual magazine fed handgun for convenience

projectkepler-ru:
> Do not remove the signalis reference revolver. Also yeah the guns kinda suck. Idk if they want them to be better or worse or not. Or not have guns at roundstart at all. I've no clue really. As is they suck and are just a worse wespe.

FINALLY someone gets its a signalis reference

projectkepler-ru:
> > additionally the cck pistol still arent ideal
> 
> Gun balance is out of scope of this PR. They're a roundstart sidearm, not murderhobo guns.

gotcha

Redrover1760:
> > their damage multiplier reduce the viablity of using different ammo type at all
> 
> I agree here, the .65 multiplier on damage in tandem with the MASSIVE firing delay should be rolled back at _least_ slightly. The HOS's (now Warden's) varient also just, is better with a .75 damage mult, and lower firing delay.
> 
> > and making them, at best, a very bad early game weapon for spider and dragon, and at worst, something to throw away as soon as you buy a wespe
> 
> I think the Warden should do their job and hand out weaponry that is more appropriate to the situation is these cases. I think the alert level pinning is silly, considering they come loaded with flat-heads (EDIT: These are not less-than-lethal like I previously thought, but just neutered lethals), but I also understand wanting to force people to use disablers and batongs (Read: Non-Lethal) on green.

They aren't level alert pinned at the moment, as far as I'm aware.

VickiMorris:
> They aren't level alert pinned at the moment, as far as I'm aware.

https://github.com/Bubberstation/Bubberstation/blob/master/modular_zubbers/code/modules/security/security_glock/gun.dm

The HoS's varient isn't, the one regular security officers get is.

Redrover1760:
R10 probably got all of 10mm ammo nerfed to 30 damage instead of 40 damage (ansem rounds) for no reason because thanks skyrat.

Formerxboxgamer:
The bubber Glock is really really bad. I had no idea it had hidden modifiers on it too, that’s the definition of a noob trap. Also the firing sound is just the hl2 pistol and it comes off as dinky. It’s definitely out of scope of this PR, but we should probably rework it with a new name, coat of paint, and sound

sippykot:
> The bubber Glock is really really bad. I had no idea it had hidden modifiers on it too, that’s the definition of a noob trap. Also the firing sound is just the hl2 pistol and it comes off as dinky. It’s definitely out of scope of this PR, but we should probably rework it with a new name, coat of paint, and sound

it has a <1 damage multiplier, a significant one
imo it should be buffed but thats another PR

projectkepler-ru:
> R10 probably got all of 10mm ammo nerfed to 30 damage instead of 40 damage (ansem rounds) for no reason because thanks skyrat.

No problem hun :v

Razurath:
> This isn't the PR for balancing guns.
> 
> If you want my personal opinion, I think its good the default choice might be underpowered. It isn't meant to be good, and having an acceptably bad default means only the sweaty players are going to replace it with a wespe immediately (they will regardless).
> 
> Is it bad that security has lethals at roundstart? Probably, but said sweaty players will get an option regardless, everyone else will stick it out with what they get first.
> 
> That might say a lot more about what exists in cargo that needs reconsidered, but I don't have the stats on hand for sure for that determination (and won't be in this PR)

The entire basis of this PR is balance so saying it isn't a balancing PR is weird. Removing stuff isn't always the best option, the glock in lockers is less effectively then just bludgeoning someone to death with your baton, the beacon gives actual guns and they start with non lethals. Removing them round start makes sense, but perhaps leave them something that can be ordered on the request console or privately from cargo with sec/armory access.

Razurath:
Also you have again, I say again because this is the third PR I've seen you do this, done a removal without checking stats on stuff.

VickiMorris:
> Removing them round start makes sense, but perhaps leave them something that can be ordered on the request console or privately from cargo with sec/armory access.

You can buy two sidearm choice beacons from cargo for the low-low price of 2400cr! Which is entirely not worth it!

I don't think they're 1200cr each good, especially when imports lets you buy shotguns and SMGs for 600cr a pop. Maybe compared to a wespe or eland, but with all the options we have? There's better stuff you can get.

YakumoChen:
> Also you have again, I say again because this is the third PR I've seen you do this, done a removal without checking stats on stuff.

Yeah, I didn't check the stats on stuff because this isn't a gun balance PR. This is just removing roundstart access to excessive sidearms

Arturlang:
@Razurath The glock in the beacon should honestly be gone forever, It is literally, just, a makarov, and a makarov that can get more ammo stat-wise

Arturlang:
image
Closet glock has a damage modifier of 0.65, 0.75 on the HOS's

lessthnthree (approved):

DorfGuy1:
I would suggest that having a revolver  that is a little more strong (a little) as an option roundstart, will actually kind of move the escalating down. Think cops in like the 60s. They had revolvers cause thats all they needed.  then criminals got their wonder nines, and cops got glocks. let sec put themselves on the 1960s cop level while everyone else is in the 2500s

projectkepler-ru:
did you not check the override files
it's 25 in game for standard and 20 for ap, 35 for hP @Arturlang

Arturlang:
> did you not check the override files it's 25 in game for standard and 20 for ap, 35 for hP @Arturlang
> 
> Also it has spread of 10, which makes it very inaccurate at long range

yea, apparently c10mm has 30 damage, I don't see overrides for the rest though, doesn't that mean AP is better than normal?

PR #3063, opened 2025-02-07
Heretic blade shattering QOL - HUD instead of inhand

KazooBard:
## About The Pull Request
Have you ever had this happen to you?
brickandthehertic

If so, you know that double clicking to pick up a blade off the floor or using it inhand because you forgot to switch hands before toggling a baton uses the blade - breaking it, ending your fun, round, run, sanity . And why? Well... Just because, really. Noone likes it, it's not fun and pointless.

SO WE REMOVED THAT! Now to break a blade you use it like a spell, like your jaunt for example. Press the icon, boom. NOTHING changed about how it functions outside the way to proc it. And yes, you can bind it.

## Why It's Good For The Game

Trust me bro.

## Proof Of Testing

https://github.com/user-attachments/assets/174ac310-368d-4b35-a00b-b7cf5ca6142b
https://github.com/user-attachments/assets/9ba99057-4802-4934-b648-35b6275d30fa

## Changelog

:cl:
qol: Heretic blades are now shattered using a HUD button
/:cl:

### DIFF
--- a/code/modules/antagonists/heretic/items/heretic_blades.dm
+++ b/code/modules/antagonists/heretic/items/heretic_blades.dm
@@ -24,6 +24,21 @@
 	attack_verb_continuous = list("attacks", "slashes", "stabs", "slices", "tears", "lacerates", "rips", "dices", "rends")
 	attack_verb_simple = list("attack", "slash", "stab", "slice", "tear", "lacerate", "rip", "dice", "rend")
 	var/after_use_message = ""
+	actions_types = list(/datum/action/item_action/break_blade) //bubberstation edit start
+
+/datum/action/item_action/break_blade
+	name = "Shattering Offer"
+	background_icon_state = "bg_heretic"
+	overlay_icon_state = "bg_heretic_border"
+	button_icon = 'icons/mob/actions/actions_ecult.dmi'
+	button_icon_state = "shatter"
+
+/datum/action/item_action/break_blade/Trigger(trigger_flags)
+	if(target in owner.held_items)
+		var/obj/item/melee/sickly_blade/target_item = target
+		target_item.seek_safety(owner)
+		return
+//bubberstation edit end
 
 /obj/item/melee/sickly_blade/examine(mob/user)
 	. = ..()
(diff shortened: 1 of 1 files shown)


Majkl-J (commented):
Do we really need an extra UI button for preventing a fumble similar to accidentally smacking someone because you forgot combat mode?

zeroisthebiggay:
this does slow down breaking your blade considerably and assigning a hotkey is likely to unassign with each new blade
unless there's a hotkey which when pressed will break your inhand blade and isn't attached to the blade itself then i don't like this

also does this break being able to smash a heretic's blade as a nonhheretic
making sure

destrucktoid:
In the situation where you genuinely need to break your blade to escape, you don't have time to fumble for a UI button or to remember which button you had it assigned to, especially when there is a risk of you either forgetting to assign a button to it or just forgetting which button it's on. Not to mention that with how many spells heretics have nowadays you don't have enough buttons to assign them all and this will conflict for a button spot.

Making it so you can't accidentally break a blade like this isn't the right call in my opinion and as a result i'm personally opposed to the change. After all, the amount of times i've accidentally broke blade mid fight can be counted on one hand. It's rare to do it and it's not usually a big deal when it happens. It's definitely not round-ending as the PR implies.

KazooBard:
> Do we really need an extra UI button for preventing a fumble similar to accidentally smacking someone because you forgot combat mode?

I have talked to quite a few mainstay heretic players, of which none so far have disagreed with the change. From what I have heard, most prefer this to the alternative, as it happens mostly when trying to pick up the blade off the floor. I have a much different experience personally with blades breaking when I didn't mean it. 

I can count the amount of times I haven't broken the blade quickly enough, but of the amount of times I tried to pick it up off the floor and clicked it's icon twice whilst picking it up is one that I have lost track of.

VickiMorris:
> but of the amount of times I tried to pick it up off the floor and clicked it's icon twice whilst picking it up is one that I have lost track of.

If that's the issue, couldn't you make it so there's a half-second (or so) timer when you equip the blade before you can shatter it in hand? Long enough to avoid accidentally procing it, while keeping the same behavior and quick access to a blade break?

projectkepler-ru:
This is stupid

StrangeWeirdKitten:
making it a hud button to remember to click kind of does the opposite of breaking your blade in a rush.

PR #3076, opened 2025-02-11
Readds Advanced hypospray

Vanilla1040:
## About The Pull Request

This PR Readds the Advanced hypo in its already nerfed state to the ~~medical lathe~~ moved to imports and also readds the Large Hypobox that goes on your belt slot, the problem was already solved and i think its silly how you are forced to buy a whole tech kit JUST for the hypo as the PR which removed it from being printable did NOT add it into imports and locked it to said tech kit, its still locked behind research of course.

## Why It's Good For The Game

This stops Medical staff from being forced to fork out 2500 credits, or a 20 minute department order cooldown for a hypo upgrade on a tool that has already had the initial problem solved, ~~allowing for the progression/upgrade path to be returned.~~ This includes re-adding the large hypo box which was also removed.

## Proof Of Testing

Complies and works on localhost

## Changelog

:cl:
add: Readds the Large Hypobox being Printable in the Medical Lathe
add: Adds Advanced Hypospray into imports
/:cl:

### DIFF
--- a/modular_skyrat/modules/company_imports/code/armament_datums/deforest_medical.dm
+++ b/modular_skyrat/modules/company_imports/code/armament_datums/deforest_medical.dm
@@ -238,6 +238,10 @@
 	item_type = /obj/item/device/custom_kit/deluxe_hypo2
 	cost = PAYCHECK_COMMAND * 2
 
+/datum/armament_entry/company_import/deforest/equipment/advanced_hypospray
+	item_type = /obj/item/hypospray/mkii/piercing
+	cost = PAYCHECK_COMMAND * 7
+
 /datum/armament_entry/company_import/deforest/equipment/afad
 	item_type = /obj/item/gun/medbeam/afad
 	cost = PAYCHECK_COMMAND * 5
--- a/modular_skyrat/modules/hyposprays/code/autolathe_designs.dm
+++ b/modular_skyrat/modules/hyposprays/code/autolathe_designs.dm
@@ -63,23 +63,6 @@
 	)
 	departmental_flags = DEPARTMENT_BITFLAG_MEDICAL
 
-/datum/design/hypomkii/advanced
-	name = "Hypospray Mk. II Advanced"
-	id = "hypomkii_adv"
-	build_type = PROTOLATHE | AWAY_LATHE
-	materials = list(
-		/datum/material/plastic = SHEET_MATERIAL_AMOUNT * 5,
-		/datum/material/glass = SHEET_MATERIAL_AMOUNT * 3,
-		/datum/material/silver = SHEET_MATERIAL_AMOUNT,
-		/datum/material/titanium = SHEET_MATERIAL_AMOUNT,
-		/datum/material/diamond = HALF_SHEET_MATERIAL_AMOUNT,
-	)
-	build_path = /obj/item/hypospray/mkii/piercing
-	category = list(
-		RND_CATEGORY_TOOLS + RND_SUBCATEGORY_TOOLS_MEDICAL_ADVANCED,
-	)
-	departmental_flags = DEPARTMENT_BITFLAG_MEDICAL
-
 // Hypospray upgrade
 /datum/design/hypomkii/deluxe
 	name = "Hypospray Mk. II Deluxe Upgrade"
@@ -107,14 +90,12 @@
 
 /datum/techweb_node/medbay_equip_adv/New()
 	design_ids += list(
-		"hypomkii_adv",
 		"hypokit_deluxe",
 	)
 	return ..()
 
 /datum/techweb_node/alien_surgery/New()
 	design_ids += list(
-		"hypomkii_adv",
 		"hypomkii_deluxe",
 	)
 	return ..()
--- a/modular_skyrat/modules/hyposprays/code/autolathe_designs.dm
+++ b/modular_skyrat/modules/hyposprays/code/autolathe_designs.dm
@@ -1,3 +1,4 @@
+// Hypovials
 /datum/design/hypovial
 	name = "Hypovial"
 	id = "hypovial"
@@ -8,18 +9,10 @@
 	)
 	build_path = /obj/item/reagent_containers/cup/vial/small
 	category = list(
-		RND_CATEGORY_INITIAL,
 		RND_CATEGORY_EQUIPMENT + RND_SUBCATEGORY_EQUIPMENT_MEDICAL,
 	)
 	departmental_flags = DEPARTMENT_BITFLAG_MEDICAL
 
-/datum/techweb_node/medbay_equip/New()
-	design_ids += list(
-		"hypovial",
-	)
-	return ..()
-
-/// Large hypovials
 /datum/design/hypovial/large
 	name = "Large Hypovial"
 	id = "large_hypovial"
@@ -29,6 +22,7 @@
 	)
 	build_path = /obj/item/reagent_containers/cup/vial/large
 
+// Hypospray cases
 /datum/design/hypokit
 	name = "Hypospray Case"
 	id = "hypokit"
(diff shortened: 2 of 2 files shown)


Lazhannya:
Removing Hypos from the lathe was a really pointless change to begin with, as all it does is making medical staff's job harder for no reason. And as was said in the PR it makes playing medical either a race for credits to get a completely overpriced tech kit for 2500 credits, or completely clogs up the department orders - which makes any other order impossible to get through and completely kills variation or reaction to various situations if there isn't a CMO with medical budget available.

Especially since the effect of having the advanced sprays be instantaneous has already been fixed and moved to antags-only.
Thus re-adding them would simply unburden medical staff, make treatment of patients with modsuits and similar more accessible (or even possible at all, if its a fused suit) and encourage various cargo orders again, rather than simply focusing on spamming tech kits.

NargaInDisguise:
> So what is the real point in keeping them out?

Crew-side medical powercreep. If someones stuck in a modsuit, it's of their own volition. Either take them out, or if they're entombed, get threadrippers. If you want to use chems on them, just use patches or pills, or the piercing syringe that everyone forgets exists because of the hyposprays. These hypos just shouldn't have ever been so accessible, even if they're now nerfed.

Vanilla1040:
Thing is Narga, lets say a secoff is dying on the floor in a MOD, its faster to use said Hypo to inject them with stuff that will safe them, taking off the MOD takes about as long as raising my first born child, by the time they could die which means a trip to medbay which means more time spend treating them instead of getting them back up there and then, no one carrys around pill bottles 24/7 filled with patches or pills, patches can work yes but pills most the time wont as any Mask and MOD will prevent the user from taking said pill making that point invalid, piercing Syringe deals damage, it may not be much but it can be a difference in them dying from the shot as they are already in crit and them living.

Vanilla1040:
direct injection dont no but no one carrys 10 beakers around with chems, i thought they ment them being used from a syringe gun my bad,

aKromatopzia:
> direct injection dont no but no one carrys 10 beakers around with chems, i thought they ment them being used from a syringe gun my bad,

I said point to where in the code -- because this is a hallucinatory statement.

Vanilla1040:
image

Bubberstation\code\modules\projectiles\projectile\bullets\dart_syringe

Moostard:
The only valid point for the nerf was the near instant inject for advanced hypos. But removal was not needed.
Now you only see them carried by people who were first to unbox the cargo department order or taking a lot of time to do bounties to get an entire ungodly overpriced kit with it.

They were a part of research, which aims to unlock better tools for all departments.
Hyposprays make carrying and applying medicine easier, advanced ones bypass thick suits, reducing the time and tedium of stripping someone.

There is no powercreep in it.
People clearly enjoy having them as medical for reasons stated above.
Not much different from carrying beakers of chems and a piercing syringe, but that should not be a reason to not allow hyposprays to exist.

Arturlang:
advanced hyposprays should not exist in autolathes, a alternative way to get them is fine, it just makes syringes just way too useless otherwise, normal hyposprays are fine. You can add a cheaper import for advanced hyposprays if you want, but it should not be trivially acquirable

Arturlang:
And that's why normal hyposprays still exist and are plentiful, piercing should be difficult to obtain, or inconvenient via something like piercing syringes

Moostard:
The word "useless" should be replaced with "inferior".
Just like screwdrivers, wrenches and crowbars become inferior to power tools once research gets the node. Those powertools can be printed en masse with engi techfab. You never see an engineer with normal tools if they visits the fabricator once after that.
Inconvenience and tedium just for the sake of it does not sound like a fun idea. Who said it **should** be that way?

This is a similar thing but instead of tools and power tools, we got syringes and hyposprays.
It is a tool to store and apply medicine with the advanced version making you skip the stripping part if someone is wearing a modsuit or similar.

YakumoChen:
> Thing is Narga, lets say a secoff is dying on the floor in a MOD, its faster to use said Hypo to inject them with stuff that will safe them, taking off the MOD takes about as long as raising my first born child, by the time they could die which means a trip to medbay which means more time spend treating them instead of getting them back up there and then, no one carrys around pill bottles 24/7 filled with patches or pills, patches can work yes but pills most the time wont as any Mask and MOD will prevent the user from taking said pill making that point invalid, piercing Syringe deals damage, it may not be much but it can be a difference in them dying from the shot as they are already in crit and them living.

Being in a modsuit is a tradeoff, there isn't a reason we need a super flexible bypass for these systems, as opposed to the more stringent ones that exist (using patches)

whatisthisplace:
> Thing is Narga, lets say a secoff is dying on the floor in a MOD, its faster to use said Hypo to inject them with stuff that will safe them, taking off the MOD takes about as long as raising my first born child, by the time they could die which means a trip to medbay which means more time spend treating them instead of getting them back up there and then, no one carrys around pill bottles 24/7 filled with patches or pills, patches can work yes but pills most the time wont as any Mask and MOD will prevent the user from taking said pill making that point invalid, piercing Syringe deals damage, it may not be much but it can be a difference in them dying from the shot as they are already in crit and them living.

Why cant you just flip the hypo from inject to spray??

Vanilla1040:
It can fun fact, Left Click is Spray, Right Click is inject which does the exact same thing apparently

Arturlang:
> It can fun fact, Left Click is Spray, Right Click is inject which does the exact same thing apparently

It's a different in inject speeds/method type, which matters for some chemicals

Arturlang:
@Moostard you are correct in that it makes medical gameplay more efficient and quick, but the issue here is that hyposprays are simply better, quicker and have more utility than syringes, and are capable of shutting down antags, even if advanced is no longer instant, it will still be used with it's rather quick inject while the victim is not moving too much or stunned from a moment.
Thus while I don't mind the adv hypo existing, I'd rather it not be as easily accessible as powertools are to engineering, especially when it bypasses a major feature like suits and thick clothing preventing injection, which is the main way a lot of antags protect themselves from death/sleep-mixes

aKromatopzia:
> Sir in what way is powertool a sidegrade, they always universally perform faster and have secondary function

larger bulk means they cant fit in small containers, they must be worn on a belt. It's a small thing, but it makes it so someone cant carry both medical and engi upgraded tools at once.

But also; the other thing about making it not-printable is scarcity. Advanced hyposprays shouldn't be allowed to be ubiquitous, because that becomes the comparison standard. See; Skyrat surgery speed bonus (and removal of).

projectkepler-ru:
> larger bulk means they cant fit in small containers, they must be worn on a belt. It's a small thing, but it makes it so someone cant effectively carry both medical and engi upgraded tools at once.
> 
> But also; the other thing about making it not-printable is scarcity. Advanced hyposprays shouldn't be allowed to be ubiquitous, because that becomes the comparison standard. See; Skyrat surgery speed bonus (and removal of). There are tools which powercreep doesn't really matter such as with the experimental scanner and the bluespace version; and medical/engi tools to some extent aren't all that required to carry all 16 from both depts on you at all times. But a hypospray? Readily accessible for distribution to everyone, so that everyone has ease-of-access to high-tier chems?


A. ok I'd agree, if not for the fact you actually want to get surgical toolset implant 9/10 over just med tools as soon as you can because it work faster. 
B. actually a valid take for once,  sure we don't want everyone to have ease of access to things, I dont disagree, I simply don't see how this is thing in particular inherently a bad thing because it's still a hypospray and that means you have to; break into med, steal it, play as med.  And I have never in my life seen anyone except the guy who added it ask med for it

Which reminds me of how bloodsucker player want all of sec ballisticoption gone just so they can teleport into armoury, take the coat and then never worry about laser
Am I wrong in this comparison? probably yes, but I'm pointing out the author knows what they are doing

lessthnthree (changes requested):
After some discussions, the large hypospray case is fine printable. Remove "hypomkii_adv" from the techweb nodes, making the advanced hypospray a cargo import only.

In modular_skyrat/modules/company_imports/code/armament_datums/deforest_medical.dm around line 229 add:
```
/datum/armament_entry/company_import/deforest/equipment/advanced_hypospray
	item_type = /obj/item/hypospray/mkii/piercing
	cost = PAYCHECK_COMMAND * 7
```

lessthnthree (approved):

PR #3083, opened 2025-02-13
Test Disable Lua for Crash Reasons? [DNM]

nevimer:
DO NOT MERGE THIS

Trying to find out what's causing crashes, one thing at a time. This guess goes to auxlua. Will need to remove the .so files aswell.


Majkl-J:
Well we know this isnt it now.

PR #3113, opened 2025-02-16
Actually buffs funny cyborg health number

StrangeWeirdKitten:
I guess I didn't realize ``/mob/living/silicon`` has a child of ``/mob/living/silicon/robot`` which has its health set to 100.

image


NargaInDisguise:
Huh, I guess this explains why I was so confused when testing damage slowdowns.

PR #3123, opened 2025-02-18
Adjusts some story teller thresholds

Odairu:
## About The Pull Request
Affects Gamer, Fragile, and Default Andy
## Why It's Good For The Game
Gamer:
I feel with the way that gamer is portrayed as the high intensity mechanics antag v crew sort of teller, Major track should run more than just 1 time, this change has it so it will run 2 times in a normal round (one that doesn't get continue voted multiple times)

Default Andy:
As it stands Major and Ghost set still only run 1 time per normal round, major will run twice if the round gets extended, this change would push the major occurance from around 2hrs 20 minutes (right before the shuttle call) to a little bit below two hours so it doesn't happen right as everyone's wrapping up whatever they're doing and getting ready to end the round.

Fragile:
I believe that much like Chill, Major shouldn't really run on this teller, bumped up the point threshold for it so it only runs if the round gets extended.
## Proof Of Testing
Just number changes, it works :^)
<details>

</details>

## Changelog
:cl:
balance: changed point values on tellers so they run a bit earlier/later depending on teller
/:cl:


ReturnToZender:
Actually seems surprisingly reasonable!

NargaInDisguise:
Yeah honestly all of these make complete sense

Odairu:
the TL;DR is
Gamer is a bit more chaotic
Andy is the same chaos, things just happen earlier, but the amount of things is unchanged
Fragile is a bit less chaotic

BurgerLUA:
Not sure if the gamer should be increased when it comes to event intensity. If the gamer is too peaceful, the amount of antags should increase instead.

Odairu:
> Not sure if the gamer should be increased when it comes to event intensity. If the gamer is too peaceful, the amount of antags should increase instead.

its moreso so more unique things happen, rolling major twice just means 1 more event over the course of a whole round, and major only has 4(?) antags as part of its event pool, with things like rad-storm, immovable rod, etc being more common
I think its just borers blob abductors and revenant that will naturally spawn in major.

Cephalopod222:
Good storyteller tweaks, Andy spawning shit right at the end of the round was always *screams-incoherently*

lessthnthree:
I think instead of eliminating major from fragile instead do an adjustment to the tracks.

Blob, pirates, perhaps others should be moved from Major to Crewset. If you look at the rest of the events in the major track (rod, rad storm, wormholes, anomaly, cme) I don't think they're intense enough to eliminate entirely when fragile sends what, one per round?

NargaInDisguise:
> ~I think instead of eliminating major from fragile instead do an adjustment to the tracks.~
> 
> ~Blob, pirates, perhaps others should be moved from Major to Crewset. If you look at the rest of the events in the major track (rod, rad storm, wormholes, anomaly, cme) I don't think they're intense enough to eliminate entirely when fragile sends what, one per round?~
> 
> I thought about it, my suggestion can be its own PR.

Absolutely do not put these down to crewset. Crewset runs CONSIDERABLY more often on almost every storyteller, with things like traitors, spies, heretics, lings and bloodsuckers. Do not put blob or pirates or anything else on there, they would become considerably more common. Storyteller doesn't need to be made more hostile towards players.

lessthnthree:
> > ~I think instead of eliminating major from fragile instead do an adjustment to the tracks.~
> > ~Blob, pirates, perhaps others should be moved from Major to Crewset. If you look at the rest of the events in the major track (rod, rad storm, wormholes, anomaly, cme) I don't think they're intense enough to eliminate entirely when fragile sends what, one per round?~
> > I thought about it, my suggestion can be its own PR.
> 
> Absolutely do not put these down to crewset. Crewset runs CONSIDERABLY more often on almost every storyteller, with things like traitors, spies, heretics, lings and bloodsuckers. Do not put blob or pirates or anything else on there, they would become considerably more common. Storyteller doesn't need to be made more hostile towards players.

Yeh, I meant ghostset, the one with a timer default of 7000.

It was more a general "events that have actual players behind them should be separated from major PvE events" where right now Blob and Pirates are alongside those other ones.

Arturlang:
Does fragile really need nerfing? Not really seen it run much of anything at all

coffeecoffeecoffee:
I have basically never seen Fragile run anything of any major consequence in my entire time playing already, yeah. Everything else here sounds great to me personally!

PR #3151, opened 2025-02-24
Micro Squishing & Quirk (for abusing small sprites and being SMALLER)

nevimer:
## About The Pull Request

Adds some balance around small sprites. You get trampled easily if you're smaller than normal, by pixel scale and not mutations. 
## Why It's Good For The Game

Micro Balance, arguable enterprise resource content.

## Proof Of Testing
<details>

</details>

## Changelog
:cl:
add: Three new micro-sized traits. They make you smaller at a cost.
balance: Micro characters get squished easily. They also have to duck under people. 
/:cl:


modular_skyrat/modules/customization/datums/dna.dm
```diff
@@ -175,6 +175,15 @@ GLOBAL_LIST_EMPTY(total_uf_len_by_block)
 	holder.transform = holder.transform.Translate(translate_x, translate_y)
 	holder.maptext_height = 32 * features["body_size"] // Adjust runechat height
 	current_body_size = features["body_size"]
+	if(current_body_size < 1) // BUBBER EDIT BEGIN
```
nikothedude, line 178: i really hope youre not saying "ANYONE THAT IS BELOW 100% SCALING AT ALL GETS OWNED"

because thats hilariously stupid

nevimer: you're right, it should be 80% which is the min

KateAnbolho: It loos like that's exactly what its, what lmao

nikothedude:
also i again desparately ask stop speedmerging your prs let people give fuckin feedback before you use them as guniea pigs

nevimer:
it turns out the component that TG has is buggy as heck and there isn't a simple way of fixing it- it catches like everything in its wake.

SovietSpaghetti:
I'm all for nerfing small sprites, but I don't quite think this is the way to do it. For one, this, as far as I can tell, only takes into account character creation sprite size. It does not take into account character height, species (teshari/dwarf), genetics (gigantism/dwarfism/acromegaly), and also doesn't work on borgs (cat/roomba model + shrink module being the biggest offender). If anything the most this seems to do is cause incidents that you're going to accidentally damage someone because they're shorter than you, especially in situations where a bunch of people are crawling over each other (like with CRAB-17).

If anything, a debuff to small sprites should take into account changes to sprite size from stretching effects like dwarfism, dwarfs, and character height, and only account for melee damage taken, as that's the only area they have an actual advantage. A scaling increase to taken melee damage depending on how much your sprite is shrunk would probably be the least annoying method of doing this. Even then, it's kinda unfun to apply a mechanical disadvantage to something that is ultimately (mostly) just used as a flavor decision.

Iamgoofball:
We shouldn't be letting people make their sprites this small to begin with. I'm sorry, but the people addicted to enterprise resource planning to 2d spaceman may need to just use the theater of the mind for this one, it is not worth the constant compromises we have to make with the game design to satisfy an incredibly niche fetish.

ArrisFairburne:
> it is not worth the constant compromises we have to make with the game design to satisfy an incredibly niche fetish.

Not everyone who plays as/with tiny characters has a fetish for that, you know.

xiangfei-chen:
So, upon testing this, I've discovered the following. Anyone under .9 sprite scale moving into the tile of a person of 1.0 scaling or higher has to roll to avoid being squashed. They also get tripped when they drop anything. Micro characters can also trample each other. They can also be stunlocked by being walked over repeatedly. 

Suffice to say, I think this needs a closer look.

Iamgoofball:
Fair point. Still, though. Those players often aren't setting themselves so small they're nigh unclickable.

nevimer:
> So, upon testing this, I've discovered the following. Anyone under .9 sprite scale moving into the tile of a person of 1.0 scaling or higher has to roll to avoid being squashed. They also get tripped when they drop anything. Micro characters can also trample each other. They can also be stunlocked by being walked over repeatedly. 
> 
> Suffice to say, I think this needs a closer look.

not applicable by the time you made this comment

nevimer:
This has been changed to a QUIRK. You can still be small without taking it. Taking it will opt you enable you to be smaller than normal limits, and you will receive the penalty of being crushed by others when bumped.

HometownFunky:
Would it be possible to make this give something equally mechanically punishing (like an unavoidable hardstun of varying lengths or smt) instead of the bone crushing- without having to take pacifist. Or maybe make the bone crushing opt in for both parties? My concern is that walking around and breaking tiny folks' bones is not everyone's cup of tea, especially for a quirk only one side has to take.

Crystal9156:
https://i.gyazo.com/23fd9355dcb7bca05770a19fc058fcf5.png
https://i.gyazo.com/2e540f147531792836cda4f6c9985cbb.png
https://i.gyazo.com/d7480e4986c46e11ca004522c7b99545.png
Behold, three fracs and softcrit in under five minutes of shift start.

CEO-Of-Dogborg-Corp:
This is just too silly, either way it is fetish content that can be used for combat. This seems out of place on a server with antags.

Bombermansam:
I hated the original pr but I like the quirk, its funny and it can allow for some RP situations. The are some problems with the current implementation of the quirk.

1. Simple mobs can and will crush you. Even a cleanbot or a medibot is enough to knock you down and crush you. While funny, I can see people who want to be small not taking this quirk for this very reason.
2. People with the quirk can run into each other, knock themselves down and crush themselves. If this quirk is meant to portray a larger creature crushing a smaller one then why can micro people crush each other?
3. The crushing is excessive. For people that want to use this quirk to be small, they will be constantly injured or dead. Imagine how many times people bump into you normally in a round. Now imagine everytime they do it takes 20% off your health with a chance of wounding. If this quirk is trying to solve the problem of small sprites having an advantage in combat then why dont you have the damaging portion of the quirk check for an active combat mode on the person doing the squishing? You can still have the micro person fall to the ground if bumped into when the person bumping into them is not in combat mode (think of them jumping out of the way), but the micro person should not be crushed(damaged) unless the person bumping them has combat mode on. This would allow you to keep the balance purpose of this quirk while also giving the people that are not using this quirk to fight some quality of life.

Tl;Dr: If you add a check so that people with this quirk can be knocked down but not crushed unless the other person has combat mode on, it will improve the QOL of this quirk.

AXLFAN2000:
I've played with one of the quirks a bit, I think it's absolutely hilarious and has a lot of funny character potential, I love it.

Things I noted as feedback:
-  pixelshifting to the edge of a tile should prevent knockdowns from happening IMO.  Since it lets people share your tile normally, it's not very intuitive that small people get sucked to the middle of the tile and squashed anyways
-  they should be more effected by chems IMO, like have to drink less to get really drunk.  Tiny body weight!! 
-  other characters with the quirk shouldn't crush them (or at least not do the damage)
-  all four times I was squished (with no other quirks that would add to the chance), I got a fracture along with the sizeable damage  --  I personally think it would be better if that wasn't 100% guaranteed (or if it isn't a guarantee already, then just being less frequent would be good I think since the damage is already a huge hit)

StrangeWeirdKitten (changes requested):
This should honestly be combined into one quirk with a tgui thing that lets you choose your size shrinkage and the effects you get. There's no need to have three seperate quirks for what's basically the same thing.

code/controllers/subsystem/processing/quirks.dm
```diff
@@ -45,6 +45,7 @@ GLOBAL_LIST_INIT_TYPED(quirk_blacklist, /list/datum/quirk, list(
 	list(/datum/quirk/overweight, /datum/quirk/obese),
 	list(/datum/quirk/dominant_aura, /datum/quirk/well_trained),
 	list(/datum/quirk/equipping/entombed, /datum/quirk/badback),
+	list(/datum/quirk/micro, /datum/quirk/micro/smaller, /datum/quirk/micro/smallest),
```
StrangeWeirdKitten, line 48: should also conflict with oversized

StrangeWeirdKitten (changes requested):

CatoChristopherMrow:
I love this quirk, as it's finally a middle ground between people who wanna play small characters and those who have to fight small characters. To the people screaming it's weird and it's a fetish, really? The only way you could POSSIBLY conceive of someone wanting to play a character that's not the exact average height is that it must be their fetish? Are tesh all just micro fetishists?

ReturnToZender:
"I'm into that actually" applies here, it's totally a fetish and there's nothing wrong with that, if you look at a small creature and go "that's fetish content" in anything other than a positive way I have REALLY bad news to you with regards to the rest of the server

CEO-Of-Dogborg-Corp:
"Joking" about intentionally making people uncomfortable, awesome.

plsleavemealon:
Something that would definitely benefit this quirk, would be making it so when you pixel shift up against a wall, that you wont get crushed by someone walking past.

Currently, when someone pixel shifts up against the wall, and they are normal sized, another person will just walk past without issue (they wont be bumped/moved).

However, someone with one of these quirks will be crushed while trying to do the same.

CatoChristopherMrow:
Fullmerge this already zamn

TealSeer (approved):

PR #3152, opened 2025-02-24
Revert "Lewd Interactions are subtle, not subtler (#2889)"

nikothedude:
## About The Pull Request
This reverts commit f4f4621826338e75f65f3920d59ac6ec1dffa31c, PR https://github.com/Bubberstation/Bubberstation/pull/2889.
## Why It's Good For The Game

Every time I ghost, I am forced to bare witness to non-stop sex. While yes, subtler isn't a requirement for ERP, we should at least encourage using subtler, as oftentimes people don't want to see exactly in detail how you are getting railed.
## Proof Of Testing

It's a revert. I dont think I need to test this.

</details>

## Changelog
:cl:
add:Lewd verbs are now subtler. No longer will you have to see people getting absolutely railed as a ghost.
/:cl:

### DIFF
--- a/modular_skyrat/modules/interaction_menu/code/interaction_datum.dm
+++ b/modular_skyrat/modules/interaction_menu/code/interaction_datum.dm
@@ -96,7 +96,7 @@ GLOBAL_LIST_EMPTY_TYPED(interaction_instances, /datum/interaction)
 	// We replace %USER% with nothing because manual_emote already prepends it.
 	msg = trim(replacetext(replacetext(msg, "%TARGET%", "[target]"), "%USER%", ""), INTERACTION_MAX_CHAR)
 	if(lewd)
-		user.emote("subtle", null, msg, TRUE)
+		user.emote("subtler", null, msg, TRUE)
 	else
 		user.manual_emote(msg)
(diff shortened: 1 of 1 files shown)


lessthnthree:
@nikothedude I would like you to actually provide proof of testing.

It was changed when subtler was given range to an entire room, and entire rooms were thus hearing interactions. You may need to adjust something else, such as the default subtler range.

sippykot:
Disable 👏 global 👏 emotes 👏 - no one is forcing you to watch people fucking as a ghost, its your choice.

StrangeWeirdKitten:
If you're going to revert my PR, you can at least just tell me what the issue is first. If it's something to do with global emotes (something we've set to be default disabled for over a year and a half now so it's not really been a reported issue)

I don't see subtle at all when I'm ghosting personally, so I'm going to assume that's the case. I'm not saying you have to bug fix my code, but I have a feeling it just needs an extra observer check in the emote chain to fix your issue. But I'd appreciate just telling me.

nikothedude:
if global emotes being disabled is asctually the default, then i guess theres no issue here, since i somehow opted into this pain

PR #3156, opened 2025-02-25
Jerk Quirk

nevimer:
## About The Pull Request

Adds the Jerk Quirk, a quirk that makes you rudely shove anyone in your way. You will at random shove someone, even unrelated people, when someone is in your way of movement. Jerk. You can't shove people without a free hand, however, and not rapidly either. 

## Why It's Good For The Game

A bit of flavour, maybe a catch 22 for taking it. 

## Proof Of Testing

<details>

</details>

## Changelog

:cl:
add: Added the Jerk Quirk
/:cl:

### DIFF
--- a/modular_zubbers/code/datums/quirks/positive_quirks/defensive.dm
+++ b/modular_zubbers/code/datums/quirks/positive_quirks/defensive.dm
@@ -1,0 +1,37 @@
+/datum/quirk/defensive
+	name = "Jerk"
+	desc = "You are rude and impatient, and want people to get out of your way. "
+	icon = FA_ICON_FROWN_OPEN
+	medical_record_text = "Patient has anger issues."
+	value = 6
+	gain_text = span_notice("You feel like punching someone.")
+	lose_text = span_notice("You feel calmer.")
+	quirk_flags = QUIRK_HIDE_FROM_SCAN
+	COOLDOWN_DECLARE(shovingspam)
+
+/datum/quirk/defensive/add(client/client_source)
+	. = ..()
+	RegisterSignal(quirk_holder, COMSIG_LIVING_MOB_BUMP, PROC_REF(on_bump))
+
+/datum/quirk/defensive/proc/on_bump(mob/living/bumped)
+	if(!COOLDOWN_FINISHED(src, shovingspam))
+		return
+	if(quirk_holder.get_active_held_item())
+		return
+	var/list/problems = oview(1, quirk_holder)
+	var/list/actual_problems = list()
+	for(var/mob/living/problem in problems)
+		if(problem == quirk_holder)
+			continue
+		if(istype(problem, /mob/living))
+			actual_problems |= problem
+	var/mob/living/intheway = pick(actual_problems)
+	quirk_holder.disarm(intheway) // yes we have to hunt and pec
+	COOLDOWN_START(src, shovingspam, 1 SECONDS)
+
+/datum/quirk/defensive/remove()
+	. = ..()
+	UnregisterSignal(quirk_holder, COMSIG_LIVING_MOB_BUMP)
+
+
+
--- a/tgstation.dme
+++ b/tgstation.dme
@@ -8740,6 +8740,7 @@
 #include "modular_zubbers\code\datums\quirks\neutral_quirks\oversized_quirk.dm"
 #include "modular_zubbers\code\datums\quirks\neutral_quirks\pet_owner.dm"
 #include "modular_zubbers\code\datums\quirks\neutral_quirks\waddle.dm"
+#include "modular_zubbers\code\datums\quirks\positive_quirks\defensive.dm"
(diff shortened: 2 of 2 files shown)


modular_zubbers/code/datums/quirks/positive_quirks/defensive.dm
```diff
@@ -0,0 +1,37 @@
...
+	var/mob/living/intheway = pick(actual_problems)
+	quirk_holder.disarm(intheway) // yes we have to hunt and pec
+	COOLDOWN_START(src, shovingspam, 1 SECONDS)
+
+/datum/quirk/defensive/remove()
+	. = ..()
+	UnregisterSignal(quirk_holder, COMSIG_LIVING_MOB_BUMP)
+
+
+
```
Arturlang, line 37: Excess spaces at the end of your file

modular_zubbers/code/datums/quirks/positive_quirks/defensive.dm
```diff
@@ -0,0 +1,37 @@
+/datum/quirk/defensive
+	name = "Jerk"
+	desc = "You are rude and impatient, and want people to get out of your way. "
+	icon = FA_ICON_FROWN_OPEN
+	medical_record_text = "Patient has anger issues."
+	value = 6
```
Arturlang, line 6: kinda a huge value for something not that incredibly negative

modular_zubbers/code/datums/quirks/positive_quirks/defensive.dm
```diff
@@ -0,0 +1,37 @@
...
+	lose_text = span_notice("You feel calmer.")
+	quirk_flags = QUIRK_HIDE_FROM_SCAN
+	COOLDOWN_DECLARE(shovingspam)
+
+/datum/quirk/defensive/add(client/client_source)
+	. = ..()
+	RegisterSignal(quirk_holder, COMSIG_LIVING_MOB_BUMP, PROC_REF(on_bump))
+
+/datum/quirk/defensive/proc/on_bump(mob/living/bumped)
+	if(!COOLDOWN_FINISHED(src, shovingspam))
```
Arturlang, line 17: This should use the same cooldown that manual shoving uses or this could be abused

Arturlang (changes requested):

ReturnToZender:
An autoattack like this is a pretty notable positive, but I'm not sure exactly how strong it would be. My suggestion would be that you need to have an open hand _selected_ rather than available. This would be interesting to be able to roleplay like, a bull's charge.
This feels like it has the potential to be a positive trait that actually changes something about your character, which is something that positive traits currently lack, so I'm excited nonetheless.

- Curious how this combos, if at all, with sleeping carp. Carp is already so strong that basically every traitor takes it, and theoretically you could combo a crashing wave kick with an unarmed attack.
- Make sure to check other antag abilities for niche interactions

ReturnToZender:
image
My new OC just dropped

modular_zubbers/code/datums/quirks/positive_quirks/defensive.dm
StrangeWeirdKitten: I was watching him code it in a stream. This is a justified value imo

modular_zubbers/code/datums/quirks/positive_quirks/defensive.dm
StrangeWeirdKitten: Cooldown is more for the fact it auto shoves and runs the proc fast. It already has the shoving cooldown but it shouldn't be as fast as shoving

zeroisthebiggay:
this seems like a pretty big net positive since you can shovefight by playing like TGMC xenos

Razurath:
Would this trigger scarp combos?

latin313:
So far my experience with this has been.. honestly terrible. I feel like this should have some sorta downside like slowing down the user. have had a couple shifts now where a prisoner or someone else would just fly around on drugs/using wrestling to harass the crew and security couldnt do much. Something to mention is you auto-shove people even while laying down or in crit

StrangeWeirdKitten (changes requested):
Ensure you can't do it while != CONCIOUS and if you're laying down.

modular_zubbers/code/datums/quirks/positive_quirks/defensive.dm
```diff
@@ -0,0 +1,43 @@
...
```
StrangeWeirdKitten, line 17:
```suggestion
/datum/quirk/defensive/proc/on_bump(mob/living/bumped)
        SIGNAL_HANDLER
        
	if(!COOLDOWN_FINISHED(src, shovingspam))
```

RyszardParys:
The humble combat intent in my pocket:

3likeee

CatoChristopherMrow:
I'm unsure since I havn't checked, but this works on the same cooldown as normal shoving right? You can't doubleshove someone by running into them with this quirk then intentionally shoving them with rightclick to insta fullstun someone, right?

PR #3166, opened 2025-02-27
Removes illegal tech flag from agent ID and military belt

YakumoChen:
## About The Pull Request
As title
## Why It's Good For The Game
It's too easy to find on station for easy illegal tech node
More items may be added later if they're easily found (such as in smuggler satchels)

Illegal tech should only be reasonably procured from actual antagonists, the node should be rare.
## Proof Of Testing
<details>

</details>

## Changelog
:cl:
balance: Agent ID no longer gives illegal tech
balance: Syndicate military belt no longer gives illegal tech
/:cl:


StrangeWeirdKitten:
I need a better reason why this is good for the game then "it should be more rare"

IgiariValkyr:
> I need a better reason why this is good for the game then "it should be more rare"

Illegal tech gets unlocked 9/10 rounds it seems like. It may as well be actually in the tech tree at that point without anything unlocking it. So I'm guessing this is to stall out the reliance on illegal tech by making fewer things actually grant it.

YakumoChen:
> I need a better reason why this is good for the game then "it should be more rare"

as a hidden tech node, it should not have simple and easy in-round requirements that can consistently be found in smuggler satchels that randomly generate on the station. Hidden tech nodes are intended to be only unlocked uncommonly on rounds, preferably from an item generated by an actual antagonist.

The same can be said for alien tools, but we have ruin generation turned up so high that that specific ruin generates far more commonly than it probably would on TG.

YakumoChen:
> I still need some reasons why then just "it should be rarer" to be quite fair. Like illegal tech really doesn't do much except give the occasional borg a stun arm.

Hard disagree, I provided a few examples of items that are problems to be consistently available every round. Advanced camera makes anyone in security effectively an AI, MCR unlocks include the "better" shotgun blast that does a pretty extreme amount of damage, and a laser beam that always sets the target on fire

Borgs generally should not be getting illegal tech or chameleon modules either every round, and they've been reported to cheese illegal tech for these modules.

NargaInDisguise:
> > I still need some reasons why then just "it should be rarer" to be quite fair. Like illegal tech really doesn't do much except give the occasional borg a stun arm.
> 
> Hard disagree, I provided a few examples of items that are problems to be consistently available every round. Advanced camera makes anyone in security effectively an AI, MCR unlocks include the "better" shotgun blast that does a pretty extreme amount of damage, and a laser beam that always sets the target on fire
> 
> Borgs generally should not be getting illegal tech or chameleon modules either every round, and they've been reported to cheese illegal tech for these modules.

Again I think we should literally just make it a policy issue for borgs to be getting illegal tech for no reason.

PR #3179, opened 2025-03-02
Gives ghosts access to LOOC

DaylightE:
## About The Pull Request

Ghosts can now send and receive LOOC messages.
Ghosts show their ckey instead of character name in LOOC.

## Why It's Good For The Game

The pros:

- Scenes where one person ends up dying can now have people talk about their experience and share feedback afterwards.
- Scenes where one person ends up dying won't have the living person confused about why their partner isn't answering their LOOC anymore.
- Dead people can have 1:1 chats about how the round went with the living after the shuttle has docked with CC.

The cons:

- You could use this to metacommunicate, but that hasn't stopped the living from getting LOOC, OOC, AOOC, SOOC etc. And ghosts can already use regular OOC or discord messages if they want to be cheeky.
- You could hear the LOOC from lewds you are observing, but LOOC messages don't usually contain full context for what's going on and shouldn't offend anyone too badly. 

## Proof Of Testing

<details>

image

</details>

## Changelog

:cl:
add: Ghosts can now send and receive LOOC messages.
/:cl:

### DIFF
--- a/modular_skyrat/modules/verbs/code/looc.dm
+++ b/modular_skyrat/modules/verbs/code/looc.dm
@@ -40,12 +40,6 @@
 		if(is_banned_from(ckey, BAN_LOOC))
 			to_chat(src, span_warning("You are LOOC banned!"))
 			return
-		if(mob.stat == DEAD)
-			to_chat(src, span_danger("You cannot use LOOC while dead."))
-			return
-		if(istype(mob, /mob/dead))
-			to_chat(src, span_danger("You cannot use LOOC while ghosting."))
-			return
 
 	msg = emoji_parse(msg)
 
@@ -78,10 +72,6 @@
 		if (is_holder)
 			admin_seen[hearing_client] = TRUE
 			// dont continue here, still need to show runechat
-
-
-		if (isobserver(hearing) && !is_holder)
-			continue //ghosts dont hear looc, apparantly
 
 		// do the runetext here so admins can still get the runetext
 		if(mob.runechat_prefs_check(hearing) && hearing.client?.prefs.read_preference(/datum/preference/toggle/enable_looc_runechat))
--- a/modular_skyrat/modules/verbs/code/looc.dm
+++ b/modular_skyrat/modules/verbs/code/looc.dm
@@ -79,10 +79,14 @@
 			// I wish it didn't include the asterisk but it's modular this way.
 			hearing.create_chat_message(mob, raw_message = "(LOOC: [msg])", runechat_flags = EMOTE_MESSAGE)
 
+		var/sender_name = src.mob.name
+		if (istype(mob, /mob/dead))
+			sender_name = key
+
 		if (is_holder)
 			continue //admins are handled afterwards
 
-		to_chat(hearing_client, span_looc(span_prefix("LOOC[wall_pierce ? " (WALL PIERCE)" : ""]:</span> <EM>[src.mob.name]:</EM> <span class='message'>[msg]")))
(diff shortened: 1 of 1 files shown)


Iamgoofball:
Please obscure the name of the ghost if they're a ghost; otherwise this reveals information about who's dead.

ReturnToZender:
I feel as if there are better solutions to the issues mentioned here, and more cons than are listed.

DaylightE:
> I feel as if there are better solutions to the issues mentioned here, and more cons than are listed.

You're free to list any ideas or concerns, I just found it odd coming from a more antag-heavy server where LOOC was the only OOC channel available to everyone.

ReturnToZender:
> You're free to list any ideas or concerns, I just found it odd coming from a more antag-heavy server where LOOC was the only OOC channel available to everyone.

There are many cases where LOOC can be abused, but not be deliberately intended to be. However, I don't think that this idea should be thrown out outright, as there's some things I'd really like to see.

> Being hounded with advice in a crisis situation by ghosts. Everyone wants to help, but having that many voices in your head can hurt.

I think making a system by which Mentors can LOOC while ghosted would be really sick to alleviate this, and give mentors more tools to help other players.

> After round-end, ghosts can use LOOC.

I think this would be an acceptable addition, because why not?

The listed positives relate mainly to ERP, I think? Unless it's code for being permitted to salt after being killed, but I think a system where you can still LOOC for a set time after death may not be entirely terrible.
The main reason most servers don't permit LOOC for an observer/ghost is because people would be annoying with it. I do think there's a lot of room for improvement with regards to communication in general, though.

Swiftfeather:
This doesn't seem like the right call from an administrative perspective.

DaylightE:
Ghosts now use their ckey as a name instead of character name.

Vanilla1040:
ANY ghost having access to sending and getting LOOC is a bad idea and is restricted to admins only for a good reason, however i do agree of at least giving it to mentors maybe.

Hemlock32:
I dunno
I've staffed for a server that gave normal player ghosts access to LOOC.
I don't think we ever saw it being abused for anything other than being salty after dying

Anything this could be abused for, could be done through a dozen other means that are harder to track, or, more public

If anyone abuses this, I doubt they were exclusively acting in bad faith because of a mechanic like this, and, would've found other methods
In the rare cases it is abused instead of those alternate methods, its almost easier for staff to deal with I'd imagine.

Zenitheevee:
I mean you're kinda really missing the big con of post death salt LOOC  

While i can agree that, In more tame environments (Of both types of the nature), This is honestly quite a nice thing, Able to have light OoC Chatter while dead, GG's after a duel, ETC. But honestly it comes with the big issue of: Some people have no self control and rules only apply when the breaks are found out. Lack of reports can often cause the issues to be overlooked, plainly because some people do see a lack of action as a reason to not further report and further feed into the cycle.

And while I will immediately start by stating SS13 has a filter for certain groups here, quite frankly i dont think the lovebirds in the interlink need a ghost typing moodkilling shit in LOOC. Nor do i think we need death salt LOOC's being a problem

StrangeWeirdKitten:
hell no

StrangeWeirdKitten:
> ANY ghost having access to sending and getting LOOC is a bad idea and is restricted to admins only for a good reason, however i do agree of at least giving it to mentors maybe.

On the topic of this, mentors already have a thing that lets them look at people and obviously talk to them. Maybe, but that's the only merit I can see from this PR.

Hemlock32:
> I mean you're kinda really missing the big con of post death salt LOOC
> 

Not much different than someone sitting in their corpse salting in LOOC, aside from being able to move
Its also, really just, a giant "Ban me" sign the person is waving around

Like I said, I staffed for a server with this sorta thing
Its _really_ not that bad

PR #3183, opened 2025-03-04
Bio-RCD Concept for Medical as a Research Unlockable

nevimer:
## About The Pull Request

The goal of this device is to be a replacement or supplement for lategame surgery and the SAD. This device will do the following:
- Print limbs, appendages, features, and other externals of someone you sample.
- Perform a surgery (actually just installs), placing the newly created and selected limb onto the mob you target.
- Consume Synthflesh, or other reagents to create a desired appendage. 
- Use a TGUI to select what you want.
- Read the prefs of the sampled mob, to generate their "designs". 
- Generally be open to changes and revision as needed.

edit: discord folks say they'd rather it be a limb spitter than an autosurgeon, so we may veer to that
This PR is not even close to finished. Let me know your thoughts.
## Why It's Good For The Game

I want medical to combat the SAD as a catch all fix for problems, and generally make life easier for medical and their patients. This device will basically operate as a sci-fi body restructuring type device. 

## Proof Of Testing
<details>

</details>

## Changelog
:cl:
add: medical bodypart rcd
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/medical_rcd/medical_rcd.dm
+++ b/modular_zubbers/code/modules/medical_rcd/medical_rcd.dm
@@ -1,0 +1,106 @@
+
+
+/obj/item/construction/medical
+	name = "medical grade rapid-construction-device (RCD)"
+	icon_state = "plumberer2"
+	inhand_icon_state = "plumberer"
+	lefthand_file = 'icons/mob/inhands/equipment/tools_lefthand.dmi'
+	righthand_file = 'icons/mob/inhands/equipment/tools_righthand.dmi'
+	worn_icon_state = "plumbing"
+	icon = 'icons/obj/tools.dmi'
+	slot_flags = ITEM_SLOT_BELT
+	banned_upgrades = RCD_ALL_UPGRADES & ~RCD_UPGRADE_SILO_LINK
+	matter = 200
+	max_matter = 200
+	drop_sound = 'sound/items/handling/tools/rcd_drop.ogg'
+	pickup_sound = 'sound/items/handling/tools/rcd_pickup.ogg'
+	sound_vary = TRUE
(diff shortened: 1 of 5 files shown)


ReturnToZender:
Heard a request for feedback. Interesting idea, needs to pass tests. Not sure how this would slot into the existing medical set

shellspeed1:
I like the idea. I think it would be a good idea to tie in DNA disk compatibility and have the same also integrated into the limb grower.

YakumoChen:
I'd prefer this doesn't replace surgery as it ruins the doctor job if they have a magic replace tool (and potentially trivialises a lot of other treatments in the process

Like the RCD this should probably be much less efficient than the limb grower so that the latter is still the economic choice.

If the SAD is being used as a magic fix device (like treating wounds) that should definitely be addressed as unintended.

KateAnbolho:
> I'd prefer this doesn't replace surgery as it ruins the doctor job if they have a magic replace tool (and potentially trivialises a lot of other treatments in the process

Feedback was already given in the discord to axe the surgery part, surgery dosen't even take very long in 99% of circumstances when it's three steps, or just six if you're removing a limb for esoteric reasons

DorfGuy1:
> > I'd prefer this doesn't replace surgery as it ruins the doctor job if they have a magic replace tool (and potentially trivialises a lot of other treatments in the process
> 
> Feedback was already given in the discord to axe the surgery part, surgery dosen't even take very long in 99% of circumstances when it's three steps, or just six if you're removing a limb for esoteric reasons

IMO, it should work like the organizer MOD module, where you have the organ manipulate surgery 'open' on the patient, and then you activate it to shoot the organs into them, swapping them out. You could have it where you have a prosthetic replacement surgery up, activate it, to swap or replace missing limbs.  IDK if this is retreading old ground, but if we already have it one way, on the mod module, why not have it this way on the OCD (organic construction device)
We should totally call it the OCD by the way.

PR #3271, opened 2025-03-09
Fixes 516 character previews being small.

StrangeWeirdKitten:
This is really awful code and basically a splint, but the root cause is a byond bug. Hopefully just a temporary fix. Hopefully


nikothedude:
every temporary solution is a permanent one 🙏

StrangeWeirdKitten:
if not I guess I'll cherry pick it from TG because I opened the PR there too

IgiariValkyr:
Doesn't fix other previews like medical/security records being small.

PR #3295, opened 2025-03-12
being prone (army crawling, borers) now adds the TRAIT_IGNORE_ELEVATION trait

Ghommie:
## About The Pull Request
What it means is that there shouldn't be any more sprite stuttering* when crawling under racks or tables due to the elevation elements attached to them.

*Under normal cirumstances, it should be offset vertically but for some reason the offset is reset, perhaps from some of your modular changes, leaving only the animation in.

## Why It's Good For The Game
Fixing a nominal issue.

## Proof Of Testing
It works for xeno larvas upstream. It'll work here as well.

## Changelog

:cl:
fix: You no longer jitter when crawling under tables and the likes as a borer or from of army crawling.
/:cl:

### DIFF
--- a/modular_skyrat/master_files/code/datums/components/prone_mob.dm
+++ b/modular_skyrat/master_files/code/datums/components/prone_mob.dm
@@ -10,7 +10,7 @@
 
 	if(hand_blocking)
 		parent.add_traits(list(TRAIT_HANDS_BLOCKED), type)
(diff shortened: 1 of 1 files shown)


BurgerLUA:
Very nice, didn't think this was fixable without a massive overhaul

Majkl-J (approved):
I have no clue what modular changes there are but this is elegant and nice

PR #3298, opened 2025-03-13
Dullahan Sec Module

IgiariValkyr:
## About The Pull Request

Dullahans don't have a dedicated Security Module. Here's one made from a mix of the Meka Sec Module for the single-shoulder cape, and the existing Dullahan Peacekeeper Module.

## Why It's Good For The Game

Sec Modules have been seeing a LOT of use lately for some reason, despite the module itself supposed to have being been disabled. Dullahans were left out from Sec Modules in their initial PR, so this lets you be a sexy floating head and still beat people.

## Proof Of Testing

dreamseeker_2025-03-12_20-51-53
dreamseeker_2025-03-12_20-51-58
dreamseeker_2025-03-12_20-52-19

## Changelog

:cl:
add: Dullahans now have a Security Module.
/:cl:


IgiariValkyr:
I still say this should not be subject to freeze because it only touches _our_ files and not anything from the upstream.

Majkl-J:
> I still say this should not be subject to freeze because it only touches _our_ files and not anything from the upstream.

The freeze was a general one to any sort of features so we had time to focus on bugfixing.

PR #3302, opened 2025-03-13
Resprites icewalker wall torches and adds them to ashwalker ruin too

ExusA:
## About The Pull Request
This changes the flares on the walls of the ashwalker ruin to the icewalker wall torches, and resprites wall torches.
As a side-effect, you can't pick them up off the walls anymore. Because of that they're now easier to smash so you can still get rid of them if you want.
## Why It's Good For The Game
It looks better, I can only hope!
## Proof Of Testing
<details>
Lavaland:

wall torch

Icemoon:

walltorch_icemoon

Torch

</details>

## Changelog
:cl:
image: Resprited primitive wall torches.
del: Primitive wall torches are easier to smash.
add: The ashwalker base now has wall torches instead of flares.
/:cl:


YakumoChen:
Should give them a very minor fire particle effect like bonfires have

Also freeze

YakumoChen:
This also likely creates a sprite disparity with craft able, wielded torches we have in the game

ExusA:
Oh no, my mistake about the Freeze.
I am struggling to understand some of these points, if you could maybe clarify...

> This also likely creates a sprite disparity with craft able, wielded torches we have in the game
I'm not sure what you mean by this exactly. They will be different from the 'firebrand' you can craft for sure, but that's a different object, isn't it? Do you mean the flame effects should be similar for both items?
> They are not flares, they are child objects of flares.
Well, no, you're right. Is there something I should be doing differently because of that?
> Also 👎 for removing functionality, mechanics before sprites
Assuming you mean the functionality of taking them off the wall to hold, I could try add that back in some way if you think it's important, but I might struggle since I'm just a spriter.

ExusA:
Also since there's a freeze should I just close the PR anyway?

ExusA:
Oh right, I see what you mean now. What you're suggesting would be a good way to fix the issue, but I don't have the ability to do it properly yet and don't want to bumble around learning how since I opened this during a freeze anyway, so I'll close it at least for now.

PR #3339, opened 2025-03-18
You can now put a full size reference picture on the examine panel

StrangeWeirdKitten:
## About The Pull Request

I also centered the tab titles so it looks a little straighter

o4s526oYFJ

image

## Why It's Good For The Game

Cause I want to throw art on the examine panel. 

It'll be a little slow switching back due to ByondUI. Sadly

## Proof Of Testing

See above

## Changelog
:cl:
qol: Adds the ability to put a reference picture on the examine panel.
/:cl:

### DIFF
--- a/tgui/packages/tgui/interfaces/ExaminePanel.jsx
+++ b/tgui/packages/tgui/interfaces/ExaminePanel.jsx
@@ -76,20 +76,19 @@ export const ExaminePanel = () => {
         {page === 'main' ? (
           <>
             {art_ref && (
-              <Button
+              <Button>
                 color={data.art_ref_nsfw ? 'pink' : 'green'}
-                tooltip={
-                  data.art_ref_nsfw ? 'NSFW Reference Image' : 'Reference Image'
-                }
-                icon="image"
-                style={{
+                tooltip=
+                {data.art_ref_nsfw ? 'NSFW Reference Image' : 'Reference Image'}
+                icon="image" style=
+                {{
                   position: 'absolute',
                   bottom: '10px',
                   right: '10px',
                   zIndex: 1,
                 }}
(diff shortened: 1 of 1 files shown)


ArrisFairburne:
good preview gif, shark jumpscare

tgui/packages/tgui/interfaces/ExaminePanel.jsx
```diff
@@ -50,165 +58,253 @@ export const ExaminePanel = (props) => {
...
+                      <Stack.Item>{formatURLs(character_ad)}</Stack.Item>
+                    </Section>
+                  )}
+                </Box>
+              </Stack.Item>
+            </Stack>
+          </>
+        ) : (
+          <div>
+            <Button
```
Arturlang, line 282: 'content' is deprecated, put it as a children of <Button>
that is, <Button> childrenGoHere </Button>

tgui/packages/tgui/interfaces/ExaminePanel.jsx
```diff
@@ -50,165 +58,253 @@ export const ExaminePanel = (props) => {
...
-                      tabIndex === 2
-                        ? resolveAsset(headshot_nsfw)
-                        : resolveAsset(headshot)
-                    }
-                    height="250px"
-                    width="250px"
-                  />
-                </Section>
-              </>
+        {page === 'main' ? (
```
Arturlang, line 76: using ternary to choose the page isn't that great, maybe add functions within ExaminePanel that return the different pages? Then you could add error handling if it's not one of the valid choices as using strings is easy to mess up.

Arturlang (changes requested):

PR #3343, opened 2025-03-19
Removes the martyr and hijack objs, again. Also removes some heavily disruptive spy OBJs

Maia-J:
## About The Pull Request

See name, removes the martyr and hijack objs from traitor and spy

## Why it's good for the game

Giving these out for free on a random chance on the most common antag types is a bad idea. Especially when it gives the potential justification to bypass the station integrity rule. Also it's a mess to admin + we aren't trying to be Manuel, despite what it might seem like at times.

## Changelog
:cl:
del: Removes the martyr and hijack objs from traitor and spy
/:cl:

### DIFF
--- a/code/modules/antagonists/spy/spy.dm
+++ b/code/modules/antagonists/spy/spy.dm
@@ -160,6 +160,7 @@
 				cage_the_jailbird.no_failure = TRUE
 				objectives += cage_the_jailbird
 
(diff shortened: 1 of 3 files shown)


Majkl-J:
These were erronously added with an upstream, if you feel like we should have them, make a PR to readd them and argue for why we should have them. I shouldn't need to explain this

PR #3362, opened 2025-03-22
Exploitables no longer runtime on examine and don't expose antags.

Maia-J:
## About The Pull Request

Seems the logic here was reading variables from the person being examined.. not the person examining.

Yes this meant that souless mobs would runtime, and anyone could meta an antag because antags would show the option to view exploitables when others would not.
## Why It's Good For The Game

Holy shit how did I miss this in a review
## Proof Of Testing
I tested it, no runtimes related to examination
image

## Changelog
:cl:
fix: Exploitables no longer show up when you examine an antag, exposing them
fix: Exploitables now check the variables of the correct mob
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/examine.dm
+++ b/code/modules/mob/living/carbon/examine.dm
@@ -329,7 +329,7 @@
 		. += flavor_text_link
 
 	//BUBBER EDIT ADDITION BEGIN - EXAMINE RECORDS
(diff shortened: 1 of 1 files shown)


ReturnToZender:
We should totally make public runtimes a thing so we can start cleaning up a lot of the loose runtimes on the server rn

nikothedude:
how did i miss this

PR #3364, opened 2025-03-22
Twinkaine: The Scarfacening (probably TM it for balance)

sippykot:
## About The Pull Request

Say hello to my little friend: cocaine rework; new recipes, no toxin damage, movement speed buff is back (but weaker).

## Why It's Good For The Game

Cocaine as it stand is useless, except as a very potent poison - the nerf to it made it both poisonous and rather useless. - There's 0 reason to make it except to poison someone or immediately ship it off to cargo. This PR aims to bring cocaine back to its former glory (of actually being an usable drug) but now requiring a rather advanced chem (nitric acid) to make it, and making the whole production process more expensive and complex.

## Proof Of Testing

I complied and tested it locally, i swear :3

## Changelog

:cl:
add: new reagents - coca powder
add: new reagent - toxic coca paste intermediary
add: recipes for cocaine and crack changed a lot, now a multi-step process
add: cocaine is a painkiller now, as it is back on Earth
balance: cocaine is worthwhile to make again and not laced, but much harder to make
balance: you cant stack cocaine and crack in yourself
balance: crack is stronger than cocaine but has lower OD (15)
fix: cooling cocaine into powder no longer deletes half of it
/:cl:


sippykot:
ok im stupid, gimme a sec

sippykot:
I didnt test it after adding the last recipe and messed up a bunch of "/", ill fix it and compile it again.

sippykot:
ok here we go, sorry that was embarassing

DorfGuy1:
As a player who frequently makes and manufactures all of the drugs (In game) thank you. Not sure about the nitric acid tho. Maybe nock it down one level in complexity? Like Right now it'd require at least two reaction chambers for the required chemicals, then some other things. I think having acetone to make it would be complicated enough. If you leave it this complicated and have the move speed be less than before I think it won't be worth it.  Either make it less complicated and keep it at the level of speedbuff you have, or make it more speedy and keep the level of complexity. That's my two credits.

sippykot:
> As a player who frequently makes and manufactures all of the drugs (In game) thank you. Not sure about the nitric acid tho. Maybe nock it down one level in complexity? Like Right now it'd require at least two reaction chambers for the required chemicals, then some other things. I think having acetone to make it would be complicated enough. If you leave it this complicated and have the move speed be less than before I think it won't be worth it. Either make it less complicated and keep it at the level of speedbuff you have, or make it more speedy and keep the level of complexity. That's my two credits.

I guess i could make it a bit more speedy? But like, it really has no downsides. There's no organ damage from using this as long as you don't OD. Its really strong even with the nerfed speed buff. Its really just better than meth. Also, you don't need much nitric acid to make a lot of it; 10u of nitric acid will get you uhh... 75u of crack in the end?

sippykot:
Honestly i thought about it maybe giving no slowdown from damage for the scarface larp but idk

sippykot:
the only feedback i got was about it being not strong enough... so its buffed slightly - also crack effects now consistently stronger, not just run speed

sippykot:
could probably buff coca leaf and coca tea... come to think of it the status effect reductions on them are pretty pitiful compared to even normal tea, but idk

modular_skyrat/modules/morenarcotics/code/cocaine.dm
```diff
@@ -78,14 +101,130 @@
...
+	M.add_mood_event("zoinked", /datum/mood_event/stimulant_heavy, 1, name)
+	M.AdjustStun(-15 * REM * seconds_per_tick)
+	M.AdjustKnockdown(-15 * REM * seconds_per_tick)
+	M.AdjustUnconscious(-15 * REM * seconds_per_tick)
+	M.AdjustImmobilized(-15 * REM * seconds_per_tick)
+	M.AdjustParalyzed(-15 * REM * seconds_per_tick)
+	M.adjustStaminaLoss(-2 * REM * seconds_per_tick, 0)
+	if(SPT_PROB(20, seconds_per_tick))
+		M.emote(pick("scream","twitch","shiver"))
+	..()
```
Arturlang, line 165: Doesn't this inheret from /datum/reagent/drug/cocaine/on_mob_life, thus is a really effective antistun? -15+-18

Arturlang (changes requested):

modular_skyrat/modules/morenarcotics/code/cocaine.dm
sippykot: ...wait, does it stack? shouldnt it just overwrite?

sippykot: besides, i specifically made it so you cant stack these 3

sippykot:
Coca tea removed because i dont feel like its worthwhile
paste and crack aren't children of cocaine anymore
also OD on cocaine and such now should check if your heart is organic

modular_skyrat/modules/morenarcotics/code/cocaine.dm
Arturlang: as in, it inherents the functionality from cocaine's proc, as if you had both reagents in terms of the reagent process

Arturlang (changes requested):
Even if you prevent these from stacking with each-other, you can still stack it with other stimulants, especially when freebase cocaine gives you antistuns and speed without any downsides, needs some better balancing, honestly, remove the speeed buff and make freebase also have some sort of downside, potentially, make cocaine extremely easy to get addicted to? (easier than it is now)

PR #2989, opened 2025-01-22
nameless quirk

StrangeWeirdKitten:
## About The Pull Request

Adds the nameless quirk with the option to have a prefix. Without a prefix, you will spawn is as your [job] #[number]. If you don't have a job or prefix, you have some random prefixes I threw together that it chooses from.

## Why It's Good For The Game

Adds an option for characters to spawn in as a random number. Adds to the bubber roleplay experience. You can be mass produced or something.

## Proof Of Testing

image

## Changelog

:cl:
add: Nameless Quirk
/:cl:

PR #3392, opened 2025-03-28
Revert "nameless quirk"

StrangeWeirdKitten:
Reverts Bubberstation/Bubberstation#2989

Removing this because I don't see it used (outside of a very select few) and it needs to not be just a quirk.

🆑
del: removes nameless quirk
/:cl:


ReturnToZender:
I actually saw one guy who uses this because everyone hates him. That said, I do think this would be better as a toggle in the names menu

amsy2:
I mean, unless it's clogging code or smth, can we at least keep it in until replacement is made? I know its niche but is real nice when it works. Also for future iterations, would be nice if name could be random but random chosen by player.

PR #3403, opened 2025-03-31
Fix brain trauma surgery

lessthnthree:
## About The Pull Request

Fixes the highest level of trauma surgery leaving leftover traumas. Applies a status effect instead of the random roll of a new trauma.

## Why It's Good For The Game

Having to do lobotomy surgery over and over until you hit the 25% RNG to make it successfully go away isn't a good gameplay experience. Better to just apply a status effect for a while instead.

## Changelog

:cl: LT3
fix: Lobotomy/blessed lobotomy should correctly remove all surgery based traumas
balance: Lobotomy/blessed lobotomy now provides a predictable status effect instead of a trauma RNG roll
/:cl:


IgiariValkyr:
I believe the lobotomy is a chance and not a cure outright is because a healthy person can undergo it to roll for special traumas that can be extremely unbalanced otherwise. IE: Tenacity, which disables soft and hard-crit; or Bluespace Entanglement which lets you see carp portals on the regular; all the way to "soft antag" traumas like Obsession.

NargaInDisguise:
> I believe the lobotomy is a chance and not a cure outright is because a healthy person can undergo it to roll for special traumas that can be extremely unbalanced otherwise. IE: Tenacity, which disables soft and hard-crit; or Bluespace Entanglement which lets you see carp portals on the regular; all the way to "soft antag" traumas like Obsession.

personally I'd rather have "lobotomy gives you a bad debuff for 15 minutes but cures the absolutely dogshit afflictions" because we can trust the admins to moderate people out if they're fucking rolling lobotomies like a trauma-gacha on an RP server

lessthnthree:
> > > I believe the lobotomy is a chance and not a cure outright is because a healthy person can undergo it to roll for special traumas that can be extremely unbalanced otherwise. IE: Tenacity, which disables soft and hard-crit; or Bluespace Entanglement which lets you see carp portals on the regular; all the way to "soft antag" traumas like Obsession.
> > 
> > 
> > Those traumas were already not available from lobotomy. It is currently only the 'mild' (annoying) traumas that are given on the roll, so there really isn't a reason to do it over giving a status effect.
> 
> When were they removed? Because when I looked last night, lobotomy code still pulls from Special traumas, which is all of those and about 8 others.

Are you looking at modular_skyrat/modules/medical/code/trauma_surgery.dm

It's been like that for 8 months.

PR #3438, opened 2025-04-04
Nerfs super syndicat and makes it no longer a freedom implant.

pixelkitty286:
## About The Pull Request
- This PR makes the new syndicat ears item have a 1 minute cool down when used and the normal ears have the same thing.

- This also ports a fix that was done on monkey station: https://github.com/Monkestation/Monkestation2.0/pull/6017

## Why It's Good For The Game
- After watching a round on here and getting complaints about this from SPLURT I understand now why 20 seconds is too short for such a item like this if your gonna be a cat your gonna be a cat for a long while.

- Also to quote monkey station:
"It will make the cat not able to bypass any security non-lethal means, which in turns mean security won't be forced to lethal them everytime they see it."

## Proof Of Testing
Should work only changed the time values.

Edit: Compiled and tested locally

<details>

</details>

## Changelog
:cl:
balance: Super cat ear cool downs are longer and removed the freedom implant function.
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/spells/spell_types/shapeshift/kittyform.dm
+++ b/modular_zubbers/code/modules/spells/spell_types/shapeshift/kittyform.dm
@@ -20,7 +20,6 @@
 
 //Traitor's Super cat spell
 /datum/action/cooldown/spell/shapeshift/kitty/syndie
-	cooldown_time = 120 SECONDS
 	name = "SYNDICATE KITTY POWER!!"
 	desc = "Take on the shape of an kitty cat, clad in blood-red armor! Gain their powers at a loss of vitality."
 	possible_shapes = list(/mob/living/basic/pet/cat/syndicat/super)


YakumoChen:
Why shouldn't an item that costs 3/4ths of your TC not also function as a freedom implant?

NargaInDisguise:
> Why shouldn't an item that costs 3/4ths of your TC not also function as a freedom implant Tripling the cooldown is also a bit much
> 
> This goes against precedents set for other transform powers such as wizard, changeling, and miner transforms.

Because it makes them completely uncontainable with no counterplay other than to kill them and never revive (unless you can remove this item from them somehow, but I'm not aware of if you can)?
Wizard is a much weightier antag than Joe Traitor. Changelings are almost always RR'd immediately when they break cuffs, and the lesser dragon TF is both fairly rare to ever get (1/4 chance from a megafauna) and also not all that like, great at escaping?

This item used to let you completely bypass all stuns, slows and cuffs on a 20 second cooldown. It was absolutely absurd the level of protection it gave you. AND, for the record: The freedom implant had a limited number of uses as well, and still cost a ton of TC. 
SSC either needs to lose its ability to break cuffs, a MUCH MUCH longer cooldown, lose the ability to vent crawl to escape everything, etc. It was just an overbloated ability. The ability to ventcrawl is already worth a ton.

YakumoChen:
Freedom implant is 5 TC (1/3rd the cost) also makes you kill on sight, and has exactly enough uses to make security kill you on sight

The vent-crawling will make you also kill on sight anyway (security isnt likely to weld every vent), and the cat has low health, I don't agree with these reasons 

Sleeping carp and stands also costs you almost all your TC, are unremovable and make you a menace even when pacified, and we don't add arbitrary restrictions to those either.

NargaInDisguise:
> Freedom implant is 5 TC (1/3rd the cost) also makes you kill on sight, and has exactly enough uses to make security kill you on sight
> 
> The vent-crawling will make you also kill on sight anyway (security isnt likely to weld every vent), and the cat has low health, I don't agree with these reasons
> 
> Sleeping carp and stands also costs you almost all your TC, are unremovable and make you a menace even when pacified, and we don't add arbitrary restrictions to those either.

You literally dust if you die with a stand.
You can't punch people while in cuffs with scarp, deflect, or do anything while incapacitated. Scarp is it's own seperate piece of shit balance wise that I wish was nerfed, but thats out of the scope of this conversation.
Now you can't transform then ventcrawl away while in cuffs with SSC, which was one of its biggest pieces of contention.

There is a reason ventcrawling is so extremely limited to specific antags. It was combined with breaking out of cuffs tied into the same ability. It just did too much to justify it being on an antag that spawns up to 6+ times a round.

PR #1513, opened 2024-05-14
Tajaran Tackling

RadiantFlash7:
## About The Pull Request
Grants Tajarans a tackling bonus like moths.

## Why It's Good For The Game
Tajarans currently get unique eyes, which are incredibly weak to flashes and welding, in exchange for minorish night vision capabilities.
However, they were also given unique sensitive ears which will pretty much always be deafened from flashbangs. Theres no real upside to this- so I figure making Tajaran ears apply a bonus to tackling would be a good idea. 
Plus, cat pouncing memes is pretty great.

## Proof Of Testing
It Just Works ™️ 
But in seriousness, I tackled a few mobs on a test server. The change seems to be working as intended.

## Changelog

:cl:
balance: gave tajarans better tackles to compensate for weak ears
/:cl:

PR #3443, opened 2025-04-05
Removes tackle buff for having cat ears

Swiftfeather:
## About The Pull Request

Title, manual revert of #1513 

## Why It's Good For The Game

Tackle chance could be brought to 100% and people have reportedly been exploiting this fact.

## Proof Of Testing

I think it will complie 

## Changelog

:cl:Swiftfeather
balance: Cat ears no longer provide a tackle bonus
/:cl:

### DIFF
--- a/code/datums/components/tackle.dm
+++ b/code/datums/components/tackle.dm
@@ -445,11 +445,6 @@
 	var/obj/item/organ/wings/sacker_wing = sacker.get_organ_slot(ORGAN_SLOT_EXTERNAL_WINGS)
 	if(sacker_wing)
 		attack_mod += 2
-	// BUBBER EDIT START - Taj tackle bonus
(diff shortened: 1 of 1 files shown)


projectkepler-ru:
can we lose the flashbang sensitivity then

Swiftfeather:
> Alternatively the tackle can just be moved to an isspecies() check instead of cat ears. Which is a really weird way of handling it. If a istajaran() check doesn't exist, then one can easily be made where we have all those things at.

People will switch their species to tajaran to retain this bonus since species can be changed while hardly impacting character appearance. I see no path forward keeping any sort of tackle bonus since it offers a distinct combat advantage for those vile enough to gamer it.

Reducing a tajaran negative trait or finding another positive for them can be done independently of this pr.

Odairu:
Tbh I like organs/parts having bonuses that arent species specific, so I can steal those organs from other people and obtain them myself :^)

PR #3445, opened 2025-04-05
You can cuff people to the tram rails #90293

lessthnthree:
## About The Pull Request

https://github.com/tgstation/tgstation/pull/90293

image

Handcuffed people can be buckled to the tram rails.
If the tram hits them then they will die badly.
If they're not shockproof and are buckled to a powered rail, it will zap them repeatedly for a little bit of damage (but not a lot because you probably want them to die from something else).
Unbuckling someone from a powered rail is also pretty dangerous.

While doing this I fucked around a bit with how the "is the tram deadly" calculation worked because I didn't like how a deadly tram cuts your head off but does literally no other damage. Now it pulverises and beheads you.

## Why It's Good For The Game

image


DorfGuy1:
https://media.istockphoto.com/id/164303473/vector/a-vector-image-of-a-villain-that-is-wearing-a-cape.jpg?s=612x612&w=0&k=20&c=fTWt0aorkQuRn5osWfNXLN_4PUcQ1vLm_x3aGUQ41S0=

PR #3477, opened 2025-04-10
Increase wildcard slots

Cyprex:
## About The Pull Request
Increases wildcard slots:
Normal cards: 4->10 common slots
Silver cards: 6->8 common slots. These have 2 command and 1 private command slot, which can both carry common accesses

## Why It's Good For The Game

I believe the current amount of wildcards is still fairly restrictive- giving someone access to part of a department similiar to what a job would have and, say, janitorial is often not possible with only four slots. 

Having insufficient slots is very frustrating for both the HoP and the person asking for access. Having to issue a silver ID just to have enough slots is very awkward and some HoP players do not consider this option- Silver IDs can be seen as reserved for members of command.

The numbers for silver may look odd, these can now carry 11 common accesses(8+2+1), so they are still strictly superior to grey cards. I did not increase this further because some silver ID users(command staff) have a lot of trim slots already, and I do not want them to have AA on a silver card.

## Changelog
:cl:
balance: More wildcard slots on IDs
/:cl:

### DIFF
--- a/code/__DEFINES/id_cards.dm
+++ b/code/__DEFINES/id_cards.dm
@@ -11,13 +11,13 @@
  */
 
 /// Wildcard slot define for basic grey cards. Only hold 4 common wildcards.
-#define WILDCARD_LIMIT_GREY list(WILDCARD_NAME_COMMON = list(limit = 4, usage = list()))
+#define WILDCARD_LIMIT_GREY list(WILDCARD_NAME_COMMON = list(limit = 10, usage = list())) //BUBBER EDITS: More slots
 /// Wildcard slot define for Head of Staff silver cards. Can hold 6 common, 2 command and 1 private command.
 #define WILDCARD_LIMIT_SILVER list( \
-	WILDCARD_NAME_COMMON = list(limit = 6, usage = list()), \
+	WILDCARD_NAME_COMMON = list(limit = 8, usage = list()), \
 	WILDCARD_NAME_COMMAND = list(limit = 2, usage = list()), \
 	WILDCARD_NAME_PRV_COMMAND = list(limit = 1, usage = list()) \
-)
+) //BUBBER EDIT: More slots
 /// Wildcard slot define for Captain gold cards. Can hold infinite of any Captain level wildcard.
 #define WILDCARD_LIMIT_GOLD list(WILDCARD_NAME_CAPTAIN = list(limit = -1, usage = list()))
 /// Wildcard slot define for select Syndicate-affiliated cards. Can hold infinite of any Syndicate level wildcard. Syndicate includes all station accesses.


DorfGuy1:
can you remind me how these trims and access levels work now?

code/__DEFINES/id_cards.dm
```diff
@@ -11,13 +11,13 @@
  */
 
 /// Wildcard slot define for basic grey cards. Only hold 4 common wildcards.
-#define WILDCARD_LIMIT_GREY list(WILDCARD_NAME_COMMON = list(limit = 4, usage = list()))
+#define WILDCARD_LIMIT_GREY list(WILDCARD_NAME_COMMON = list(limit = 10, usage = list())) //BUBBER EDITS: More slots
```
projectkepler-ru, line 14: 10 is INSANELY high, that covers 3 important department

sippykot:
this whole thing honestly doesnt really have place on the server? like... what is the point of the slot restrictions? - if you want to give someone AA you can print out multiple cards
this feels like just something we have over from tg with no changes despite tg being much different, and trying to mechanically enforce something that should be a roleplay/rules concern

NargaInDisguise:
> this whole thing honestly doesnt really have place on the server? like... what is the point of the slot restrictions? - if you want to give someone AA you can print out multiple cards this feels like just something we have over from tg with no changes despite tg being much different, and trying to mechanically enforce something that should be a roleplay/rules concern

Because theres literally no reason anyone on the station NEEDS this many accesses besides THE captain? Giving people AA SHOULD be hard because you LITERALLY NEVER should do it. Paramedics already abuse their incredibly obnoxious list of accesses when they're antags and everyone constantly complains about it, we don't need *even more* people to have that, tbh.

sippykot:
> > this whole thing honestly doesnt really have place on the server? like... what is the point of the slot restrictions? - if you want to give someone AA you can print out multiple cards this feels like just something we have over from tg with no changes despite tg being much different, and trying to mechanically enforce something that should be a roleplay/rules concern
> 
> Because theres literally no reason anyone on the station NEEDS this many accesses besides THE captain? Giving people AA SHOULD be hard because you LITERALLY NEVER should do it. Paramedics already abuse their incredibly obnoxious list of accesses when they're antags and everyone constantly complains about it, we don't need _even more_ people to have that, tbh.

yeah but... why try to do this like this? we have rules and stuff

aKromatopzia:
> yeah but... why try to do this like this? we have rules and stuff

Why doesn't medical just **start** with advanced chems in the printer, why does chem even need to mix things together? **Because it's the narrative baseline. ** It is the ground upon which the world is built on, it is the incomplete answer so that you can be bothered to come up with a novel solution. Let's remove **every** speck of friction for "qol". You should in future avoid giving your input anywhere.

sippykot:
> > yeah but... why try to do this like this? we have rules and stuff
> 
> Why doesn't medical just **start** with advanced chems in the printer, why does chem even need to mix things together? **Because it's the narrative baseline. ** It is the ground upon which the world is built on, it is the incomplete answer so that you can be bothered to come up with a novel solution. Let's remove **every** speck of friction for "qol". You should in future avoid giving your input anywhere.

1) go drink some water
2) being able to administratively give someone more access as a captain/hop =/= mechanical buffs to medical (???)
all the tg stuff does is make the hop or captain have to give people a 2nd card if they want to give them more access than fits

YakumoChen:
Maybe they should think about what accesses are important to them if they want more

sippykot:
> Maybe they should think about what accesses are important to them if they want more

but like, there is nothing actually preventing you from getting more access, you just need to carry an extra id card around

YakumoChen:
only if the HoP is willing to do that

nevimer:
I feel like that if you want to get this much access, you should step your greytide game up. The HOP has silvers. Maybe you can social engineer them with a "lost id" bit and get more access that way. Or get a wallet from loadout and carry 4 cards...

aKromatopzia:
> being able to administratively give someone more access as a captain/hop =/= mechanical buffs to medical (???)
all the tg stuff does is make the hop or captain have to give people a 2nd card if they want to give them more access than fits

> I feel like that if you want to get this much access, you should step your greytide game up. The HOP has silvers. Maybe you can social engineer them with a "lost id" bit and get more access that way. Or get a wallet from loadout and carry 4 cards...

What Ozzo said is what I mean; it's not the "mechanical buffs", it's the idea that "since you can already take 10 steps to do something I would like if we only needed 1". Bypassing any requirement of consideration, having the slightest bit of inconvenience in needing to handle 2 ID cards, to have more accesses than a non-command member except maybe the janitor would ever need is an irregularity that shouldn't be normalized.

YakumoChen:
The HoP shouldn't be giving all access to anyone

Nobody needs that much access, not even the janitor (they have keys)

PR #3490, opened 2025-04-11
Fix admin chat buttons to cancel/refund storyteller events not working

TealSeer:
## About The Pull Request

The chat messages that get sent when a storyteller event is about to run try to call `Topic` on the scheduled event datum but there isn't actually a `Topic` proc defined for them

## Why It's Good For The Game

It's mean to trick admins with fake buttons

## Proof Of Testing
<details>

image

</details>

## Changelog
:cl:
admin: the cancel and refund buttons in storyteller event notifications will actually do something now
/:cl:


Odairu (approved):

Shadow-Quill:
This probably should check the user to make sure they have admin rights

PR #3494, opened 2025-04-13
Stowaway Job

nevimer:
## About The Pull Request
Adds a stowaway job (not really a job)
You as a stowaway are a non-crew inhabitant of the station. 
## Why It's Good For The Game

Requested by players. 

## Proof Of Testing
<details>

</details>

## Changelog
:cl:
add: Stowaway job (anti-job) added. 
/:cl:


SapphoQueer:
Does this prevent you from taking an engineer slot (or other jobs) as a stowaway? It'd be weird if you could not be on the manifest and still spawn with a toolbelt+rcd. It says "command loadouts" only.

Also make it be a random locker for extra soul.

Swiftfeather (changes requested):
Why not make stowaway a job trait in the same vein as sec vet and bridge assistant? 

Same people always being the 'stowaways' will wear quickly.

nevimer:
> Does this prevent you from taking an engineer slot (or other jobs) as a stowaway? It'd be weird if you could not be on the manifest and still spawn with a toolbelt+rcd. It says "command loadouts" only.
> 
> Also make it be a random locker for extra soul.

If you select a job, it will equip the gear, delete your id, pda, bank accounts, and manifest records. It won't announce you joining. It will also immediately free the slot you took, as if you never joined. Then it will teleport you to a maintenance location, and you are on your own from there.

Arturlang:
This feels like it should be it's own job with it's own loadout

SovietSpaghetti:
Really should probably be a job so you don't get random noncrew with sec loadouts or just picking their jobs to game in the first place.

nikothedude:
this feels roundabout, just make it a job or a ghost spawner

YakumoChen:
This comes off as a License To Grief role since security can just arrest you for trespassing on the station and take your stuff

Is there any reason to take this trait except to be Greytide± ?

AstroTonic:
> This comes off as a License To Grief quirk since security can just arrest you for trespassing on the station and take your stuff
> 
> Is there any reason to take this trait except to be Greytide± ?

There are ways to go about without immediately resorting to just greytide, and those who don't know how will only filter themselves out. This isn't a license to begin with, nor a self-antag pass. Picking this trait should make you aware that you forfeit your "crew rights" whilst still confining to server rules.

Fn-OS:
Just blacklist people with the quirk from playing anything but assistant. Easiest fix ever

nevimer:
> Just blacklist people with the quirk from playing anything but assistant. Easiest fix ever

That was attempted, but it would also stop you from having some job gear. Might be better in the end.

coffeecoffeecoffee:
Just blacklist certain jobs. I think some of them like cargo tech stowaways or doctor stowaways could be good, genuine fun without breaking things.

YakumoChen:
> There are ways to go about without immediately resorting to just greytide, and those who don't know how will only filter themselves out. This isn't a license to begin with, nor a self-antag pass. Picking this trait should make you aware that you forfeit your "crew rights" whilst still confining to server rules.

Outcome 1: Security catches you, creates FREE conflict with security since you break the law by existing, GOTO Outcome 2

Outcome 2: You go to HoP and create potential headaches for command / security by again, existing, depending on the job you want

Outcome 3: You persist in the round unnoticed with complete zero access (including maint) with your clothes and maybe basic tools, fate worse than playing Assistant, you exist to be security's problem (Outcome 1)

Make it make sense

DOUBLE JEPROADY: The quirk is not round limited so potentially half the crew could be unlisted, especially whenever this is (test)merged making the round hell

YakumoChen:
The worst part is that it's going to be me who gets bwoinked when I try to send the Undocumented Assiatants to El Spacidor when they get caught commiting space law violations, not them

Arturlang (changes requested):
Worth a try and i want to see how players interact with this, but this absolutely needs to be a limited job and the player needs to be able to have a way to open maint doors, so they dont get stuck, cant rely on them getting the loadout toolbox

SovietSpaghetti:
> > Just blacklist people with the quirk from playing anything but assistant. Easiest fix ever
> 
> That was attempted, but it would also stop you from having some job gear. Might be better in the end.

I don't see what you would need job gear for, or why you would have it if you're a stowaway. 

> Worth a try and i want to see how players interact with this, but this absolutely needs to be a limited job and the player needs to be able to have a way to open maint doors, so they dont get stuck, cant rely on them getting the loadout toolbox

It could also just be a job that starts with a toolbelt, maybe with the shitty improvised tools.

Fn-OS:
Should definitely be a ghostrole or limited spawns thing

Kajortoq:
I feel this would need to be done in joint with a sec policy statement since sec will (and should) round up non-manifest to either perma for being essentially boarders/nonpersons or give jobs, which defeats the purpose. If we just start giving everyone not on manifest a job, why not non-stowaways?

KathrinBailey:
Why a quirk and not just a seperate job or a ghost role?

nevimer:
Now as a job!

SovietSpaghetti:
It seems like every stowaway spawns in the same place. Is that intended?

nevimer:
> It seems like every stowaway spawns in the same place. Is that intended?

It should be using the same logic a blob does when it's overmind is created - find a safe spot in maints.

Fn-OS:
Cardboard ID's don't seem to open maints (on some maps at least), making it impossible to scrounge around or hide away without hacking

There's probably an ID that gives maints/assistant access somewhere in code without showing name/assignment/whatever

coffeecoffeecoffee:
As security: Handling this was incredibly fun actually. 10/10

CatoChristopherMrow:
Add the testmerged label considering this is testmerged please.

YakumoChen:
> Cardboard ID's don't seem to open maints (on some maps at least), making it impossible to scrounge around or hide away without hacking
> 
> There's probably an ID that gives maints/assistant access somewhere in code without showing name/assignment/whatever

Cardboard IDs aren't IDs. Not having an ID is the entire poing

VickiMorris:
On round 7945, I spawned as a Stowaway on Ice Box, and was placed in the RnD Server room, the cold side. Is there a way to set certain areas as excluded from spawning?

MajManatee:
It seems like as a stowaway, you can spawn anywhere? Ive had one spawn on the bride on Moon Station, and couldn't get out, then was arrested within seconds of arriving. Might be an idea to prevent that sort of thing?

NanoCats:
I just want to chime in that this is one of the things Monkestation added, and when it was added there it was basically just a big fat "Make a lot of headaches for Sec" feature.. 

They also made one where you started the round wanted for a random crime, and that fucking sucked so bad man, 10 people would spawn in wanted and you can't just haul them in because people would interfere with the arrest and they'd whine.

NanoCats:
> Yesterday I spawned in an enclosed pool of fire ants, died promptly within a minute, and had admins IC issue me when I reported it. 10/10 would play again. image

This is why you turn your suit sensors on

NanoCats:
> > It seems like every stowaway spawns in the same place. Is that intended?
> 
> It should be using the same logic a blob does when it's overmind is created - find a safe spot in maints.

Don't blobs typically spawn in a spot selected by the player which has almost no safety net? (Read: blob in the SM that dusts instantly), I imagine the automatic blob spawn is almost never used.

modular_zubbers/code/modules/jobs/stowaway.dm
```diff
@@ -0,0 +1,42 @@
+/datum/job/assistant/stowaway
+	title = "Stowaway"
+	description = "Be a stowaway crewmember."
```
BurgerLUA, line 3: Should change this description to make it clear that you're not crew and don't count as one.

modular_zubbers/code/modules/jobs/stowaway.dm
```diff
@@ -0,0 +1,42 @@
...
+
+	var/mob/living/carbon/human/human_holder = spawned
+	var/datum/bank_account/account = SSeconomy.bank_accounts_by_id["[human_holder.account_id]"]
+	account?.Destroy()
+	var/turf/T
+	if(LAZYLEN(GLOB.blobstart))
+		var/list/blobstarts = shuffle(GLOB.blobstart)
+		for(var/_T in blobstarts)
+			T = _T
+			break
```
BurgerLUA, line 33: You should probably use the build in proc ``find_maintenance_spawn(atmos_sensitive = TRUE, require_darkness = FALSE)`` instead of the custom code here. Blob spawn turfs are meant for blobs who don't need to breathe or worry about temperature, and generally have no guidelines upstream in terms of where they should be placed in terms of safety, since only blobs use them. I'm pretty sure on some maps there are blob spawns inside atmos tanks.

Also fun fact: On metastation, there are some blob spawn locations in erp-heavy areas or extremely sketchy, such as:
- The library backroom.
- Some bathroom stalls.
- The detective's closet.
- Evidence Room storage.

nevimer: you are right

modular_zubbers/code/modules/jobs/stowaway.dm
nevimer: done

modular_zubbers/code/modules/jobs/stowaway.dm
BurgerLUA: :sunglasses:

IgiariValkyr:
This is just a license to tide and does nothing to add to the station. Oh hey there's a non-crew, guess they're getting arrested and sat in perma for the whole round.

Fn-OS:
> Ghost roles can be told to go home and kicked back off the station
> 
> 
> 
> Stowaways aren't inherently from anywhere off site, it is a lot harder to justify deporting them back to where they came from.
> 
> 
> 
> It's not practical to give them a space suit and force them to leave at gunpoint and likely to result in an ahelp.
> 
> See also: [this post](https://github.com/Bubberstation/Bubberstation/pull/3494#issuecomment-2800171055)

Could always kick them off to lavaland or send them through the gateway

If someone ahelps for being kicked off, that's an IC issue and something you acknowledge will happen eventually when you play that role

Stowaway is a fun role to play, and allows for more gimmicks and rp to happen.

PR #3498, opened 2025-04-13
restores RCD vending, moves RCD's to lockers

nevimer:
## About The Pull Request
moves RCD's from engineers backpacks to their lockers roundstart - to accommodate the stowaway change aswell as keep a tab on infinite station resources. 
## Why It's Good For The Game

Above point

## Proof Of Testing
<details>

</details>

## Changelog
:cl:
balance: reenabled the RCD vendors and moved RCD's from spawning in backpacks to lockers. 
/:cl:

### DIFF
--- a/code/game/objects/structures/crates_lockers/closets/secure/engineering.dm
+++ b/code/game/objects/structures/crates_lockers/closets/secure/engineering.dm
@@ -18,10 +18,13 @@
 	new /obj/item/extinguisher/advanced(src)
 	new /obj/item/storage/photo_album/ce(src)
 	new /obj/item/storage/box/skillchips/engineering(src)
-	new /obj/item/storage/box/gas_miner_beacons(src) // SKYRAT EDIT ADDITION
-	new /obj/item/construction/plumbing/engineering(src) //SKYRAT EDIT ADDITION
-	new /obj/item/circuitboard/machine/rodstopper(src) //SKYRAT EDIT ADDITION
-	new /obj/item/card/id/departmental_budget/eng(src) //SKYRAT EDIT ADDITION
+	// BUBBER EDIT BEGIN
+	new /obj/item/storage/box/gas_miner_beacons(src)
+	new /obj/item/construction/plumbing/engineering(src)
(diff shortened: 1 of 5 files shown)


Odairu:
not every map has enough lockers to accomadate for how many engi slots we have
not the BIGGEST of deals, but could be very annoying on high pop being forced to wait on research/use the first 2 department orders on them

lessthnthree (changes requested):
Because there's still also the RCD added to the EngiVend in modular_skyrat/modules/modular_vending/code/engivend.dm the RCD appears twice on the vending machine.

lessthnthree (approved):

TealSeer:
> not every map has enough lockers to accomadate for how many engi slots we have not the BIGGEST of deals, but could be very annoying on high pop being forced to wait on research/use the first 2 department orders on them

RCD is base research now thanks to our beloved skyrat overlords

trefoils:
Why are we punishing the lowest played dept further for a trait? Just make a new load out for stowaway copy the engineers shit over minus the rcd. Stop punishing engineering for no reason. The fucking rcd makes like five walls if that before going out, do you think that tiny amount of iron is going to make a difference? If so maybe should look into removing the gold from the vault and the starting silver mining gets. the lockers are not near enough for the job slots, the vending machine is over priced as fuck.  Engineering loses a small bit of qol for no reason no gain and only a loss.   RCD's are expensive to print from a lathe as well, mats that could be used towards a rebuilding effort in the event of coming on during an emergency now have to be funneled into making the damn rcd in the first place.

DorfGuy1:
is this the same person who did this for the portable crew scanners? Do you just have thing for vending machines? nobody likes using vending machines any more than they have to, cause everything has  like a 400% markup

AngryDoorknob:
If I've understood this correctly, this would only subtract from other's experience, especially engineering players, since they won't know if they can even get an rcd (without having to go other places to get one). Infinite is also in error, as rcd matter storages are not that big. I can understand that it is 'just a minor inconvenience', but these can very rapidly add up to being a very annoying waste of time, especially if you have to do it every shift.

PR #3515, opened 2025-04-14
We have a propigating NaN

StrangeWeirdKitten:
Is it a gas issue or is someone making a sun in a can. I strive to know. This is debugging code.


Vanilla1040:
i know how to make a sun in a can and im pretty sure its NOT intentional as i dont think a small canister could withhold 700000
 kelvin

Vanilla1040:
if the burn can has like the TINIEST amount of oxygen in it, even if its just 1 mol of oxygen and heated plasma in it, it just keeps "burning" rising the temp more and more to a nearly endless amount until the oxygen mol reaches 0.01% then it FINALLY stops heating up

PR #3548, opened 2025-04-18
Non-living beings no longer glow from being irradiated.

BurgerLUA:
## About The Pull Request

Non-living beings no longer glow from being irradiated.

## Why It's Good For The Game

Massive GPU lag on the client from the amount of radiation filters any time something is irradiated is not good. This PR makes it so that radiation is no longer visible on things that aren't living. You can still check if something is irradiated by scanning it with a Geiger counter.

## Proof Of Testing

Untested.

## Changelog

:cl: BurgerBB
del: Non-living beings no longer glow from being irradiated.
/:cl:

### DIFF
--- a/code/datums/components/irradiated.dm
+++ b/code/datums/components/irradiated.dm
@@ -61,7 +61,7 @@
 	))
 
 /datum/component/irradiated/Destroy(force)
-	var/mob/living//parent_movable = parent //BUBBERSTATION CHANGE: MOVABLE TO LIVING
+	var/mob/living/parent_movable = parent //BUBBERSTATION CHANGE: MOVABLE TO LIVING
 	if (istype(parent_movable))
 		parent_movable.remove_filter("rad_glow")
 
--- a/code/datums/components/irradiated.dm
+++ b/code/datums/components/irradiated.dm
@@ -61,7 +61,7 @@
 	))
 
 /datum/component/irradiated/Destroy(force)
-	var/atom/movable/parent_movable = parent
+	var/mob/living//parent_movable = parent //BUBBERSTATION CHANGE: MOVABLE TO LIVING
 	if (istype(parent_movable))
 		parent_movable.remove_filter("rad_glow")
 
@@ -149,7 +149,7 @@
 	)
 
 /datum/component/irradiated/proc/create_glow()
-	var/atom/movable/parent_movable = parent
+	var/mob/living/parent_movable = parent //BUBBERSTATION CHANGE: MOVABLE TO LIVING.
 	if (!istype(parent_movable))
 		return
 


nevimer:
You shall test this PR....
I detect unforseen issues, namely with radiation events.

CatoChristopherMrow:
What do you MEAN you didn't test it??? We don't test on production. (ignore test merges)

BurgerLUA:
I see a downvote from a maintainer, which scares me, so I will better explain this PR.

Items that are irradiated have zero functionality. As it stands, it currently has no mechanical purpose other than telling you "this item was exposed to radiation", which can is information that can be learned anyways with the Geiger counter.

There was a bunch of irradiated items in medical one shift and it dropped my framerate to absurdly low levels. This is a common thing to happen when you have too many filters + overlays on a screen. I don't know if /tg/ has a similar issue with this, but bubbers certainly does considering the amount of overlays one mob makes (markings, special species organs, modular clothing, etc).

Given this, radiation effects for items needs to be removed. Other mechanics can convey the same information as the glowing, without the massive performance impact.

SapphoQueer:
Should probably be pr'd upstream if you can reproduced there.

BurgerLUA:
> Should probably be pr'd upstream if you can reproduced there.

I can't reproduce upstream as erp is against the rules.

BurgerLUA:
Upstream does not have this issue so it's a bubbers issue. As mentioned above, bubbers has a ton of stuff for clients to process, so lag occurs from being overloaded.

CatoChristopherMrow:
Was this TMed with the proof of testing section being "untested" and no comment about it being tested by the author?

BurgerLUA:
> Was this TMed with the proof of testing section being "untested" and no comment about it being tested by the author?

yes

(To be fair these are very simple edits. I've submitted PRs that I haven't tested on this codebase and other coderbases before that contained more complex edits that didn't have any issues. At the same time, I've tested things that worked but caused other issues with other features. Coding is fickle.)

plsleavemealon:
Wait...irradiated items do not irradiate people anymore? That is shocking to me, I am surprised they even let the items become irradiated if they no longer have any function to it.

BurgerLUA:
> Wait...irradiated items do not irradiate people anymore? That is shocking to me, I am surprised they even let the items become irradiated if they no longer have any function to it.

Previously they did for quite a while, but it was removed because there was a lot of exploits involving them (put a bunch iof machine parts next to the supermatter and now you have a rped of death)

BurgerLUA:
I actually never tested this but it seems the server did for a month so this seems good to me.

Shadow-Quill (approved):

PR #3567, opened 2025-04-19
[MODULAR] Modulator Gasmask: Hide your voice for cheap!

VickiMorris:
## About The Pull Request

Partial port of: https://github.com/BeeStation/BeeStation-Hornet/pull/9186 made with permission from original author.

<details>

image
image

</details>

Adds a craftable gasmask that sets your voice to **"Unknown"** and gives you a robotic voice (custom 'say' verb of "states", and robotic voice span), giving a generic but very obvious method of disguising your voice. It is crafted from basic parts that can be found in maintenance/tool storage/at an autolathe, with the hardest part probably being the 'classic' gasmask.

## Why It's Good For The Game

As it stands, the only methods to disguise your voice (that I am aware of) are: Have an ID with a custom name, and a chameleon gasmask (unlikely if you aren't a traitor with an uplink). Be a changeling. Be a heretic and use the shadow cloak ability (extremely obvious). Remove your ID and use sign language (annoyingly specific, can't be used while cuffed, holding items, or over radio)

This gives a lot more leeway on disguising your identity, and gives a new way of removing the meta-tell of "I know it's X because I heard their voice" without needing to be a specific antag or spend TC on it. I don't picture this being a security issue, as use of one could very easily fall under 'Refusal to Identify' or 'Impersonation', and it is very obvious what is being done.

## Proof Of Testing
<details>

Crafting

https://github.com/user-attachments/assets/91bc69d4-9d70-4344-a3c8-967cfe4f3f06

</details>

## Changelog
:cl: Impish_Delights
add: Voice modulator gasmask, a craftable gas mask that robotizes and disguises your voice!
/:cl:


modular_zubbers/code/modules/clothing/mask/gasmask.dm
```diff
@@ -0,0 +1,63 @@
...
+
+	if(H.wear_mask == src && modulate_voice)
+		H.special_voice = modulated_name
+		T.temp_say_mod = "states"
+	else
+		H.special_voice = previous_special_name
+		T.temp_say_mod = initial(T.temp_say_mod)
+		previous_special_name = null
+
+	return
```
softcerv, line 55:
```suggestion
		return 
	
	H.special_voice = previous_special_name
	T.temp_say_mod = initial(T.temp_say_mod)
	previous_special_name = null
```

modular_zubbers/code/modules/clothing/mask/gasmask.dm
```diff
@@ -0,0 +1,63 @@
+/obj/item/clothing/mask/gas/modulator
+	name = "modified gas mask"
+	desc = "An older model of gas mask issued for use on station to help protect against airborne hazards. This one appears to be \
+		heavily modified, and the filter assembly has been replaced with a voice modulator to make the wearer sound more robotic."
+	clothing_flags = BLOCK_GAS_SMOKE_EFFECT | MASKINTERNALS
+	max_filters = 0
+	var/modulated_name = "Unknown" //Our SPEAKING name while our voice is cloaked. Defaults to "Unknown", but can be set by var editing.
+	var/previous_special_name = null //Used to store a special name if they wearer has one before equipping the mask.
+	var/modulate_voice = TRUE //If we are actually using the voice changer or not.
```
softcerv, line 9:
```suggestion
	///Our SPEAKING name while our voice is cloaked. Defaults to "Unknown", but can be set by var editing.
	var/modulated_name = "Unknown"
	///Used to store a special name if they wearer has one before equipping the mask.
	var/previous_special_name = null
	///If we are actually using the voice changer or not 
	var/modulate_voice = TRUE 
```

modular_zubbers/code/modules/clothing/mask/gasmask.dm
```diff
@@ -0,0 +1,70 @@
...
+	var/modulated_name = "Unknown"
+	///Used to store a special name if they wearer has one before equipping the mask.
+	var/previous_special_name = null
+	///If we are actually using the voice changer or not 
+	var/modulate_voice = TRUE 
+
+
+/obj/item/clothing/mask/gas/modulator/examine(mob/user)
+	. = ..()
+	. += span_notice("Its voice modulator is currently [modulate_voice ? "enabled" : "disabled"]. <b>Alt-click</b> to toggle it.")
```
TealSeer, line 17: Would like this to be a context tooltip when you mouse over the item instead of in the examine text (the alt click to toggle part)

modular_zubbers/code/modules/clothing/mask/gasmask.dm
```diff
@@ -0,0 +1,70 @@
...
+/obj/item/clothing/mask/gas/modulator/dropped(mob/user)
+	. = ..()
+	UnregisterSignal(user, COMSIG_MOB_SAY)
+	update_voice(user)
+
+/obj/item/clothing/mask/gas/modulator/proc/update_voice(mob/user)
+	if(!ishuman(user))
+		return
+
+	var/mob/living/carbon/human/H = user
```
TealSeer, line 44: no single letter variables, make it something more verbose

modular_zubbers/code/modules/clothing/mask/gasmask.dm
```diff
@@ -0,0 +1,70 @@
...
+	var/obj/item/organ/tongue/T = null
```
TealSeer, line 45: this whole thing can be simplified to something like 
```suggestion
	var/obj/item/organ/tongue/tongue = H.get_organ_slot(ORGAN_SLOT_TONGUE)
	if(!tongue)
		return
```
and then you avoid needing to check the tongue var every time further down

TealSeer (changes requested):

modular_zubbers/code/modules/clothing/mask/gasmask.dm
VickiMorris: I specifically want to avoid an early return if the user doesn't have a tongue, as it prevents their name from being disguised.

TealSeer (approved):

VickiMorris:
Not an expert, but I think this is the flaky error I've been seeing pop up in the github channel.

TealSeer:
> Not an expert, but I think this is the flaky error I've been seeing pop up in the github channel.

it is yes, feel free to disregard

Vanilla1040:
Funny side note

You can pick up a potted plant and talk over comms no matter what you have on gas mask or not ID equipped or not it will show you as "Unknown"

VickiMorris:
> Funny side note
> 
> You can pick up a potted plant and talk over comms no matter what you have on gas mask or not ID equipped or not it will show you as "Unknown"

I did see this while code diving! This is because it applies the "Unknown" trait which wholly hides your identity (similar to heretic Shadow Walk, or the tot Stealth MOD). This also makes you unable to be examined, and your mouse-over name is always "Unknown", even if you have an ID.

GoreDem:
You can't actually use the gas masks from firelockers to make this 👍

VickiMorris:
> You can't actually use the gas masks from firelockers to make this 👍

This was intentional. I can change it to use that as well and generate an alt-version that uses those textures, if that's wanted.

PR #3568, opened 2025-04-20
[TM ONLY] Fix shuttles making negative pipenets

Shadow-Quill:
Copy of https://github.com/tgstation/tgstation/pull/90721 so we can TM it.


TealSeer:
Early pulled

PR #3596, opened 2025-04-22
S.W.A.T part 4

CatoChristopherMrow:
## About The Pull Request

Makes the following SMGs require two hands to use:
Shaytan
Sinado
## Why It's Good For The Game
Generally, I think it's unreasonable to be able to one hand a full auto gun (especially a ballistic one) with a shield in the other hand. Helps make these guns a bit more reasonable. 
## Proof Of Testing
I snuck into your home at night and tested this. It works.
<details>

</details>

## Changelog
:cl: UvvU
balance: Made most SMGs require two hands.
/:cl:


CatoChristopherMrow:
I will note, this also helps reduce the shield meta, which I am personally not a fan of.

projectkepler-ru:
not opposed

code/modules/projectiles/guns/energy/stun.dm
```diff
@@ -45,6 +45,8 @@
 	name = "disabler smg"
 	desc = "An automatic disabler variant, as opposed to the conventional model, boasts a higher ammunition capacity at the cost of slightly reduced beam effectiveness."
 	icon_state = "disabler_smg"
+	weapon_weight = WEAPON_HEAVY //BUBBER ADDITION
```
projectkepler-ru, line 48: these can be placed in modular folder

code/modules/projectiles/guns/energy/stun.dm
```diff
@@ -45,6 +45,8 @@
...
+	w_class = WEIGHT_CLASS_BULKY //BUBBER ADDITION
```
projectkepler-ru, line 49: ditto

YakumoChen:
This PR also only has code for the disabler (is this supposed to be a draft?)

The premise doesn't really make any sense at all when the majority of long guns in the game aren't two-handed, and SMGs are generally not even long-arms.

CatoChristopherMrow:
Ouuugh fuck my PR fucked up you're so right. One moment.

CatoChristopherMrow:
Alright, fixed. I should note that this also makes the disabler SMG bulky.

As for the changes, sure, they're the same on PAPER, but in practice they're alot different. Landing your first shot matters ALOT with disablers, as it makes every subsequent shot easier, so disabler SMGs in practice, are alot better because of this. Additionally, they're harder to dodge, as there's twice as many projectiles to do so. This is a pretty small nerf to them to stop them from being just better. I've talked to a good few of the more mechanical inclined players about them, and we can all agree that disabler SMGs are a bit overturned.

Shadow-Quill:
> disabler SMGs

A note about these - while they do give you double the shots, it _also_ halves the stamina damage each bolt does, so it turns out to be able to do effectively the same as a normal disabler, it just tends to feel better because semi-auto

CatoChristopherMrow:
> > disabler SMGs
> 
> A note about these - while they do give you double the shots, it _also_ halves the stamina damage each bolt does, so it turns out to be able to do effectively the same as a normal disabler, it just tends to feel better because semi-auto

It is infact, as I previously stated, more effective due to how slowdown and the amount of projectiles work. Again, on paper it's the same, in practice not so much.

YakumoChen:
@StrangeWeirdKitten might want to comment on this again in any case

CatoChristopherMrow:
Upon testing, the disabler downed in 5 shots. The SMG in 9.

YakumoChen:
That's probably because we use 135 hp instead of 100, so the way the numbers add up aren't going to be exactly as fair as tg

CatoChristopherMrow:
> That's probably because we use 135 hp instead of 100, so the way the numbers add up aren't going to be exactly as fair as tg

I'm aware of why it's happening, and I'm glad we can agree that it's not as fair as on TG. Hence, why it requires two hands here.

YakumoChen:
I don't think the difference of one shot for a gun that does the same thing one of two different ways justifies making it abysmally bulky and unwieldy by comparison.

YakumoChen:
The second order effect of making it not fit in a bag (as a fully non-lethal gun) is that they're going to have to ditch their likely only non-lethal gun as soon as the armoury opens, presuming they have a HoS or warden that doesn't just let them carry extra disablers

CatoChristopherMrow:
That's a weird presumption to make.

YakumoChen:
Sometimes the warden doesn't let officers have bags full of guns. Shocker, I know.

VickiMorris:
It's been a bit since I played as a seccie. Don't you start with a default disabler in your backpack? To get to this, they'd have to order a disabler SMG, I presume turn their old disabler in to the warden or their locker, then when armory opens, decide to just, not grab the disabler they dropped off.

This is a personal thing too, but I feel like a lot of people don't see "full-automatic" as an upgrade, when it is. The gun being easier to use and keep sustained shots on target makes it an upgrade over a regular disabler, even when it has the same theoretical damage.

CatoChristopherMrow:
This is hardly to the floor.

CatoChristopherMrow:
> Making it bulky and require both hands is a very substantial nerf on a weapon that isn't really much of an issue, at all. I also don't mind non-lethal sources being a little easier to use versus their non-lethal counterparts, especially being paired with things like shields. Yes, by making it unwieldy, you are effectively making it more undesirable to ever use as a trade off for what is very minor balance concerns or gains. I'll have the lethal SMG bulky modifier looked at and discussed separately since lethals are a big of a different story, but the disabler SMG is not something that needs a nerf and there's not really been anything said to convince me otherwise, sorry.

Yaknow what, I'll leave the whole disabler SMG discussion for way down the road. For now I'll change just to just the lethal SMGs when I get the chance.

CatoChristopherMrow:
Ok, disabler SMG removed from this PR. I'll get you yet, my nemesis...

NanoCats:
Isn't that like the whole point of an SMG over a fully fledged rifle

DorfGuy1:
Yeah,  right now the meta option is the CMG or the Romulus, or it used to be a at least. Shield and SMG is cool and tactical. Also shield is so used because of the crazy ap of a lot of antag weapons. And Scarp. What I'm getting at is that the shield is a symptom of the overall arms race and meta hugging between antag and sec players, where both sides have conditioned the other to feel like they absolutely must powergame (And use the same boring shit each round)  to win. At least that's the impression I get, from seeing both sec and antags use similar loadouts frequently. (Grain of salt. I don't know if we collect stats on what antags use what in a convient, readable form for this to be a statistically accurate statement, and we definitely don't on the loadouts of sec officers. Such a thing might be useful tbh.)

PR #3605, opened 2025-04-23
[TM Only] Supermatter anomaly/hallucination fixes #90788

lessthnthree:
TM of some anomaly/hallucination fixes for delams.


lessthnthree:
Merged

PR #3427, opened 2025-04-02
Coom 2.0 - lets you cum on things

sippykot:
## About The Pull Request

I was looking thru the code for ways to cum into beakers... and turns out we already had one. It just was impossible to use and kind of broken when you did use it. With this, you can now cum into reagent containers and make cum decals on floors by targeting other things. It comes with a 6 sec channel time by default so its not too spammable (can add a long emote cooldown timer or lengthen the channel time if it turne out to be an issue).

## Why It's Good For The Game

As it currently stands, there is no way to actually cum into beakers. There's the milking machine, but it's extremely awkward to use and doesn't fit most rp scenarios. We already had a code intended for this, it just wasn't used. Tremendous NSFW maintenance bar rp benefits?

## Proof Of Testing

I did test it. Quite a lot. Too much while trying to figure out sex code and why it wasn't working. In the end I had to cut the condom interaction because i just couldn't get it to use one properly, but anyways, it should work.

## Changelog
:cl:
add: adds a *fap emote, letting you jerk off onto a target - you can finally cum into reagent containers
fix: the old previously unused coom hand_item had a bunch of issues that should be fixed now
/:cl:

PR #3614, opened 2025-04-23
Removes practically unused genital fluids

Shadow-Quill:
## About The Pull Request

Sexual fluids are basically unused in the code, outside of the milking machine.
Considering the obscurity (and the issues where they can just become a problem with no benefit (_eyes that one bubber gif_)), I felt it best to flat out remove them.

Orgasm outcomes are unchanged, as those use unrelated decals.

## Why It's Good For The Game

Removal of code used basically nowhere else.

## Proof Of Testing

Nyet.

## Changelog

:cl:
remove: Orgasm reagents have been removed.
/:cl:


CatoChristopherMrow:
Ok but it IS used by some ERP items. I feel like there's not a strong enough justification for removal here. If someone uses them outside of ERP, that's a administration issue, not a code issue. Also what do you MEAN no testing?

Shadow-Quill:
> some ERP items

They're used by one; exclusively the milking chair..

> Also what do you MEAN no testing?

image

MothionTechnologies:
The fluids are not just used by the milking machine (at least for cum). #3427 gives the ability to also directly fill beakers and some people do use that function for ERP.

Shadow-Quill:
> #3427 gives the ability to also directly fill beakers [...]

and it was closed three days ago

plsleavemealon:
> > #3427 gives the ability to also directly fill beakers [...]
> 
> and it was closed three days ago

It was closed because it was an emote. It was not closed because of its function, so if the author found a way to do the same function with it being an emote it would be fine.

sippykot:
why is the fact that code is only used in one place a justification to remove it...?

sippykot:
> > > #3427 gives the ability to also directly fill beakers [...]
> > 
> > 
> > and it was closed three days ago
> 
> It was closed because it was an emote. It was not closed because of its function, so if the author found a way to do the same function without it being an emote it would be fine.

well, its back now #3677

SpaceCatSS13:
>Orgasm outcomes are unchanged, as those use unrelated decals.
This is not true, one of the outcomes allows your to fill a container

PR #3618, opened 2025-04-24
Unifies coco powder to one reagent instead of two

KazooBard:
## About The Pull Request

This PR aims to solve this issue by making any ways to obtain the non-crafting-friendly hot chocolate powder give the regular hot chocolate powder instead
yeahs
## Why It's Good For The Game
Unintuitive, now with less hassle, now with more ease, harms noone, makes people happy. 

I swear if hot chocolate causes the destruction of the server or controversy on merge I will eat a sheet of paper. 
## Proof Of Testing
Basic replacement, trust me bro
## Changelog
:cl:
qol: unifies the two existing coco powder related reagents into one

/:cl:

### DIFF
--- a/modular_skyrat/modules/colony_fabricator/code/appliances/chem_machines.dm
+++ b/modular_skyrat/modules/colony_fabricator/code/appliances/chem_machines.dm
@@ -96,12 +96,13 @@
 	show_ph = FALSE
 	base_reagent_purity = 0.5
 	// God's strongest coffee machine
+	//bubber edit: Coco instead of powdered chocolate
 	dispensable_reagents = list(
 		/datum/reagent/water,
 		/datum/reagent/consumable/powdered_milk,
 		/datum/reagent/consumable/sugar,
 		/datum/reagent/consumable/powdered_lemonade,
-		/datum/reagent/consumable/powdered_coco,
+		/datum/reagent/consumable/coco,
 		/datum/reagent/consumable/powdered_coffee,
 		/datum/reagent/consumable/powdered_tea,
 		/datum/reagent/consumable/vanilla,
@@ -114,6 +115,7 @@
 		/datum/reagent/consumable/nutraslop,
 		/datum/reagent/consumable/enzyme,
 	)
+	//bubber edit end
 
 	/// Since we don't have a board to take from, we use this to give the dispenser a cell on spawning
 	var/cell_we_spawn_with = /obj/item/stock_parts/power_store/cell/high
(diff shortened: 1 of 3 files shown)


Swiftfeather:
We're no longer a skyrat downstream, you don't really need to include the bubber edits in skyrat files.

KazooBard:
> We're no longer a skyrat downstream, you don't really need to include the bubber edits in skyrat files.

Noted for future reference! Will do!

sippykot:
uhhh i dont think this is a good idea - the powdered coco and coco powder are NOT the same thing
one is actual coco and the other is pretty much vending machine sloppa
just like powdered tea and coffee are not the same as tea and coffee powder...
the names are kinda confusing but you get the point

plsleavemealon:
> uhhh i dont think this is a good idea - the powdered coco and coco powder are NOT the same thing one is actual coco and the other is pretty much vending machine sloppa just like powdered tea and coffee are not the same as tea and coffee powder... the names are kinda confusing but you get the point

We have purity levels to reflect this. The purity of the output of the "vending machine" (sustenance dispenser) is 50%, which is lower than what you'd get from good quality cocoa beans.

sippykot:
> > uhhh i dont think this is a good idea - the powdered coco and coco powder are NOT the same thing one is actual coco and the other is pretty much vending machine sloppa just like powdered tea and coffee are not the same as tea and coffee powder... the names are kinda confusing but you get the point
> 
> We have purity levels to reflect this. The purity of the output of the "vending machine" (sustenance dispenser) is 50%, which is lower than what you'd get from good quality cocoa beans.

i... guess? still its weird to make it do real coco but not coffee and tea

plsleavemealon:
It does make real coffee and tea, its just lower purity results for it and requires a slight extra step.

sippykot:
> It does make real coffee and tea, its just lower purity results for it and requires a slight extra step.

so does coco lol
tea leaves are used for cooking

plsleavemealon:
> > It does make real coffee and tea, its just lower purity results for it and requires a slight extra step.
> 
> so does coco lol tea leaves are used for cooking

Then if anything, those should also be unified alongside this.

DorfGuy1:
> > We're no longer a skyrat downstream, you don't really need to include the bubber edits in skyrat files.
> 
> Noted for future reference! Will do!

You should include "Bubber edit" While leaving as many of the "Skyrat Edit" comments in as possible. That way when someone downstreams us and makes edits they can do the same, so people can appreciate just how old some parts of the code are

PR #3646, opened 2025-04-26
Implement nanite scanner gate detection

TealSeer:
## About The Pull Request

Add a new signal to nanites to get their cloud ID, implement the nanite detection function on scan gates to match cloud ID

## Why It's Good For The Game

This was half implemented now its full implemented

## Proof Of Testing
<details>

image

</details>

## Changelog
:cl:
add: scanner gates can now be set to detect certain nanite cloud IDs
/:cl:

### DIFF
--- a/code/game/machinery/scanner_gate.dm
+++ b/code/game/machinery/scanner_gate.dm
@@ -399,7 +399,7 @@
 			var/new_cloud_id = params["new_cloud"]
 			if(!isnum(new_cloud_id))
 				return
-			nanite_cloud = clamp(new_cloud_id, NANITE_MIN_CLOUD_ID, NANITE_MAX_CLOUD_ID)
+			nanite_cloud = clamp(new_cloud_id, NANITE_MIN_CLOUD_ID+1, NANITE_MAX_CLOUD_ID)
 			. = TRUE
 		// BUBBER EDIT ADDITION END - NANITES
 
--- a/code/game/machinery/scanner_gate.dm
+++ b/code/game/machinery/scanner_gate.dm
@@ -320,6 +320,10 @@
 			"specie_name" = capitalize(format_text(specie.name)),
 			"specie_id" = species_id,
 		))
+	// BUBBER EDIT ADDITION BEGIN - NANITES
+	.["min_cloud_id"] = NANITE_MIN_CLOUD_ID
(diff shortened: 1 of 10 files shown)


code/game/machinery/scanner_gate.dm
```diff
@@ -378,6 +390,14 @@
 						detect_gender = "female"
 			. = TRUE
 		//SKYRAT EDIT END - MORE SCANNER GATE OPTIONS
+		// BUBBER EDIT ADDITION BEGIN - NANITES
+		if("set_nanite_cloud")
+			var/new_cloud_id = params["new_cloud"]
+			if(!isnum(new_cloud_id))
+				return
+			nanite_cloud = clamp(new_cloud_id, 1, 100)
```
Arturlang, line 398: Can you use a shared define for max and min clouds?

Arturlang (changes requested):

code/game/machinery/scanner_gate.dm
TealSeer: would need to change it in the other code and pass it in static data so the chamber control panel can use it too, that alright with you?

tgui/packages/tgui/interfaces/ScannerGate.jsx
```diff
@@ -320,8 +320,9 @@ const ScannerGateNanites = (props, context) => {
             <NumberInput
               value={nanite_cloud}
               width="65px"
-              minValue={1}
-              maxValue={100}
+              minValue={min_cloud_id + 1}
+              maxValue={max_cloud_id}
+              step={1}
```
Arturlang, line 325: Do you know if changing step to 1 does anything here? Does this prevent from fractions or? Checked tgui-core and it mostly seems to affect dragging 
image

Arturlang: https://github.com/tgstation/tgui-core/blob/0aa77815c306dccfc43cb004006711d5a66c0bfc/lib/components/NumberInput.tsx#L134

TealSeer: I threw that in because dragging was broken on that input

Arturlang (approved):

PR #3653, opened 2025-04-28
Fixes synth organ emp spam

ArrisFairburne:
## About The Pull Request

Fixes a bug where synth organs could be EMP spammed to destroy them entirely almost immediately. They now require a 10 second wait between EMP uses to deal organ damage again.
## Why It's Good For The Game
From what I can tell the synth organs were intended to have this mechanic -- it's just that the synth organs aren't coded well. The cooldown was defined in the organ code for a long time, and always marked as "so we can't EMP spam to kill people", it just didn't actually touch anything of significant note or do anything of actual value.

This will prevent situations where EMPs (in this case, easily-spammed EMPs like the EMP flashlight) just explode all of a synth's organs. It won't remove their EMP weakness in general, though. 
## Proof Of Testing
I tested this locally and it works, the cooldown check now actually affects if the organ takes damage or not.
## Changelog
:cl:
fix: Fixed the EMP damage cooldown on synth organs not activating. Synths now require a 10 second pause between EMPs to deal further damage to them.
/:cl:


Vanilla1040:
thank fuck finally, thank you

ReturnToZender:
This is nothing short of amazing, great work!

TealSeer (approved):

PR #3675, opened 2025-04-30
Restricts guns from being able to be held on the security belt

KazooBard:
## About The Pull Request

Wands, syringeguns + variants, the reagent dartgun, AR rifles and the romtech flechette  gun are no longer able to be stored on the security belt.

## Why It's Good For The Game

The goal of this PR is to prevent egregious cases of gunstacking, return predictability to their storage and by proxy return the syndicate holster to a better position. 

Firstly, a note: If I knew the PR would go through, as it did not in the past, I would have made this a full security belt weapons removal. The notion, that a person can hold 4 normal sized weapons on their belt slot is simply insane. 80 disabler shots, and whatever it is that the person brings in the backpack - the larger storage item of both.

Without further ado, allow me to explain why this item is problematic in it's current state.

----------------------------------------------------------------------------------------------------------

# Issue 1. Intention vs Outcome, the egregious cases of the generic typepath.

The largest issue that the item currently has is it being universal. I have manually selected items which cannot be held in it, but I still fear it is not enough. The generic typepath it uses, being /obj/item/gun, means that ANYTHING normal-sized that is a gun can be fit there, even, if it's not technically a gun, nor would it logically make sense from a game design standpoint.

**In practise,** Wizard Wands can be fit in the security belt, the belt used to store security equipment. Plasma cutters and PKA's can be stored in the security belt. Syringe guns which aren't bulky, as well, and so can be the reagent dartgun.

The reasoning is, that it's not only nonsensical, but a kick to the groin to the unspoken rule that belts generally respect eachother's personal space and a true omnibelt isn't *really* a thing. This applies to crew belts. 

However, if someone were wise enough to say "well what about X antag belt, it goes against the rule by being universal or allowing gun stacking".

Yes. That was the idea. Antagonist tools are always better versions of what the crew has.  It's working as intended. This? This is an oversight, nobody would have wanted sec belts to be able to store a rare antagonist's weapons unless it were a clear as day lack of forethought.

# Issue 2. Trampling the Security belt(bubber edition's) established rules.
Our belts are capable of storing medium sized wepaons, so handguns, mostly handguns and the 2 unique lasers. Clearly, the less firepower a person can keep on the belt at once, the better, as no aces remain hidden up anyone's sleeves ready to deal devastating damage and leave enough space for copies of itself. 

This helps prevent a single person going munchkin but allows for accessible storage and not having to minmax every slot, letting a player have a comfortable idea of "It'll fit somewhere, surely". Now, anyone can have 4 disablers, that's not very fun, true, but compared to main weapons such as the sol, the laser, the sindano, etc, none of the crew handguns have a comparable damage output.
I believe firmly that this is on purpose. 

Now, the odds of two guns, who do not fit the norm of sidearms with their DPS, however do with their size - being an outlier and deserving exclusion are very high considering that wands can be stored on the belt. 

The Flechette rifle and the AR's are powerful guns. They do not compare to their neighbours. It doesn't shoot like a sidearm, look like a sidearm, but somehow is classified as one by the belt. 

# Issue 3, an item killed by the secbelt.

As previously stated, not only is being too broad in what you can hold bad because items which by no means fit the category find their way onto it causing powercreep, munchkinism by stacking items which are sized medium for a reason, 
And not only has this slip-up caused not-so-hypothetical tools to enter it and actively oppose the natural idea the item seems to currently hold in our meta, but also:
It kills other items. 

Why would anyone buy the 2tc syndicate holster, with inbuilt chameleon, if not to save 2 slots off gun storage or keeping it low-profile, when a 4 leather item exists to do the same thing. If normal-sized weapons were to not fit in the belt either, it'd become really apparent really quickly, that storage is CRUCIAL to preparations for combat. Decisions of what should stay and what shouldn't are thrown out the window. 

To prevent it going off-topic from how gunbelts kill sec to how gunbelts shouldn't exist, I will simply end this, by asking, if you have ever seen the holster used. I've seen it on other servers (tg), rarely. Yet here it is nonexistant, as whilst it's unpopular due to a slim advantage of those two slots, it's job is done better by a piece of leather, and that is NOT how work input vs reward output area supposed to work.

I am very much so open to hear ideas of staff, coders, maintainers and players. I will edit the items appropriately. I just need an official version of "what now, boss?" to follow

## Changelog
:cl:
balance: A plethora of /obj/item/gun typecache weapons are gone from the secbelt.
/:cl:


KazooBard:
Apologies for the drop in quality, it is late and I wanted to finish this

SapphoQueer:
Apologies for the drop in quality he says, as he drops a wall of text that makes sense.

KazooBard:
Might I add in the current state i have NOT disabled mining/engi tools off belts. Do let me know if you'd like them go as well.

The same thing goes for the temperature gun, as it's a unique type and barley used - so some sorts of redemption would be nice

sippykot:
if you made the secbelt unable to hold pistols, no one would use it lol

modular_zubbers/code/game/objects/items/storage/belt.dm
```diff
@@ -0,0 +1,27 @@
...
+		/obj/item/melee/baton,
+		/obj/item/radio,
+		/obj/item/reagent_containers/spray/pepper,
+		/obj/item/restraints/handcuffs,
+		/obj/item/restraints/legcuffs/bola,
+		/obj/item/stock_parts/power_store/cell/microfusion, //SKYRAT EDIT ADDITION
+	))
+	atom_storage.open_sound = 'sound/items/handling/holster_open.ogg'
+	atom_storage.open_sound_vary = TRUE
+	atom_storage.rustle_sound = null
```
TealSeer, line 26: This is all redundant copy paste, this whole thing can just be your `cant_hold` line added to the original `/obj/item/storage/belt/security/Initialize()` with a bubber comment

KazooBard: Done

code/game/objects/items/storage/belt.dm
```diff
@@ -335,6 +335,7 @@
 	inhand_icon_state = "security"//Could likely use a better one.
 	worn_icon_state = "security"
 	content_overlays = TRUE
```
TealSeer, line 337: this isn't in `Initialize()`, it won't work in here

TealSeer (approved):
code looks good

YakumoChen:
> if you made the secbelt unable to hold pistols, no one would use it lol

It holds Batons so security always has a reason to use it.
Not even starting to mention the main function of belts are convenient access to job tools (in this case, like handcuffs and various stun weapons)

It's obvious why every ss13 server has secoffs prefer their belt

code/game/objects/items/storage/belt.dm
```diff
@@ -376,6 +376,16 @@
...
 	))
+	// BUBBER EDIT ADDITION BEGIN - Restrict belt items
+	atom_storage.cant_hold = typecacheof(list(
+	/obj/item/gun/ballistic/automatic/rom_flech,
+	/obj/item/gun/syringe,
+	/obj/item/gun/chem,
+	/obj/item/gun/ballistic/automatic/ar,
+	/obj/item/gun/magic/wand,
+	/obj/item/gun/energy/recharge/kinetic_accelerator,
+	/obj/item/gun/energy/plasmacutter))
```
Arturlang, line 387: Honestly I think we should just get rid of /obj/item/gun in general, this specific blacklist feels like it doesn't handle the issue even that much.

YakumoChen: Big agree. Officers don't need a reason to hold a bunch of guns on their belt without a specific holster belt or the naked slot

KazooBard: > Honestly I think we should just get rid of /obj/item/gun in general, this specific blacklist feels like it doesn't handle the issue even that much.

Well I'm happy to hear! I'll do that, then

KazooBard: @Arturlang

plsleavemealon:
So long as I can hold a _sidearm_ like a pistol/revolver on my belt I'm fine with this. I will be sad if we lose the ability to store a revolver, pistol, or standard disabler on there though.

DorfGuy1:
My question is....ages and ages and ages ago, tg had holsters like bay. They attached to your suit and you could remove them from your holster with a hotkey. It was an accessory. Then we got rid of attachements for your jumpsuit, mostly, and somewhere along the line holsters that could be used in your accessory slot got removed too. They were present on some tg downstreams, but not on ours anymore. What ever happened to those, and can they come back? Also so many suits have completely arbritrary restrictions on what you can put in them...has nt never heard of a shoulder sling? There is so much arbritrary suit cant hold this cringe that this will be anti qol. To a degree.

YakumoChen:
> Antags regularly body security, no matter how many guns they have.

Then they don't need to have 7 pistols on their belt if it doesn't matter

> Having 4 pistols to do a new york reload is also funny and should be allowed.

Store them in your backpack then

Arturlang (approved):

DorfGuy1:
> > Antags regularly body security, no matter how many guns they have.
> 
> Then they don't need to have 7 pistols on their belt if it doesn't matter
> 
> > Having 4 pistols to do a new york reload is also funny and should be allowed.
> 
> Store them in your backpack then
> 
> Or use the dedicated holster belt

What about the neck slot holsters. I swear that was a thing once upon a time.

YakumoChen:
> > > Antags regularly body security, no matter how many guns they have.
> > 
> > 
> > Then they don't need to have 7 pistols on their belt if it doesn't matter
> > > Having 4 pistols to do a new york reload is also funny and should be allowed.
> > 
> > 
> > Store them in your backpack then
> > Or use the dedicated holster belt
> 
> What about the neck slot holsters. I swear that was a thing once upon a time.

next slot didn't even exist until relatively recently and even then its only ever been for cosmetics.

Ashrollen:
> > Antags regularly body security, no matter how many guns they have.
> 
> Then they don't need to have 7 pistols on their belt if it doesn't matter
> 
> > Having 4 pistols to do a new york reload is also funny and should be allowed.
> 
> Store them in your backpack then
> 
> Or use the dedicated holster belt

You know your opinion was always most toxic considering you change with dragon that made it so easy to win against sec (Oh yes i still remember your change with cool armor boost button dear). In your place i would rather excuse yourself out of conversation since whatever you spit usually leaves bad aftertaste mr.Igotsomehowfuckedbysecletsnerfthemagain. As was said 7 pistols don't matter if guy can't utilise them but at least guy has fun in storing them, so you got problems with others having fun? Glorious opinion. Store in backpack what? Its a place for my 5 thrownable batons and will always will and stay like this considering right now its easier to just go yolo and throw 5 33% chance of stun batons (that will be countered with single EMP or one MOD addon). Dedicated holster? Mhhhhm dedicated shittier holster that can't do anything, id rather carry sabre on my belt than use such imbalanced way to carry guns especially small arms that have near zero fucking AP. As for your guys standart situation i would suggest to rise skill of antags (that still arent able to read what gear/skill of theirs does what exactly) and stop trying to nerf what exist already in word only.Though you know? Fuck it nerf shit faster more you do such stuff guys faster entire sec base will move to something like ss14 where holster is an interesting convenient item to store not only 1 fucking pistol but also other items.

PR #3679, opened 2025-05-01
ethereal-revive-bugfix

plsleavemealon:
## About The Pull Request
Somehow Ethereals revives were changed back to TG's version. This fixes it back to bubbers version which gives them magical traumas instead of absolute traumas. Thus this is a bugfix, restoring how it originally was for bubber.

## Why It's Good For The Game

Getting a trauma that cannot be removed via any method is exceptionally annoying, and often just leads to the player immediately cryoing. Ethereals cannot control whether they revive via their passive. It is not something they can refuse without just flat out going DNR. A magical trauma is very difficult to remove, but not impossible, thus it is balanced.

## Changelog

:cl:
fix: ethereal revives gives magical trauma instead of permanent
/:cl:

### DIFF
--- a/code/modules/unit_tests/ethereal_revival.dm
+++ b/code/modules/unit_tests/ethereal_revival.dm
@@ -42,7 +42,7 @@
 	TEST_ASSERT(length(current_traumas) == 1, "Ethereal failed to gain trauma when reviving.")
 
 	kill_and_revive(victim, respawn_heart)
-	TEST_ASSERT(length(victim.get_traumas()) == 2, "Ethereal failed to gain additional trauma on second revival.")
+	//TEST_ASSERT(length(victim.get_traumas()) == 2, "Ethereal failed to gain additional trauma on second revival.") //commenting out for now, to allow ethereals to gain a magical trauma instead of permanent one
 
 	instant_crystallise(victim, respawn_heart)
 	victim.heal_and_revive()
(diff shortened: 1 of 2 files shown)


plsleavemealon:
Unsure why it is throwing an error for this. It should be applying the trauma to the individual, but maybe it is doing the A-heal after the trauma is applied?

TealSeer:
It's failing because the revive proc its using in the unit tests can cure up to magic resilience level traumas, so they're getting healed from it. You probably need to modify the unit test to just not check that bit

code/modules/unit_tests/ethereal_revival.dm
```diff
@@ -42,7 +42,7 @@
 	TEST_ASSERT(length(current_traumas) == 1, "Ethereal failed to gain trauma when reviving.")
 
 	kill_and_revive(victim, respawn_heart)
-	TEST_ASSERT(length(victim.get_traumas()) == 2, "Ethereal failed to gain additional trauma on second revival.")
+	//TEST_ASSERT(length(victim.get_traumas()) == 2, "Ethereal failed to gain additional trauma on second revival.") //commenting out for now, to allow ethereals to gain a magical trauma instead of permanent one
```
Shadow-Quill, line 45:
```suggestion
	//TEST_ASSERT(length(victim.get_traumas()) == 2, "Ethereal failed to gain additional trauma on second revival.") //Bubber Edit - commenting out for now, to allow ethereals to gain a magical trauma instead of permanent one
```

Shadow-Quill (changes requested):

Shadow-Quill (approved):

PR #3703, opened 2025-05-04
blinking is now config toggleable

Arturlang:
## About The Pull Request
blinking is now config toggleable for perf reasons, may be chugging quite a bit, kept support to toggle it ingame as well

## Why It's Good For The Game
lag fixes i hope

## Proof Of Testing

<details>

</details>

## Changelog

:cl:
config: added config options for toggling blinking
/:cl:

### DIFF
--- a/code/modules/surgery/organs/internal/eyes/_eyes.dm
+++ b/code/modules/surgery/organs/internal/eyes/_eyes.dm
@@ -461,6 +461,8 @@
 /// Animates one eyelid at a time, thanks BYOND and thanks animation chains
 /obj/item/organ/eyes/proc/animate_eyelid(obj/effect/abstract/eyelid_effect/eyelid, mob/living/carbon/human/parent, sync_blinking = TRUE, list/anim_times = null)
 	. = list()
+	if(!CONFIG_GET(flag/blinking)) return // BUBBER EDIT - CONFIG BLINKING
+
 	var/prevent_loops = HAS_TRAIT(parent, TRAIT_PREVENT_BLINK_LOOPS)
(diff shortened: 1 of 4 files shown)


code/modules/surgery/organs/internal/eyes/_eyes.dm
```diff
@@ -461,6 +461,8 @@
 /// Animates one eyelid at a time, thanks BYOND and thanks animation chains
 /obj/item/organ/eyes/proc/animate_eyelid(obj/effect/abstract/eyelid_effect/eyelid, mob/living/carbon/human/parent, sync_blinking = TRUE, list/anim_times = null)
 	. = list()
+	if(!CONFIG_GET(flag/blinking)) return // BUBBER EDIT - CONFIG BLINKING
```
TealSeer, line 464: might be better to put this below in `animate_eyelids`, will just mean it gets interrupted a little earlier in the call stack

SapphoQueer:
Think upstream'd accept this?

NargaInDisguise:
> Think upstream'd accept this?

They wouldn't let lizards choose to not have green blood, so probably not. They don't like adding prefs for anything.

sippykot:
can we pick between normal and async blinking?

NargaInDisguise:
> can we pick between normal and async blinking?

Also this, let people choose what blink (none, normal, async) they use, that'd be nice

code/modules/surgery/organs/internal/eyes/_eyes.dm
Arturlang: I just realized that this also breaks rapid eyeblinks ugh

PR #3713, opened 2025-05-05
Adds gloryhole portals and stuck in wall content.

SpaceCatSS13:
## About The Pull Request

Adds the LustWish portal bore to the vendor, an item that allows you to connect two locations via a portal that can act either as a gloryhole or a route for stuck in wall content. These portals allow for remote scenes using subtler or interact menu verbs.

Portal2
Portal1V2 

## Why It's Good For The Game

I've heard requests for both portal play and gloryhole content, this scratches both itches sufficiently in my opinion

## Proof Of Testing

<details>
<summary>
Screenshots/Videos</summary>

Both modes pictured

Portal3V2

</details>

## Changelog

:cl:
add: The Lustwish Portal Bore has been added, your lower half can now be in dorms while your head performs your job
/:cl:


Vanilla1040:
this needs the god against nature lable 100%

Adrian16199:
Yes... but also it would be funny tbh if you could do the "im stuck" meme if the sprite suddenly was halved from waist up. 
Yes im givin ideas because its funny.

SovietSpaghetti:
can we not

AstrumLernaean:
Based

BurgerLUA:
I should be able to use a banhammer to smack a portal'd part in case someone eventually abuses this.

SovietSpaghetti:
Okay, elaborating a bit further on what I meant. I don't hate the kink side of this, but I have concerns considering the fact people can just leave these around in public for anyone to use. I get that the whole point is public free use or whatever, but I would really prefer not to have random glory holes/portal entries left around the station as has been in the past when this kind of stuff has been implemented on other codebases. Even a disclaimer or something when examining the item saying "please do not leave these things out in public places and expose everyone to your kink" would be sufficient.

BurgerLUA:
> Okay, elaborating a bit further on what I meant. I don't hate the kink side of this, but I have concerns considering the fact people can just leave these around in public for anyone to use. I get that the whole point is public free use or whatever, but I would really prefer not to have random glory holes/portal entries left around the station as has been in the past when this kind of stuff has been implemented on other codebases. Even a disclaimer or something when examining the item saying "please do not leave these things out in public places and expose everyone to your kink" would be sufficient.

Current rules prevent this. If not, I will complain enough to have those rules covered.

Frites01:
Definitely needs to be a feature
Glory hole

Aroliacue:
This is incredible, thankyou coderman

Only true bubboids will merge this PR

lessthnthree:
This is the Wallening we deserve.

lessthnthree (approved):

lessthnthree (approved):
This is the Wallening we deserve.

SpaceCatSS13:
Known bugs that I'm working on:

- Underwear/moth wings randomly re-appear on the head side of a portal - cannot replicate
- Some emotes restore parts of the body on the head end

Kepteyn:
Firstly, thank you so much for making this, it's great! I've had a play with it, and it was quite fun and works pretty well!


As for some feedback...

1) both my partner and I kept accidentally sending the subtler to the regular 1 tile radius and thus did not recieve it. We caught this, but given the inability to see the other person, it might be tricky to tell if this has happened normally. I would suggest having it so that, if there's a portal you can interact with, that the first option on the list be the portal one, rather than the regular radius option, so hitting enter and enter again sends it through the portal instead. In occassions where there is no portal, the default would still be the regular one.

2) It would be nice for there to be a way to deconstruct it, possibly by a certain click combination to take it down. This might give you back the borer so you can replace the portals. This would both allow you to clean up after a scene, correct a mistakenly placed portal, and take them down if people put them in dumb places (an issue that some people have mentioned here).

3) It'd also be nice to have the option for it to appear as just a hole, as opposed to a portal. Perhaps a toggleable option, allowing you to change appearances? This is more of something I'd personally like and think would be cool, rather than major feedback though - I'd rather a hole to a portal.

For those last two, I would suggest making it so that right-clicking on the portal, rather than just switching gloryhole and stuck-in-wall modes, opens up a menu like toggling arousal currently does, from which you can choose to change mode, deconstruct, or change appearance.


Again, a massive thank you for making this!

Snajper2022:
I noticed that while using a portal while being in the stuck mode. Any changes to the expose genitalia caused the character full body to be exposed fully.

It would also be nice to have the portals cyborg friendly if possible.

Adrian16199:
The only thing I could add to this PR is proposin that the "stuck in wall mode" would be bit different.

Make the lower part of body cut at the hips. Its bit weird seeing only a head pop out from the other side or how it just looks like someone has their head stuck.

Thats my initial nitpick as soon as i even tried it to just find out how it looks the "stuck in wall" mode.

SpaceCatSS13:
>Snajper2022

> I noticed that while using a portal while being in the stuck mode. Any changes to the expose genitalia caused the character full body to be exposed fully.

I'll have a look into this, it didn't come up in testing I've got some ideas how this might have happened

> It would also be nice to have the portals cyborg friendly if possible.

I would be really nice but without creating different split point for each model it would look really poor 


>Adrian16199

> Make the lower part of body cut at the hips. Its bit weird seeing only a head pop out from the other side or how it just looks like someone has their head stuck.

I did want to do this however it significantly increases the complexity and I was not confident I would be able to get it working in a time reasonable enough for me to actually finish this PR

> Thats my initial nitpick as soon as i even tried it to just find out how it looks the "stuck in wall" mode.

I'll add something to the description detailing this soon.

Krypandenej:
attempting to use an oversized character in the stuck in wall mode may result in: 
image

I do wish the larger sprite size for genitals also transferred, specifically for gloryhole mode!

SpaceCatSS13:
> attempting to use an oversized character in the stuck in wall mode may result in
> I do wish the larger sprite size for genitals also transferred, specifically for gloryhole mode!

I'm aware of this issue and I know that its not exclusive to the portals, I'm going to look into a potential fix for this in the coming days.

modular_zubbers/code/game/objects/structures/lewd_portals.dm
```diff
@@ -0,0 +1,447 @@
...
+			eye_organ.refresh(call_update = FALSE)
+			current_mob.overlays_standing[BODY_LAYER] += eye_organ.generate_body_overlay(current_mob)
+			current_mob.apply_overlay(BODY_LAYER)
+
+/obj/structure/lewd_portal/post_unbuckle_mob(mob/living/unbuckled_mob)
+	UnregisterSignal(current_mob, list(COMSIG_MOB_POST_EQUIP, COMSIG_HUMAN_UNEQUIPPED_ITEM, COMSIG_HUMAN_TOGGLE_UNDERWEAR, COMSIG_MOB_HANDCUFFED, COMSIG_MOB_EMOTE, COMSIG_EMOTE_OVERLAY_EXPIRE, COMSIG_HUMAN_ADJUST_AROUSAL, COMSIG_HUMAN_TOGGLE_GENITALS))
+	visible_message("[current_mob] exits the [src]")
+	current_mob = null
+	mob_scale_manager = null
+	qdel(relayed_body)
```
TealSeer, line 192:
```suggestion
	QDEL_NULL(relayed_body)
```
QDEL_NULL also sets the var to null after calling qdel, we want this because otherwise the reference to it stays in `relayed_body` and it can never fully clean up

Elli-Skala:
WARDEN OUT OF BRIG

PR #3729, opened 2025-05-06
Remove Skyrat lights off override

lessthnthree:
## About The Pull Request

Removes a Skyrat add that turns off lights by default in rooms that have light switches

## Why It's Good For The Game

There's no reason to make you hunt for the lights shift start. It also means the janitor can't tell what rooms have broken lights at a glance unless someone hit the switch.

## Changelog

:cl: LT3
del: Removed lights being turned off by default in various locations
/:cl:


IgiariValkyr:
Did this still leave the lights off map helper?

lessthnthree:
> Did this still leave the lights off map helper?

No map helpers were removed, so presumably.

Issue #3740, opened 2025-05-07
Health Scanners don't work on NIF implanted individuals


DuffCreeper:
### Issue Summary:

Scanning someone who is implanted with a NIF and using a health scanner on them or themselves will provide no results

Scanning someone who isn't implanted with a NIF will provide a read out as expected

### Round ID:

8102

### Reproduction Steps:

Grab a health scanner/advanced health scanner
Scan someone who is implanted with a NIF
Get no results

Shadow-Quill:
relevant runtime
```[2025-05-07 15:56:09.485] RUNTIME: runtime error: type mismatch: Standard Type NIF (/obj/item/organ/cyberimp/brain/nif/standard) += "<img class=\' icon icon-base_n..."
 - proc name: healthscan (/proc/healthscan)
 -   source file: code/game/objects/items/devices/scanners/health_analyzer.dm,338
 -   usr: Name (/mob/living/carbon/human)
 -   src: null
 -   usr.loc: the floor (150,93,2) (/turf/open/floor/iron/smooth_edge)
 -   call stack:
 - healthscan(JohnSmith (/mob/living/carbon/human), JohnSmith (/mob/living/carbon/human), 1, 0, 1)
 - the health analyzer (/obj/item/healthanalyzer): interact with atom(Name (/mob/living/carbon/human), JohnSmith (/mob/living/carbon/human), /list (/list))
 - JohnSmith (/mob/living/carbon/human): base item interaction(JohnSmith (/mob/living/carbon/human), the health analyzer (/obj/item/healthanalyzer), /list (/list))
 - the health analyzer (/obj/item/healthanalyzer): melee attack chain(JohnSmith (/mob/living/carbon/human), JohnSmith (/mob/living/carbon/human), /list (/list))
 - JohnSmith (/mob/living/carbon/human): ClickOn(JohnSmith (/mob/living/carbon/human), "icon-x=21;icon-y=19;left=1;but...")
 - JohnSmith (/mob/living/carbon/human): Click(the floor (150,93,2) (/turf/open/floor/iron/smooth_edge), "mapwindow.map", "icon-x=21;icon-y=19;left=1;but...")
 - DuffCreeper (/client): Click(JohnSmith (/mob/living/carbon/human), the floor (150,93,2) (/turf/open/floor/iron/smooth_edge), "mapwindow.map", "icon-x=21;icon-y=19;left=1;but...")```

Shadow-Quill:
[...]

PR #3741, opened 2025-05-07
Fix health analyzer runtime on cybernetic organs (#90900)

Shadow-Quill:
should fix https://github.com/Bubberstation/Bubberstation/issues/3740


lessthnthree:
Merged into master.

PR #3768, opened 2025-05-09
More Dullahan Cyborg models

Ziiro:
## About The Pull Request

This adds a new set of Dullahan Cyborg Models with four legs instead of two.

## Why It's Good For The Game

I saw these over on Nova ( https://github.com/NovaSector/NovaSector/pull/5399 ) and I thought  they were pretty cool. They actually differ slightly in the upper body from the standard Dullahan models as well, so they're not just a slight modification of the baseline sprite.

## Proof Of Testing
Spawned in a bunch, transforming both into normal models as well as using chameleon modules. The overlays for having the panel open as well as cell in/out are all present. Also tested the hat offset.

dreamseeker_9sgoLjXq7n

## Changelog
:cl:
add: More Dullahan cyborg models have been added!
/:cl:


Ziiro:
This is a draft because I'm trying to find or create one for Secborgs and research borgs so it's all of the. If it's okay those are missing then this is probably ready.

Ziiro:
Found the sprites, but they aren't quite ready for prime time yet as they lack a few states. I'll call this one complete, and I the SecBorg and SciBorg versions can be added at a later date.

ArrisFairburne:
> I'll call this one complete, and I the SecBorg and SciBorg versions can be added at a later date.

Given that we killed secborgs (again) I don't think that y'all will ever need to make a sec sprite for these. Maybe for the (probably) very rare instance that a min decides to turn HOS AI on manually..?

Shadow-Quill (approved):
These look pretty nice.

PR #3121, opened 2025-02-18
Precise syringe injections

sippykot:
## About The Pull Request

lets you pick lower injection amounts from syringes

## Why It's Good For The Game

No one uses syringes because of how much better hypos are, and being able to precisely dose medications makes them a bit more usable.

## Proof Of Testing

It compiled, i dont expect any problems from it, considering its just a simple edit... scream at me and beat me with a rolled newspaper if its broken

## Changelog

:cl:
add: Syringes have been fatfinger-proofed, letting you inject in 1u and 2u increments
/:cl:

PR #3806, opened 2025-05-13
The droppering

sippykot:
## About The Pull Request

Hey, so, remember #3121 ? I was meaning to make a 2nd part to this concerning droppers so they aren't half-useless right now.
Could re-enable pipettes that got removed during a _barber shop tweak pr_ but eh. Less bloat.

## Why It's Good For The Game

gives the dropper something to do, makes it so you can split reagents down to 0.1u without having to resort do dumb stuff like printing 0.1u pills or diluting them, splitting and then deleting the dilutant with a chemmaster...

## Proof Of Testing

simple overwrite

## Changelog
:cl:
add: droppers now can transfer as little as 0.1u, 0.2u and 0.5u reagents at a time
/:cl:


sippykot:
ah yes, the famous map error for a pr tweaking an object property...

Shadow-Quill (approved):
You can ignore the map CI error, it's unrelated.

PR #3810, opened 2025-05-13
TC change revert, as god intended

MajManatee:
## About The Pull Request

Does something that should have been reverted long ago...
Sets the TC traitors get back to 20, from 25

## Why It's Good For The Game

Nothing is balanced around having 25 TC. 

We have discounts to allow for variation in your plan, and giving you the opportunity to play with things you wouldn't normally do. Having 25 TC was intended to allow for gimmicks, but that is not how the system is used. People that want to do gimmicks have just made opfors, or ahelped, and gotten what they wanted that way.  The current system just allows people to be more powerful than intended, and we keep seeing the same "GIMICKS" done all the time (Desword, Modsuit, Glue)

The change itself was made improperly. None of the costs for gear were changed (See crates still costing 20, Kits cost 20). The uplink spawn location cost was never adjusted to account for having more, so you just lose 5 TC for having your uplink spawn in your head. 

In the end, I don't really care since I just buy an emag and finish my objectives in 10 seconds, then goof off, but this PR is open to encourage POLITE discussion, and try to fix this nonsense one way or another.

## Proof Of Testing

<details>

num change

</details>

## Changelog

:cl:
balance: Reverts the TC change from long ago, back to 20.
/:cl:

### DIFF
--- a/code/__DEFINES/antagonists.dm
+++ b/code/__DEFINES/antagonists.dm
@@ -333,7 +333,7 @@ GLOBAL_LIST_INIT(human_invader_antagonists, list(
 #define MAROON_PROB 30
 
 /// How many telecrystals a normal traitor starts with
-#define TELECRYSTALS_DEFAULT 25 // BUBBER EDIT - GIMMICK TRAITOR
+#define TELECRYSTALS_DEFAULT 20
 /// How many telecrystals mapper/admin only "precharged" uplink implant
 #define TELECRYSTALS_PRELOADED_IMPLANT 10
 /// The normal cost of an uplink implant; used for calcuating how many


amenitykitten:
while i agree that having 25 TC is a bit much, this pr feels like it was made in bad faith and i do not like how it is worded.

regardless, agree with the change.

sippykot:
Yeah, uhhh... This is a very bad change. Why the hell are people trying to nerf traitor even more, it's already the weakest and worst antag that revolves around having to powergame and grind because you only get a small amount of non-renewable resources.
We have rounds that are... 2-3 times longer than tg. Giving traitor the same amount of TCs for the time is not a good idea, considering a good bit of traitor items are consumable. Saying that having traitor be piss-weak is ok because you can just beg admins for free stuff is not a solution to anything, and giving people 5 tc less is not going to stop them from rampaging across the station if they want to, since to do that you have to powergame/grind anyways. It's just making it worse for people who want to play traitor without sweating.

PR #3818, opened 2025-05-14
Security Bodycams [WIP]

Ziiro:
## About The Pull Request

This ports bodycams from fulpstation: https://github.com/fulpstation/fulpstation. They can be ordered from cargo, attached to outerwear, and destroyed with EMPs.
## Why It's Good For The Game

Gives something for the Warden/AI to keep an eye on - to keep track of individual officers. These are useful, however they can be easily deleted with a single EMP so they're completely counter-able by most antags. It should lead to some fun interactions.

## Proof Of Testing
<details>
doesn't yet actually that's why it's WIP. working on it. it's a port but I have to retrofit some stuff.

</details>

## Changelog
:cl:
add: Bodycams have been added to cargo for security to order! Expensive and fragile to EMPs, but very useful on keeping tabs on officers!
/:cl:


Ziiro:
So they attach, can be uninstalled, swiped with IDs to turn off/on. The issue is that it doesn't seem to register for *actually* showing camera vision radius and I can't figure out why.

Test case: Carbon A puts bodycam on sec armor. Walks into maint. AI can track them via the cams (they show up on the track list) but it's still overlaid with static.

BurgerLUA:
> So they attach, can be uninstalled, swiped with IDs to turn off/on. The issue is that it doesn't seem to register for *actually* showing camera vision radius and I can't figure out why.
> 
> 
> 
> Test case: Carbon A puts bodycam on sec armor. Walks into maint. AI can track them via the cams (they show up on the track list) but it's still overlaid with static. 

 
Might be because of how the body camera is attached. Try making the body camera attached to the suit instead of the upgrade item.

If that doesn't work, it is likely because it needs to be attached to the mob. Maybe some other code exists that does this (but runners, maybe)

Rykka-Stormheart:
For realism they should buzz (message in chat)/beep (audible sound) every x minutes (configurable to turn on/off the buzz or beep), just like IRL bodycams) when they're turned on. Helps reduce leaving them on all the time. <: 

(The IRL purpose of that is to notify/remind officers that the bodycam is on, and it's configurable by the department to buzz or beep every x minutes, along with a visible indicator light on the camera housing.)

Ziiro:
That's a fair point. There's a related PR for fulp that creates a jammer that works on radios and cameras, but even then that would only be one antag type that would have access (traitor) - so going to close this up for now. Thank you!

sippykot:
> That's a fair point. There's a related PR for fulp that creates a jammer that works on radios and cameras, but even then that would only be one antag type that would have access (traitor) - so going to close this up for now. Thank you!

@Ziiro , we have craftable jammers from tg though? last time i checked - you could just make sure these works for bodycams also

BurgerLUA:
Damn wish this wasn't closed :(

PR #3867, opened 2025-05-21
Quirk cost rebalance

plsleavemealon:
## About The Pull Request

Rebalances the point costs for numerous quirks

## Why It's Good For The Game

Apathetic 4:2) This quirk is actually more of a negative quirk then a positive one. There are so many more positive mood bonuses in the game that are easy to achieve than negative ones, that taking this quirk actually ends up penalizing you more and makes managing your mood arguably harder than without it.

Light step 4:2) A very simple quirk that does remarkably little. Decreases damage from stepping on sharp things (barefoot stepping on glass) and caltrops. Also decreases the chance your clothing gets bloody from stepping on blood, which is barely a factor in things. Both of these effects are so incredibly minor that you usually wont even notice it. The final effect is decreasing your footstep noise, this...doesn't actually appear to do anything at all. Definitely overpriced.

Chip Connector 4:2) This quirk is arguably a negative quirk disguised as a positive one, as if you have this quirk and are EMPed your skillchips will fall out. This means that you could potentially lose extremely difficult to replace skillchips such as the chef skillchip for kitchen CQC, roboticists skillchip, etc. Not only should this quirk be reduced in cost, but it should also give a skillchip alongside it.

Poster Boy 4:2) This quirk does give a solid mood bonus from your posters, but these are 1-shot items so if they get destroyed they are gone. And if you put up your poster in a bad spot...tough luck. Simply overpriced for what it does.

Signer 4:3) It is strange to me that signer costs more then telepath, when telepath gives you perfect stealth communication to someone in LOS. Though I suppose signer is able to communicate easier on the radio thanks to translator gloves.

Strong Stomach 4:2) Does anyone take this quirk? Anyone at all? It has its niche I suppose, but its a niche that doesn't seem to come up all that often. I suppose it does reduce the stun time from vomiting, but again...that doesn't really happen that often.

Tagger 4:2) An esoteric quirk, as it gives a mechanical benefit to improving room quality...but this mechanic doesn't seem to be fully implemented or at least not in a proper manner. Its real benefit is from doubling the uses from your art supplies and that frankly, isn't really a powerful benefit justifying 4 point cost.

Bilingual 4:2) Honestly, surprisingly few languages are used ingame with the lions share of people that use a secondary language often speaking Draconian, with most people just having these traits purely for roleplay as they do not mechanically do much. Which is somewhat surprising to me. Lowering the cost of this quirk could allow for more language expression. Of them I am most tentative about this one

Linguist 4:2) Same reason as bilingual.

Spacer 5:4) This quirk is usually more of a hindrance then a boon. It does grant a small stamina regen bonus when it actually is being helpful. If our anti-nausea medications didn't suck then I'd be much more hesitant to promote reducing its cost.

Drunken Resilience 8:7) This quirk should be blacklisted from any synthetic species, as it is completely nonfunctional for any species with non-organic limbs. The healing it provides is very miniscule, but the biggest reason why I advocate a point cost reduction is because it only works for organic bodyparts. If it worked for all bodyparts, regardless of type then at 8 points it is justified. Also, its description should be modified to warn that it only works on organic bodyparts. 

## Changelog

:cl:
balance: rebalanced many positive quirk costs
/:cl:


ReturnToZender:
Seems like most, if not all of these changes can be taken upstream rather than overriden

The "why it's good for the game" section seems to cover what the PR is doing rather than what the benefit of adjusting point costs here is
I also don't know why it's suggesting changes that aren't included in this PR

Arturlang (approved):
Code is good, thank you for modularizing it

Arturlang:
@ReturnToZender Might be very hard to get quirk rebalances to /TG/ plus some of these changes are based on our specific quirks

plsleavemealon:
@ReturnToZender the reason why the suggestions were included is both to justify the cost changes, and also to provide ideas for others on improvement.

If I knew how to implement these suggestions I would do so, and for something like Drunken Resilience I'd want to discuss further whether or not it should be restoring all limbs, or if it should be just focused on the organic ones. I'd be able to easily tweak it to have it heal all limbs regardless of synthetic or not, as I can just remove the check from it for organic limbs...though I am not sure how to modularize it if I were to do that tweak.

If it were to limit it from all non-organic species instead of making it heal all limbs, I am not actually sure which ones are considered non-organic other then the super obvious ones (IE the synth variants), and I am not certain how to blacklist species from quirks.

plsleavemealon:
As for why its good for the game and why it includes what the PR is doing...its just a formatting thing. It was easier to include it there. It was either basically axe the "why its good for the game" section, or the "about the PR" section, and I chose the latter as I felt the explanation for why it was good for the game took up the largest part of it.

PR #3889, opened 2025-05-24
Agile Mechs

ArrisFairburne:
## About The Pull Request
Adds a new ultra-light combat exosuit chassis type: the McCloud mech.
Features/Todo:
- [x] Fragile speedster stats. Fastest movement speed among exosuits and outspeeds players normally. Lowest armor and integrity stats among mechs. High energy use from movement forces the user to regularly return to recharge points.
- [x] Two modes: bipedal and fighter jet. Biped mode makes the mech function like most other mechs. Jet mode can only be switched to while in low-pressure environments (space/lavaland); this subjects it to momentum-based movement, but allows the mech to move very fast. 
- [x] (Jet mode only) Crashing into objects will damage both the other object and the mech.
- [x] Scope toggle that allows the user to see slightly ahead of the direction their mech faces. This is just so that you can move closer to max speed without constantly bumping things, really.
- [x] (Biped mode only) Getting hit by disabler shots will temporarily slow down the mech.
- [x] (Jet mode only) Momentum while moving is added to any bullets/lasers/etc. it shoots.
- [x] Small explosion when the mech loses its entire integrity meter. This significantly damages the pilot, but not enough to instantly crit them from full HP.
- [x] Mech sprites
- [x] Related mech chassis parts/circuits/research
- [x] Make pressure system check independently of mech movement
- [x] Bugfix to scoping to make it auto-unscope when the player exits the mech
- [x] Why is the jet mode switch borked, fix that
- [x] Less-annoying sound effect for mech walking since it's so rapid
- [x] Sound effect for mode swapping
- [x] Properly implement mech collision damage to victim
- [x] Mech subtypes

## Why It's Good For The Game

I submitted a hackmd with this as a feature and it was approved. But to explain here too, I think this would be great for the game because "space fighter jet pilot" and "mech that transforms" are both common science fantasy tropes that are not used at all in SS13. Adding this to the game would enable more storytelling for daring pilots and space-flying mercenaries, in addition to filling a "gimmickless lightweight war mech" niche that isn't really filled in the game as-is.

## Proof Of Testing
<details>

</details>

## Changelog
:cl:
add: Added a new ultra-light mech chassis type, the McCloud, as well as related research and mech construction components
/:cl:


ArrisFairburne:
> I like this idea a lot!! I especially like it being affected by disabler fire- maybe the electrical impulse disrupts internal systems and forces it to slow down and recalibrate.

😳 thank youuu

> How does it handle transitioning from low pressure to high pressure while still in fighter mode? I assume it just forces it back to bipedal. It could be cool to have it skid along the ground for 1/2/3 tiles (depending on its speed) when it transitions, either if forced to by pressure or if the pilot does it themselves while a floor is under them.

Yeah right now if it enters anything above lavaland pressure, it will switch to biped form. I like the skidding idea, but I do not want to code that, the newtonian move code scares me and I don't want to touch that short of doing the bare minimum for my thing.

Also I need to to-do a rewrite the pressure check to be independent of the move code, right now you can enter normal atmosphere areas and still keep the crazy fast momentum so long as you don't press a movement key.

> Maybe for turning from bipedal to jet requires a run-up? Like, you press the button, and then you need to move 3 tiles in a straight line to actually transform? It could also be interesting if the fighter jet mode always has to keep moving and can't stand still, but that might be annoying.

I like the concept of having to get a run-up to build momentum then shift into jet mode, but having it as a requirement for anytime you want jet mode would be a severe limitation. Almost every transition from indoors to space is checkpointed with airlocks, so you can't really have the straight runway takeoff from inside to outside. You'd need to holofan the airlock, bolt them open, and only then be allowed to runway up, or leave the station to find a sufficient space to do this on, and I'm not sure I want to saddle the mode (which is already only usable in specific context) with an activation requirement. 

And, yeah, jet fighter mode having constant movement would definitely be a nuisance. It's probably fine for this to be able to hover around -- real life fighter jets need to move forward cause that's how they generate lift to avoid falling, whereas space jets probably wouldn't need to consider gravity.

> Sorry for the idea-guy comment, I iust think this idea is really cool 😅
> 
> Maybe upstream would enjoy this too? I'm not sure. Either way, this sounds kickass, can't wait to see it!!

Could be good for TG too, yeah -- would be nice to modify the TG code to include my changes instead of adding bubber edits. Most of the reason why I'm doing it here is cause I don't like using my GBP lol. The other thing is that I noticed in the code that the design direction for mechs is intentionally very slow and sturdy and I'm not sure if that's a must-do for them, since "fast low-armor mech that's balanced around having to retreat often for recharging" is basically the exact opposite of how they're otherwise designed.

MajManatee:
I just think "McCloud" is a silly name, that's my only comment.

PR #3912, opened 2025-05-27
Adds AI-Uplink brain to the techweb as a late game technology

Ziiro:
## About The Pull Request

This adds the AI-Uplink brain to the techweb for a steep cost, and high pre-reqs.

## Why It's Good For The Game

This is a very fun and interesting piece of tech, but having it currently locked to only appear on one ruin on one map (Out of 13!) behind a megafauna feels like it makes it **too** rare. Adding it to the techweb, but requiring both alien techs, and illegal research, as well as bananaium (obtainable through space ruins, or very expensive through cargo) should still limit it and keep it uncommon.

## Proof Of Testing

Went through and spun up a test and actually went through the research steps to get it to make sure it's accessible. Illegal tech, alien tech, ordinance, etc. (but I forgot to take a screenshot after I finished all the research, but it does work!)

image

## Changelog
:cl:
balance: The AI uplink has been added to the tech web as a difficult-to-get research.
/:cl:

### DIFF
--- a/modular_zubbers/code/__DEFINES/techweb_nodes.dm
+++ b/modular_zubbers/code/__DEFINES/techweb_nodes.dm
@@ -1,5 +1,6 @@
 //Onstation Tech
 #define TECHWEB_NODE_EMPATHY_IMPLANT "empathy_implant"
+#define TECHWEB_NODE_AI_UPLINK_BRAIN "ai_uplink_brain"
 
 //Offstation Syndicate Tech
 #define TECHWEB_NODE_INTERDYNE "interdyne_tech"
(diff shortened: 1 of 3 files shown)


NargaInDisguise:
NONONONO
ABSOLUTELY _**NOT**_

This thing should NEVER EVER be researchable.
This gives AIs WAY too much power. This item is fundamentally gamebreaking and even having it on icebox AT ALL is excessive. It genuinely never should've been added to the game.

Shadow-Quill:
After chatting with another maintainer, we don't agree with the balance shift this would cause.

Everything that's required to unlock the uplink brain (as-is in the PR) is able to be done with cyborgs.
Find a smuggler satchel for illegal tech,
sciborg(?) does xenoarch for alien tech, 
mediborg preps a valid body for the brain, 
and 5 bananium is only 50k credits from cargo.

That, in turn for giving the all-access near-omnipresent AI hands, which negates one of the biggest restrictions AIs have — relying on other people.

PR #3919, opened 2025-05-28
Adds partial language integration for our language menu + makes partial translation more consistant

nikothedude:
## About The Pull Request

Majority of code taken from https://github.com/DopplerShift13/DopplerShift/pull/588

Title.

Pretty simple behavior. Adds a 0-100% slider to language prefs. The lower it is, the less you understand of the language. Additionally, partial translation can never get less likely based on the commonness of the word - only more common.

Of note is that jobs do in fact still enforce language requirements with this. The minimum understanding for most important jobs is 90% (this is stupid and if it was up to me I'd kill all language requirements but Oh Well)
## Why It's Good For The Game

Partial languages are cool. Also, in testing on doppler, partial langs proved to be a lot less predictable and seem a lot less useful without the max() around translate prob.
## Proof Of Testing
<details>

image

image

</details>

## Changelog
:cl:
add: Language menu now has a input for partial languages
qol: Partial language translation is now never less likely based on commonness of the phrase
/:cl:


SapphoQueer:
This is amazing given humans cant take common second language. Love it.

nikothedude:
> might have performance implications

how so? what part of this?

nevimer:
> > might have performance implications
> 
> how so? what part of this?

not so, but indirectly. It will have to make a unique runechat for each hearer. AFAIK

TealSeer:
> For the record, if anyone wants to TM this PR in the future, _if the bug happens again_, the only precaution that has to be taken next round is telling players to clear their languages
> 
> The mutant colors appear not to stick, and while languages are put in a broken format, clearing languages resets them

Sorry I've been meaning to get around to investigating this properly, because I'm not sure why it broke on live only

nikothedude:
> Possible clue to what happened on live by the by, in some people's save files, I see their language pref entry was updated to the new format with the separate `flags` and `knowledge` entries, but the version of their save file didn't bump from 8 to 9. No idea how that would play into colors breaking but it's something (and probably points to a problem with migration code more than this specific PR)

Very weird

nikothedude:
still works, still available to tm. still dont know why that happeend

PR #3858, opened 2025-05-19
Minor Protean Adjustments & Fixes

Ziiro:
## About The Pull Request

Making a few things more obvious/readable with Proteans, as well as making some small adjustments here and there:

- Proteans can now regenerate their nanite slurry blood, but this makes them very resource hungry to do so
- Visible messages for protean repairs, retreating/emerging
- Proteans that are in mod suit form are now immune to pressure/hot/cold to prevent death/damage from the vacuum of space.

## Why It's Good For The Game

Regarding new visible messages: The readability of repairing/helping a new race for both the roboticists/doctors healing them, as well as the players playing them.

Regarding the blood changes: Currently, although Proteans have nanite slurry for blood, they have no way to actually create more. This allows them to create more, at an exchange rate slightly better than a chemists making it (1 Iron sheet + 1 Gold bar = 20u of Slurry). **I'm definitely open to adjustments in numbers here**, but I'm trying to balance the fact they can only store 10 sheets of metal at a time, and blood volume is 560u. So although having a harsher exchange rate might be good in terms of keeping chemists relevant for nanite slurry, it would really suck to have to eat 40 iron to regenerate blood.

Regarding idle iron consumption adjustment: Proteans are very hungry for material if they need repairs, and I think tuning that dial back just a bit for if they're idle/not needing repairs makes downtime less punishing.

Suit_traits: this is a bug fix to prevent proteans from dying when retracted into their suit, in space. Seemed a bit silly for a space-capable suit to be harmed by space.

## Proof Of Testing

<details>
<summary>Screenshots/Videos</summary>
WIP. Still trying to get a few things to work. Namely, the spans on a reviving/revived Protean. I'd also really like to solve the issue where a protean never fully "die" dies releasing them to the ghost realm, but I'm still spinning on that.
</details>

## Changelog

:cl:
add: Proteans can now regenerate their nanite slurry blood by eating materials.
add: Proteans have had a few helpful messages added to their revival mechanics and objects.
balance: Protean idle material consumption has been decreased
fix: Proteans in suit form no longer take pressure or temperature damage.
spellcheck: Cleaned up a few things related to Proteans
/:cl:

PR #3928, opened 2025-05-29
Proteans can now be sacrificed by heretics

KazooBard:
## About The Pull Request

Previously: Protean becomes modsuit on death. Is unsacrificable.

With the change: Protean in modsuit can be sacrificed, causing them to revive, reform, get cuffed and then sent as would be a typical carbon.

## Why It's Good For The Game

Not being able to be sacrificed is not an intended game mechanic, is a bug and unfairly robs heretics out of their hard work.

## Proof Of Testing

obraz
I am not sure still if all is good and for whether 1 line is not that big of a deal to not have mondularized or if I should modularize it, too
## Changelog

fix: Proteans can now be sacrificed by heretics in modsuit form
/:cl:

### DIFF
--- a/code/modules/antagonists/heretic/knowledge/sacrifice_knowledge/sacrifice_knowledge.dm
+++ b/code/modules/antagonists/heretic/knowledge/sacrifice_knowledge/sacrifice_knowledge.dm
@@ -89,7 +89,11 @@
 		// Otherwise if it's neither a target nor a cultist, remove it
 		else if(!(sacrifice in heretic_datum.sac_targets) && !IS_CULTIST(sacrifice))
 			atoms -= sacrifice
-
+	//BUBBER EDIT, NO HIDING FROM THE DARK GODS!
+	for(var/obj/item/mod/control/pre_equipped/protean/protean_item in atoms)
+		for(var/mob/living/carbon/human/hiding_target in protean_item.contents)
+			atoms += hiding_target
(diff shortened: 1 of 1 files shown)


SovietSpaghetti:
And what if they die in the minigame? Do they get sent back to the station in a random room as a MODsuit unable to use comms and appearing fine on sensors? What happens with the cursed organs?

projectkepler-ru:
> And what if they die in the minigame? Do they get sent back to the station in a random room as a MODsuit unable to use comms and appearing fine on sensors? What happens with the cursed organs? Can they fix the traumas?

they can regenerate organ and worst case scenario they'll auto revive after like 10 mins

Ziiro:
I think this is a hacky solution affecting only one antagonist. What about traitors? What about changelings? Etc. This is something that needs to be solved at the species level imo

Arturlang (approved):
Fine solution until we figure out something better

Ziiro:
> > I think this is a hacky solution affecting only one antagonist. What about traitors? What about changelings? Etc. This is something that needs to be solved at the species level imo
> 
> Got ideas as to how? I'm not quite sure how

Been chewing on either having them actually die when they're forced into their suits or trying some fuckery with death signals on #3858 but I'm out of town and it's on hold for the moment. Having them actually die does remove something unique about them but it also solves a lot of problematic interactions

PR #3932, opened 2025-05-30
Opens up the Southern Half of the Interlink

Mewchild:
## About The Pull Request

Changes all public shutters on the interlink to pre-opened variants. Removes all medical and surgical kits from the medbay to avoid unfair advantage. Makes all vending machines free, as is the standard for the interlink. NT Rep and Blueshield office have remained unchanged, but they only have a set of CC clothing and a clipboard within them

## Why It's Good For The Game
It's been a long requested item, for the rest of the interlink to be opened for scenes.

## Proof Of Testing
image

</details>

## Changelog
:cl:
map: The interlink is now (almost) fully accessible to crew.
/:cl:

### DIFF
--- a/_maps/bubber/automapper/templates/centcom/interlink_adminoffice.dmm
+++ b/_maps/bubber/automapper/templates/centcom/interlink_adminoffice.dmm
@@ -22,7 +22,7 @@
 /turf/open/floor/wood/large,
 /area/centcom/central_command_areas/admin/interlink)
 "e" = (
-/obj/machinery/modular_computer/preset/id/centcom,
+/obj/machinery/modular_computer/preset/command,
 /turf/open/floor/carpet/executive,
 /area/centcom/central_command_areas/admin/interlink)
 "g" = (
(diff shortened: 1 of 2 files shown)


Rykka-Stormheart:
I still desire interlink + ghost cafe server combo where off-duty people can ICly mingle and have a zero-intensity zone to work their jobs. Bubberchill.........

Rykka-Stormheart:
That said though this looks very nice!

Mewchild:
> The sleepers should probably go too, there's a reason those are only on the emergency shuttles and in some ruins. Also you seem to have removed tables at the bar and kitchen according to mapdiffbot but that doesn't reflect the changes in the files?
> 
> Nice though

I didn't touch the tables. There _were_ shutters over the bar and kitchen, which I replaced with pre-opened variants. I'm personally okay with the sleepers remaining, as it would be nothing but a boon for individuals evacuating the station to continue RP somewhere safe; They can only treat minor injuries, not wounds, and does not drop a circuit if deconstructed.

PR #3978, opened 2025-06-06
Removes getting blackface from Voidwalkers (as well as the "cursed" trait you get) because I died to it once.

BurgerLUA:
## About The Pull Request

Removes the "cursed" trait you get from Voidwalkers that gives you forced pacifism, makes you black, and worst of all: steals your underwear.

## Why It's Good For The Game

On top of voidwalkers being a midround ghost antag that is mute, extremely evasive, picks fights, and is based around ambushing crewmembers; losing to a Voidwalker applies a lobotomy-resistant trauma that makes all your limbs black and forces you to be pacifist.

This is not good for bubbers because you're being punished for being caught in an ambush, and the only "counter" to an ambush is to pre-emptively gear up for one, which is something I don't think staff want players to do.

In nearly all encounters I've had with voidwalkers, the crew (this includes security) isn't very eager to engage with them because of how awful they are to fight and how awful the consequences of dying is.

Also fun fact: This trait previously made you mute.

## Proof Of Testing

Untested.

## Changelog

:cl: BurgerBB
del: Removes the "cursed" trait you get from Voidwalkers that gives you forced pacifism, makes you black, and worst of all: steals your underwear.
/:cl:


BurgerLUA:
Even if /tg/ didn't remove it, it's not good for Bubbers. Those features only exist because it is meant as an alternative to round removal while also punishing validhunters. Bubberstation doesn't have round removal, or too many validhunters, so it's trying to solve a problem that doesn't exist.

BurgerLUA:
> There really should be some replacement for this instead of just removing it, otherwise you can just keep kidnapping people over and over and over with no consequence. Mayhaps simply cahnge around the trauma itself to not be so bad?

There is a bit of a consequence already. You have to play a minigame that heals you, and after that, you're teleported to a random location on the station.

BurgerLUA:
> There really should be some replacement for this instead of just removing it, otherwise you can just keep kidnapping people over and over and over with no consequence. Mayhaps simply cahnge around the trauma itself to not be so bad?

There is a bit of a consequence already. You have to play a minigame that heals you, and after that, you're teleported to a random location on the station.

BurgerLUA:
> There really should be some replacement for this instead of just removing it, otherwise you can just keep kidnapping people over and over and over with no consequence. Mayhaps simply change around the trauma itself to not be so bad?

To elaborate on this further, the current consequence for getting caught is being forced to play a minigame that usually doesn't work (if you're in crit, it wont work). After the minigame, you get teleported to a random location on station.

I thought about replacements, but the issue is that many people don't know how to cure this to begin with. When I asked how to cure it, people said only legion cores could (when the cure was actually either that or a lobotomy, (or apparently someone is saying that dying on land is a cure). It's wildly esoteric, and I rather not have another condition that you have to jump through hoops and consequences to cure.

PR #3985, opened 2025-06-07
Increases the minimum players for blob to spawn from 20 to 50 because I saw a suggestion in #suggestions about it once.

BurgerLUA:
## About The Pull Request

Increases the minimum players for blob to spawn from 20 to 50.

## Why It's Good For The Game

20 players minimum for blob to spawn is absurd. Note that this does not mean 20 crew, but 20 living human non-afk players on the server.

## Proof Of Testing

Untested.

## Changelog

:cl: BurgerBB
balance: Increases the minimum players for blob to spawn from 20 to 50.
/:cl:


BurgerLUA:
Alternative PR that solves the root problem: #3988

PR #3988, opened 2025-06-07
Storyteller Playercount Calculation Overhaul

BurgerLUA:
## About The Pull Request

Significantly overhauls how players are counted when running events. It now considers where the player actually is, what (who) they are doing, and if they are actually a crewmember.

## Why It's Good For The Game

Previously, events would use a proc that'd get every single active living player, regardless if they were crew or not or where they were in the map. This includes people on the interlink, ashwalkers visiting the station for """roleplay""", or player-controlled medibots roaming around the station looking for people to say "fuck you" to. 

Now, it uses a calculation that actually counts players who are engaging in the game so we don't get situations where blob ran because technically there are 20 players playing (5 are on interlink, 3 are in dorms, 2 are ashwalkers, and 10 are  actually on the station).

A player is considered active if they are:
- In the living mob list.
- They are not AFK.
- They signed up as a job, and not an off-station role.
- On a station z-level.
- Not in dorms.

## Proof Of Testing

Active check works. I can't really test anything else because I don't have friends.

## Changelog

:cl: BurgerBB
balance: Significantly overhauls how players are counted when running events. It now considers where the player actually is, what (who) they are doing, and if they are actually a crewmember.
/:cl:


modular_zubbers/code/modules/events/ev_roleplay_check.dm
```diff
@@ -3,7 +3,7 @@ GLOBAL_LIST_EMPTY(dorms_areas)
 
 /area/station/commons/dorms/New()
 	. = ..()
-	GLOB.dorms_areas += src
+	GLOB.dorms_areas[src] = TRUE
```
TealSeer, line 6:
```suggestion
	GLOB.dorms_areas |= src
```
This doesn't need to be an associative list, `|=` will prevent duplicate entries

modular_zubbers/code/modules/events/ev_roleplay_check.dm
```diff
@@ -40,12 +40,15 @@ GLOBAL_LIST_EMPTY(dorms_areas)
...
-	var/area/player_area = get_area(player_turf)
 
 	if(station && !is_station_level(player_turf.z))
 		return TRUE
-	if(dorms && istype(player_area, /area/station/commons/dorms))
-		return TRUE
+
+	if(dorms && length(GLOB.dorms_areas))
+		var/area/player_area = player_turf?.loc
+		if(player_area && GLOB.dorms_areas[player_area])
```
TealSeer, line 51:
```suggestion
	var/area/player_area = get_area(player)
	if(dorms && player_area && (player_area in GLOB.dorm_areas))
```

modular_zubbers/code/modules/events/ev_roleplay_check.dm
BurgerLUA: I changed it to an assoc list  because of a performance benefit for the erp checking.

BurgerLUA:
This is done, but should be testmerged until confirmed working

Arturlang (changes requested):
What about in cases of rescued crew/posibrain cyborgs, are those handled? You could simply bypass this via just having a lot of rescued crew, no? Eh, mostly worried about posibrain cyborgs moreso, the other one is optional

Shadow-Quill:
> If you're dead, you're not counted. I don't think rescued crew (dead in medical) should count because there could be a medical emergency.

Clarification: Recovered crew (as I believe Artur is referring to in this instance) are an effective station ghost role, that medical can fix up from death, and upon revive, polls ghosts for a player.

See https://github.com/tgstation/tgstation/pull/87072 for the original TG pr.

PR #4003, opened 2025-06-10
Security modsuit tweaks/balance

YakumoChen:
## About The Pull Request

- Removes latent flash protection from station-side security-level modsuits (secoff, HoS, captain, Blueshield)
- - This was introduced unintentionally through the latent armour booster modules and has been removed.
- Security-level modsuits now are compatible with welding protection module again (if you want to re-add the flash protection, obviously)
- ERT/Centcom modsuits still have flash protection
- Slowdown reduction removed from the HoS's modsuit armour booster while active
- HoS modsuit now is built-in with the shove-blocker module
- - This module is present in the modsuit upstream but was overwritten with the armour booster here.

## Why It's Good For The Game

Restores some parity from upstream's intended modsuit design by removing/adding features that weren't intended.
Removal of the slowdown reduction on Safeguard was done as a tradeoff for getting the new bulwark module.

## Proof Of Testing

image

image

## Changelog

:cl:
balance: Security-grade modsuits (Secoff, HoS, Cap, BS) are no longer innately flash protected, install a welder module if you want that
balance: HoS's modsuit no longer removes all modsuit slowdown with the armour deployed
balance: HoS's modsuit is built in with the anti-shove module
/:cl:


NargaInDisguise:
This seems like a mostly pointless change since LITERALLY ALL of the people that can acquire these will also have sunglasses anyways. Also HOS modsuit having no slowdown was the point, no?

NargaInDisguise:
> > TBH modsuit armor module for crewsided armors just feels.. unnecesary in general, but this is a good PR in reducing their unintended buffs
> 
> The only reason this exists is because of the inflated armour values our modsuits have

Actually it was only explicitly added to achieve parity with the redsuit in function, IE; reduced armor but atmos sealing for space battles to have less armor.

YakumoChen:
> > > TBH modsuit armor module for crewsided armors just feels.. unnecesary in general, but this is a good PR in reducing their unintended buffs
> > 
> > 
> > The only reason this exists is because of the inflated armour values our modsuits have
> 
> Actually it was only explicitly added to achieve parity with the redsuit in function, IE; reduced armor but atmos sealing for space battles to have less armor.

Yes. Because we have inflated the armour values from their TG counterparts. On TG, modsuits outside of antags have almost no armour.

I'm the one that added armour boosters to these modsuits

projectkepler-ru:
letting hos modsuit have shove protection is so valid tbh
I always felt like the HoS Modsuit was just there to look cool but now this is good.

PR #4018, opened 2025-06-11
Speech Components are now Multilingual

Bombermansam:
## About The Pull Request
You can now define a language for speech components to output in. They are no longer limited to just outputting strings in Galactic Common. This does not give access to ALL languages, for example Codespeak, Cultspeak and Ashtongue, so you will still need to earn the privilege to speak in those languages.
## Why It's Good For The Game
Many people complained that speech components could only output in Galactic Common. This PR will make them happy.
Speech components are most often used for Universal Translator Circuits, this PR will actually make them useful for such and allow for two way communication between individuals if the circuit is made correctly.
## Proof Of Testing
It compiles and the valid languages were able to be observed while using the speech component.
<details>

image

image

image

image

image

</details>

## Changelog
:cl:
add: Language Port to Speech Components
/:cl:
## Acknowledgements
Thank you very much to Artur for helping me make the list which was a pivotal part of this PR. I was tearing my hair out trying to make it. They saved me much time and hair.


modular_zubbers/code/modules/wiremod/components/action/speech.dm
```diff
@@ -0,0 +1,36 @@
...
+
+	//List of valid languages of the speech circuit
+	var/list/circuit_languages = list()
+
+/obj/item/circuit_component/speech/Initialize(mapload)
+	. = ..()
+	for(var/datum/language/language as anything in GLOB.language_datum_instances)
+		if(language.secret)
+			continue
+		circuit_languages[language.name] = language
```
TealSeer, line 16:
```suggestion
	var/static/list/circuit_languages

/obj/item/circuit_component/speech/Initialize(mapload)
	. = ..()
	if(!circuit_languages)
		circuit_languages = list()
		for(var/datum/language/language as anything in GLOB.language_datum_instances)
			if(language.secret)
				continue
			circuit_languages[language.name] = language
```
Since this won't ever change between speech components, we can make it static and only have a single copy shared between them all

modular_zubbers/code/modules/wiremod/components/action/speech.dm
```diff
@@ -0,0 +1,36 @@
...
+
+/obj/item/circuit_component/speech/populate_ports()
+	. = ..()
+	. += language_port = add_input_port("Language", PORT_TYPE_STRING, trigger = null)
```
TealSeer, line 20:
```suggestion
	language_port = add_input_port("Language", PORT_TYPE_STRING, trigger = null)
```

TealSeer (changes requested):
It appears we also have non-modular additions to the speech component in the base file, I'm going to ask that you integrate those into this file and remove them from the base file, since you're overwriting `input_received` here, you'll be overwriting those changes unless you add them

modular_zubbers/code/modules/wiremod/components/action/speech.dm
Bombermansam: Done

TealSeer (approved):

lessthnthree (approved):

PR #4024, opened 2025-06-12
Primal stowaways

ArrisFairburne:
## About The Pull Request

Makes it so that all species selectable via the character customization are now eligible for being stowaway.

## Why It's Good For The Game

- Enables a way to play ash walkers/primitive nekos when on maps that don't have lavaland/icemoon wastes
- Enables primitive podpeople to play, period (since we don't have any maps that allow for them)
- Allows characters of these types to RP with more characters (instead of basically only ever talking to others of their species and miners) while sticking to the lore of "this species is not meant to be on the station"
- Makes the stowaway role a little bit more varied and interesting

## Proof Of Testing
<details>

GIF 11-Jun-25 10-40-06 PM

</details>

## Changelog
:cl:
add: Ash Walkers, Primitive Demihumans, and Primal Podpeople are now playable in the Stowaway role
/:cl:


ArrisFairburne:
> "Gee whiz, why did we separate the species so you can't play primals on station again?"

I don't expect primals to be that big of a balance issue cause, yes, you are getting extra survivability, but in exchange you're locked to the role that doesn't get any sort of door access, borgs don't consider you crew, and sec has the right to screw with you whenever they want even if you have no crimes.  You also don't have any home base areas that are normally afforded to these species.

If you can point to a PR that removed their playability on the station, that would be a good reference for what I might need to consider balance around.

Regardless, I'd be fine with making them ineligible for totrolls. The syndicate probably wouldn't hire them anyway.

Arturlang:
@YakumoChen ashwalkers are gonna die to lack of their preffered air, and primal podpeople have the same healing as stationside podpeople, stationside ones may even have better healing due to parent calls

nevimer:
I think this idea is fine

ArrisFairburne:
> Aren't the primal species supposed to not be on station, period, without admin approval? 
Yes, but I'm pretty sure this is only because "primal faction invades the station" is something that should only happen with min oversight and/or valid reason to retaliate. If you're a stowaway the expectation to not escalate with no reason remains.

> If we're going to make them stowaway capable, then why not just make them crew capable and remove the work around entirely? 
Ashies etc. have massive use speed prnalties to interacting with tech (and/or are completely incapable of using it). They also cannot speak galcom by default. If you late join, then you're going to be waiting at the interlink for someone else to use the terminal for you; you can't radio in to ask someone to call the shuttle since you don't speak galcom.

Being unable to speak galcom is problem enough for anyone who already plays a job, but the inability to use tech means that ashies etc. would be highly reliant on others to function in their job slot, and I don't think they should be eligible for roles that they cannot perform.

projectkepler-ru:
I should note- the only issues we had with stowaway as of late were just them using it to be a nuisance and having sec effectively have to lock them in permabrig for the whole round

If that's not something you are worried about then there's nothing wrong with it

YakumoChen:
I would rather see this race possibility apply to Prisoner rather than Stowaway.

Prisoners are already confined to being in permabrig when they spawn, and they're unlikely to be given a job that matters if they seek parole 

Stowaway just comes off as Prisoner++++ in comparison, you spawn with complete freedom and have the opportunity to rush to HoP and gain a round-impacting job if you desire

PR #4030, opened 2025-06-12
Remove all our quirk species whitelist code

### LINKED ISSUE #3732: Species quirk whitelist does not work.
### Client Version:

516.1660

### Issue Summary:

https://github.com/Bubberstation/Bubberstation/blob/d5621ec034b0c075b37882ff877519a0b53739df/tgui/packages/tgui/interfaces/PreferencesMenu/CharacterPreferences/QuirksPage.tsx#L389
Quirk whitelist doesn't properly block people from taking species specific quirks if they are not that species.

### Round ID:

_No response_

### Test Merge Information:

_No response_

### Reproduction Steps:

Attempt to add a whitelisted quirk to a species that should not be able to take it. Witness the quirk be allowed until they join in game where a runtime occurs and the quirk is removed.

TealSeer:
## About The Pull Request

Upstream now has a system for this, removes all our override code and changes the relevant quirks to use the upstream system instead

## Why It's Good For The Game

Fixes #3732 

## Proof Of Testing

Tested it on a local with sol weakness, masquerade, and hydrophillic, the latter wasn't selectable unless you were a slime, the former two weren't selectable unless you were a hemophage. Sol weakness was still removed as normal if you turned into a bloodsucker

## Changelog

:cl:
fix: fixed species-locked quirks not actually working
/:cl:


tgui/packages/tgui/interfaces/PreferencesMenu/CharacterPreferences/QuirksPage.tsx
```diff
@@ -341,7 +341,6 @@ export function QuirksPage(props) {
       positiveQuirks += 1;
     }
```
Arturlang, line 342: Might as well remove the commented out code here

Arturlang (approved):
Looks good, pretty much the same i did on my other commit, as we talked, though consider keeping the whitelist/blacklist lists

PR #4058, opened 2025-06-16
Changes the Medtech kit's bonesaw's force from 20 > 15

SapphoQueer:
## About The Pull Request

Title explains it. Goes from 6.7 hits/crit to 9
## Why It's Good For The Game
A circular saw having the damage of more dedicated melee weapons (on par with the bowie knife/cap's sabre minus the block and AP) whilst also being pocket sized - AND being included in a medkit just don't make sense. A little less damage than a regular circular saw is reasonable given it's more compact. In its current state it does more damage than the upgraded medical tools. There is very little reason for this to do as much damage as it does.
## Proof Of Testing
<details>

</details>

## Changelog
:cl:
balance: Moved the Bone Saw's force from 20 > 15
/:cl:

### DIFF
--- a/modular_skyrat/modules/exp_corps/code/gear.dm
+++ b/modular_skyrat/modules/exp_corps/code/gear.dm
@@ -19,7 +19,7 @@
 /obj/item/circular_saw/field_medic
 	name = "bone saw"
 	desc = "Did that sting? SAW-ry!"
-	force = 14
+	force = 15
 	icon_state = "bonesaw"
 	icon = 'modular_skyrat/modules/exp_corps/icons/bonesaw.dmi'
 	lefthand_file = 'modular_skyrat/modules/exp_corps/icons/bonesaw_l.dmi'
--- a/modular_skyrat/modules/exp_corps/code/gear.dm
+++ b/modular_skyrat/modules/exp_corps/code/gear.dm
@@ -19,7 +19,7 @@
 /obj/item/circular_saw/field_medic
 	name = "bone saw"
 	desc = "Did that sting? SAW-ry!"
-	force = 20
+	force = 14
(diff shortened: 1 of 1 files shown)


YakumoChen:
Who thought it was a good idea to give the overpowered gateway equipment to crew anyway?

SapphoQueer:
> Who thought it was a good idea to give the overpowered gateway equipment to crew anyway?

Paxil

Arturlang:
Another thing I dislike is that this is a small tool that outperforms normal sized surgery tools that are even superior(laser scalpel), so this is still powerful even if it does less than the bonesaw, as it is smaller, thankfully no longer outperforms antag tools entirely with it's size/damage

modular_skyrat/modules/exp_corps/code/gear.dm
```diff
@@ -19,7 +19,7 @@
 /obj/item/circular_saw/field_medic
 	name = "bone saw"
 	desc = "Did that sting? SAW-ry!"
-	force = 20
+	force = 14
```
projectkepler-ru, line 22: should up it to 15 to match survival knife which you can buy in TG and here

SapphoQueer: The reason I decided on 14 and not 15 is because the circular saw is 15, and it still works as a one whilst also fitting in a pocket. I dont see it bringing too much harm if I bump it up, but it being 1 less brute than the circular is intended.

KateAnbolho:
Clearly we need to buff the circ saw to like, 17

SapphoQueer:
> Clearly we need to buff the circ saw to like, 17

No.

PR #3636, opened 2025-04-25
Ports SPLURTs genital sprites, and butts and tummies

CydiaButt13:
## About The Pull Request
As said in title, this PR adds all of the new SPLURT sprites for breasts, penises, and balls. It also adds over butt sprites and tummy sprites. 
## Why It's Good For The Game
I believe in freedom of expression with character customization and I think this really helps that. Also big butts.
## Proof Of Testing
Arduously hammered out every bug I found until it localhosted without issue!!!!!!!
<details>
<summary>Screenshots!! NSFW!!!!!!!</summary>

stt313X

Pa1866wMhx
While I know they're large they can be smaller.
</details>

## Changelog
:cl:
add: Added Splurt Breast/Penis/Ballsack sprites
add: Added Ass and Belly sprites from SPLURT/Hyper Station
changed: Old sprites to "name (old)"
/:cl:

PR #3897, opened 2025-05-25
Adds missing/additional genital sizes & sprites.

AstrumLernaean:
## About The Pull Request

Like the title says. We found some unused sprites for genitals and decided it'd be neat to add them in along with their respective sizes. This includes cleanup to make sure the non-alt versions/older sprites that don't have up to those same sizes don't disappear and stay at the previous max size even if you choose the higher numbers. Sprite additions for penis, balls, and breast sizes are included.

## Why It's Good For The Game

More customization options don't hurt and we know a lot of people wanting these! Don't see why it'd be a bad thing, only thing we could foresee is the bigger sizes being questionable but pretty much all the risk factors of that is covered by our ERP policy/rules.
## Proof Of Testing

_Want us to post naughty bits for the world to see...?_
Yes tested numerous times on our localhost, we even accounted for the larger sizes making the older sprites disappear (for some reason), hopefully that doesn't break again somehow.
<details>
<summary>Screenshots/Videos</summary>

</details>

## Changelog
:cl: AstrumLernaean
add: More genital size options, with sprites! For your viewing pleasure.
/:cl:

PR #4071, opened 2025-06-18
Butts and bellies preferences

plsleavemealon:
## About The Pull Request

This adds in butts and bellies into the game. I limited the size options available so that people cannot make them too big. Also adds in new prefs for them too for size chemicals if/when those get ported. Also added in more sprites for breasts and testicles for future use. Those sprites (the larger testes and breasts) are not enabled ingame, they are not normally accessible and are only added so that someone trying to setup size prefs will have an easier time doing so. This is all based on the lovely work done here https://github.com/Bubberstation/Bubberstation/pull/3636 . Of note, the bellies skin tone option does not presently work. I am not sure why this is but I suspect they didn't setup the overlay for it properly. Unfortunately I am only an amateur at spriting so I don't really know how to fix that.

The images shown are the absolute largest sprite attainable ingame without modifying the files from this PR. The larger sizes could easily be enabled if someone wished (just tweaking a variable or uncommenting some lines).

## Why It's Good For The Game

More player customization! This limits the size options for the butts and bellies to be no larger then other sexual organs here. A person with them cranked to the current max settings will not be obscuring much if anything.

## Proof Of Testing

Tested locally, works just fine, barring 1 thing I don't know how to test the ERP verb for butts though.

## Changelog

:cl:
add: Adds belly and sprites
/:cl:

image
image
image
image
image
image

### DIFF
--- a/code/__DEFINES/~skyrat_defines/DNA.dm
+++ b/code/__DEFINES/~skyrat_defines/DNA.dm
@@ -69,7 +69,7 @@
 
 //In inches
 #define PENIS_MIN_GIRTH PENIS_MIN_LENGTH
-#define PENIS_MAX_GIRTH 60
+#define PENIS_MAX_GIRTH 40
 #define PENIS_DEFAULT_GIRTH 5 // a lil big but not by much
 #define PENIS_MIN_LENGTH 1
 #define PENIS_MAX_LENGTH 72


ReturnToZender:
Oh hey, sweet!

plsleavemealon:
I'm also...attempting to make size prefs. But its not easy to put it lightly. It is very annoying how little examples of alternate_appearance there is in the code. And all examples are using the basic version, whereas this would absolutely require the more "advanced" version.

plsleavemealon:
> Would it be possible to have a size 0 belly, so as to be able to configure colour and such in prefs but have it be invisible until such a time it possibly gets changed?

Pretty sure this is doable yes, though I have not added in the verb that lets you change your belly's size (or the size chems for bellies and butts in general). That would take more some more effort, which I want the maintainers to approve of this at the very least before I put any more effort into it. Because if they won't approve of even these sprites, with the limitations put in place to prevent them from getting "too big" then these sprites will never be added ever.

Size prefs are not coming. I am becoming increasingly convinced that it is completely and totally impossible.

plsleavemealon:
> Ass and belly sprites will require a pref system for people to selectively toggle their visibility if they don't want to see these before we consider adding them, as discussed previously by the other maintainers.
> 
> I'm not gonna demand you code it yourself, but It's either a decision to merge or close in its current state of affairs.
> 
> #3897

I thought that was explicitly for **hyper** sprites because hyper sprites obscure peoples vision and can cause mechanical issues in some cases. These sprites are the same size of sprites as any of the other lewd sprites we have in the game so I really don't understand the extreme refusal and stance against them.

PR #4076, opened 2025-06-18
Uncomments the experimental syndicate teleporter back into the traitor shop

KazooBard:
## About The Pull Request
This PR brings back the skyrat-removed experimental syndicate teleporter

## Why It's Good For The Game

1. Diversification of loadouts. 
The reason the traitor is seen with only 6 items ever is due to the fact that the variety of interesting options has decreased as players tried more and more and have stuck to the best options. By adding a new item, or rather, readding it, the meta will breathe a bit and the pool of common loadouts will split up into more varied encounters. 

2. Uniqueness. 
This item is not something slapped on traitor without a second thought, as it is rather something that traitor missed in it's long road to modern SS13's state - a mobility option like the rest of antagonists, this time with serious risk and reward. The unorthodox use of the item helps for easier accesibility to some more magic-related gimmicks. Budget wizards, ninja-wannabes, Disappearing looming figures with kitchen knives - list goes on. It's fresh. It's something we haven't played a thousand times.

3. Accessibility and the teleportation argument.
I hear a concern is that antagonists have a lot of teleports lately which make them annoying to fight. Yes. Antagonists lasting longer than a fight and being hard to catch has been a thing since changeling adrenaline and sting, since ethereal jaunt, heretic being released in what, 2020? The hard pill to swallow, though, is that you can take every single teleport away from every antag and they'll just steal the hand teleporter or abuse spin inverters.

However, if a traitor is to be given the option to sacrifice 8 TC, now that is an esword in price iirc, a bit more than a makarov or a whole syndicate modsuit to gain a usable, interesting and fun tool for teleportation - they will spend less TC on pure murderbone, less time stacking 50 spin inverters, less time getting game-ier tools since the price matches what they get. 

4. Difficulty of fighting.
It's not that tough. It's arguably one of the weaker teleportation tools. 

Predictable direction, chance to instantly RR yourself, unpredictability, reliance on mesons all make it less of a problem than a traitor with a hand teleporter.
Setting traps, infinite teleports, zero risk in use, lets you strand people in the middle of nowhere.
Hierophant rod lets you yoink anyone you killed to your safehouse to fullstrip and loot safely
Spin inverters work cross z level
Mining cubes are a joke balance-wise
Don't start on heretics and vamps.

So no, this won't be the downfall of balance and an overpowered teleport. It's More risk, More restrictions, actually needs an Investment and even is countered by ion rifles. It has more counterplay than EVERY single teleport ingame. There is no argument to be made against it being too strong. It also comes from another RP server originally (beestation) and is doing just fine.

## Proof Of Testing

Trust me

## Changelog

:cl:
add: readds the experimental syndicate teleporter to the uplink!
/:cl:

### DIFF
--- a/code/modules/uplink/uplink_items/device_tools.dm
+++ b/code/modules/uplink/uplink_items/device_tools.dm
@@ -85,7 +85,7 @@
 	item = /obj/item/storage/briefcase/launchpad
 	cost = 6
 	progression_minimum = 50 MINUTES //Normally this is not there but it exist to delay you just buying it and getting into everywhere before sec is prepared
-/* //Bubber edit - Moves the comment to keep the syndicate teleport commented out. Skyrat commented this out.
+
 /datum/uplink_item/device_tools/syndicate_teleporter
 	name = "Experimental Syndicate Teleporter"
 	desc = "A handheld device that teleports the user 4-8 meters forward. \
@@ -94,7 +94,6 @@
 			Comes with 4 charges, recharges randomly. Warranty null and void if exposed to an electromagnetic pulse."
 	item = /obj/item/storage/box/syndie_kit/syndicate_teleporter
 	cost = 8
-*/ //END SKYRAT EDIT
 
 /datum/uplink_item/device_tools/camera_app
 	name = "SyndEye Program"
--- a/code/modules/uplink/uplink_items/device_tools.dm
+++ b/code/modules/uplink/uplink_items/device_tools.dm
@@ -92,7 +92,7 @@
 			Beware, teleporting into a wall will trigger a parallel emergency teleport; \
 			however if that fails, you may need to be stitched back together. \
 			Comes with 4 charges, recharges randomly. Warranty null and void if exposed to an electromagnetic pulse."
-	item =
+	item = /obj/item/storage/box/syndie_kit/syndicate_teleporter
 	cost = 8
 
 /datum/uplink_item/device_tools/camera_app
--- a/code/modules/uplink/uplink_items/device_tools.dm
+++ b/code/modules/uplink/uplink_items/device_tools.dm
@@ -92,7 +92,7 @@
 			Beware, teleporting into a wall will trigger a parallel emergency teleport; \
 			however if that fails, you may need to be stitched back together. \
 			Comes with 4 charges, recharges randomly. Warranty null and void if exposed to an electromagnetic pulse."
-	item = /obj/item/storage/box/syndie_kit/syndicate_teleporter
+	item =
 	cost = 8
 
 /datum/uplink_item/device_tools/camera_app


BurgerLUA:
There was a reason this was removed.

Since this item had a very high risk/reward system, it was used exclusively by gamers. When it was used by gamers, it lead to some of the worst sort of gameplay imaginable. If a player who wasn't experienced used this, then they'd just gib and that was it for them.

Giving this sort of evasion to regular traitors is not something we should be doing. High evasion wild goose chases are awful for everyone involved, except the antag.

code/modules/uplink/uplink_items/device_tools.dm
projectkepler-ru, line 1: I take some mild offense to the "skyrat removed" part because this was actually continued and kept as our decision as far back as when I was still a maintainer

KazooBard: I am stating it as it is in the code. Skyrat removed it, we didn't change the decision. We didn't remove it.

KazooBard:
> There was a reason this was removed.
> 
> Since this item had a very high risk/reward system, it was used exclusively by gamers. When it was used by gamers, it lead to some of the worst sort of gameplay imaginable. If a player who wasn't experienced used this, then they'd just gib and that was it for them.
> 
> Giving this sort of evasion to regular traitors is not something we should be doing. High evasion wild goose chases are awful for everyone involved, except the antag.

Gamers will find other ways to game no matter what, and they will be much worse. We should not be gatekeeping items away from the populace on assumption of the current audience which includes people trying to get into antagonist roles shouldn't be allowed to try items lest someone else (who can do well without them) will benefit from it, too.

Noone starts off doing great with the teleporter and I'd know as I was there on the day it first came to beestation, it's a unique tool. Encouraging people to try new strategies even if they're risky should be the goal, as if you can argue it led to annoying chases, I can argue people DID want to learn to use it - even if poor at combat, and I can argue that it is not the bane of sec's existance with being weaker than most options. 

That is a very important thing to acknowledge. It is not stronger than most options that we have availible. 

Wild goose chases are parts of the game and it is undeniable that every antagonist, new or old will aim for this situation as it is beneficial to them. IF you want to remove wild goose chases from gameplay, then you would have to remove antagonists - not just the combatative ones, but every ounce of need to chase after someone or every way to distance yourself from someone.

BurgerLUA:
It was Skyrat that removed it and it was before Progtot was even a thing, I think. It was that bad.

projectkepler-ru:
> It might of been us who removed it, I forget. To be fair, it was removed during a time where prog-tot existed and TC was practically infinite. This needs discussion.


it was us that decided to kept it removed. I still distinctly recall having the chat with you and nevimer on if we could get the launchpad back in but kept the hand tele off until we were sure it wasn't a problem
It's not as if I think the hand tele would be a problem then but there were specific subset of people whom I felt was going to make it increasingly worse to play against. Of which, I feel those subset of players still exist.

PR #4105, opened 2025-06-22
Revert "Removes the 18+ vetted examine"

ReturnToZender:
## About the Pull Request
Reverts Bubberstation/Bubberstation#4104, which removed the vetted examine text as we no longer permit players who aren't vetted to join the game. Bubberstation is safer than it ever has been before. It may be that we don't need the code to differentiate anymore, but-- and this is almost certainly one of the only times I will use this is an argument--
**SOVL**

## Why It's Good For The Game
The vetted tag has served us faithfully for the past two years. It is Bubber's strongest warrior. Is it necessary code, anymore? Probably not. But continuing to have it present on players will remind people that the staff here ensure the environment is safe for everyone. It's not just information, it's a physical affirmation of that commitment to our players. It should stay, for at least a year or two.

## Changelog

:cl:
add: Restored the holy vetted examine text, even if it's redundant now. It can wait a bit before going to the retirement home.
/:cl:


CEO-Of-Dogborg-Corp:
> Poll is up. I'll let it run for a week

where?

AXLFAN2000:
> where?

When you're in the main menu ingame, it's at the bottom.  It should flash orange when there's a poll you haven't seen yet.
image

PR #4160, opened 2025-07-01
Fixes Liquid fires.  Finally.

Stonetear:
## About The Pull Request
At last, after over 5 years of liquid fires being inextinguishable, you can now put them out.
* Liquid fire tiles now *stay* out once you extinguish them!  They have to re-spread from an existing fire.
* Liquid fires now burn ~~100x~~ 50x times faster.  From 0.1u per tick to 5u per tick.  Fires will be noticeably shorter.
* Liquid fires now only have a 70% chance to spread to adjacent tiles.

## Why It's Good For The Game
This was broken for like 5 years.

## Proof Of Testing
Not the best screenshot but trust me bro it works ;)

Screenshot 2025-07-01 154700

## Changelog
:cl: Stonetear
balance: Liquid fires now burn out 50x faster.  Flammable liquids burn at 5u/tick, which is like 2.5u/s, per tile.
fix: Extinguishing liquid fires now actually works!  
/:cl:


sippykot:
>Liquid fires now burn 100x times faster. From 0.1u per tick to 10u per tick. Fires will be noticeably shorter.

isnt that a bit much?

Stonetear:
> > Liquid fires now burn 100x times faster. From 0.1u per tick to 10u per tick. Fires will be noticeably shorter.
> 
> isnt that a bit much?

Currently, liquid fires burn for stupidly long times.  The only real time that liquid fires gets used is when you send fuel down the trash chute because you're mad at cargo for not delivering your packages.  This makes firebombing still destructive but less persistently annoying.

I think we should go with this, and can change it if people don't like it.  Fire doesn't actually heat up the air so there's no application beyond destruction.

sippykot:
> > > Liquid fires now burn 100x times faster. From 0.1u per tick to 10u per tick. Fires will be noticeably shorter.
> > 
> > 
> > isnt that a bit much?
> 
> Currently, liquid fires burn for stupidly long times. The only real time that liquid fires gets used is when you send fuel down the trash chute because you're mad at cargo for not delivering your packages. This makes firebombing still destructive but less persistently annoying.
> 
> I think we should go with this, and can change it if people don't like it. Fire doesn't actually heat up the air so there's no application beyond destruction.

well yeah, but imo it should be 10 times faster not 100 times
10u deleting itself per tick is crazy high

modular_skyrat/modules/liquids/code/reagents/chemistry/reagents.dm
```diff
@@ -16,7 +16,7 @@
 */
 /datum/reagent/consumable/ethanol
 	liquid_fire_power = 10
-	liquid_fire_burnrate = 0.1
```
nikothedude, line 19: kinda feel like this should be a define? DEFAULT_BURN_RATE?

modular_skyrat/modules/liquids/code/liquid_systems/liquid_effect.dm
```diff
@@ -116,6 +116,9 @@
 	. = ..()
 	if(fire_state)
 		set_fire_state(LIQUID_FIRE_STATE_NONE)
```
nikothedude, line 118: wouldnt it be better to add this new line to this proc itself?

PawnAtWork:
I'm pretty sure they didn't go out because the reagents inside reached a temperature that kept them self igniting. Like oil fires in real life, unless you can smother them or reduce the temperature extremely, they ain't going out.

Stonetear:
> I'm pretty sure they didn't go out because the reagents inside reached a temperature that kept them self igniting. Like oil fires in real life, unless you can smother them or reduce the temperature extremely, they ain't going out.

Neither the temperature of the liquid (as a liquid component) nor the temperature of the turf changes during burning.  The hotspot code doesn't even work.

modular_skyrat/modules/liquids/code/liquid_systems/liquid_effect.dm
Stonetear: Upon further examination, it's right how it is.

PR #4172, opened 2025-07-03
Fixes blueshield sci access

SapphoQueer:
## About The Pull Request
The blueshield's ID had the "science" tag if you looked at it through the HOP console, but that one is specifically for circuits/experimentor/toxins/etc. 

The "research" tag in the hop console actually lets them enter science main, this means blueshield was able to enter toxins and circuits, but not the main research area.

Which is more than likely not an oversight. Being able to enter xenobio but not Actually Just The Main Department is ????

I don't blame whoever did this on skyrat given the access is the wrong way around in the code but that's an upstream problem I don't care enough to fix right now. 
## Why It's Good For The Game
Fixing oversight good
## Proof Of Testing

I loaded it up locally and it works
<details>

</details>

## Changelog
:cl:
fix: Blueshields can now properly enter science.
/:cl:


Hislyna:
it was like this for so long i thought it was a feature...

PR #4212, opened 2025-07-10
Makes it so having WALKMODE on doesn't crush people with the micro quirks

HometownFunky:
## About The Pull Request

Having WALK MODE on makes it so you don't give micros any wounds/damage. Knockdown still applies.
## Why It's Good For The Game
Giving people a way to not accidentally squish micros is good, having it tied to walkmode which everyone can turn on willingly and not just a quirk like before (Pacifism) is also good. The knockdown is still there. 
## Proof Of Testing

Tested it on a localhost
it works
<details>

</details>

## Changelog
:cl:
balance: Having WALK mode on makes you not squish micros.
/:cl:

### DIFF
--- a/code/datums/components/squashable.dm
+++ b/code/datums/components/squashable.dm
@@ -62,7 +62,7 @@
 	if(isliving(crossing_movable))
 		var/mob/living/crossing_mob = crossing_movable
 		if(crossing_mob.mob_size > MOB_SIZE_SMALL && !(crossing_mob.movement_type & MOVETYPES_NOT_TOUCHING_GROUND))
-			if(HAS_TRAIT(crossing_mob, TRAIT_PACIFISM))
+			if(HAS_TRAIT(crossing_mob, TRAIT_PACIFISM) || (crossing_mob.move_intent == MOVE_INTENT_WALK)) // BUBBER EDIT
 				crossing_mob.visible_message(span_notice("[crossing_mob] carefully steps over [parent_as_living]."), span_notice("You carefully step over [parent_as_living] to avoid hurting it."))
 				return
 			if(should_squash)


Hislyna:
I mean if you take the negative quirk, you should expect the sqiuish *shrug

AXLFAN2000:
I like this from the POV of having run into multiple people who have expressed that being on the doing the squishing end of the micro squishing makes them uncomfortable  --  it's nice to have a way to opt-out of engaging with it

Hislyna:
> I like this from the POV of having run into multiple people who have expressed that being on the doing the squishing end of the micro squishing makes them uncomfortable -- it's nice to have a way to opt-out of engaging with it

Understandable from that POV, just a very weird thing since its a negative quirk that someone chose to put on, can't just start start making people expect others do all their work and such. Otherwise we need to start talking why a neutral quirk effects a positive quirk (Which I am still against and should be removed)

HometownFunky:
Understandable, I don't think I can fix player entitlement with code tho. The intent of this PR is moreso to give people a way to opt out of squishing (at least the wound/damage part of it) if they so desire. Currently, you have no way to avoid the chance of giving a micro a wound when you move onto their tile UNLESS you willingly picks the pacifist quirk (If I am reading the code correctly); and moving onto a tile is something that happens a lot in scenes and various roleplay scenarios, which is where this is coming from. Having your limbs get a t2 crack from your friend carefully sitting on the table around you might be peak SS13 but it does kinda ruin the scene a bit, imo.

Sir-Lagsalot:
It's super frustrating to not be able to walk around someone without stepping on them - I initially assumed this is how it already worked: if you're being careful and paying attention (ie using walk mode) you step past them, like walking on wet floors. Turns out you can't even CRAWL past someone with the quirk 

Not within the scope of this PR, but would it also make sense for Light Step to have some effect? If a character is so careful they can traverse _oceans_ of blood without getting their shoes dirty, it also stands to reason they could step around a Little Guy, no?

TealSeer (approved):

PR #4228, opened 2025-07-13
Nerfs the Summon Simians spell because I died to it once.

BurgerLUA:
## About The Pull Request

Makes the Summon Simians spell summon 1 lesser gorilla instead of a random selection of 4 monkeys/lesser gorillas.
Reduces the cost of Summon Simians from 2 to 1.

## Why It's Good For The Game

Given how spammable the spell is, it can actually lag the server quite a bit because of how many overlays monkeys have. Not only that, but the lesser gorillas are very strong and can overpower an under-equipped crew quite a bit. I can see this spell being fair if wizard was the only antag type in the round and people treated wizard like nuke ops, but that isn't the case due to Die of Fate making it a mid-round and 99% of crew opting to choose the peaceful wizard route.

## Proof Of Testing

Untested. If it compiles, it werks.

## Changelog

:cl: BurgerBB
balance: Makes the Summon Simians spell summon 1 lesser gorilla instead of a random selection of 4 monkeys/lesser gorillas. Reduces the cost of Summon Simians from 2 to 1.
/:cl:


NargaInDisguise:
Honestly you've outlined the actual problem here; The die of fate.
That thing shows up constantly, crew rushes to find it in the most NRP shitfest, then metagames as fuck by dropping all their items and coming back to get them for all their old accesses (Because every character apparently knows that rolling this magical die will destroy all of your old clothes! It pisses me off to no end). Wizards need to start being treated as the threats they are and crew needs to stop being handed a free magic pass. Rolling the die of fate SHOULD GET YOU FUCKING BRIGGED. I do not get it.

Arturlang (changes requested):
too much, let it spawn 2 mobs with monkeys and gorillas in the random list of pickables

BurgerLUA:
> too much, let it spawn 2 mobs with monkeys and gorillas in the random list of pickables

Part of the reason why I removed monkeys is because they don't actually attack crew.

PR #4232, opened 2025-07-13
Replaces the Wizard Dice event with a round-start wizard role

xPokee:
## About The Pull Request
Comments out the Die of Fate event and replaces it with a round-start wizard antagonist that happens ***VERY*** rarely. Said role has a weight of 2 and requires a minimum of 35 players to be readied up for it to even roll.  The die of fate that spawns in trash piles has also been replaced with the Lavaland variant. 

_The die of fate event is still spawnable by admins._
## Why It's Good For The Game
This is going to be a controversial one, so let me run you guys through my thought process. Currently, as it stands, the wizard die event causes players to turn into mindless dice-hunting zombies that will employ various LRP tactics to get the die, ranging from rushing to cargo to get a Ouija board to simply tiding into places where they see the die. This event also runs quite frequently, leading to frustration among players in the departments that get tided into. Burger's original reason for adding the die of fate event was to "add wizard naturally into the game without a ghost role for players to idle for". This will prevent players from idling for it as it happens at round-start. This is the fairest compromise I could come up with for players to be able to roll wizard without the tiding and general tomfoolery that happens because of the dice.
## Proof Of Testing
<details>

image

</details>

## Changelog
:cl:
add: added a round-start wizard event
del: removed the wizard die mid-round event
balance: replaced the maintenance loot wizard die with the proper die of fate
/:cl:


BurgerLUA:
I think an alternative solution to this is to make the event less common, and for admins to start enforcing LRP/Powergaming rules when it comes to getting the dice. If people are rushing the dice by greytiding, I'm pretty sure that's a rulebreak.

destrucktoid:
I don't think anyone ghosts specifically because they hope to get wizard and saying as much is in my eyes disingenuous. There are lots of reasons why people ghost and don't join the round, removing antagonists from spawning as ghost roles hasn't and will not change that. We used to have lots more ghost roles than we currently do and yet we still have just as many people observing with no intention of joining the round. It hasn't fixed the problem because it was, again in my opinion, never the problem.

I agree that the dice of fate isn't the best because it incentivises people to greytide. However, this makes it so that there is only ever going to be a wizard at roundstart and there is no way for a midround wizard to appear. The nerf to the trash pile dice is also overkill in my opinion; it's rare enough to get it in a trash pile, no need to kill it even more.

As such, may I instead offer an alternative? You could instead lower the odds of the dice of fate and increase it's danger level (similar to what Burger did in his PR) and then, instead of teleporting to a random location on the station, it teleports to a random player, either in their hands or on the floor beneath them if they don't have any hands available.

This way, players still don't get guaranteed wizard from it, everyone who it teleports to gets to roll it once if they wish, when someone gets it or sees someone else get it then they can choose to announce it exists over the radio, giving security knowledge it exists and if security gets it to teleport to one of them or someone chooses to give it to them instead, then they have the ability to destroy it if they want to. This way there is no need for tiding.

xPokee:
> "Minimum of 50 players readied up" I don't think I've seen this many players ready up at roundstart on the server? This just seems like a soft removal of wizard entirely?

After seeing this I realize I made it too high, going to lower it to 35 like round-start heretic

BurgerLUA:
I don't think this should be merged because if it is, there is going to be a lot more wizard players dominating the station and a lot more rounds with bad player experiences.

The magical thing that made the dice work is three things:

- It was a mid round occurrence, which means that the station has progressed in terms of equipment/funds/research/whatever.

- It didn't always guarantee wizard, as it required someone to win a 1 in 20 dice roll.

- It was rolled by people who  normally wouldn't have wizard enabled, effectively lowering the average skill level of wizard players (making it more reasonable.)


Making wizard round start will result in extreme levels of chaos that I don't think bubberstation needs. We should be seeking nerfs/adjustments to the wizard die instead of going for the nuclear option.

PR #4234, opened 2025-07-13
The security-based antag cap is now modified by the current alert level.

BurgerLUA:
## About The Pull Request

Antag cap calculations now round down the number of antagonists, instead of rounding to the nearest whole number.

The **security-based** antag cap is now modified by the current alert level.

On code green, blue, orange, and violet, the antag to security ratio is 1.5. This is how it currently is.
On code amber, the antag to security ratio is 1.25.
On code red, the antag to security ratio is 1.
All other alert levels have a ratio of 0. This should be reserved for world-ending events.

Antagonist roles will not be removed if the alert level changes. This just  generally means that alert level increases won't spawn any new antagonists unless the antag count is lacking.

(Antag to Security ratio refers to how many antags to spawn per member in the security department. A value of 1.5 means spawn 1.5 antags for every security officer.)

(There are two antag caps here, the security based one, which is based on sec population, and the crew based one, which is based on crew population. This only affects the security based one.)

## Why It's Good For The Game

I've been trying to figure out how to intelligently control the amount of chaos in a round so that shifts aren't either a nothingburger or a insane clusterfuck of a shift. I think that this implementation is a good way to control that given that:
- If a round is out of control, code blue is used.
- If a round is very manageable, code blue or green is used.
- If a round is experiencing some minor difficulties, code amber is used.
- If shit is hitting the fan, code red is used.

This assumes that people who are setting the alert level are acting in good faith, which is generally the case. If it isn't, then it's usually very obvious if the captain is trying to reduce the amount of antags, and thus can probably be smacked by other command staff or even central command.

## Proof Of Testing

If it compiles, it werks.

## Changelog

:cl: BurgerBB
balance: The security-based antag cap is now modified by the current alert level. Antag cap calculations now round down the number of antagonists, instead of rounding to the nearest whole number.
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/storyteller/gamemode.dm
+++ b/modular_zubbers/code/modules/storyteller/gamemode.dm
@@ -236,7 +236,23 @@ SUBSYSTEM_DEF(gamemode)
 		return 0
 	if(!storyteller.antag_divisor)
 		return 0
-	return round(max(min(get_correct_popcount() / storyteller.antag_divisor + sec_crew ,sec_crew * 1.5),ANTAG_CAP_FLAT))
+
+	var/current_population = get_correct_popcount()
+
+	var/crew_cap = FLOOR(current_population / storyteller.antag_divisor,1)
+
+	var/alert_cap_multiplier = 0 //Anything above red means 0, like Delta and world ending events.
+	switch(SSsecurity_level.get_current_level_as_number())
+		if(SEC_LEVEL_GREEN to SEC_LEVEL_ORANGE)
+			alert_cap_multiplier = 1.5
+		if(SEC_LEVEL_AMBER)
+			alert_cap_multiplier = 1.25
+		if(SEC_LEVEL_RED)
+			alert_cap_multiplier = 1
+
+	var/sec_cap = FLOOR(sec_crew * alert_cap_multiplier,1)
+
+	return max( min(crew_cap,sec_cap), ANTAG_CAP_FLAT )
 
 /// Whether events can inject more antagonists into the round
 /datum/controller/subsystem/gamemode/proc/can_inject_antags()
--- a/modular_zubbers/code/modules/storyteller/gamemode.dm
+++ b/modular_zubbers/code/modules/storyteller/gamemode.dm
@@ -236,23 +236,20 @@ SUBSYSTEM_DEF(gamemode)
 		return 0
 	if(!storyteller.antag_divisor)
 		return 0
-
-	var/current_population = get_correct_popcount()
-
-	var/crew_cap = FLOOR(current_population / storyteller.antag_divisor,1)
-
 	var/alert_cap_multiplier = 0 //Anything above red means 0, like Delta and world ending events.
-	switch(SSsecurity_level.get_current_level_as_number())
-		if(SEC_LEVEL_GREEN to SEC_LEVEL_ORANGE)
-			alert_cap_multiplier = 1.5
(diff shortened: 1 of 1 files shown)


Hislyna:
So one issue I see with this is how amber and orange can be used in conjunction with eachother (because you can't use both together and is dumbbbb), cool though to prevent things just getting way too much

ReturnToZender:
If you're worried about potential abuse, It might be best to remove the middleground of Amber as Red has the strongest administrative enforcement, while leaving everything else the same

BurgerLUA:
> If you're worried about potential abuse, It might be best to remove the middleground of Amber as Red has the strongest administrative enforcement, while leaving everything else the same

We can see how it will go. I think if this PR is testmerged, people will be more aware of the current alert level.

Kroto-Ker:
Should department sec count for this? They aren't full sec but definitely still pose a real threat, I've been investigated / killed by them plenty of times as antag. 

Also do keep in mind that about 50% of antag players per round do nothing, obviously that's not every round, sometimes more do and sometimes less. But it's USUALLY about 50%. If this drastically reduces antagonist spawns, we might need to change rules around allowing people to do nothing / friendtag with their roll.

Kroto-Ker:
> Also friendtagging is already sorta looked down upon, and is straight up against the rules if you're helping security catch other antags (outside major round ending circumstances)

It's looked down upon but also extremely common. Like I said, on average about 50% of antagonists per round don't really do anything.

NargaInDisguise:
> > Also friendtagging is already sorta looked down upon, and is straight up against the rules if you're helping security catch other antags (outside major round ending circumstances)
> 
> It's looked down upon but also extremely common. Like I said, on average about 50% of antagonists per round don't really do anything.

Not doing much of anything is not friendtagging.

Sir-Lagsalot:
> > Should department sec count for this? They aren't full sec but definitely still pose a real threat, I've been investigated / killed by them plenty of times as antag.
> > Also do keep in mind that about 50% of antag players per round do nothing, obviously that's not every round, sometimes more do and sometimes less. But it's USUALLY about 50%. If this drastically reduces antagonist spawns, we might need to change rules around allowing people to do nothing / friendtag with their roll.
> 
> They're under zero circumstances actual security, so tbqh, no. Security very rarely ever deputizes people (which wouldn't count towards "sec count" anyways) and dep guards + orderly don't have actual seclathe access normally. They're also not supposed to validhunt outside their department unless its red, then everyones sec anyways and it doesn't matter.
> 
> Also friendtagging is already sorta looked down upon, and is straight up against the rules if you're helping security catch other antags (outside major round ending circumstances)

Worth clarifying: guards absolutely have access to the normal sec lathe on every map even during high pop, it's the equipment room they can't get into
Though if you didn't know this, I guess that implies the access isn't abused too much? I mostly just use it to print ammo if I've got a ballistic gun, but I know guards get access because most rounds I swing by to grab a seclite for my disabler (and a flash if I'm bouncer, since service has no guard post)

BurgerLUA:
> Should department sec count for this? They aren't full sec but definitely still pose a real threat, I've been investigated / killed by them plenty of times as antag.
> 
> Also do keep in mind that about 50% of antag players per round do nothing, obviously that's not every round, sometimes more do and sometimes less. But it's USUALLY about 50%. If this drastically reduces antagonist spawns, we might need to change rules around allowing people to do nothing / friendtag with their roll.

In my experience, departmental security are closer to non-security than security in terms of station involvement with antagonists.

PR #4246, opened 2025-07-15
Adds a quirk that makes you unable to squish micros. Makes bumping into micros STUN them for 1 second instead of just a knockdown

HometownFunky:
## About The Pull Request

Oh god I think I'm doing this correctly, but uhh. This PR does TWO things. 

First and foremost, it replaces the KNOCKDOWN (Drops items, falls down) from bumping into a micro with a STUN (Drops items, cannot move, cannot act) for the same duration of time. 

It also adds a new 0 point quirk, GENTLE STEP, that lets you walk on top of micros laying down without the risk of injury.

## Why It's Good For The Game

Saw this:
saw this

This makes the whole micro squish and bone crushing mechanic completely opt-outable. On the last micro PR it was suggested to tie this to light step, but I feel like someone shouldn't have to take a positive points quirk just to avoid seeing that type of interaction. 

The change to STUN is to allow smaller characters to trade places with larger characters, at the cost of getting STUNNED instead of just knocked down. Do tell me if this is an okay change, I can revert it easily.
## Proof Of Testing

it works
Compiles and works. I hope. I'll cry if it doesn't.

## Changelog
:cl:
add: New quirk that makes you unable to squish micros.
balance: changes the KNOCKDOWN to a STUN when bumping into micros.
/:cl:


HometownFunky:
Why is it failing : (

PR #4248, opened 2025-07-15
Adds new MOD module for proteans to buff their wearer - Protean servo module

amsy2:
## About The Pull Request

I love proteans but there's lack of stuff to actively do in MOD form, this PR gives some buttons while hanging out on back of their coworkers.

Module needs to be activated with MOD suit like other toggle modules, it will not activate if protean tries to use it in human form. Once it's on, protean inside is granted 3 abilities with shared cooldown that will give temporary buff to whoever is wearing the MOD once used. The options
-movement speed boost and clumsy trait for short duration 
-faster surgery, quick carry and faster medicine administration for moderate amount of time
-faster building and action speed for moderate amount of time 
Using one will put all on 60s cooldown

Intention of the feature is to let Protean contribute to whatever the wearer is doing at a time without big need for coordination, strong enough for effect to be noticeable, but without making it game changing. Movement speed option was added since it's probably most universally appreciated effect player might receive, but I also know it's very strong for gamer moments, hence the temporary clumsy trait to try and balance it out.

## Why It's Good For The Game

Proteans currently don't have much to do purely in MOD form, this adds a way for them to contribute to gameplay while staying true to their species gimmick. It is two whole players walking around together, feels like its fair to let them do some tasks faster then single person.

## Proof Of Testing

image
image
image

## Changelog

:cl:
add: Added protean servo module for their MOD that lets them buff their wearer 
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/customization/species/proteans/protean_servo_module.dm
+++ b/modular_zubbers/code/modules/customization/species/proteans/protean_servo_module.dm
@@ -122,7 +122,6 @@
 		ADD_TRAIT(thing, TRAIT_NODROP, PROTEAN_SERVO_TRAIT) //to prevent items droping out of wearer's hands
 		RegisterSignals(thing, list(COMSIG_ITEM_DROPPED, COMSIG_MOVABLE_MOVED), PROC_REF(clear_servo_trait)) //In case you somehow drop it anyway
 	owner.add_traits(list(TRAIT_RESTRAINED),PROTEAN_SERVO_TRAIT) //main drawback to balance out this particular effect
-	return ..()
 
 /datum/status_effect/protean_servo_movement/on_remove()
 	. = ..()
--- a/modular_zubbers/code/modules/research/designs/mechfab_designs.dm
+++ b/modular_zubbers/code/modules/research/designs/mechfab_designs.dm
@@ -171,6 +171,3 @@
 		/datum/material/glass = SHEET_MATERIAL_AMOUNT,
 	)
 	build_path = /obj/item/mod/module/protean_servo
-	category = list(
-		RND_CATEGORY_MODSUIT_MODULES + RND_SUBCATEGORY_MODSUIT_MODULES_GENERAL
-	)


Hislyna:
Cool as fuck

Hislyna:
> I'm not really a fan of metal based buffs / debuffs, especially with the state of our material economy. There's no shortage. It's just you have some or you don't have any due to the amount vents give. If it being power is a concern, locking it in the tech tree would serve a better purpose as a time gate.

True true, I just love the idea of metals doing stuff ;p

Hislyna:
> > If it being power is a concern, locking it in the tech tree would serve a better purpose as a time gate.
> 
> I want the module to be easy to get as basic thing proteans can do if they want so I'd rather just tweak the speed buff. Reduced speed modifier just now and added 1s to duration, I'll be keeping an ear out for any feedback about the power level and its use to balance it out if that's ok.

Its also not free for proteans, takes up some of their space, so a good way to balance it is just the amount of complexity it has!

amsy2:
Yeah also that, lot of space to balance it out, I still feel like it's fair purely couse it's two player requiring action

SapphoQueer:
This is the kinda thing that would be good for proteans to be ideally, but its also adding even more onto an already really strong species.

I wish they leaned into the support aspect more.

Kroto-Ker:
How fast is the movement speed buff compared to other similar speed buffs? I worry about security having access to an ability like that, something usually reserved for antagonists or stimulants. Especially since it's meant to be used with a modsuit that can be armored. Clumsy is a somewhat good mitigation but certain attacks do still work through it.

amsy2:
@Kroto-Ker Unless I missed something reading the code, light-pink stable extract or sphere module have movement speed modifier of -0.5 , this module has -0.4 which is cocaine or DNA vault buff and lasts 7 seconds

Action speed is at -0.35, best comparison i can find is kronkaine which gives -0.5. Honestly was pretty iffy about that one but I just couldn't figure out better way to buff building speed (construction speed buff from stuff like tinkerer gloves does not apply to most construction things ironically enough) but it do be two players in one body so felt justified

KazooBard:
I don't think this PR is balanced well enough to be added in it's current state. The movement speed is where my qualm lies the most with how clumsiness and a 60s cooldown for 7 seconds of movement speed increase are too strong for a roundstart thing that Is Just There.

**The downside of clumsiness can be succesfully avoided in it's entirety by:**
Most murder-happy weapons such as eswords, armblades, heretic blades, spells are unaffected last time I checked
Pairing it with martial arts also works
Pairing it with most decent melees will also do it wonders, not even starting on forged ones

On the crew and security side:
The bokken whilst rather niche still functions as a stunbaton lite last I checked with dealing stam on hit (0.4 movement speed is worth that I'd say). 
Bolas, throwables, grenades (maybe grenades had some sort of clumsy interaction)
Shove combos, which are not present in light pink's buff due to a lack of aggrograb

Comparing it with the stabilized light pink I'd say they're about evenly cut with clumsiness not being all that bad when you still retain a wide arsenal of options. 

**The cooldown + duration:**
Yes, it's only 7 seconds, the ttk for most of the aforementioned weapons is however 5-7 hits with 7 hits being a little less than 7 seconds if the player reliably times it. That means for the approximate duration of an entire fight with 1 person you are at an enormous advantage if you do anything but run in a straight line at them. Guns are a big part of the game but melee is just as reliable if not more in places, especially when it comes to actually hitting someone at extra speeds (which goes both ways, actually).
It's still a large enough boon to be the make-or-break on it's very own. I don't think this sorts of power should be handed out for no more effort than partnering up, as the amount of coordination to fighting 2v1 in opposition to having a backpack protean 2v1 is greatly increased (For proof name 1 fight sec has without friendly fire, bodyblocking eachother, etc.).
This means more gain for less work.
Furthermore, I wouldn't say 60s is a tiny bit above the standard fight durations between sec and antags as is. Furthermore, god forbid backup doesn't arrive in 60s because this benefit is universal enough to help with defense as well as offense. You cannot go wrong with this. 

**Concluding:**
I don't think what functionally looks like temporary xenobio extract is fit for a roundstart boon a race can grant anyone, especially when it's speed of all things. Speed is the strongest statistic in the game. It's why beartraps and bolas end fights while meth makes security players sigh in sync. 

I say either leave the speed out as the others are probably going to be fine, or make the speed thing induce a way longer cooldown, I am confident in saying 8-10 minutes is justified for something that I will underline again after reading through the PR 3 times sounds to be a roundstart thing. There's a chance the cooldown is actually relevant with this duration.

amsy2:
@Kroto-Ker @KazooBard 
Combat was never primary purpose of the module, speed movement is just something that is universally enjoyed by crew with daily activities which is why I aimed to add it, something that you can click button for and be noticeable and useful no matter the department. Clumsy as downside was supposed to make it focused towards non combat uses, though I didn't know it's bypassed by that many weapons. 

Increasing the CD to 8-10 min feels bit too long, though I'm trying to have the cake and eat it too so question to you is: What do you feel would be appropriate downside for current effect and either current CD or closer to 5min? I just need ideas, I don't engage in combat often so I'm bit lacking in knowing stuff that affects it. First thought was to just give Pacifism like light-pink extracts do but I'm well aware that one can be bypassed in many ways too. Ofc there's also coding aspect too, much easier to just slap a trait then code custom behaviours but I will try

Hislyna:
> Alright, I'll look into no item usage today then. I considered making you drop items while its active but that felt like potentially too annoying to coordinate, didn't even think of just preventing item use

Just make it pacify you like the light pink slime does with speed

amsy2:
Ok, I ended up giving wearer full on restrained for duration of the movement buff but the status effect also gives nodrop to held items for duration so they don't just drop when Protean sues the ability without warning

projectkepler-ru:
i wont lie i like the idea of making protean the ultimate TF2 Comp Medic experience where  I need to not only choose who I stick to but also make sure they're actually actively useful or that I'm not wasting opportunity cost

but I am also a highlander player so bite me!

latin313:
It would be really cool to have some sorta "Assisted Exosuit" one as well, mainly focusing on things like cargo or heavy-lifting.

amsy2:
@latin313 Speaking form experience? Thought 20 sec was enough but out of all buffs surgery one shouldn't be an issue to buff up

PR #4292, opened 2025-07-22
Gives proteans the expanded storage module roundstart.

SapphoQueer:
## About The Pull Request
Gives proteans the expanded storage module roundstart.
## Why It's Good For The Game
The default mod storage is extremely limiting for something that is Not Leaving Your Back, Ever. The other way of having a modsuit you can't remove is entombed, they **DO** get this expanded storage.

Either you have to go to cargo to order one roundstart for 100 of your roundstart credits (added tedium for something that shouldn't have to be the case). Or you wait for robotics to research it. (takes forever.) The wait would be a lot more reasonable if it were a choice for you to wear it with being able to swap back to your regular bag, however, this is not the case for proteans.
## Proof Of Testing
yeah i loaded it on local and it was expanded
<details>

</details>

## Changelog
:cl:
add: Proteans now start with the expanded storage module instead of the regular one.
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/customization/species/proteans/_proteans_species.dm
+++ b/modular_zubbers/code/modules/customization/species/proteans/_proteans_species.dm
@@ -141,7 +141,7 @@
 
 	var/obj/item/mod/module/storage/storage = locate() in species_modsuit.modules // Give a storage if we don't have one.
 	if(!storage)
-		storage = new()
+		storage = new /obj/item/mod/module/storage/large_capacity()
 		species_modsuit.install(storage, owner, TRUE)
 
 	if(outfit.backpack_contents)


SapphoQueer:
> I consider this one of their downsides, which is why I've purposely not given them the expanded module regardless if it's orderable via cargo.

I think it's a bit of a weird downside to have, I get the intent for having lower storage, however what it does in practice is very minor. It's not fun gameplay to feel like you have to go to cargo/robotics every shift just to get the equivalent of a regular backpack. This is very rarely optional, backpacks carry a signficantly larger amount of stuff. Regular mod storage units have a maximum of 7 items and a total of 15 maximum combined weight, compare this to the backpack, which has 21 in both of those numbers. I do not reasonably consider it fun gameplay for any side here, protean, robotics, whatever, to have to go out of their way just to match a default backpack.

SapphoQueer:
Adding it to goodies with the other PR

PR #4300, opened 2025-07-24
Reduces damage taken from disposals by 60%

SapphoQueer:
## About The Pull Request
title.

## Why It's Good For The Game
Disposals as it stands is quite horrible to be shoved within, what is a potential emergency exit for dire situations, is currently a one-way-ticket to crit, and on some stations with particularly long pipes (nebula...), death. This change causes it to still be dangerous- it definitely isn't a get-out-of-jail-free card with these changes, but it leaves it as a somewhat viable option. It's fun to use it as a "oh shit" option, or a way to get rid of an assailaint without damn-near killing them. There still being damage means it cannot be spammed either. Would you rather get fucked up by pipes in an effort to escape john antag, or get REALLY fucked up by john antag?

There was a conversation a few days ago in one of the discord channels (either contributorbus or balance-channel) about this, and the vast majority of people there agreed that if not a full removal, a reduction would be nice.

There additionally are a few ruins (and a bitrunning domain) that use disposals, in its current state the clown one will leave you on red health, from full health.

## Proof Of Testing
1 number change

<details>
<summary>Screenshots/Videos</summary>

</details>

## Changelog

:cl:
balance: Reduces damage taken from disposals by 60%
/:cl:

PR #4304, opened 2025-07-25
Gets Rid Of Disposals Damage

SapphoQueer:
## About The Pull Request
No more half measures walter

## Why It's Good For The Game
A vast amount of the reasoning from #4300 still applies here, if you want to non-lethally eject someone deep into your department you would usually just shove them down disposals, but alas, they shall die or go into crit if you do that in its current state. 

Some have stated the argument of "being spammable" however, if an antagonist is repeatedly doing this as an emergency exit to escape security... there's only one place they could possibly pop out in, and that will make it easy for security to grab them. 

Another argument I have overheard against this is that it would allow people to break into cargo with ease. However, on the vast majority of maps (idk about biodome and I can't be fucked to check), the room the tubes lead into is one seperated from the rest of cargo, typically something which holds a few belts, shipping supplies, and a second tube to maintenance disposals. 

Upstream content such as select bitrunning maps as well as ruins are not designed with damage in mind, these will do significant damage to you, and in the case of the clown ruin leave you near-death.

This also opens up things like people with RPDs using their disposal tools to do more than 1: fix disposals when it breaks or 2: bounty cube transit system. As it can transport people, there's a lot of fun potential to create under-floor fast travel methods. Maps like metastation already have a lesser version of this, with the prisoner chute at perma. Yes, this would be comparable to the transit tubes that already exist, however, with those it is typically very clear where they are going, and they can be *instantly* exited. The inherent downside to having the path be hidden and inescapable is that it can be abused by antagonists to lead people into some form of trap. Disposals pipes can still be escaped from by pressing the resist button with both a chance to stop, and then a doafter, so infinite loops wouldn't be an issue either.

Also shoving clowns into disposals is very funny

## Proof Of Testing

https://github.com/user-attachments/assets/0917e020-1de0-4cbe-aad9-3560f775d563

<details>

</details>

## Changelog

:cl:
balance: Erases disposals doing damage entirely.
/:cl:


IgiariValkyr:
I think we should stop babying people who get thrown into disposals. These are hard metal pipes, not padded slides in a fucking kiddy playplace. Think about people who use disposals to escape security. Do you really want them to come out of the other side perfectly fine now?

KateAnbolho:
> I think we should stop babying people who get thrown into disposals. These are hard metal pipes, not padded slides in a fucking kiddy playplace. Think about people who use disposals to escape security. Do you really want them to come out of the other side perfectly fine now?

it's not really babying when it was added on with skyrat, not tg

SapphoQueer:
> I think we should stop babying people who get thrown into disposals. These are hard metal pipes, not padded slides in a fucking kiddy playplace. Think about people who use disposals to escape security. Do you really want them to come out of the other side perfectly fine now?

I cover this in the big wall. If you use it to escape security, there are only two possible locations for you to pop out of. A locked room in cargo, or a room in maintenance. It makes you an extremely easy target for any sec team that's even quarter-organized. 

> I agree with its damage but it can be fixed for things like the mail module on the modsuit and bitrunning avatars. No, I don't think you should be able to flush yourself down disposals without some consequence.

I think I'd agree more with this if there weren't applications for tubes beyond just the main disposals loop. The transportation machinations example I pointed out in my large wall of text above is just one of many things, and it's even one already supported by more than one map. The image in the dropdown below is just one example, I know voidraptor also has something like this but smaller. The range of applications for The Tubes can be greater than just disposals, yet it is severely hindered by the damage.
<details open>
<summary>Dose this contane a mysteryy..... click to find out...</summary>
image
<br>
</details>

Razurath:
It seems weird, but I've never been on a server that didn't have damage when going through disposals. It just always made sense that if you go into the pressurized trash chute inside a metal pipe, it's gonna fuck you up without protection of some kind. Even if it does wounds rather then flat damage. If you want transport in a tube, we have the glass transit tubes for that. They don't go under flooring, but they are also made for safe transport.

Stonetear:
This is a good change.

Possible alternatives:
* Make it so that any helmet that provides head protection will also protect you from damage.  (Assistants can easily grab the firefighter helmet)
* Activated modsuits with the disposals selector mod protect you from damage.
* Make it so that disposals damage can't make you more hurt than a certain amount.

SapphoQueer:
> Then you could build your projects that were hampered by pipe damage, and the greyshirt who jumps down disposals in security can still take (some) damage.

I like this idea in concept, however 

1: I have no clue how to do that lmao
And
2: I think the predictability of "there is only one place you can end up" is plenty of punishment in the vast majority of cases.

NanoCats:
Don't get me twisted I'm in favor of just removing it, however; even if we didn't remove disposal damage I'd want it reworked to be damage that is in line with the physics implied.

1. If the damage is from pressure, wearing a space suit should nullify it
2. If the damage is from impact with corners, melee armor should defend against it

Currently, disposal damage is "true" damage, which means it'll just fucking _GET_ you no matter what.

SapphoQueer:
I will get to the rusted pipes eventually:tm:

PR #4305, opened 2025-07-25
Ports oversized sprite toggle - click behind your own sprite.

KazooBard:
## About The Pull Request

Characters with the oversized trait get an ability that changes their sprite to a normal size **on their own screen**. This allows them to fight as normal and avoid hitting themselves, whilst maintaining their character's looks for everyone else.
## Why It's Good For The Game

An oversized sprite can reach 2 tiles of height, meaning that anyone directly above your character is unable to be interacted with without using the alt click menu on the tile itself. This goes for structures, machines, tables, even wall-mounted equipment. This is unnecessary for a neutral quirk (which arguably has more downsides anyhow). 

Ranging from situation to situation, this can either be a mild annoyance in filling out paperwork or cost you fights if you commit the sin of wanting a big character as a security memeber, antag, or existing nearby a monkey, as fighting anything above you becomes impossible once they attempt to exploit your stature to hide behind your sprite. 

## Proof Of Testing

https://github.com/user-attachments/assets/38a86642-bfff-400b-b32d-3d06e4fedcc3

obraz

I see no runtimes past the saddles and workbench which have been there for ??? time already

## Changelog

:cl:
qol: Oversized characters can shrink their sprite on their own screen (to 100% size)
/:cl:

### DIFF
--- a/modular_zubbers/code/datums/quirks/neutral_quirks/oversized_quirk.dm
+++ b/modular_zubbers/code/datums/quirks/neutral_quirks/oversized_quirk.dm
@@ -30,6 +30,8 @@
 	human_holder.blood_volume_normal = BLOOD_VOLUME_OVERSIZED
 	human_holder.physiology.hunger_mod *= OVERSIZED_HUNGER_MOD //50% hungrier
 	human_holder.add_movespeed_modifier(/datum/movespeed_modifier/oversized)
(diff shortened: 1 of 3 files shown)


SapphoQueer:
Stores the large size to be used as the small one, but only once. Any alterations to the sprite when in small-vision doesn't appear as you do them. Might want to find a way to get it to be dynamic?

PR #4308, opened 2025-07-25
heretics are getting a rework so let's make them more likely

Arturlang:
## About The Pull Request
gives heretic a weight of 9, above lings by 1, and ups them to max 4,
make sure this is TM'd with the rework cause it's a bit pointless withotu it.
TM only, unless...?

## Why It's Good For The Game
i wanna play heretic and not RNG the fuck out of it, since the rework may get TM'd

## Proof Of Testing

<details>

</details>

## Changelog

:cl:
add: heretic weight is now 9 and has a maximum of 4 heretics
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/storyteller/event_defines/crewset/heretic.dm
+++ b/modular_zubbers/code/modules/storyteller/event_defines/crewset/heretic.dm
@@ -4,10 +4,10 @@
 
 	antag_flag = ROLE_HERETIC
 	antag_datum = /datum/antagonist/heretic
-	weight = 3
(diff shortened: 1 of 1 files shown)


projectkepler-ru:
finally

I can roll heretic

Kroto-Ker:
Will we keep the removed heretic midrounds of the rework? Personally I think that change fits better on TG than here.
Even if we have increase heretic weight, they only have a chance to spawn roundstart when at least 35 people are readied up which, from what I've seen, only occurs about once a day if at all.

PR #4316, opened 2025-07-27
Adds Security Missile Launchers (real)

BurgerLUA:
image

## About The Pull Request

Adds Security Missile Launchers. These are special missile launchers that have a custom snowflake code active radar system that seeks out targets based on their perceived density, somewhat similar to how IRL radar works.

Codewise, while the missile is in flight, it will scan targets in a cone in front of it, and do a weighted random selection of which target to home in on. For walls and turfs, it is based on its heat capacity (generally, stronger materials have higher radar signature). For objects and mobs, objects with higher health are more likely to be chosen, thus it should only be used in very wide open areas, such as space, as it is extremely likely the missile will target a nearby vending machine, a wall, the neighbor's dog, or even the floor.

Once the missile hits, it will create a custom snowflake explosion that calls the explosion proc on non-living objects and just apply bomb damage to living objects. This is to ensure that the missile launcher doesn't delimb and ruin an antag's round. Against living targets, it applies a 2 second knockdown.

One missile launcher will exist round start, in the armory, along with 4 missiles.  Additional missile launchers can be ordered for 50,000 credits each. A box of 5 missiles can be ordered for 10,000 credits each.

Keep in mind that you can't just walk up to someone and point blank a target with it, avoiding the drawbacks of the weapon. The missile must travel for some time before it actually arms. Hitting someone too close will cause toolbox-tier damage, and stun them for 2 seconds.

## Why It's Good For The Game

The intent here is to actually give security some fighting chance against space tiding. As it currently stands, security is at a MASSIVE disadvantage in space, due to:
- Limited security modsuits round start.
- Jetpacks being able to travel faster than bullets/lasers/etc.
- Antagonists being designed to 1v2 security officers, and win.
- Antagonists having built-in or easy-to-access EVA capabilities.
- Inherently low view range, making scouting difficult and easy to get caught in an ambush.
- Easy to access modsuit modules, jetpacks, and modsuits themselves via cargo and robotics.
- In space, no one can hear you complain about being stripped and left for dead on another z-level.

The intent of a missile launcher is to even the playing field against spacetiding antagonists without having to directly nerf spacetiding antagonists. While this won't stop spacetiding, it will require spacetiders to dodge missiles and stick closer to the station in order to bait missiles into hitting walls.

Don't expect this missile launcher to work well in hallways, or near multiple targets. As the American Declaration of Independence once said:

> "We hold these truths to be self-evident, that all men are tracked by missiles equally."

## TODO

- [x] Move all the code to the modules folder instead of keeping everything in one file.
- [x] Add custom effects like smoke, light, and a missile ignition sound.
- [x] Add a reload time to the rocket launcher so you can't just rapid fire them. (Not possible, added a fire delay of 2 seconds).
- [x] Fix the angle code and figure out if an angle of 0 means north or east.
- [x] Add a (nerfed) backblast to it.
- [x] Make the weapon obtainable via armory.
- [x] Make the ammo obtainable via cargo + armory.
- [x] Add emag interaction to the rocket launcher (radar will instead loop around back to the user after being fired).
- [x] Make a better ammo icon sprite.
- [x] Make the missile run radar seeking code the moment it is created so it isn't a normal missile for 1 second.
- [x] Balance the missile (make it slower, deal less damage, add an HP cap for mobs so large HP mobs like space dragons aren't dicked, etc)
- [x] Add a debug variant that has like a magazine of 50 rockets for lag testing.

## Proof Of Testing

Tested extensively :tm:

https://github.com/user-attachments/assets/780ff8d0-dc5d-445f-b34c-0e3c3878faa9

## Changelog
:cl: BurgerBB
add: Adds Security Missile Launchers. These are special missile launchers that have a custom snowflake code active radar system that seeks out targets based on their perceived density, somewhat similar to how IRL radar works.
/:cl:


projectkepler-ru:
What the fuck burger

KateAnbolho:
but secmods get ions by default now

Razurath:
Should give this a minimum engagement range. If you point blank someone it will still hurt, but the safety of the user it won't be able to explode. This will keep folks from trying to run up and point blank an antag hoping they can tank the explosion better.

BurgerLUA:
> Should give this a minimum engagement range. If you point blank someone it will still hurt, but the safety of the user it won't be able to explode. This will keep folks from trying to run up and point blank an antag hoping they can tank the explosion better.

Agree. Will implement this.

BurgerLUA:
> Should give this a minimum engagement range. If you point blank someone it will still hurt, but the safety of the user it won't be able to explode. This will keep folks from trying to run up and point blank an antag hoping they can tank the explosion better.

Implemented. Close range impacts will result in duds.

sippykot:
50k is like "cargo buys this to flex" territory

nikothedude:
casual burger w

BurgerLUA:
As I said in discord, there aren't really many good identifiers for a void ling other than its type. I want to avoid using semi-expensive checks like istype for checking if it's a void walker.

nikothedude:
that sounds like overoptimization, checking an istype every tick is really negligible if its just caused by one entity that will exist for a fraction of a round

BurgerLUA:
Better idea thought up. I could just implement a space tech bullshit thing that deals more damage to armor (like modsuits) while dealing less damage to organic material.

BurgerLUA:
I discovered during testing a bit ago that the missile launcher barely deals damage to simplemobs because of how explosions work. Humans take a lot of damage from explosions compared to simplemobs, so this should be balanced against Void Walker and Space Dragon.

Jungling:
Make it do a stamina damage proc in an aoe around the hit point, only work in low/no pressure (aka space). Also make it emaggable to work ON the station so antags can interact with it.

BurgerLUA:
> Make it do a stamina damage proc in an aoe around the hit point, only work in low/no pressure (aka space). Also make it emaggable to work ON the station so antags can interact with it.

I have future plans when this gets merged.

I was going to design a Security Engineer role that also has access different payload types and whatnot, including non-lethal and quick breach repair options.

babamandaa:
a two second stun for hitting a dud rocket at melee range? 

i-think-you-should-leave-tim-robinson

PR #4331, opened 2025-07-29
(READY FOR MERGE) Adds a new species: Lycans (and cursekin)

xPokee:
## About The Pull Request
Adds a new species, lycans. They start as a cursekin, but can transform between lycan form and cursekin form at will via an action button, with a three-minute cooldown between swapping forms. **You need to select ears/tail as a cursekin or your lycan form won't have them. (You can toggle them off in-game via the show/hide mutant parts verb)** The secondary mutant color for body color also changes the lycan form body color.

### Cursekin Form

In cursekin form, you are pretty much comparable to a baseline human, albeit with far more customization due to opting for mutant colors over skin tone. At this time, the one notable difference is that your body will reject any inorganic organs/implants that may be attempted to be inserted.

### Lycan Form

In lycan form is where you really start to see the differences from a human, some key differences are:
- You now have built in armor, but, are unable to wear clothes, armour, or any gear.
- Your punches do damage equivalent to a butcher cleaver (minus AP), but you cannot use guns.
- You are oversized
- You have night vision and better hearing, but this comes at a cost as you will take extra damage from flash bangs, on both the eyes and ears fronts.
- You have chunky fingers and are illiterate.
- You get the hunger increase of vulps.

### To Do:

- [x] Code in all the features outlined by the bounty issuer.

- [x] Polish the code.

- [x] Fix any remaining bugs.

- [x] Have the PR test-merged.

- [x] Fix any new bugs that come up.

- [ ] Full merge.
## Why It's Good For The Game
It's a whole new species with mostly unique mechanics not seen by other species. The transformation aspect is not seen in any other species unless you count slimes, and the cybernetic rejection is not used by other species in this exact way to my knowledge. This lets players play as a species capable of transforming between two species, which makes for some interesting possibilities.
## Proof Of Testing
Everything seemed to work on my end.
## Changelog
:cl:
add: added a new species, werewolves (and werehumans)
image: added icons for the werewolves from Monkestation
/:cl:


YakumoChen:
Do hulk punches still stun?

Regardless I very much question if we should be having a race with on-demand hulk smash (minus walls, but still works on doors) with 50% brute DR and 20% burn DR

xPokee:
> Do hulk punches still stun?

I don't believe they do, if they do, it will be removed. 

> Regardless I very much question if we should be having a race with on-demand hulk smash (minus walls, but still works on doors) with 50% brute DR and 20% burn DR

Keep in mind that when you transform into a werewolf you drop EVERYTHING and are unable to wear armor/clothes/anything (including bags), so the built-in armor is more-so that they don't instantly die given the lack of gear.

YakumoChen:
Actually thinking more. This is colossal validhunter-coded as a race and especially as a mutation genetics can just copy and pass out. 

Toggleable hulk removes almost all the parts that make hulk unappealing (it's ugly, you're stuck with it until you mutadone) and hulk is the bane of antagonists like AI that is criminally underutilised for the above reasons

xPokee:
> Actually thinking more. This is colossal validhunter-coded as a race and especially as a mutation genetics can just copy and pass out.

Werewolves are also oversized by default, and as such will not fit in DNA chambers, but if they do manage to get a workaround, I'll add whatever DNA vault mutations have so it can't be saved.

xPokee:
> What does this add to the game beyond giving players an innate ability to hulk out and be a massive combat threat? How does that contribute to roleplay?

Massive combat threat is an overstatement, while yes, they are better at unarmed combat, they are not much better at melee combat, and most of the hulk things don't apply aside from being to bash doors/windows.

As for how this contributes to roleplay, the same could be said for almost any race in the game. How does being a reskinned human contribute to roleplay?

YakumoChen:
> Massive combat threat is an overstatement, while yes, they are better at unarmed combat, they are not much better at melee combat, and most of the hulk things don't apply aside from being to bash doors/windows.

You take half damage from most things and have massively damaging unarmed attacks, that almost double if you can get someone floored, you can smash doors open, and are generally just as much if not more of a menace than hulk still is, with the added bonus of not being ugly like hulk and having it on a toggle.

SapphoQueer:
I generally really like the idea of this, even if it could have a bit of a balance pass. The concern of breaking airlocks is a valid one, but I'd rather have something with unique mechanics as opposed to Human But Different.

ReturnToZender:
This feels like its design fails to allow players to roleplay a werewolf. Presumably, the idea would be to have a command to, at will or on a trigger, switch your species to a werewolf species that is unique, from your origin species.

This implementation feels like a hulk ability on a cooldown rather than a species. Also, lycanthropy is a disease and both implementations are probably more befitting a quirk

snailomi:
To me, this idea feels more in place as a genetics infusion sort of thing- or some antagonist ability. Thinking of it, with how comparisons are drawn to hulk, it's comparable to a heavily weakened version of the wizard's 'mutation' spell. Even if that isn't the intention (I don't believe it is), I don't think the similarities can be denied. As for the statement of the species being more unique than "human but x", I might be missing something big but how does this deviate from that? It seems like "human but x(can turn into werewolf temporarily)" especially when it's said at the start of the PR that they start as a weaker _human_

YakumoChen:
I'd rather see this as an antagonist, since this is essentially what it seems like it's built to be essentially a combat species as opposed to something unique

If we're gonna push this I would rather the transformation be much less related to combat rather than just giving a human a toggled juggernaut ability 

Items should be saved to contents instead of dropped (i can already see this being the first qol they get)

This ability shouldn't be a free toggle, it should be like bloodsucker where its inevitable and they have to plan around it, with something like Blob spawning where the user can delay it at some cost if its inconvenient to do so immediately (like mid-RP)

Stonetear:
Can you try to avoid explicitly theme'ing it as werewolves so that people can make their secondary monster form anything?  (Kaiju, dragon, ect...)  😃

Razurath:
The species list doesn't actually tell you what the advantages or disadvantages of the lycan form and would be helpful if it did.

latin313:
Another issue I have found.. when transforming it doesnt take in any account of your previous size (meaning you can be micro while being 2x size and be stepped on.) but also changing back turns your sprite back to 1 times scale, even if you had oversized (meaning you could be really small but be "oversized")

Aisu9:
Two pieces of feedback.
First one: you turn back in your original form on death
Second: you can pick your wereform(meaning you pick the sprite of the form while not touching any mechanics tied to it, allowing you to be any critters with the same limitation of the transformation)

latin313:
It seems you arent actually given the "oversized" tag/quirk when you grow, meaning if you have the microquirk someone can "step" on you. It also appears that your always set to 1 times scale reguardless if you were a different size/had a size quirk originally

latin313:
> > It also appears that your always set to 1 times scale reguardless if you were a different size/had a size quirk originally
> 
> I was unable to reproduce this. I transformed while having my body size at 0.8 then upon transforming back my body size was still at 0.8.

Yeah i went to try it just now and it seems everythings working just fine, im unsure if it may have been some other PR or change that fixed it but it seems to work fine (including oversized still existing after transforming)

xPokee:
> Okay quick edit, i found out the issue. if lets say you have your size set to 1.2 but have oversized, upon spawning your at max 1.5 size as a oversized as normal. but when transforming back into cursekin form, the game will prioritize the 1.2 scale instead of the oversized quirk one.

At this point, I'm really tempted to just blacklist oversized + micro, at least until I can get a proper fix.

modular_skyrat/modules/customization/modules/mob/living/carbon/human/species/mushpeople.dm
```diff
@@ -1,3 +1,11 @@
+/datum/species/mush
+	inherent_traits = list(
```
Arturlang, line 2: if you need to add a trait or remove it, do it modularly without overriding the var, add or remove to it on New

modular_zubbers/code/modules/customization/species/lycans/_cursekin_species.dm
```diff
@@ -0,0 +1,84 @@
+/datum/species/human/cursekin
+	name = "\improper Cursekin"
+	id = SPECIES_CURSEKIN
+	inherent_traits = list(
+		TRAIT_LYCAN,
+		TRAIT_MUTANT_COLORS, // More customization options.
+	)
+	mutant_bodyparts = list()
```
Arturlang, line 8: why have this  be empty? if it's not doing anything remove it, unless the base type has some?

modular_zubbers/code/modules/customization/species/lycans/_lycan_species.dm
```diff
@@ -0,0 +1,97 @@
...
+	var/obj/item/organ/insert_organ = inserted
+	if(insert_organ.organ_flags & (ORGAN_ORGANIC | ORGAN_UNREMOVABLE))
+		return
+	addtimer(CALLBACK(src, PROC_REF(reject_now), source, inserted), 1 SECONDS)
+
+/datum/species/lycan/proc/reject_now(mob/living/source, obj/item/organ/organ)
+	organ.Remove(source)
+	organ.forceMove(get_turf(source))
+	to_chat(source, span_danger("Your body rejected [organ]!"))
+	organ.balloon_alert_to_viewers("rejected!", vision_distance = 1)
```
Arturlang, line 93: either make lycan and cursekin a parent and subtype or use a element to implement cybernetic rejection

modular_zubbers/code/modules/customization/species/lycans/organs/lycan_brain.dm
```diff
@@ -0,0 +1,55 @@
...
+
+/obj/item/organ/brain/lycan/proc/enter_beast_form()
+	var/datum/species/human/cursekin/current_wolf = owner.dna.species
+	if(!istype(current_wolf) && HAS_TRAIT_FROM(owner, TRAIT_BEAST_FORM, SPECIES_TRAIT))
+		return
+	removed_quirks = list()
+	for(var/datum/quirk/current_quirks in owner.quirks)
+		if(istype(current_quirks, /datum/quirk/micro))
+			removed_quirks += current_quirks.type
+			owner.remove_quirk(current_quirks.type)
```
Arturlang, line 15: add a list of removed_quirks and use locate to locate it from the list

modular_zubbers/code/modules/customization/species/lycans/organs/lycan_brain.dm
```diff
@@ -0,0 +1,55 @@
...
+	if(removed_quirks)
+		for(var/typepath in removed_quirks)
+			owner.add_quirk(typepath, owner.client, TRUE, FALSE)
+		removed_quirks = null
+
+/obj/item/organ/brain/lycan/proc/toggle_beast_form(mob/user)
+	set name = "Enter/Leave Lycan Form"
+	set desc = "Succumb to the rage and turn into a lycan."
+	set category = "Lycan"
+	if(user && !HAS_TRAIT(user, TRAIT_BEAST_FORM))
```
Arturlang, line 47: early return if no user instead of checking it twice

modular_zubbers/code/modules/customization/species/lycans/organs/lycan_heart.dm
```diff
@@ -0,0 +1,36 @@
...
+ * Signal proc for [COMSIG_MOB_APPLY_DAMAGE_MODIFIERS]
+ * Taken from the snail heart and modified for lycans.
+ * Reduces damage taken from brute + burn attacks.
+ */
+/obj/item/organ/heart/lycan/proc/modify_damage(mob/living/carbon/human/source, list/damage_mods, damage_amount, damagetype, def_zone, sharpness, attack_direction, obj/item/attacking_item)
+	SIGNAL_HANDLER
+	if(damagetype == BRUTE)
+		damage_mods += 0.5 // 50% damage resistance on brute attacks.
+	else if(damagetype == BURN)
+		damage_mods += 0.8 // 20% damage resistance on burn/energy attacks.
```
Arturlang, line 36: why not have the heart modify physiology damage mods?

Arturlang (changes requested):

modular_zubbers/code/modules/customization/species/lycans/organs/lycan_heart.dm
xPokee: The physiology datum says `//Stores several modifiers in a way that isn't cleared by changing species`, would that not make it so that the things don't get removed when the heart gets removed on transformation?

PR #4332, opened 2025-07-29
The full-custom blood colour quirk, again

aKromatopzia:
## About The Pull Request
Technically, this is also a bugfix for the removed portion of the evil quirk.

This is a blood colour quirk which aims to create an entirely new blood_type datum per color-modified base type, and acts as a launching pad into developing more blood stuff. 

As an additional note: In our testing injected blood wouldn't override the recipient's blood colour. However if this does happen at all in live, do state such! 

## Why It's Good For The Game

Blood colour good? Near-fully custom blood color better?

## Proof Of Testing
oh hey it works now. The three lizards in the video use red blood, cyan/vox blood, and white/nanite blood respectively.
https://github.com/user-attachments/assets/cec6465f-68de-4c6f-97f4-58061f4f7241

Not rigorously tested, but my process was to only dev the lizard type then copy to everything else... so it should be fine? Just to show it should work on blood types with existing compatible_type fields though - as in, it actually copies the list:
Screenshot_92

## What We Learnt
Trust our gut. We shouldve replaced it ages ago but thought original refactorer knew better

## Changelog

:cl:
add: blood colour quirk. it's mechanically near-identical to the normal type.
code: more blood colour presets. yum.


code/__DEFINES/~~bubber_defines/colors.dm
```diff
@@ -9,8 +9,15 @@ GLOBAL_LIST_INIT(chat_colors_by_mob_name, list(
 	"Unknown" = list("#ffffff", "#d8d8d8"),
 ))
 
+/// Blood colors
 #define BLOOD_COLOR_COPPER "#6E3B08"
 #define BLOOD_COLOR_NANITE_SLURRY "#CCCCCC"
+//alternate lizard blood
+#define BLOOD_COLOR_GREEN "#339933"
```
SapphoQueer, line 16: Instead of this, uncomment the following maybe? Lizard blood color is already set on upstream code.

https://github.com/Bubberstation/Bubberstation/blob/d06eb208e5f0f66a3242210fea965c14bd13f7e6/code/modules/mob/living/blood_types.dm#L246

Shadow-Quill (approved):

Shadow-Quill:
NB: Should take a player save backup before TMing, just in case.

Arturlang (changes requested):
Doesn't compile due to upstream, also double check that it functions, players reported it did not, also is it being both a preference option and a quirk neccesary? It could simply be a pref option

HometownFunky:
Dunno what the policy on necroposting but havn't seen anyone speak out on this, so here goes.  This PR seems to still be TM'd for roughly a month, and still working fine. It's still IN game as of round 10126. Does the autoclose mean it'll eventually get removed? I personally really enjoyed this PR and would love to see the quirk stay.

SapphoQueer:
> Dunno what the policy on necroposting but haven't seen anyone speak out on this, so here goes. This PR has been TM'd for roughly a month, and still working fine. It's still TM'd IN game as of round 10126, two days later after closing. Does the autoclose mean it'll eventually get removed? I personally really enjoyed this PR and would love to see the quirk stay.

Should've stale-b-goned, you're right. Going to ask the other maintainers if anything needed to be done before a fullmerge.

PR #4337, opened 2025-07-30
Fixes pAI and Protean MOD Pathfinder and MOD Chameleon module usage (TM only)

Stonetear:
## About The Pull Request

This makes it so that the MOD Pathfinder module and MOD Chameleon modules work when the suit isn't worn, such as by Proeans in suit form and pAIs.  This is pretty logical since those modules *should* be able to work without it.

Very slightly shitcode since `/obj/item/mod/module/used()` doesn't pass a user, and instead assumes the user is always `mod.wearer`.  It might be a good idea to just make this a testmerge while I add a bunch of `if(!mod.wearer)` checks to everything in modsuit code upstream.

I did my best to test all of the edge cases, but there's still potentially a few bugs in here.

## Why It's Good For The Game

pAIs controlling modsuits is actually kind of fun.  With the Chameleon modules, your pAI can now control your suit's appearance, and Proteans can change how they look while in suit form.  (Although the chameleon module is currently slightly broken.  Also an upstream thing.)

With the pathfinder module, pAIs can now hit the recall button.  Mechanically, this is meaningless since it's identical to the user hitting it.

## Proof Of Testing

<details>
<summary>Screenshots/Videos
proof
</summary>

</details>

## Changelog

:cl: Stonetear
fix: pAIs and Proteans can now use the Chameleon module and Pathfinder modules without the suit being worn.
/:cl:

### DIFF
--- a/code/__DEFINES/mod.dm
+++ b/code/__DEFINES/mod.dm
@@ -22,6 +22,9 @@
 #define MODULE_ALLOW_INCAPACITATED (1<<1)
 /// This module can be used while the suit is off
 #define MODULE_ALLOW_INACTIVE (1<<2)
+/// This module can be activated while unworn
+#define MODULE_ALLOW_UNWORN (1<<3) // BLUBBER ADDITION ADD
+
 
 #define UNSEALED_LAYER "unsealed_layer"
 #define SEALED_LAYER "sealed_layer"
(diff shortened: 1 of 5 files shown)


Stonetear:
Closing because I made a PR on tg that will make this obsolete.

PR #4355, opened 2025-08-02
Reworks protean regeneration

StrangeWeirdKitten:
## About The Pull Request

Makes it so when a protean takes damage, their regeneration has a delay factor to it. It will slowly ramp back up over the course of 30 seconds with a multiplier.

The timer is adjustable to feedback and gameplay. Needs a test merge
## Why It's Good For The Game

Protean regeneration is a bit substantial, I'll admit after playing a few combat roles with them especially if you have armor.

## Proof Of Testing

This is the multiplicative number. 2 * 0.1 will be 0.2 damage healed.

image

## Changelog

:cl:
balance: Protean regeneration now has a delay factor where it slowly ramps up back to normal over 30 seconds
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/customization/species/proteans/organs/protean_stomach.dm
+++ b/modular_zubbers/code/modules/customization/species/proteans/organs/protean_stomach.dm
@@ -1,4 +1,4 @@
-#define REGEN_TIME (60 SECONDS) // () are important for order of operations. Fuck you too, byond
+#define REGEN_TIME (30 SECONDS) // () are important for order of operations. Fuck you too, byond
 /obj/item/organ/stomach/protean
 	name = "refactory"
 	desc = "An extremely fragile factory used to recycle materials and create more nanite mass. Needed to facilitate the repair process on a collapsed Protean; it can be installed as a module in the rig, or as an organ."


Hislyna:
How is this an issue for Proteans who get no other regen, no other way to heal, yet Pod people are fine AND get meds to aid em

Hislyna:
Also antags have so many ways to bleed, if their regens getting nerfed, the bleed should be lessened

Hislyna:
> proteans can't bleed

Slashing them causes their blood to drop immensly, causing them to lose so much power in regenning, I cant believe that to be just from healing the damage

Hislyna:
Ah good!

Hislyna:
So are the pools of nanite slurry just cosmetic then?

Hislyna:
> How is this an issue for Proteans who get no other regen, no other way to heal, yet Pod people are fine AND get meds to aid em. This also makes mining even more impossible, where before you could at least back and survive a bit

Could this be made so that the delay is only caused by players, and not NPCs, or is that too much of a ballache to code in? I still find it odd that Pod People have a pretty considerable amount of healing too (not aaas much as proteans, but theirs is completely free) + can use chems though when proteans can't even weld, and obviously don't have chems

MrZoraman:
> proteans can't bleed

Proteans absolutely can bleed. Yesterday I played a round as a protean where I got a ninja star stuck in me and I bled out.

SapphoQueer:
This is a good change, the amount of sustain in extended fights you had as a protean with any semblance of armor was quite frankly too high for the small cost of eating some metal here and there. Especially when the mechanics that typically force people with sustain to retreat don't apply (wounds).

Could probably make the regen cost for blood be the same with this instead of 6(?)x. You wont bleed a lot but it will be CRIPPLING when it does happen by way of embed or heretic spells.

Hislyna:
This feels fine until you go up against someone with healing chems, which you have 0 counter for. Now they have a liver, how does a chem that negates this in combat sound?

PR #4396, opened 2025-08-08
Forbids some categories of uplink items from receiving discounts + doubles the amount of discounts available.

BurgerLUA:
## About The Pull Request

Changes the amount of discounts available to traitors from 4-6 to 8-12.

Forbids some categories of uplink items from receiving discounts.
Categories include:
- Spy Unique items (Mostly redundant, as all items of this category have discounts disabled)
- Special items. (So far, only a gimmicky autosurgeon implanter that appears only during a certain station trait that I think is disabled).
- Explosive items.
- Dangerous items.
- Bundles (Mostly redundant, as all items of this category have discounts disabled).
- Badassery  (Mostly redundant, as all items of this category have discounts disabled).
- Mech stuff (job-exclusive stuff not counted here) (Nukes Ops Only).
- Extra Reinforcements. (Nukes Ops Only)
- Base Keys (Nukes Ops Only)

## Why It's Good For The Game

From what I understand, discounts were meant to shake up the meta and allow more interesting stuff to happen by offering discounts to random items. Problem is, the discounts can be granted to... very insane items if RNG has its way, leading to questionably powerful antagonists.

Another problem the discounts had is that people feel obligated to buy discounted items, even if the situation isn't appropriate. For example, I recall a round where a traitor bought a syndicate bomb solely because it had a serious discount attached to it, not because they needed it or a gimmick relied on it (See: Like all the psychological studies on sales on how it affects impulse buying).

Since we have 25 TC for traitors, I think it's fair to disable discounts on certain categories.

## Proof Of Testing

Draft because untested and needs feedback.

## Changelog

:cl: BurgerBB
balance: Changes the amount of discounts available to traitors from 4-6 to 8-12.
balance: Forbids some categories of uplink items from receiving discounts.
/:cl:

### DIFF
--- a/code/modules/uplink/uplink_items.dm
+++ b/code/modules/uplink/uplink_items.dm
@@ -4,10 +4,25 @@
 	var/list/sales = list()
 	var/list/per_category = list()
 
+	/* BUBBERSTATION CHANGE START: DISCOUNT OVERHAUL
 	for (var/datum/uplink_item/possible_sale as anything in sale_items)
 		if (!(possible_sale.category in per_category))
 			per_category[possible_sale.category] = list()
 		per_category[possible_sale.category] += possible_sale
+	BUBBERSTATION CHANGE END*/
+
+
+	//BUBBERSTATION CHANGE START: DISCOUNT OVERHAUL.
+	//SEE MODULAR FILE DISCOUNT_TWEAKS.DM FOR SETTINGS.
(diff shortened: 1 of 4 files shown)


NargaInDisguise:
> > add scarp king
> 
> I actually meant to target sleeping carp in this PR but I thought that Sleeping Carp was in a different category.

kill scarp and syndicat ears discounts and my life is yours

PR #4402, opened 2025-08-08
Restores traitor to 20tc instead of 25tc because i died to it once

SapphoQueer:
## About The Pull Request
puts traitor back at 20tc

## Why It's Good For The Game
Traitor, on bubber/skyrat, has had 25tc for quite some time. In part this was due to our lack of progtot, and in other parts due to traitor being quite weak compared to security at its prime. But truth be told, 5tc/20% is a LOT more than it seems.

However, this inflated amount of currency allows for quite a few combos which are less than fun to fight against. The many scarp + (something else) combos are some of the major examples. Whilst it's a major problem player in the economy, it is not worth being the sole balancing factor for cost. Discounts, additionally, do not impact the combos all that much. The raised cost cap does allow for more nonsense. For example, if you get an esword + ebow + noslips, that's 18tc. Under normal, no-discount scenarios, this would prevent you from also getting armor. However, with the raised amount of TC we have, this allows for you also get an infiltrator mod. A full 50% damage reduction. Discounts DO impact this, if you get lucky, you can snowball even further due to this, however, it's a mere symptom of the problem, as opposed to the root cause.

While yes, in theory, you could balance the costs of EVERY single item to work with the raised limit, this would be quite the effort to actively maintain, especially with how our upstream does regularly change/add stuff that is within the realm of traitor's gear. This change will prevent some of the more ridiculous goings-on, traitor has plenty good stuff even without that extra 5.

## Proof Of Testing

<details>

</details>

## Changelog

:cl:
balance: Sets traitor and loneinf's TC count to 20 as opposed to the 25 they had before.
/:cl:

### DIFF
--- a/code/__DEFINES/antagonists.dm
+++ b/code/__DEFINES/antagonists.dm
@@ -338,7 +338,7 @@ GLOBAL_LIST_INIT(human_invader_antagonists, list(
 #define JOB_PROB 40
 
 /// How many telecrystals a normal traitor starts with
-#define TELECRYSTALS_DEFAULT 25 // BUBBER EDIT - GIMMICK TRAITOR
+#define TELECRYSTALS_DEFAULT 20
 /// How many telecrystals mapper/admin only "precharged" uplink implant
 #define TELECRYSTALS_PRELOADED_IMPLANT 10
 /// The normal cost of an uplink implant; used for calcuating how many
--- a/modular_zubbers/code/modules/opposing_force/code/equipment/antagonist_powers.dm
+++ b/modular_zubbers/code/modules/opposing_force/code/equipment/antagonist_powers.dm
@@ -6,7 +6,7 @@
 	item_type = /obj/item/uplink/standard
 	name = "Standard Syndicate Uplink"
 	description = "A mass-produced and cheap Syndicate uplink without a password and a balance of 25 telecrystals. It doesn't get more standard than this."
-	admin_note = "Traitor uplink with 25 telecrystals." // LIES. LIES. LIES!!! ITS NOT EMPTY, IT HAD 20 TC BEFORE! No more.
+	admin_note = "Traitor uplink with 20 telecrystals." // LIES. LIES. LIES!!! ITS NOT EMPTY, IT HAD 20 TC BEFORE! No more.
 
 /datum/opposing_force_equipment/antagonist_powers/tc1
 	item_type = /obj/item/stack/telecrystal
--- a/modular_zubbers/code/modules/uplink/uplink_devices.dm
+++ b/modular_zubbers/code/modules/uplink/uplink_devices.dm
@@ -1,4 +1,4 @@
 /obj/item/uplink/standard
 
-/obj/item/uplink/standard/Initialize(mapload, owner, tc_amount = 25, datum/uplink_handler/uplink_handler_override = null)
+/obj/item/uplink/standard/Initialize(mapload, owner, tc_amount = 20, datum/uplink_handler/uplink_handler_override = null)
 	. = ..()


SapphoQueer:
> another step to make traitor even worse to play if you arent powergaming <3 it already sucks

20 is plenty if you plan ahead. You can get the Vast Majority Of Meta Setups with it.

Stonetear:
Antag nerfs get suggested every few years, and people keep forgetting why furry severs have it like this to begin with:

* We have 3 hour rounds.  Your TC doesn't regenerate.
* Fluffypaws McSecondlife is unrobust and loses all of their TC instantly because they have no idea how stun combat works.  This is extremely frustrating for new players and is part of the reason so many people have antag turned off.
* Most of the items on the uplink are massively overpriced and bait unaware people into wasting their TC.
* Security on the furry servers is massively overtuned compared to tg:
* * An arsenal 3x the size of tg's with an extremely large variety of weapons.
* * Better armor than traitors.
* * Smarter and more communicative security with fewer background threats.

> Under normal, no-discount scenarios, this would prevent you from also getting armor.

Traitor armor isn't even good on the furry servers unless you run the elite modsuit.  Plus the furry servers have funny armor piercing ammo for ballistic guns when NT is supposed to be the energy weapons faction.

BurgerLUA:
this title trend needs to continue

sippykot:
all this will accomplish is have even less people play traitor

YuliceTheDruid:
> Fluffypaws McSecondlife is unrobust and loses all of their TC instantly because they have no idea how stun combat works. This is extremely frustrating for new players and is part of the reason so many people have antag turned off.

This should not ever be something we balance around. The entire stigma of 'lol furries don't know how to play' is silly and only encourages people to not learn how to play the game that they invest hundreds or thousands of hours into.

babamandaa:
I went up against the llama in question, and I genuinely think that this problem, if you even agree that it _is_ a problem, will get better when we eventually revert to 100 HP. I like the little buffer it gives contractors and bundle users, I like the extra wiggle room that it gives players when deciding their builds, and I like that it gives options that players might not normally use, like the holoparasite or powersink, more of a use case. I think that the good outweighs the bad in most cases. 

Yes, sometimes it leads to scenarios where security can get overwhelmed, but these situations are honestly pretty few and far between, as frustrating as they are. We don't remember rounds where security wins because that's just how the station functions. Outcomes outside of the norm are frustrating, and you can often feel powerless, but dropping a role's power level by nearly a quarter is an extremely heavy-handed solution to a problem people can't agree on. If things like scarp combos are _really_ that much of a problem, we can just... increase the price of scarp.

latin313:
> > I went up against the llama in question,
> 
> What does this mean

Basically we had a round where someone had a infil suit, scarp, maybe genetics?, xeno organs, and probably a few other things. They were basically impossible to stop so much so that the 4-5 sec we had? would just leave the round after dying for the 6th time as they were fully stun immune. It got so bad that the entirety of command had to just LET them kill the AI and basically have free-reign because.. they couldnt do anything (Hos and captain doubted even a red-ERT could do fuckall)

SapphoQueer:
> > > I went up against the llama in question,
> > 
> > 
> > What does this mean
> 
> Basically we had a round where someone had a infil suit, scarp, maybe genetics?, xeno organs, and probably a few other things. They were basically impossible to stop so much so that the 4-5 sec we had? would just leave the round after dying for the 6th time as they were fully stun immune. It got so bad that the entirety of command had to just LET them kill the AI and basically have free-reign because.. they couldnt do anything (Hos and captain doubted even a red-ERT could do fuckall)

I see, the "haha I can slap an infilmod which has antistun on it into any loadout" is part of why this PR is made, but scarp isn't the sole perp as I've explained in both of the big walls. We'll see how this plays out, given it's testmerged.

PR #4424, opened 2025-08-10
Add: Isotropic Stability Quirk (Radiation Resistance)

Floofies:
## About The Pull Request

This PR is a paid commission which adds a new quirk named "Isotropic Stability", which provides radiation resistance.

Isotropic Stability prevents many of the side effects of radiation from progressing normally, including mutations, collapsing, vomiting, paralysis/weakness, and hair loss. The probability for burn splotches is reduced down to 25%. Toxin damage is increased to 10, but only increases from continued exposure to dangerous radiation, and the initial 10 points of toxin damage received from becoming irradiated is retained.

image

### Quirk Changes to Radiation Effects

Documented here is a list of the default effects for irradiation, and the tweaks added by the quirk.
Side Effect|Default|With Quirk
-|-|-
Toxins|2 every 25sec| 10 every 25sec when near radiation.
Burns|11 every 30-60sec|-75% probability added
Hair Loss|7.5% probability every 2sec|None
Vomiting|0.5% probability every 2sec|None
Collapse|0.5% probability every 2sec|None
Mutation|0.5% probability every 2sec|None

### Granular Change List
- Created `modular_zubbers/code/datums/quirks/positive_quirks/isotropic_stability/isotropic_stability.dm` with `/datum/quirk/isotropic_stability`.
- Created `modular_zubbers/code/datums/quirks/positive_quirks/isotropic_stability/irradiated_override.dm`
  - Added variable `exposed_to_danger` which is `TRUE` if the mob was recently exposed to irradiating levels of radiation.
  - Added override `/datum/component/irradiated/InheritComponent()` which sets `exposed_to_danger` to `TRUE` when the mob becomes irradiated more than once.
- Added `TRAIT_RAD_RESISTANCE`
  - Added define to `code/__DEFINES/~~bubber_defines/traits/declarations.dm`
  - Added global to `code/_globalvars/traits/_traits.dm`
  - Added global to `code/_globalvars/traits/admin_tooling.dm`
- Edited `code/datums/components/irradiated.dm`
  - Added define `RADIATION_TOX_DAMAGE_FROM_QUIRK`
  - Added conditional to `/datum/component/irradiated/Initialize()` to return early if the mob has `TRAIT_IRRADIATED`. Prevents double-whammies.
  - Added conditional to `/datum/component/irradiated/Destroy()` to return early if the parent mob lacks the component.
  - Added conditional to `/datum/component/irradiated/process()` to call `process_tox_damage()` and return early if the mob has `TRAIT_RAD_RESISTANCE` and `exposed_to_danger` is `TRUE`. Bypasses `species.handle_radiation()`, which prevents irradiation from causing hair loss, vomiting, mutation, and collapsing.
  - Added conditional to `/datum/component/irradiated/proc/process_tox_damage()` to apply 10 toxin damage if the mob has `TRAIT_RAD_RESISTANCE`.
  - Added conditional to `/datum/component/irradiated/proc/give_burn_splotches()` for a 75% chance to return early if the mob has `TRAIT_RAD_RESISTANCE`.

## Why It's Good For The Game

The quirk allows radiation to be slightly less inconvenient and adds randomness to its lethality.

Users of the Isotropic Stability quirk have reduced side effects of becoming irradiated, however active dangerous radiation sources will still metabolize into toxin damage, and burns may occur more randomly.

The quirk provides a unique and impactful effect which reminds me of "Ghouls" from Fallout games. Players could design species which are similarly resistant to radiation, and perform slightly differently in roles that expose them to dangerous levels of radiation.

Additionally, because of the reduction in side effects, radioactive foods/drinks such as Nuka Cola are substantially less detrimental to drink for users of this quirk. Toxin damage from being irradiated will still increase, but only as long as such substances are in their blood.

## Proof Of Testing

<details>

image
image
Screenshot 2025-08-10 174648
Screenshot 2025-08-10 174424

</details>

## Changelog

:cl:
add: Added new Isotropic Stability quirk.
/:cl:


StarHorder:
bestie you cooked with this fr fr

NotRyan2766:
les goooo

Elli-Skala:
the radioation microlaser taking another L, first synths now this

ReturnToZender:
What are the impacts of this quirk? Does the radioactive microlaser actually even use radiation? I need to know more about what the perk does and what its potential impacts will be in order to form an opinion on it

ReturnToZender:
Also, what does this perk offer flavor-wise over the rad resistance genetics mutation?

ReturnToZender:
Removed previous comment due to it being inaccurate, I feel like this would be amazing as a genetics mutation rather than a quirk, and would provide the same level of flavor while contributing more to the server environment and having something resembling opportunity cost

NotRyan2766:
> What are the impacts of this quirk? What does it do, _specifically?_ Does the radioactive microlaser actually even use radiation? Does Nuka Cola actually even give you radiation? I need to know more about what the perk does and what its potential impacts will be in order to form an opinion on it

Nukacola irradiates you, we learned this from testing it and in large dose may continue to damage you. The quirk itself removes some of the bad effects of radiation, such as hairloss, mutations, and burns but will not stop toxic damage from increasing if you are near a radioactive source.

NotRyan2766:
> Removed previous comment due to it being inaccurate, I feel like this would be amazing as a genetics mutation rather than a quirk, and would provide the same level of flavor while contributing more to the server environment and having something resembling opportunity cost

It would be neat as a genetics mutation and definitely could have a use there, however I ask them to make it a quirk so I could make more realistic creatures that thrive in radiation, since we lack a species that does that's organic.

AstrumLernaean:
> > HOLY PEAK BLESS YOU FOR MAKING THIS QUIRK
> > Honestly, yeah 4 points is probably too cheap for a quirk like this given how -actually- useful it is, We'd say something more along the lines of 6 point cost would probably do better, maybe even 8, though it may be a bit of a stretch. I guess at that rate it depends how quickly that toxin damage builds up? Not sure how that part works but this is an amazing idea, a cost adjustment may be necessary for sure though.
> 
> 2 tox -> 4-5 tox (You absorb more radiation than most people) 10 starting tox 0% -> 25%-35% chance of burns (Before you took no burns at all) 4 points -> 6-8 points Does this seem fair to you?

6 points may be good if other things are getting balanced alongside it, but that's our own thoughts.

How would the 25% chance of burns work exactly? Every time you'd take damage from being irradiated, it rolls a 25% chance of giving you burns?

NotRyan2766:
> we suppose that if the concern is engineers using this as a way to just.... not use a modsuit, for some reason, the tox damage in active radiation could be increased significantly (admittedly 4 every 25 seconds is quite low, if the burn resistance was still there, of course). IF burn resistance is still being considered, 10 tox every 20 or 15 seconds would likely be a much bigger threat/disincentivise it being used as a way to avoid using modsuits to counter being irradiated.

Well, the increase of tox will be adjusted then but in return, you keep the resistance to burn. Also mind that, even when you have the irradiated effect, you will not take tox damage if you avoid dangerous places. Without the quirk, you would take toxins permanently until flushed.

The quirk is always checking the room like a geiger counter, the moment it detects *any* source of radiation in range, you will tick up the toxins.

AstrumLernaean:
> The quirk is always checking the room like a geiger counter, the moment it detects _any_ source of radiation in range, you will tick up the toxins.

That sounds like it'd be a fairly good balance if the tox damage were aggressive enough, you're basically trading taking high damage from being irradiated, for taking high damage only when -near- active sources of radiation.

PR #4468, opened 2025-08-16
LTO - RimPoint / Wawastation

lessthnthree:


BurgerLUA:
Has Wawastation changed at all since it was removed? The map had a lot of gameplay issues.

NargaInDisguise:
Wawa was pretty universally hated, no?

lessthnthree:
> Has Wawastation changed at all since it was removed? The map had a lot of gameplay issues.

Yeh this is to get it in basemaps to test how it's fared after some more time in the oven.

Hislyna:
Two major issues for Tcomms Spec, one is their ID can't control the AI Sat turrets (Thats where tcomms is and they should by default have access like all other maps), the second being they don't have a console to check messages by default. A third could be their consoles are in the freezing room aswell, but I play synths so thats not an issue.

lessthnthree:
These are all seemingly TG related, but will see about fixing it upstream.

ChaosCzar:
Mapping issue with Rimpoint, looks like this is the PR adding it back in. The engineering department guard office has an exosuit control console and a science guard locker, neither of which are for engineering. Nor can an engineering guard use them.

SapphoQueer:
image
This one airlock is AA

BurgerLUA:
I think there needs to be more navigate verb destinations placed (the autoname one) as it was difficult to find some areas based on the name.

Other than that, the only complaint I have is how antag sided this map is. Tons of trees to hide or cover behind, along with tons of open areas plus a catwalk system that isn't really seen or heard by when it comes to members of the public.

lessthnthree:
> Adding some comments of my own, namely armory and perma: There is no lethal ballistics option present in armory by default (no autorifles and the shotgun locker has only rubbers/beanbag), and elsewise it's lacking an ion rifle. Otherwise, I appreciate the extra sets of bulletproof armor, especially with the latest change to health.
> 
> As for perma, please for the love of goodness do not have it be the funny double-doors, ideally at least not for the actual entrance into perma itself. In addition, there are no genpop lockers (unless I somehow missed them), and you have to go upstairs to the labor shuttle dock to get a prisoner ID card from there. It would be greatly appreciated if it just got the standard perma entrance of door and the like holographic checkpoint entrance. Elsewise, the map, aesthetically, looks great, I think the little nature features like the rain and lightning and water are fun little elements, and I'm looking forward to seeing it get ironed out and made part of the rotation.
> 
> Edit because I remembered a lil later: There is just, no brigmed? And then the brigmed locker is just shoved in the security locker room. Really need to add that, please.

Noted, will have a look.

re: Secmed, this is a TG map adapted for Bubber so I just threw the secmed locker in the locker room as it didn't have anywhere else apparent. Could configure a space, but as is I didn't want it to delay getting the map in rotation.

ArrisFairburne:
Rimpoint might lack an ERT shuttle docking area, I tried spawning an ERT in and they couldn't take their shuttle to the station.

ArrisFairburne:
Also does science ordnance have any oxygen canisters laying around anywhere? I couldn't find any last time I played.

Krypandenej:
I was asked to relay this by Niko:

image

nikothedude:
Nightmares are overpowered on rimpoint (paradoxically)

1. The outside is lit up by solarlights; they light a large area and give the nightmare tons of room to maneuver by just breaking one
2. The rain defeats the nightmare's main counter, which is fire
3. The lack of space lets the nightmrae go LITERALLY ANYWHERE (issue with all planet maps but its terrible here)
4. Solarlights cant be rapidly repaired with light replacers(?) or at all(??) (I dont know for sure)

Solarlights could be made to auto-repair after x time, rain could refuse to douse nightmares, and solarlights could be made to easily repair

Otherwise, I suggest disabling nightmrae on rimpoint until a better solution is found

PR #4481, opened 2025-08-17
[Skyrat gun replacement] Balances the prices for guns in cargo

KazooBard:
## About The Pull Request

This touches only guns that are left. All changes in Why Its Good For The Game. This is part of the gun overhaul revolving around skyrat guns being gone. Report any inconsistencies and odd things you spot in here, please.

## Why It's Good For The Game

LASERS: The default was set to 300 credits per laser rifle, but the energy gun and carbine were priced at an absurdly high cost of either 800 or 1000 if my memory serves right. Since they all are undeniably competing with eachother with factors such as ammo count, spammability and being used for upgrades, it is only fair to put them all at one price. 

Energy gun, Laser Gun, Laser Carbine all cost 700. We've a bit more money than TG, so putting it at just a tad more will not hurt - making buying guns very early on for players a bit harder earlier on, but overall, it will be easier than TG due to our longer rounds which thus include bigger paychecks overall.

The crates for the guns are priced at 1400 each, with the energy gun crate now having 3 instead of 2 energy guns.

In addition, hellgun kit is now 250 instead of 100, which is pocket change for a weapon of it's quality.

SHOTGUNS:

Renoster shotgun crate now has 3 shotguns instead of 2 and costs 3000 instead of 6000 credits - 500 less than the combat shotguns. The pattern you will see is combat shotguns being a bit pricier than renosters, renosters separated from woodstock by a bigger leap than the former and finally the double barrel will be much cheaper.

Renoster Shotgun's price is now  1400

Woodstock Shotgun's price is now 1200

Double-Barreled Shotgun's price is now 1000. I was considering 800 but the ammo can be printed at any hacked lathe which isn't teh case with neither lasers nor most guns around this pricetag

MURPHY:
Singlepacks availible for 500
Singlepack magazine availible for 200
Crate costs 800 with 3 pieces inside of 2 inside, they're cheap and reliable and bad against anything armoured, so no shit they'll be good for mass-buying. And 2 for 800 was a bad deal.
Crate of ammo has 4 mags for 600 now, in case you want to arm a lot of assistants for cheap.

The colt is untouched because it's a near-unique weapon with really good burst. Yes, same as the double-barrel, but this gun has none of its weaknesses in range, storage, it can reload with speedloaders, is generally way better as a gun and need I go on about why I am sure colt's price is fine? 

PLASMA:
Untouched, I'll PR it on the plasma-tweaks branch when the PR for making the sharpshooter useful drops. 

WT550:
Singlepack for 800, before you stone me for making it more expensive than the lasers, remember about EMP's, reloading on the go, how it's got a larger firerate so dodging it is more tedious and how bleed wounds need immediate split attention vs burn wounds which are a Problem For Later.
Singlepack magazine for 250. 20 bullets is a lot, after all.
Crate is 3 for 2000, Ballistics in general tend to be more oppressive due to having the ability to reload on the go, call it the 'Elite' option.
Ammo crate is 800 for 4
Smuggled crate is 800 for 6 with a 2/2/2 reg/ap/inc spread. It's riskier since smuggled required either hacking or emag, so you cannot order it directly and have to go through that trouble and any eyebrow-raises you get from lawful good captains, officers, hos'es, cargo techs for hacking the console. 

## Proof Of Testing
<details>

</details>

## Changelog

:cl:
add: WT and Murphy availible in singlepack, as well as having singlepack magazines up for purchase
balance: A lot of gun prices were adjusted to make sense in comparison to other weapons
/:cl:


Arturlang (changes requested):
it doesn't build, also, non-modular changes, make sure you're not changing TG values

modular_zubbers/code/modules/cargo/packs/goodies.dm
```diff
@@ -19,6 +19,36 @@
 	access_view = ACCESS_WEAPONS
 	contains = list(/obj/item/gun/energy/e_gun/mini)
 
+/datum/supply_pack/goody/wt550_single
+	name = "WT-551 Autorifle Single-Pack"
+	desc = "An NT-security grade autorifle, it comes with excellent heating and poses no health-related risks for the user. Comes as a single-pack with one WT-551 locked and loaded.
```
xPokee, line 24:
```suggestion
	desc = "An NT-security grade autorifle, it comes with excellent heating and poses no health-related risks for the user. Comes as a single-pack with one WT-551 locked and loaded."
```

xPokee (changes requested):

xPokee (approved):

PR #4512, opened 2025-08-22
Removes IRCDs

xPokee:
## About The Pull Request
Does exactly as the title says, removes IRCDs from the code. 
## Why It's Good For The Game
Engineers already start with RCDs, and Engineering can print RCDs from their lathe, yet this is one of the most common orders to be placed *immediately* by Engineering upon round start. They serve no purpose aside from being slower and having 60 more ammo. This is arguably just a noob trap, given that a fully upgraded regular RCD can easily do the exact same job, but *better*. Quite frankly, the fact that *anyone* can also buy these from Cargo for 1900 credits (easily obtainable from doing a few bounties) is absurd, along with the fact that they were made into a ***DEPARTMENTAL ORDER*** for Engineering.

Without them existing, engineering will have an actual reason to want to get RCD upgrades, and most importantly, crew will have to go through engineering to get one instead of being able to just ***order one***.
## Proof Of Testing
It was a simple removal, it should work fine as long as checks pass.
## Changelog
:cl:
del: removed IRCDs from the code entirely
/:cl:


trefoils:
We still want rcd upgrades, the reason these are ordered is mostly for atmos techs who don't an rcd round start and dont want to spend 200 credits on one or wait 3 years for science to research them. You can also ignored the fact that these let us make machines round start, which include like apcs,frames etc. Which further just kind of quashes the noob traps argument as this just makes it easier for new engineers.  Noob trap not really its maybe slightly slower if at all like do a side by side i've done it before theres like 0 difference. If you want speed you build by hand. You still have to upgrade these with disruption and cool down upgrades 
I dunno why you woke up and suddenly hate these, but its fine its doing no harm.

lessthnthree:
The shift is three hours, you can use a normal RCD until upgrades are done. Or... build using regular tools.

ChaosCzar:
I mean, they're ordered at the start of shift for a reason. Also- "Engineering will have a reason to want to get RCD upgrades"

Engineering isn't in control of researching upgrades, though. They're at the discretion of science for that.

ChaosCzar:
Also engineers DON'T spawn with RCDs here anymore, that was changed.

Mitryll:
> The shift is three hours, you can use a normal RCD until upgrades are done. Or... build using regular tools.

Revert https://github.com/Bubberstation/Bubberstation/pull/3498 because otherwise you get turbo giga shafted if you're not a roundstart engi. Idrc about using the baby basic RCD until upgrades but you legit cannot if the few that exist roundstart get yoinked. This is the entire reason people order the IRCD asap anyways.

Razurath:
> > Revert #3498 because otherwise you get turbo giga shafted if you're not a roundstart engi. Idrc about using the baby basic RCD until upgrades but you legit cannot if the few that exist roundstart get yoinked. This is the entire reason people order the IRCD asap anyways.
> 
> Unless I misread the code, regular RCDs are literally printable from the engineering protolathe at round start.

They have to be researched first.

PR #4534, opened 2025-08-24
Moves the Blueshield role to being a station trait

xPokee:
## About The Pull Request
A revival of https://github.com/Bubberstation/Bubberstation/pull/1186.
## Why It's Good For The Game
By creating an environment where the blueshield’s presence isn’t guaranteed, the role gains a newfound importance, uniqueness and gives room for parallels, so players can contrast to really see the impact they make on the rounds.

This will additionally encourage security-command cohesion through mutual dependence without a middle-man and go as far as to let antagonist action against the captain carry more weight and let the captain flex their equipment and mechanical knowledge (or call for sec as always, really)

While an SOP solution would improve the current player frustrations with blueshield’s confusing spot on the roster by defining it further, this PR will provide the environment needed for an examination of its true place in the game, the ways to settle it firmer than pure theory ever could and lastly - let the players catch a break and appreciate the more special role.
## Proof Of Testing
See the attached screenshot.
<details>
<summary>Screenshots/Videos</summary>
image

</details>

## Changelog
:cl:
add: added a blueshield locker beacon
map: replaced the blueshield offices on various maps
refactor: refactored blueshield to be a station trait
/:cl:

PR #4535, opened 2025-08-24
Saving Blueshield [alternative to #4534]

ReturnToZender:
## About The Pull Request
This PR changes the Blueshield's starting loadout to use only a telescopic baton, and changes out their security HUD for a medical HUD.

Included ranged weapons is a serious force multiplier for redshields, and the medical HUD will be a lot more useful for them while allowing them to still have some pretty cool sunglasses. 

This PR does not need to be merged, but I think it should earnestly be considered if Blueshield is on the chopping block. This role is genuinely one of our most popular roles.

## Why It's Good For The Game
Honestly, it isn't, and neither is making it a station trait. Blueshield is an extremely popular command role, and administratively, it requires very little oversight, ourselves having to ban only a handful of characters for misconduct. Having anyone around to defend you is a force multiplier, sure, but the Blueshield is a serious increase in force. As opposed to making it a station trait, this PR seeks to make it so that the Blueshield is only equipped with a telescopic baton instead, which should massively reduce their ability to extend their conflict further out from Command

## Proof Of Testing
Compiled and tested

<details>

</details>

## Changelog

:cl: ReturnToZender
del: The Blueshield now no longer receives a sechud or an e-gun, receiving a telescopic baton instead.
/:cl:


KazooBard:
Don't blueshields have weapons permits so they can just buy the items removed from loadout every round?

ReturnToZender:
> Don't blueshields have weapons permits so they can just buy the items removed from loadout every round?
Right, I forgot to remove their weapons permit, which is why I took the e-gun

PR #4562, opened 2025-08-28
Improves disposals code so it actually deals damage to you on direction changes and nerfs disposals damage.

BurgerLUA:
## About The Pull Request

Improves disposals code so it actually deals damage to you on direction changes.
Nerfs disposals damage so that it has a 20% chance to apply instead of an 80% chance to apply.
You no longer take disposals damage if you're in hard crit or dead.
Disposals damage when it procs deals 4 damage instead of 5.

## Why It's Good For The Game

Making disposals damage work as intended is always good.

With the new 100hp change, people are feeling the (disposals) burn, so the damage was reduced.

## Proof Of Testing

Draft because I literally coded this from scratch in the github online code editor.

## Changelog

:cl: BurgerBB
balance: Nerfs disposals damage so that it has a 20% chance to apply instead of an 80% chance to apply. You no longer take disposals damage if you're in hard crit or dead. You take 1 less disposals damage when it applies, now, a total of 4 per turn
fix: Improves disposals code so it actually deals damage to you on direction changes.
/:cl:

### DIFF
--- a/modular_skyrat/modules/hurtsposals/code/pipe.dm
+++ b/modular_skyrat/modules/hurtsposals/code/pipe.dm
@@ -1,8 +1,12 @@
 // Make disposal pipes hurt you if you ride them.
 
+/obj/structure/disposalpipe
+	// Set this to true if you want a disposal pipe to deal no damage.
+	var/padded_corners = FALSE
+
 /obj/structure/disposalpipe/transfer_to_dir(obj/structure/disposalholder/holder, nextdir)
 
-	if(!holder || !holder.hasmob || holder.dir == next_dir || !prob(80)) //Disposals optimization.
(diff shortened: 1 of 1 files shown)


BurgerLUA:
Turns out that the disposal code did proc 20% of the time and not the 80% I thought it was.

lessthnthree:
> Nerfing damage plus making it only a 20% chance, makes disposals actually, entirely non-threatening and barely hurts at all. I feel like 20% is a bit too much of a change. From taking damage 4/5ths of the time, to taking damage 1/5th of the time, atop lowering damage? Feels like this might as well be removing disposal damage at that point

An example of a 600 tile trip on metastation:

Current:
image

This PR:
image

We go from 240 damage, certain death to 72, which with our reverted HP of 100, is about half way towards crit.

BurgerLUA:
I chose the values because I did a worst case scenario calculation on Moonstation and average case scenario is enough to put you into crit.

PR #4566, opened 2025-08-29
Splitsicle station

ArrisFairburne:
The non-viability of extended mining operations on the icemoon surface in Nanotrasen's permitted areas continue to rise, between increasingly unlivable atmospheric conditions far from the poles, the suspicions of sabotage from Cybersun operatives and the logistically-outdated mass-produced research facilities used for these operations. 

As such, Nanotrasen has opted to deploy the experimental Semi-Amphibious Bifurcated Research Vessel -- commonly known as "Splitsicle Station" to its workers -- to one of the seas of plasma near the south pole of the icemoon, as a staging area to carry on continued lavaland mining.

## About The Pull Request

Adds a new station to the game: Splitsicle Station. This station is planetary, set on the icemoon in a lake of liquid plasma; the defining trait is that it's split into two separate sub-stations, connected by transit tubes.

## Why It's Good For The Game

This map's purpose is to address issues with Icebox while retaining flavor associated with the icemoon planet. Issues that are intended to be addressed include:
- **Icebox's station layout is too similar to other maps.** Lots of our existing maps are designed in the same way: "central square hallway hub with shorter hallways coming off of it that go to departments". Metastation, Deltastation, and Icebox are all designed this way. This is also not considering the fact that we run the map that Icebox is directly based on, Boxstation, which has a mostly-identical station layout. Replacing Icebox with this map will reduce how similar the maps are.
- **It's too easy to exist outside on Icebox.** With three different z-levels, the ability to bare-handed dig through snow tiles, and a breathable atmosphere, the only real inhibition to escaping the station on Icebox is staying warm, which is generally trivial. This station's designed to bring back threat to going outside, with its large plasma lake making movement highly limited without existing infrastructure or specialized equipment to deal with it.
- **Icebox's multi-z implementation is disruptive.** Sub-departments like xenobiology, ordnance, and botany are split between two z-levels; having to navigate up and down between these areas is awkward and hurts gameflow. Splitsicle is a single-z level and so shouldn't have this issue.
- **Icebox mining is disruptive.** Let's be honest, having the whole screen shake from chain gibtonite explosions is a nuisance.

## Proof Of Testing

WIP

image

## Changelog

:cl:
map: Nanotrasen operations on Icemoon have shifted towards the plasma seas on the south side of the planet.
/:cl:


ReturnToZender:
It better have a tram

NezzieNazzie:
I, I am split on this. 

On one hand, this looks neat, on the other, I cannot imagine this being fun to play with high ping because you're one laggy frame away from getting borderline RR'd

NargaInDisguise:
Ok right off the bat, "Replacing Icebox with this map will add more variety in player choice for maps to play on?"
Absolutely ridiculous. Icebox is a great map by itself. You wanting to replace a map completely is a bit harsh.

This map seems incredibly bland and awful for crew. I'm also mildly annoyed by the complete removal of icebox mining in this map because, icebox mining IS GOOD and FUN. It is my favorite mining and I hate that this uses lavaland mining. The wording in your PR is also sorta hinting towards wanting Icebox removed, and that seems really shitty.

Onto other complaints;
-A single atmos breach or bomb seems like it'd make this station uninhabitable really fast.
-It takes, ONE destroyed transit tube to completely cut half the station from the other and/or kill 30+ in an emergency with no counterplay because transit tubes are so awful.
-Sure, you're right, existing outside icebox isn't really hard at all, but there has to be a better way than just making everything plasma. You're removing all of icemoon and space ruins just for a giant plane of plasma.
-Just remove gibtonite shaking other Z-levels? Icebox mining isn't disruptive, it's LITERALLY only that.
-Icebox has had plenty of changes to remove some of the problematic multi-z things, and some patches that changed blob cheesing as well. This map seems like its going to be an absolute paradise for blobs.

AstrumLernaean:
Before we get to anything else, it'd be cool if this was simply in addition to the maps we have, rather than a replacement, especially in its... current state. Icebox has its own thing going for it and lots of people find it unique in its own right and like it for there own reasons, (there's a reason it gets voted as often as it does, although map variance has been pretty wide lately imo), WE like Icebox ourselves too! 
This is.... we're putting this nicely... not something worthy of replacing it, at best.
The station design itself has a neat flair to it and looks cool, however there is (several) pretty fundamental issue with this station that we feel should be addressed:

**The plasma.**
Plasma is significantly more dangerous than lava, and the way the station is built directly on top of the plasma ocean seems to give no leeway for accidents, station integrity destruction, or well, anything. There's not much consideration for antag spawn locations (space dragon, ninja, lone op, infiltrator, etc.). The transit tubes especially feel like a significantly dangerous game to play as they will drop you out immediately and directly into the plasma lake if at all compromised because transit tubes are goofy (which is basically guaranteed to kill you, plasmaman you, and effectively soft RR you). This in particular is a potential mass RR waiting to happen.
The plasma is too dangerous and the design of the station itself (with no ground around the edges or underneath the transit tubes whatsoever) is a nightmare waiting to happen realistically.

There's more we could go on about, but we'd like to suggest that adding grounding around some portions of the lake is maybe a better idea, rather than just a.... plane of deadly liquid plasma? Especially adding icemoon grounding around the station itself would likely be a much better idea, make it a sort of "island" station surrounded by plasma, though even that might still just be a start to what can be done to improve the map. 
This on its own is rather bland, poorly thought out and not crew friendly (and ironically not antag friendly either) whatsoever (not that it should be 100% crew safe, but this is rather overkill).

As for the arguments made...

> _**"Icebox's station layout is too similar to other maps."**_

Icebox is really, realistically, only ACTUALLY similar in layout to Box station, which itself has been (recently!) overhauled in some parts to make it more unique and interesting, which is pretty awesome! Icebox itself has received lots of changes overtime to do just the same for it, making it the map that's enjoyed by many today. This is not a great argument for replacing Icebox outright.

> _**It's too easy to exist outside on Icebox.**_

It's even easier to exist space, or hell even Moonstation. Moonstation is MUCH easier, you don't even need an air tank or special clothing for anything other than sandstorms. Not a great argument imo, especially when the rebuttal for a "better design" is to essentially strip the ability to even exist outside the station without significant hassle or effort compared to every other map we have (and would it even be worth it when the entire map is just plasma anyway?).

> _**Icebox's multi-z implementation is disruptive.**_

And so is every other multi-z map. But we're not replacing those.

> _**Icebox mining is disruptive.**_

Admittedly, we can't say much to this one, as it's a fairly opinionated statement. However we will say that Icebox offers a unique flair on its mining being on the same active level as the station, and not on lavaland, and is unique for that, we've heard of lots of people that enjoy mining on icebox, and some others that don't, in the end it's a matter of opinion, however this is not an argument or justification for an outright replacement of the map from our point of view.

Mitryll:
Be antag

Get gravity MOD

Get ion thrusters

You are now playing on server admin mode.

PR #4572, opened 2025-08-30
[SR REPLACEMENT] Buffs/Reworks plasma marksman

KazooBard:
## About The Pull Request

Special thanks to offwrldr/Bangle for the unique sound effects of the explosion. 

plasma3

https://youtu.be/niNd51UAtBk

## Why It's Good For The Game

I will _**attempt**_ to write short instead of creating an essay here.

With the Murphy, the leeway for crew pistols lessened, and frankly my core belief is that when something is added, it should add interesting new gameplay elements, **things that are interesting to both fight with and against**. Items that deviate from the regular gameplay loop of shooting but in X manner and allow for expansion in the depth of combat - and with all that in mind, our goal was to deliver on that.  To **keep the core of the weapon's mechanics, expand upon them and integrate it with their faction's lore as a cherry on top.**

The plasma marksman (sharpshooter) is in an utterly horrid state. I could go on to list how it's impossible to shoot someone walking away from you in a straight line, how even if you hit them - your shots will deal 3 damage per round on most antagonist armours, how it's huge clip is an issue in the face of now-reduced HP values.

But it doesn't have to be that way. It wasn't just another military weapon, it's part of a charming faction with their own gimmicks shown through their equipment and a concept that just fits SS13 flavour-wise, but not in gameplay. 

The old weapon was capable of use almost exclusively on the defense, as any gun would outperform it on the offense, and even when on it's home turf - the weakness to armour wasn't helping it's case as "THE cheap self-defense option for second amendment enjoyers". 

### Why did we keep the magazine?
Magazine weapons have an element of pressure that needs a careful balancing act to avoid becoming a plain spam weapon that loses it's depth and becomes oppressive against targets due to the lack of need to break in it's shooting and how easy it is to hoard ammo in a box. The combination of consistency, wounds, a bag of holding full of rounds (hyperbolically) and raw damage output are frankly "ass" to fight. 

What makes this weapon different, however, is that the amount of rounds isn't actually the true depth of the magazine. As an ingenious and frankly wonderful suggester pitched, the gun's able to function normally as it's projectile speed is high, however, careful dodging or misses from the wielder are punished as the gun drops in reliability - it wears down, and the consistent theme you will see across it, is "How Badly Do You Want It?". The buffer is less of an issue on defense, where a regular civilian would need it most, but while it ups the difficulty of offensive kills and chasing after a target, say, an attacker, to finish them off - it doesn't prevent it. 

In this way, the gun is defense-biased, with a possibility for offensive action that is dependant on how well can the user cater to their gun, rather than having it mold to fit their needs exactly in every situation ever. This will prevent it from becoming oppressive, it will make split-second decisions all the more valuable, and lastly, players who manage to plan ahead to lead fights to terrains where they know the downsides aren't as prominent are rewarded with a much smaller penalty to consistently landing shots, and so will those who use it in tandem with tools to slow their opponents down. 

### What is the point of the overclock mode?

The overclock mode is once again, a question, of How Badly Do You Want it? There's two situations where it is primarily going to be used, one is when you *need* to pick someone off, and one is when you know you have no chance. 

Let's start with the first:
Now, even preparation can go awry when your target rolls up in freshly-bought armour, or they manage to escape your kill-room, or you miss too many shots. This is where you can double down:

You gamble, that as you take 25 damage over the course of 10 seconds, with the improved projectile speed and technically increased armour penetration you will put the person you want dead into crit before your gun explodes and people will hear the commotion. Whether it's worth it, whether you believe yourself to be able to do it, or whether you have the HP or escape routes to commit to this further is dependant on you and you only.

In the second situation, meanwhile:
You risk going down earlier than you wish, but leaving a calling card of where you've been, and now remember, whether people hear it and whether people come is not guaranteed - and what's not guaranteed, is that by injuring yourself and announcing that you're out of better options, the assailant won't disengage, reset, and get you again. It's a careful game of chicken and who managed to out-position who, as well as who can more accurately gauge their own damage output versus what they risk taking by comitting to a kill, or the overclock.

The practical example of the second situation is fighting a redsuit, while the first would be attempting to use the gun to assasinate a person (conveniently disposing of the murder weapon).

**That sounds like the old backup on the sechailer!**

A one time use, baitable, force-able by EMP, non-global and non-pinpoint accurate backup. It's a conjunction of putting it in your hands and in people's hands around, so no matter whether people show up or not - you gave yourself at least a bit of advantage, and at least left a trail. 

### Conclusions:
Merge my PR.

## Changelog

:cl:
balance: The plasma marksman has been overhauled. It now comes with an overcharge mode and progressive projectile speed.
/:cl:


Cephalopod222:
He cooked

ChickyRooster:
Snipin's a good job mate

KazooBard:
> The power packs for the guns don't seem to be rechargeable, despite saying so in their description.

This is essentially a rework, and that descrpition went under my radar. I did not change the charging mechanic for the magazines, nor the description until now, so it seems it simply just said that. 

Anywho, it's not mentioned as a part of the rework, so I'll just make the description reflect the reality of it.

KazooBard:
Happens with all projectile weapons, WT, Lasers, etc etc

Issue #4559, opened 2025-08-27
SAD sometimes generating garbage characters


lessthnthree:
### Issue Summary:

After (presumably) an upstream, some players when trying to use the SAD come out as horrific monstrosities that aren't the selected character slot.

### Reproduction Steps:

Use the SAD as an unlucky one

lessthnthree:
It might be caused by people getting their DNA scrambled in a radstorm.

Syilumi:
remember you gotta clean the SAD in-between uses ,':3 and keep metal objects out of it

PR #4575, opened 2025-08-30
Disables radiation storm until the SAD is fixed

lessthnthree:
## About The Pull Request

Disables radiation storm until https://github.com/Bubberstation/Bubberstation/issues/4559 is fixed so it quits generating a flood of ahelps


lessthnthree:
This ended up not being an easy apparent fix, so might as well merge for now until it's fixed given no ETA.

PR #4216, opened 2025-07-11
[Ready] Beepsky can no longer see you if you haven't moved or interacted with something for 4 seconds or more.

BurgerLUA:
## About The Pull Request

Beepsky can no longer see you if you haven't moved or interacted with something for 4 seconds or more.

## Why It's Good For The Game

Prevents Beepsky from interrupting security when they're trying to engage in erotic roleplay with a criminal in the hall.

## Proof Of Testing

Tested in a previous PR made ages ago. Should still work.

## Changelog

:cl: BurgerBB
balance: Beepsky can no longer see you if you haven't moved or interacted with something for 4 seconds or more.
/:cl:

PR #4598, opened 2025-09-03
Revert "[Ready] Beepsky can no longer see you if you haven't moved or interacted with something for 4 seconds or more."

Iamgoofball:
Reverts Bubberstation/Bubberstation#4216

This is ridiculous and makes Beepsky ineffective at his job. If I'm sitting at the bar Wanted, Beepsky and ED-209 will walk right by me without even glancing.

Officers who are talking to someone they aren't ready to arrest should be setting their target to Suspicious. Wanted is a very specific flag that means WANTED. If you haven't made the decision to arrest someone yet, don't set them to Wanted. Set them to Suspicious. You can freely set someone's security status by shift clicking them with HUDglasses and your security ID on.

Security being unwilling to properly use their tools is no reason to kneecap a perfectly functional mechanic.

### DIFF
--- a/code/modules/mob/living/simple_animal/bot/secbot.dm
+++ b/code/modules/mob/living/simple_animal/bot/secbot.dm
@@ -494,14 +494,6 @@
 		if((nearby_carbons.name == oldtarget_name) && (world.time < last_found + 100))
 			continue
 
-		//BUBBERSTATION CHANGE START: BEEPSKY IS A DINOSAUR NOW. CAN'T SEE YOU IF YOU DON'T MOVE.
-		if(nearby_carbons.client) //We have a client. We're a player that uses move_delay.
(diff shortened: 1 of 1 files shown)


BurgerLUA:
On paper, yes security should be setting arrest status only when they want to arrest people, but in practice it is set to tell officers "HEY THIS GUY IS AN ACTUAL CRIMINAL." and thus leads to situations where Beepsky interupts roleplay or whatever is going on, ruining the situation.

Suspicious unfortunately does not fix this. Suspicious is just used as a note for security officers telling them to watch for them.

lessthnthree:
What about beepsky still reporting them on radio if they're in the stationary state. Then if security actually wants the guy, they know where they are.

BurgerLUA:
I don't really have a horse in this race. I only implemented it because some antagonist/security players requested it as it was interrupting their arrest procedure. If you want to do a full revert, go ahead.

PR #4601, opened 2025-09-03
Reduces storyteller sec/antag cap to 1.15

nikothedude:
## About The Pull Request

Title.

From now on, each secoff will open one antag slot - with two opening once 8 total sec members are on.

The value is mostly arbitrary and taken from timsweeney's suggestion of something between 1-1.15.

## Why It's Good For The Game

Sec's power is in numbers. As it stands, in the worst case scenario, sec will always be outnumbered 50%. This is, obviously, not ideal. We want antags to be outnumbered by sec, not the other way around.

Of course, some antags wont engage, which is why the number should still be on the higher side to account for more passive antags.

There was allegedly a discussion in staff chat about this, ending in the conclusion that reducing this might be a good idea.

## Proof Of Testing

<details>

image

</details>

## Changelog

:cl:
balance: Storyteller antag/sec cap is now 1.15
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/storyteller/gamemode.dm
+++ b/modular_zubbers/code/modules/storyteller/gamemode.dm
@@ -243,7 +243,7 @@ SUBSYSTEM_DEF(gamemode)
 		return 0
 	if(!storyteller.antag_divisor)
 		return 0
-	return round(max(min(get_correct_popcount() / storyteller.antag_divisor + sec_crew ,sec_crew * 1.5),ANTAG_CAP_FLAT))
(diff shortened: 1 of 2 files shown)


nikothedude:
It might also be wise to reduce this across storytellers, but antag_divisor reduces antag counts accordingly

SapphoQueer:
> There was allegedly a discussion in staff chat about this, ending in the conclusion that reducing this might be a good idea.


There was a talk, but there was no clear conclusion.

nikothedude:
> > There was allegedly a discussion in staff chat about this, ending in the conclusion that reducing this might be a good idea.
> 
> There was a talk, but there was no clear conclusion.

i stand corrected

PR #4616, opened 2025-09-05
Fixes new Dragon borg text offset, and wide borgs offset [Bounty]

john-pawson:
## About The Pull Request

Dragon borgs, due to their sprite being bigger than any other sprite included, used to have their text offset to the side and appear from their waist, not only that, but they also missed the TRAIT_R_TALL trait, that even shorter borgs like NiKa had.

image

This PR adds a new trait exclusive to the Dragon that should fit any other borgs that are made the same size (If there's more made of the same size, considering how these sprites are the only exception so far.) this trait also has a higher height, so the text appears properly above their head, and contains the properties of the wide trait (Since it's made for 64x64 sprites).

image

On the process of the PR, I also managed to centre the text of the borgs that used TRAIT_R_WIDE, so when they speak their messages won't appear offset to the side of where they're standing.

image 
image

## Why It's Good For The Game

No more offset texts! Now the messages appear where they should.

## Proof Of Testing
This one's only mostly included to demonstrate that I did not break the TRAIT_R_TALL text offsets, the rest of the proof is on the screenshots in the description.

<details>
image 
image

</details>

## Changelog

:cl:
fix: fixed wide borgs text offset
fix: fixed "Dragon" borg text offset
/:cl:


NargaInDisguise:
smolraptors finally fixed too im assuming, yippee

john-pawson:
> Looks good except for this
 Is it supposed to be over its face?
It's the same height as it used before, simply that now the text comes from the centre instead of being offset to the left while talking.
Basically, yes, it comes out from it's face, it's just that now it does it from the centre of the face instead of the side of the face.

xPokee:
> > Looks good except for this
> > Is it supposed to be over its face?
>
>  It's the same height as it used before, simply that now the text comes from the centre instead of being offset to the left while talking.
> Basically, yes, it comes out from it's face, it's just that now it does it from the centre of the face instead of the side of the face.

Alright, I'll go ahead and merge this then.

PR #4627, opened 2025-09-06
Reduces how certain Security roles are counted towards the antag cap because I observed a round for one minute once.

BurgerLUA:
## About The Pull Request

Reworks how certain Security roles are counted towards the antag cap. Some security roles are not counted towards the antag cap as much as officers.

### Current Values

Note that before this PR, the value for all roles was 1.5

Security Officer: 1.5
Warden: 1
Security Medic: 1
Head of Security: 1.5
Captain: 1
Corrections Officer: 0
Detective: 1.5

## Why It's Good For The Game

Previously, all security roles counted as 1.5 for the purposes of antag caps, which means that for every security person, it would add 1.5 antags (final value is rounded up) to the hard antag cap.

This was a bit of a problem if the manifest was weird, such as having 1 warden, 2 correction officers, and 1 security medic. This meant that with those roles, the hard antag cap was 6, even though that cursed combination of roles doesn't really justify that number.

### Warden

Warden is supposed to stay in brig and they have the responsibility of monitoring prisoners and monitoring the brig. They can't really afford to roam around like a regular officer, dealing with threats.

### Security Medic

Security Medics, while officers+, still have the extra responsibility of treating other officers who are down as well as extra prisoners.

### Captain
Captains are not supposed to be security officers, and generally don't do security stuff anyways.

### Corrections Officer
This is literally an ERP role. This role was added because wardens kept going warden to fuck prisoners. COs are not respected and don't go out validhunting antags.

## Proof Of Testing

Untested. If it compiles, it werks.

## Changelog

:cl: BurgerBB
balance: Reduces how certain Security roles are counted towards the antag cap. Some security roles are not counted towards the antag cap as much as officers.
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/storyteller/security_cap_calculations.dm
+++ b/modular_zubbers/code/modules/storyteller/security_cap_calculations.dm
@@ -14,7 +14,7 @@
 	sec_antag_cap = 1
 
 /datum/job/security_medic
-	sec_antag_cap = 1
+	sec_antag_cap = 1.25
 
 /datum/job/detective
-	sec_antag_cap = 1.5
+	sec_antag_cap = 1.25


nikothedude:
why are detectives 1.5? they're not a combat role and are generally discouraged from making arrests if they don't have to

BurgerLUA:
> personally, I think Captain too should count as "0" Captains should generally **never** do Sec work , as they aren't Sec at all, /tg/ adding the Captain to the security list, is just standard /tg/ behaviour (to not use the r-word) and should not be counted as any sort of valid change. 

I'm conflicted about this because I've seen captains in combat roles before.

KateAnbolho:
> > personally, I think Captain too should count as "0" Captains should generally **never** do Sec work , as they aren't Sec at all, /tg/ adding the Captain to the security list, is just standard /tg/ behaviour (to not use the r-word) and should not be counted as any sort of valid change.
> 
> I'm conflicted about this because I've seen captains in combat roles before.

Captains can and do often lead from the front. The antag cap isn't just for tators, but like stuff that the captain WOULD fight. (blob, xenos, and the like)

PR #4670, opened 2025-09-13
Adds a borg obedience module

Tamamitsune:
## About The Pull Request

Adds a brand new cyborg module to the exofab and lustwish; the borg obedience module. As a sister module to the dominatrix module, it functions identically the same, but instead of applying the dominant aura trait, applies the well trained trait. This also makes it so the quirk functions on silicons properly.

## Why It's Good For The Game

Some borgs want to have the funny toys but not the dominant aura, or would prefer to just be able to have access to this trait, even if it may be a negative one on paper, not to mention it's a bit odd that only the one side of this two sided quirk is available.

## Proof Of Testing

It compiles, only one dominatrix or obedience module can properly be used at the same time, borgs react how they should with the quirk.
<details>

submissable
snapsnap

</details>

## Changelog
:cl:
add: Obedience cyborg module added, a sister module to the dominatrix module that applies the well trained quirk
add: Obedience module available in the lustwish and exofab with the other utility borg modules
code: Updated the dominant aura quirk to work on well trained silicons
/:cl:

### DIFF
--- a/modular_skyrat/modules/modular_items/lewd_items/code/lewd_machinery/lustwish.dm
+++ b/modular_skyrat/modules/modular_items/lewd_items/code/lewd_machinery/lustwish.dm
@@ -44,6 +44,7 @@
 				/obj/item/spanking_pad = 4,
 				/obj/item/tickle_feather = 8,
 				/obj/item/borg/upgrade/dominatrixmodule = 5,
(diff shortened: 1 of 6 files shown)


xPokee (approved):
As long as checks pass this is fine

PR #4686, opened 2025-09-15
Ports big funny hoodies from Doppler

theselfish:
## About The Pull Request

image

## Why It's Good For The Game

Customization good, also being a cat with a cat ear hoodie would be Funny I Think.

## Proof Of Testing
<details>

</details>

## Changelog
:cl:
add: Finally, cats can have four sets of ears with a brand new hoodie. (Also comes in earless variety)
/:cl:


icons/map_icons/clothing/head/_head.dmi
xPokee, line 1: Why was this changed?

icons/map_icons/clothing/suit/_suit.dmi
xPokee, line 1: Same with this one.

icons/map_icons/clothing/head/_head.dmi
theselfish: GAGS items point to these files so in mapping they aren't invisible. Look at other gags items, I dunno why it has to be tg icon folders but that's the way every base I've seen handles it.

xPokee (approved):

icons/map_icons/clothing/suit/_suit.dmi
OrionTheFox: Yep, they're auto-generated and it's 100% intended that they're non-modular. @xPokee (sry for ping just figure you'd want the explanation)

These files are pre-assembled GAGS icons, used for the purposes of Mapping and Loadouts.
They're only updated when manually launching a test server - Live doesn't generate the icons (because that would mean doing it every round)

It is 100% intended (and **expected**) that these files are updated; They're already unique to each server, build directly off GAGS items.
If these updates don't go through, mapping/loadouts will display items as Errors or outdated.

PR #4688, opened 2025-09-15
[QOL]Ports Nova's Rack Mechanics

BasilTamaya:
Because being able to choose where you place things on racks is objectively better than forced center placement.

## About The Pull Request
Tin. The shelf sprites have separate shelves but force objects to be on the center currently. Functionally, it works the same as how I coded it on Nova (I am the coder of the original), but I still tested it.

## Why It's Good For The Game
Would be nice if players could actually choose where they put their items on for fancier placement, for RP or for better or worse.

## Proof Of Testing

<details>
image

</details>

## Changelog

:cl:BasilTamaya
qol: Racks now have a precise placement vs center placement mechanic, making shelfs and racks fancier.
/:cl:

### DIFF
--- a/modular_skyrat/modules/aesthetics/rack/code/rack.dm
+++ b/modular_skyrat/modules/aesthetics/rack/code/rack.dm
@@ -7,6 +7,44 @@
 	icon = 'modular_skyrat/modules/aesthetics/rack/icons/rack.dmi'
 	icon_state = "shelf"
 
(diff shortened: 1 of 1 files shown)


OrionTheFox:
Honestly, this is QOL enough that it might just go well upstream. 
Racks are technically theirs after all, and making them behave nicely like this might also help them finally get to respriting them 🥲

PR #4719, opened 2025-09-21
Re-Enables Mold

KateAnbolho:
## About The Pull Request

It brings mold back, removing an override that disabled mold from being spawned. 

## Why It's Good For The Game

With the (presumed) death of Fleshmind, Bubber is sorely lacking when it comes to external threats. Blob seems to be exceptionally rare, and spiders/xenos are eh. Bringing mold back would let us have a true PvE threat that people can fight (or not, I guess)

## Proof Of Testing

works on my machine 

## Changelog

:cl:
add: Brings mold back

/:cl:


Hotmud9:
I miss fleshmind...

BurgerLUA:
Mold needs a lot of work before it should be enabled again, tbh.

PR #4757, opened 2025-09-29
[TM Only] Give someone the spare

ReturnToZender:
## About The Pull Request
This PR makes it so that crewmembers are the lowest possible choice on the chain of command, codewise. This means that if someone is selected as Acting Captain roundstart, they will be given the spare ID immediately.

## Why It's Good For The Game
In extreme lowpop scenarios, an acting Captain is selected, but sometimes they can't do anything. This would mean that the acting Captain selected receives a spare ID code in the process, automatically.

## Proof Of Testing
image

## Changelog

:cl: ReturnToZender, Burger (code help)
add: Extreme acting captains will now receive the spare ID code when they are selected
/:cl:


theselfish:
@Iamgoofball said this was just a config change. Why not change that?

theselfish:
> > @Iamgoofball said this was just a config change. Why not change that?
> 
> Investigation has shown that this is not just a config change, the system is just not working in extreme scenarios, and that's why we designed this

No complaints from me! I do think, that for comedy the Clown should come after the HoS but that's just personal preference. 

Great PR! :)

NargaInDisguise:
Really theres two ways this could possibly go; Passing the buck to someone wholly unprepared to suddenly be the literal captain after only selecting to be a crewmember, and then cryo'ing because they didn't want this, or it works out decently okay. I don't think this will be great but it's certainly an option to try out.

theselfish:
> Really theres two ways this could possibly go; Passing the buck to someone wholly unprepared to suddenly be the literal captain after only selecting to be a crewmember, and then cryo'ing because they didn't want this, or it works out decently okay. I don't think this will be great but it's certainly an option to try out.

I think you, and some others have this oddly negative view on Captain and Command in general. One I personally don't understand as the 'responsability' of Captain is way overblown.

I think people will be confused, yes. But, it might make some future Captain players. People should be pushed out of their comfort zone, and I think this is a great way to do so.

Arturlang:
why not give them the cookie + bridge access?

KateAnbolho:
The issue with not wanting to deal with AC as a head is enough to make me not ready a bunch tbh, I just wanna deal with the rest of what my job needs and not have the 3 extra targets on my back

NargaInDisguise:
Being a captain, acting captain, or even part of command are all things that your character has to be qualified for IC. They also come with massive upped responsibilities and very much do paint a target on your back. I won't pretend like most of command barely or doesn't seem that way to begin with, but both this and the new PR (https://github.com/Bubberstation/Bubberstation/pull/4763) simply pass the buck on potentially completely unqualified characters that never should be part of command.

Arturlang:
https://github.com/Bubberstation/Bubberstation/pull/4760 should be enough, we can reopen this if it's not

PR #4760, opened 2025-09-30
Lone operative can't spawn below 30 players

ReturnToZender:
## About The Pull Request
This prevents Lone Operative from spawning on extreme lowpop
## Why It's Good For The Game
Prevents lowpop players being punished for the NAD overall
## Proof Of Testing
Compiles, currently testing

</details>

## Changelog
:cl: ReturnToZender
add: Lone operatives can no longer spawn below 30 players.
/:cl:

PR #4763, opened 2025-09-30
[TM Only] Expands the Chain of Command to every job.

theselfish:
## About The Pull Request
Alternative to https://github.com/Bubberstation/Bubberstation/pull/4757

https://github.com/Bubberstation/Bubberstation/pull/4757 has an issue where random jobs are getting the biscuit when jobs who should get it exist on station.
## Why It's Good For The Game

It really isn't but we've got to fix this Lone Ops drama somehow.

I don't think this PR should exist. This drama/discourse should not exist.

But people refuse to pick up the disk. So now, players are being forced to do it. You made your own bed.

## Proof Of Testing

image

## Changelog
:cl:
add: There must always be a Captain in the Bridge.
/:cl:


ItzRiumz:
I will preface this and say I’m not sure this is really needed since the whole lone op on lowpop issue seems to be solved with the pop requirement pr. If we are going to merge this there is some questionable placements when it comes picking which jobs become acting captain.

I’m confused on some of these placements for the chain of command. Specifically warden being below the veteran advisor and the bridge assistant. The warden is usually expected to lead the security department if the hos is dead or if they don’t have one. They also have access to pretty much the entire security department which makes them a good acting hos in the first place. The veteran advisor should be below the warden with the bridge assistant either being above the veteran or below them.

The mime should absolutely be above shaft miners. It’s silly to think a shaft miner will care about the disk at all compared to the mime. Especially since their job content involves them leaving the station.

ItzRiumz:
> Actually. I think the Advisor and Assistant should be above the Warden, they have less to do compared to them. HoS is at the bottom of the default chain for this exact reason. They typically have a lot more to worry about.

I don’t think the fact that both of those jobs having less to do should factor in when it comes to this. While the hos is at the bottom of the chain they’re still above both the bridge assistant and the veteran.
the veteran is basically a glorified security officer while the bridge assistant is a coffee jockey for command members. It absolutely makes more sense that the warden would still be above those two when they’re the acting hos.

theselfish:
> CoC is a policy thing, why is this a PR anyway?

https://github.com/Bubberstation/Bubberstation/pull/4757 was causing random players to be given the biscuit even if existing CoC members existed. This, as stated in the PR body, fixes that, at the cost of losing the randomness and making it a hard structure.

Realisticly, a command member will probably show up. This just ensures someone has the disk up to that point.

Arturlang (approved):
Better than it being random

Arturlang:
https://github.com/Bubberstation/Bubberstation/pull/4760 should be enough, we can reopen this if it's not

PR #4753, opened 2025-09-29
The penalties for not securing a disk no longer occur if there is no captain (or acting) around because I heard people died to it once.

BurgerLUA:
## About The Pull Request

The penalties for not securing a disk no longer occur if there is no captain (or acting) around.

## Why It's Good For The Game

While securing the disk is a staple of Space Station 13, so is taking a stapler to the face.

This PR is meant to affect lowpop times when there is not even a captain assigned to secure the disk, which is pretty common. Players shouldn't be punished for playing on lowpop when 90% of the population are probably there to get their 3am gooning session in.

## Proof Of Testing

If it compiles, it works.

## Changelog

:cl: BurgerBB
balance: The penalties for not securing a disk no longer occur if there is no captain (or acting) around because I heard people died to it once.
/:cl:

PR #4775, opened 2025-10-02
Revert "Lone operative can't spawn below 30 players"

StrangeWeirdKitten:
Reverts Bubberstation/Bubberstation#4760

This was merged way too quickly in like 2 hours. So I'm going to revert it. I am also reopening #4753 which should be merged after this revert is merged.

🆑 
del: Reverts the 30 pop minimum for Lone Operative
/:cl:


NargaInDisguise:
So, why NOT have a min-pop for loneop though? Is it really so important to space station to have a loneop show up at 15 pop? 15-20 minimum doesn't seem like it would hurt, no?

Issue #4788, opened 2025-10-05
Powerator selling over it's limit


TheEmber:
### Issue Summary:

When powerator connected to grid, power of which exceeding 5 MW, it capable of selling more than 5 MW, which apparently, should be limit for it. It occurs when it powered via tesla coils. Not sure about SMES though.

<img width="525" height="139" alt="Image" src="https://github.com/user-attachments/assets/a9a348a1-9430-4687-b62c-7a51c0b37658" />

### Round ID:

9713

### Test Merge Information:

[#4468: 'LTO - RimPoint / Wawastation' by lessthnthree at commit b7621862ce](https://github.com/Bubberstation/Bubberstation/pull/4468)
[#4584: '[TM ONLY] finally eat the stupid honey sweet roll' by trazodont at commit 32bfdaa81e](https://github.com/Bubberstation/Bubberstation/pull/4584)
[#4627: 'Reworks how certain Security roles are counted towards the antag cap because I observed a round for one minute once.' by BurgerLUA at commit 63fa5ca274](https://github.com/Bubberstation/Bubberstation/pull/4627)
[#4572: '[SR REPLACEMENT] Buffs/Reworks plasma marksman' by KazooBard at commit e2f8888ca9](https://github.com/Bubberstation/Bubberstation/pull/4572)
[#4759: 'Cryogenics now sends an alert when the last member of command goes to Cryo' by nevimer at commit 5cffbc8cd2](https://github.com/Bubberstation/Bubberstation/pull/4759)
[#4745: 'Removes some Bay references from the Credits' by theselfish at commit 58edf1dba3](https://github.com/Bubberstation/Bubberstation/pull/4745)
[#4736: 'Restores Per-Map Shuttles' by theselfish at commit aff324270f](https://github.com/Bubberstation/Bubberstation/pull/4736)
[#4332: 'The full-custom blood colour quirk, again' by aKromatopzia at commit 9ea37d02ad](https://github.com/Bubberstation/Bubberstation/pull/4332)
[#4005: 'Adds an option to have four eyes + Multitool changing Robotic Eye Color ' by Spaghetti-bit at commit e19b2b30aa](https://github.com/Bubberstation/Bubberstation/pull/4005)

### Reproduction Steps:

1. Create power source which produces over 5 MW of power using tesla coils.(Occurs with both, SM and Fish generator)
2. Connect powerator, set power to sell to 5 MW
3. It gonna eventually sell more than it's limit.

[...]

PR #4789, opened 2025-10-05
Fixed bug with powerator?

TheEmber:
## About The Pull Request
Should fix [Powerator selling over it's limit #4788](https://github.com/Bubberstation/Bubberstation/issues/4788)
All simple. If power in grid bigger than max_power, it will use max_power value. If lower - it gonna use power in grid value. Looks sane.
## Proof Of Testing
Uhhhh... I tried to reproduce it, but I kinda struggled to fall into this condition with attached_cable.newavail() > max_power.
I think it will work though!
Update: These screenshots should do I think
<details>
image
image

</details>

## Changelog
:cl:
fix: Fixed powerator being able to sell more power than it supposed to
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/power/powerator.dm
+++ b/modular_zubbers/code/modules/power/powerator.dm
@@ -178,7 +178,7 @@
 		return
 
 	if(!attached_cable.avail(current_power))
-		if(!attached_cable.avail())
+		if(!attached_cable.newavail())
 			return
 		current_power = min(max_power, attached_cable.newavail())
 	attached_cable.add_delayedload(current_power)


modular_zubbers/code/modules/power/powerator.dm
```diff
@@ -180,7 +180,7 @@
 	if(!attached_cable.avail(current_power))
 		if(!attached_cable.newavail())
 			return
-		current_power = attached_cable.newavail()
+		current_power = min(max_power, attached_cable.newavail())
```
TealSeer, line 183:
```suggestion
		current_power = min(max_power, attached_cable.avail())
```
Ditto

modular_zubbers/code/modules/power/powerator.dm
```diff
@@ -180,7 +180,7 @@
...
 	attached_cable.add_delayedload(current_power)
```
TealSeer, line 184:
```suggestion
	attached_cable.add_load(current_power)
```
Similarly, machines are supposed to use `add_load()` instead of `add_delayedload()`

TealSeer (changes requested):
I think what's happening here is it's using a mix of avail and newavail, these aren't supposed to be mixed, machines are supposed to use one and non-machines are supposed to use another, having the `min()` in as a final sanity check should help if somehow `attached_cable.avail(current_power)` returns false but actually has more than the amount being requested

TheEmber:
Okay... I say I dont know how to fix issue with multiple powerators but this keeps happening consistently in rounds. Changing newavail for avail not helps with that and need to rewrite that part of code completely. But I will open PR for that once I figure out how it should be done.

TealSeer (approved):

PR #4793, opened 2025-10-08
Adds Bridge Assistant to Chain of Succession

Mewchild:
Per a conversation in staff, Bridge Assistant should be the lowest level on the chain of succession, given the bridge is their workspace and they're technically command. Wiki will be updated to reflect this upon merge.

🆑
add: Bridge assistant added to chain of succession, at the very lowest level.
/:cl:


Aisu9:
can they still roll antagonist? because that would be a huge problem.

Mewchild:
They can, but that's by design. As it stands, BA is _already_ first in line to get the spare ID simply because, in the event of no command, any crewmember can request the code via the command console. It just so happens that the BA is the only role that has access to it, by default. This PR changes nothing except cutting out the middleman.

PR #4797, opened 2025-10-08
RBMK Overhaul, Fixes, and Roundstart Engine Selection

BurgerLUA:
## TL;DR for engineers new/returning to the RB-MK

At T1, the RB-MK2 is cooled openly, like a supermatter. Generally speaking, the hotter it runs, the more power is generated. The more power it generates, the hotter it becomes. The gimmick here is that you're meant to find a sweet spot temperature for your setup and go with it to maximum power generation.

RB-MK2s melt down when too much power is generated. Melt downs are (generally) prevented when the safety system is enabled. When safeties trigger, the reactor does get damaged, which can easily be fixed with a welder. Sometimes the rod might get stuck if the reactor is too damaged when safeties are triggered.

If the RB-MK2 melts down, you should immediately decrease power generation. You can do this by blasting the machine with super cold air and disabling overclock settings. You can forcibly stop a meltdown by disabling the RB-MK2 forcibly, but this will cause the machine to suffer massive damage that may or may not explode violently in your face depending on the damage of the machine and how far in the meltdown stage it is.

You should not be turning off safeties or messing with the vent settings or overclocking it unless you know what you're doing.

You should at most build an RBMK in a small 9x9 circle, with an empty space for the sniffer. Building too many RB-MK2 units close together will increase tritium consumption, which in turn increases power generation and heat generation.

## TL;DR for admins

Expect engineers to immediately melt down the RB-MK2 when they get their hands on it. Feel free to untm this PR or get a maintainer to un tm this if this ruins too many rounds (either through bugs or just general fuckery).

### Major Changes

Nearly every aspect of the RB-MK2 was tweaked in terms of power generation, heat generation, and tritium consumption. It should be somewhat balanced now. It is now normal (and intentional) to run the RB-MK2 hot instead of doing the coldest possible temperature.

Replaces the roundstart station supermatter with an engine selection beacon. This beacon cannot be moved, and allows you to select the engine to use for the shift. Currently supported engine types are RB-MK2 and The Supermatter Crystal (large/standard).

Adds RB-MK2 upgrade disks that can be researched. This can be researched after energy manipulation. Current upgrades are: Safety Improvements, Overclock Upgrade, and Auto-Vent upgrade.

Adds RB-MK2 voidouts. RB-MK2 voidouts occur during meltdowns when there is no tritium left to process. This is a bad thing and you shouldn't do this. This also fixes the infinite meltdown state bug.

Adds Supermatter interaction with the RB-MK2. You can construct an RB-MK2 next to a supermatter and fill the fuel canister with hyper-nobellium (along with tritium for fuel) to directly energize the supermatter to absurd levels. At least 30 moles of hyper-noblium is required for this (does not get consumed).

Building too many RB-MK2 units close together will increase tritium consumption, which in turn increases power generation and heat generation.

Building too many RB-MK2 units (more than 12) may trigger very negative events. (Subject to change).

### Minor Changes
RB-MK2 research was moved to the energy manipulation techweb instead of it having its own node.

You can no longer overclock the RB-MK2 without the Overclock upgrade.

Tweaked the sensitivity/spamminess of the RB-MK2 sniffer.

Improved the sprites of the RB-MK2 sniffer.

## Why It's Good For The Game

RB-MK2 needed code improvements and this PR aims to do this. With bug fixes added and new features added, this should be worthy of acting as an alternative to the supermatter.

## Proof Of Testing

Tested engine deployment and cancelation.
Tested normal operations and meltdown operations/fixes/etc.

## Changelog

:cl: BurgerBB
add: Replaces the roundstart station supermatter with an engine selection beacon. This beacon cannot be moved, and allows you to select the engine to use for the shift. Currently supported engine types are RB-MK2 and The Supermatter Crystal (large/standard).
add: Nearly every aspect of the RB-MK2 was tweaked in terms of power generation, heat generation, and tritium consumption. It should be somewhat balanced now. 
/:cl:

### DIFF
--- a/modular_zubbers/code/game/machinery/burger_reactor/reactor.dm
+++ b/modular_zubbers/code/game/machinery/burger_reactor/reactor.dm
@@ -434,14 +434,11 @@ GLOBAL_VAR_INIT(active_rbmk_machines, list())
 	safeties_max_power_generation = max(initial(safeties_max_power_generation),round(max_power_generation*(safeties_upgrade ? 0.9: 0.75),25000))
 
 	//Requires x4 servos
-	//Servos increase the strength of the fans, forcing out gas at a higher rate, leading to better cooling.
+	//Servos increase the strength of the fans, forcing out gas at a higher rate.
 	var/vent_pressure_multiplier = 0
-	var/heating_divisor = 1
 	for(var/datum/stock_part/servo/new_servo in component_parts)
 		vent_pressure_multiplier += new_servo.tier * 0.25
-		heating_divisor += (new_servo.tier-1)/(3*4*2) //At T4, this value should be 1.5.
 
-	heating_multiplier = initial(heating_multiplier) / heating_divisor
 	vent_pressure = initial(vent_pressure) * vent_pressure_multiplier
 
 /obj/machinery/power/rbmk2/ui_interact(mob/user, datum/tgui/ui)


BurgerLUA:
I have some very spicy/major changes stashed. Will probably need to get maintainer approval on what the consequences will be for having too many reactors present on the map. Will do that tomorrow.

PR #4850, opened 2025-10-24
Adds a training collar

sunnyariess:
## About The Pull Request
Adds a new Funny Sex Item in the form of the "training collar", a way for a person to subject themselves to the "well-trained" quirk from one person in particular- and for the other party to recieve the "dominant aura" quirk for that person.

Works by activating in-hand, and binding to a person chosen by list. I'm arguing that this is a better method than the lock collar, because you do not have to hand it off to your partner once/to set up. It's one-and-done.
image

When bound and worn, it also disables the "well-trained" quirk (while keeping the moodlet), this is to simulate solo ownership. Once the collar is taken off- or if there is no owner as it's worn, well-trained returns.
image

Works on borgs!
image

Also adds it to loadout and lustwish.
image
image

Lastly, it has logging for it being worn, and an owner being chosen.

Code shamelessly stolen from the subtle text interface and the well-trained and dominant aura quirks. Feel free to crucify me. Please tell me if the file's in the right spot.
## Why It's Good For The Game
Non-obtrusive sex item. https://www.youtube.com/watch?v=IPBnrIAWDeU
## Proof Of Testing
Compiles. Tested extensively with the "mobs without client" checks commented out.
<details>

</details>

## Changelog
:cl: sunnyaries
add: Adds a "training collar" for people who want the well-trained experience, but find it too much.
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/lewd_machinery/sub_collar.dm
+++ b/modular_zubbers/code/modules/lewd_machinery/sub_collar.dm
@@ -30,6 +30,7 @@
 		owner = tgui_input_list(user, "Pick an owner", "Owner Selection", in_view)
 		if(!owner)
 			return FALSE
+		user.log_message("[user] set the [src]'s owner as [owner]", LOG_ATTACK)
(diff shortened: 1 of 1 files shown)


lessthnthree (changes requested):
See review comments

PR #4865, opened 2025-10-28
Stunswords cannot be crafted with cargo sabres, renfair claymores and bone swords

KazooBard:
## About The Pull Request

Upstream PR blacklisted almost all non-true variants, including a lot of unique ones - this is a very powerful weapon reserved for rare instances, NOT FOR MASS PRODUCTION. 

And with them not having all of the items this PR targets, it slipped by for us. They (items) shouldn't have been allowed to do this to begin with. 

Scream at me for formatting change on the override.

## Changelog

:cl:
fix: Stunswords cannot be crafted with cargo sabres/bone swords/renfair swords
/:cl:


SapphoQueer:
Screenshot_20251031-230041_Discord

Ashrollen:
Seems like sec 1984 continues. Don't worry guys server will probably die anyway but this...This is a good joke that i can throw on its funeral.

SapphoQueer:
> Seems like sec 1984 continues. Don't worry guys server will probably die anyway but this...This is a good joke that i can throw on its funeral.

What? This fixes an oversight based on tg not accounting for skyrat melee weapons we have when making the crafting recipe. It is not intentional for sec to get a baton with more damage than an esword with little to no effort.

PR #4888, opened 2025-11-02
Medical Protolathe combat/high tier implant removal

IronDragoon:
## About The Pull Request

Removes high tier organs and combat aimed implants from the medical protolathe. 

## Why It's Good For The Game

I believe this helps provide a greater level of responsibility and need on roboticists. It feels far more thematic that the robo would be the one to produce and install these. Medical already has plenty of responsibilities and this feels sensible for both gameplay balance as well as for RP sense. 

Often you will find crew going to medbay as first port of call for implants, and I feel that having roboticists take back this action as more part of their primary role, as opposed to an occasional thing helps flesh it out more and make it distinct. 

## Proof Of Testing

I tested it, it removes the stuff and runs

## Changelog

:cl:
balance: removed tier 3 organs and combat focused/specialized implants from the medbay protolathe
/:cl:


code/modules/research/designs/medical_designs.dm
```diff
@@ -449,7 +449,7 @@
 	category = list(
 		RND_CATEGORY_CYBERNETICS + RND_SUBCATEGORY_CYBERNETICS_IMPLANTS_HEALTH
 	)
-	departmental_flags = DEPARTMENT_BITFLAG_MEDICAL
+	//departmental_flags = DEPARTMENT_BITFLAG_MEDICAL//Removing to only allow robotics to fit these
```
Arturlang: make it clear it's a bubber edit, or modify it in a modular file

Arturlang (changes requested):

BasilTamaya:
> There's potentially an argument for the combat implants but tier 3 organs being gone is pretty questionable at best.

T1 and T2 were left alone as standard organ replacements. They exist to keep people alive if needed, but most players aren’t lining up for them.

Tier 1 is the basic prosthetic replacement. Tier 2 adds basic quality-of-life fixes like halving EMP vulnerability from 80 to 40 and lowering oxygen requirements for breathing for the lungs, improved toxins filter slightly better than normal for the liver, the heart has blood replication at T2 and slows down bleeding but with a toxins side effect for doing so, and the stomach at T2 digests faster and clears your sickness quicker.

Tier 3 on the other hand introduces EMP resistance that no lower tier cyber organ has, inclusive of better durability and atmospheric tolerance for the lungs, the liver processes toxins far more efficiently than the previous two, the stomach clears out grossness quicker as well, and the heart **gives a ten second grace where you can still run while critically injured once every five minutes because it gives trait_nosoftcrit.** Those benefits are **not medical requirements.** You'd only think about these if you wanted them in the first place, which is why it's being gated behind a fabricator like every other elective augmentation.

The goal of this PR is to gate non-essential upgrades so Robotics actually has more meaningful gameplay. We have a Cyberneticist role as well as a Biomechanical Engineering role as alt titles for people to roleplay a ripper doctor or whatever, but it can't be done because 90% of the server stops at **medical as their one stop shop for every augmentation.**

Yes, people visit Robotics for modsuits, but that’s about where it ends. The operating table sees more use reviving the rare dead synth every 4th round, and it doesn't help when most of the synths are in security getting treated by combat medic roles because Security for whatever reason has a whole operating room for that.

Roboticists start with surgical tools and augmentation resources for a reason. This change lets them use those tools for more than emergency repairs (which never sees the light of day) or revival, and gives them a proper role in cybernetic augmentation instead of leaving all of it to Medical.

NargaInDisguise:
> > There's potentially an argument for the combat implants but tier 3 organs being gone is pretty questionable at best.
> 
> Why? Can you please explain instead of just making a claim like that, because I disagree

Instead of talking about "combat implants" and "tier 3 organs" like I was originally going to, I'm actually going to go over the specific changes in your code that I think are confusing.

Tier 3 liver, heart, lungs and stomach specifically are like, literally never asked for by normal crew, these are only ever printed when they're available for someone that has had organs removed or damaged, and this is just going to push them into semi-obscurity by way of making them robotics only.

The reviver implant is literally designed to imitate a sci-fi pacemaker, no? This is quite literally a medical item.

Same with the breathing tube.

Everything else I can understand, even if I don't agree with the opinion, but those I feel are specifically still MEDICAL items through and through.

IronDragoon:
I have adjusted to add T3 organs back into the med lathe. I took out the jump-starter and nv eyes, I think the list looks more balanced as an inbetween here.

modular_skyrat/modules/medical_lathe_nerf/code/machines/medlathe_implant_denylist.dm
```diff
@@ -0,0 +1,63 @@
...
+/datum/design/cyberimp_thrusters
+	departmental_flags = DEPARTMENT_BITFLAG_SCIENCE
+
+/datum/design/cyberimp_welding
+	departmental_flags = DEPARTMENT_BITFLAG_SCIENCE
+/datum/design/cyberimp_welding/moth
+	departmental_flags = DEPARTMENT_BITFLAG_SCIENCE
+
+/datum/design/cyberimp_nv
+	departmental_flags = DEPARTMENT_BITFLAG_SCIENCE
```
theselfish: Trailing new line.

theselfish:
> Medbay has direct access to stasis beds which makes upgraded organs much safer to stuff into people, as well

MedBay isn't meant to be doing surgery on stasis beds.

PR #4919, opened 2025-11-14
[TM ONLY] Revert "Prevents SSwardrobe from messing with modsuit parts (#93779)"

TealSeer:
This is causing crashes, revert until proper fix


TealSeer:
https://github.com/Bubberstation/Bubberstation/commit/8b66328e4d2f9ce91c0c6e991dedd99643c351ca

PR #4925, opened 2025-11-15
Heretic now requires a chaplain in order to roll because I saw someone die to it once.

BurgerLUA:
## About The Pull Request

Heretic now requires a chaplain in order to roll. There must be at least 1 chaplain present in order for heretic to roll.

## Why It's Good For The Game

Heretic's design is very balance dependent on a chaplain existing, much like how blob is very dependent on engineers existing (blob requires engineers to roll) and how virus events are dependent on medical existing (viruses require medical to roll).

While holy water/null rods can be obtained by breaking into the chapel, it is very easy for a heretic to just steal the null rod.

## Proof Of Testing

Untested. I can't really test this solo.

## Changelog

:cl: BurgerBB
add: Heretic now requires a chaplain in order to roll. There must be at least 1 chaplain present in order for heretic to roll.
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/storyteller/event_defines/crewset/heretic.dm
+++ b/modular_zubbers/code/modules/storyteller/event_defines/crewset/heretic.dm
@@ -18,3 +18,11 @@
 /datum/round_event_control/antagonist/solo/heretic/midround
 	name = "Midround Heretics"
 	roundstart = FALSE
+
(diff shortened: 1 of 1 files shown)


Yobrocharlie:
You should also add chaplain to the job blacklist for heretic,
because the creator of heretic is a very,
smart,
and,
intelligent,
person and they allowed people to roll both chaplain and heretic at the same time.

PR #4926, opened 2025-11-15
Fix compile errors on 516.1670+

TealSeer:
## About The Pull Request

All the instances of numbers being used as keys in assoc lists were just regular lists with the indexes written out for readability, comments all these out so it stops erroring
## Why It's Good For The Game

Compile on new versions yes

## Proof Of Testing

Yes

## Changelog

N/A


nevimer:
interesting

PR #4948, opened 2025-11-22
removes the speed boost from cybernetic raptoral bodyparts

SapphoQueer:
## About The Pull Request
Teshari legs do not by default get the speed boost, I don't see why a roundstart robotic equivalent augment that *gives* quirk points should also make you faster when normal tesh legs dont.

also does it for the (unobtainium) digitigrade ones for futureproofing maybe idk

## Why It's Good For The Game
uhhh probably an oversight from skyrat ngl

## Proof Of Testing
a bunch of 1 line changes

<details>

</details>

## Changelog

:cl:
fix: Cybernetic raptoral parts no longer speed you up.
/:cl:

### DIFF
--- a/modular_skyrat/modules/bodyparts/code/teshari_bodyparts.dm
+++ b/modular_skyrat/modules/bodyparts/code/teshari_bodyparts.dm
@@ -115,7 +115,6 @@
 	base_limb_id = SPECIES_TESHARI
 	brute_modifier = TESHARI_BRUTE_MODIFIER
 	burn_modifier = TESHARI_BURN_MODIFIER
-	speed_modifier = -0.1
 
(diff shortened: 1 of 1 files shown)


bunny232:
as the person who likely causes this pr you should also remove the speed bonus from 
obj/item/bodypart/leg/left/digitigrade/teshari and the right counterpart if we do not want teshari having extra speed.
edit: should also probably change the burn mod to 1 instead of .9 so they are inline with regular roundstart cybernetics

SapphoQueer:
Did the change, not particularly worried about the burn/brute modifiers. It's fine really. This pr JUST kills the movement speed bug.

PR #4969, opened 2025-12-01
deletes all job-specific bunnysuits

axietheaxolotl:
## About The Pull Request

does what it says in the title. tested, and it works.

to be cordial, I have kept the greyscale and custom versions of these. so if you really wanted to make them again, you can just make them in-loadout with GAGs, they just won't have any built-in resistances.

i also kept the tailcoats, as while I personally don't like them. they're at least justifiable and can't find any ontological reasons for them to not exist beyond personal dislike and mild sprite criticisms.

if this is too extreme for the maints i'd be willing to tone this down to just deleting them from vendors or locking them behind hacking. this was just my first instinct as I don't really think they really carry any value.

## Why It's Good For The Game

there's a fundamental problem with these that I feel I don't need to mention, but ignoring the giant fetish content shaped elephant in the room (address me) there's several non-moral problems with these:

- one, they're bloated. not only is their one for EVERY job in the game. many have alternates (for example, lawyer has FOUR). including that, there's both syndie, and wizard ones. ever seen them? probably not
- two, they're not very good looking. some are okay, and some are even mildly creative. this doesn't change the fact most are tacky, poorly shaded cubes that get muked up by a bunch of other items that people tend to wear. they don't have any mix-and-match potential, you either wear the full-set, or you don't.
- three, like I said before they're incredibly tacky. they're devoid of most taste and look like cheap Halloween costumes.
## Proof Of Testing

wtf do I even put here the compile log? idk it works ok and I tested it on my machine check the runtimes chud
<details>

</details>

## Changelog
:cl: axietheaxolotl / viro
del: job-specific bunnysuits have been removed, greyscale ones remain in game and are still in the loadout.
/:cl:

### DIFF
--- a/_maps/bubber/automapper/templates/lavaland/lavaland_persistence.dmm
+++ b/_maps/bubber/automapper/templates/lavaland/lavaland_persistence.dmm
@@ -4541,7 +4541,6 @@
 /obj/item/clothing/under/syndicate/skyrat/baseball,
 /obj/effect/spawner/random/exotic/syndie,
 /obj/effect/spawner/random/exotic/syndie,
-/obj/item/storage/box/syndibunny,
 /obj/effect/spawner/random/entertainment/plushie,
 /obj/effect/spawner/random/entertainment/plushie_delux,
 /obj/structure/closet/generic/wall{


NargaInDisguise:
Look, I personally really don't like the look of them, but you're most certainly reaching for the "fetish content" tag.
That, and its up to maintainers to decide if somethings actually taking up a bunch of space and load time from existing, not us.

You should definitely bring this up to maintainers first before making a PR like this.

SapphoQueer:
Really not a fan of a full removal. Instead of deleting them, maybe put the entire outfit in a box, and letting you buy the "bunny kit" box from a vendor? Reduces bloat in the vendor. Lets people still buy it.

axietheaxolotl:
I think some of you are forgetting that if you really want a department-coloured bunnysuit, I have left ways to still get them. It'll just be the greyscaled one, really, there's a benefit to this as you can get it roundstart, something you can't do with the snowflaked ones.

> This seems... silly. And, honestly I'd like to address that elephant in the room. If silly clothes like these are creating a moral quandary, you might as well remove the _**entire vending machine that caters to similar things.**_ Nobody puts these things on that doesn't _**want**_ to wear them, so the whole moral high ground thing that is alluded to is silly, like I said earlier. If the time comes where a few outfits become the difference between the server running in a stable fashion or not, then yeah, maybe they should go. For now, though? Seems like a salt pr from someone with a vendetta against silly clothes.

insane whataboutism but ignoring that there's a strong distinction between the autodrobe, something that exists in only two places on the map, and bunnysuits and ears clogging up every single vendor in the game. also silly clothes are not the same as obvious fetish content that people are required to interact with.

> Look, I personally really don't like the look of them, but you're most certainly reaching for the "fetish content" tag. That, and its up to maintainers to decide if somethings actually taking up a bunch of space and load time from existing, not us.
> 
> You should definitely bring this up to maintainers first before making a PR like this.

what do you think PRs are for exactly?

regardless, i'm creating an alternate PR to just lock them behind the contraband wire as I think that's a more digestible change, even if I think it's a bit of a slap on the wrist.

theselfish:
I disagree with this for biased, obvious reasons.

But I should add, some, a lot, of these items are unique or have unique properties.The pendant is just a necklace, the Syndibunny suit is armored. A lot of them aren't just plain variants.

I also agree with what other players say, 'you choose what you wear'.

PR #4985, opened 2025-12-06
Disables borers because I died to it once.

Owai-Seek:
## About The Pull Request

I made the weight 0 and made the minimum player count 999. I'm assuming that'll stop it from happening.

## Why It's Good For The Game

Borers are an unmaintained antagonist, they need a ton of work to bring them into an acceptable or balanced state. Something something "Improve, don't remove", but if leaving these enabled as-is causes a lot of headache and disruption considering how outdated they are.

As it stands now, they can give themselves free respawns by hiding eggs, act as free stimulant machines, dump infinite death chems into people at a moments notice, develop immunity to their primary weakness, and gain stupid amounts of health.

If a borer is bored or hates you, enjoy getting brain damage and organ failure while they pump an endless slurry of chemicals in your body until they get yanked out by an overworked medbay. It's not fun being taken out of a round for 30 minutes because something click-dragged itself into you while you were stunned by a sec officer or slipped on a tile of water. It's even less fun when they can do it again the moment you leave medbay because there was an egg stored somewhere.

## Proof Of Testing

This is one of those things that can't really be tested locally, I don't think, but it's just a numbers change and I read the datums so I knew to change minimum player counts instead of weights.

</details>

## Changelog

:cl:
del: Borers are (probably) disabled.
/:cl:

### DIFF
--- a/modular_skyrat/modules/cortical_borer/code/cortical_borer_antag.dm
+++ b/modular_skyrat/modules/cortical_borer/code/cortical_borer_antag.dm
@@ -93,8 +93,8 @@
 /datum/round_event_control/cortical_borer
 	name = "Cortical Borer Infestation"
 	typepath = /datum/round_event/ghost_role/cortical_borer
-	weight = 3
-	min_players = 20
+	weight = 10
+	min_players = 999
 	max_occurrences = 0 //should only ever happen once. If they were enabled, that is.
 	category = EVENT_CATEGORY_ENTITIES
 	description = "A cortical borer has appeared on station. It will also attempt to produce eggs, and will attempt to gather willing hosts and learn chemicals through the blood."
@@ -146,8 +146,8 @@
 	midround_type = LIGHT_MIDROUND
 	pref_flag = ROLE_BORER
 	min_antag_cap = 1
-	weight = 0
-	min_pop = 999
+	weight = 3
+	min_pop = 20
 	repeatable = TRUE
 
 /datum/dynamic_ruleset/midround/from_ghosts/cortical_borer/create_ruleset_body()


NiklasNeighbor:
I believe you’re supposed to set max_occurances to 0 to disable an antag

modular_zubbers/code/modules/uplink/uplink_items/dangerous.dm
```diff
@@ -4,6 +4,6 @@
 			learn new chemicals through the blood if old enough. Be careful as there is no way to get the borer to pledge allegiance \
 			to yourself. The egg is extremely fragile, do not crush it in your hand nor throw it. \
 			The egg is required to sit out in the open in order to hatch. (Cannot be hidden in closets, etc.)"
-	progression_minimum = 20 MINUTES
+	progression_minimum = 150 MINUTES // If someone wants to spend 10-20 TC for a borer egg after two and a half hours, that's their business.
```
xPokee: If you're disabling borers, remove the traitor item as well, unless you really want them to still be accessible in _some_ way. However, having them only buyable ten minutes before a potential shuttle call is the wrong choice here.

modular_skyrat/modules/cortical_borer/code/cortical_borer_antag.dm
```diff
@@ -146,8 +146,8 @@
 	midround_type = LIGHT_MIDROUND
 	pref_flag = ROLE_BORER
 	min_antag_cap = 1
-	weight = 3
-	min_pop = 20
+	weight = 0
+	min_pop = 999
```
xPokee:
```suggestion
```
If weight is at 0 you don't need to add this.

Owai-Seek: > 	var/weight = 10 //The weight this event has in the random-selection process.
> 									//Higher weights are more likely to be picked.
> 									//10 is the default weight. 20 is twice more likely; 5 is half as likely as this default.
> 									//0 here does NOT disable the event, it just makes it extremely unlikely

I put the minimum count to 999 specifically because the round control datum and it's variables says weight being zero does not stop it from spawning.

modular_zubbers/code/modules/uplink/uplink_items/dangerous.dm
Owai-Seek: Commented this out, left the actual code in so if it gets fixed later, they don't have to dig.

CB35SSC:
"Improve, don't remove" literally --  yet again I have to scream at coders for taking the easy way out. I absolutely adore borers and do not wish to see them gone. The roleplay potential -- which you are probably ignoring -- is too high to simply remove them entirely.

PR #5004, opened 2025-12-11
[TM Only] Clausophobia only triggered by genuine Santa

lessthnthree:
## About The Pull Request

Adjusts clausophobia to only trigger in the presence of Santa Claus, rather than any Christmas themed clothing.

## Why It's Good For The Game

We hand out santa hats to everyone, it's a death sentence

## Changelog

:cl: LT3
code: Only the real Santa Claus will now trigger clausophobia
/:cl:

### DIFF
--- a/modular_zubbers/code/datums/components/fearful/sources.dm
+++ b/modular_zubbers/code/datums/components/fearful/sources.dm
@@ -1,7 +1,7 @@
 // Adjusted clausophobia check to look for Santa themself rather than the clothing
 /datum/terror_handler/simple_source/clausophobia/check_condition(seconds_per_tick, terror_buildup)
 	. = ..()
-	if (!.)
+	if(!.)
 		return
 
 	var/certified_jolly = FALSE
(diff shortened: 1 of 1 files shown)


Sir-Lagsalot:
> Why not just make it a holiday-tied thing instead of a year-round thing? That way we still have, as the cool kids call it, 'sovl'. You could probably make it tied to the actual Santa mob from the event that summons Santa instead of just Santa clothes as well if the concern is it being super annoying to people who intentionally chose a negative quirk.

I can personally vouch for it being a lot more debilitating than it sounds, especially given it's a negative quirk on-par with something like Smoker. It's a bit much having it straight-up just kill you.

code/datums/quirks/negative_quirks/claustrophobia.dm
xPokee: Line 3 needs to be replaced, see the suggestion below.
```desc = "You are terrified of small spaces. If you are placed inside any container, locker, or machinery, a panic attack sets in and you struggle to breathe." // BUBBER EDIT CHANGE - Clausophobia Removal - Original: "You are terrified of small spaces and certain jolly figures. If you are placed inside any container, locker, or machinery, a panic attack sets in and you struggle to breathe."```

xPokee:
> I can personally vouch for it being a lot more debilitating than it sounds, especially given it's a negative quirk on-par with something like Smoker. It's a bit much having it straight-up just kill you.

That's why I'm saying don't remove it entirely but make it not tied to the clothes but to the Santa event mob instead. There'd be one of them per round, instead of potentially every player in the round having it at round-start.

Issue #5050, opened 2025-12-22
Heretic robes can be worn non-functionally, and still work


Shadow-Quill:
### Issue Summary:

The robes heretics get can be changed to be worn "non-functionally" around the neck.
They cannot be removed by third parties while on the neck, and still work as a focus when the hood is up.

They should be changed so they _CANNOT_ be swapped to non-functional neck wearing, given their use.

### Reproduction Steps:

Get heretic robes, ctrl+shift+click robes to toggle into non-functional mode.
Put robes on neck.
Toggle on hood, see it can be used as a focus.
Use a second body to strip off the robes.
Can't remove robes.

PR #5053, opened 2025-12-23
Disables all Heretic Robes to be swapped to non-functional mode (neck slot)

shayoki:
## About The Pull Request

Could fix: https://github.com/Bubberstation/Bubberstation/issues/5050

This PR adds all heretic robes to the `only_functional` blacklist that makes the robe unable to be switched to Non-Functional mode. This is a soft hack-fix but ultimately should be implemented anyways. There is an issue preventing specifically heretic robes from being stripped if equipped on the neck slot; thus making the wearer put the hood on whenever for the focus.

This PR prevents all heretic robes to be used on the neck slot and must be worn on the suit slot.

Additionally, I've added a message sent to your chat if you attempt to swap a blacklisted piece of suit into non-functional mode.

## Why It's Good For The Game

This keeps the forced functional blacklist updated with items that should remain functional due to its mechanical use.

## Proof Of Testing

<details>

https://github.com/user-attachments/assets/7b6e0d9b-69b9-41c9-943f-5ada9c634b68

</details>

## Changelog

:cl:
fix: Fixed heretic robes functioning on the neck slot.
code: Added a message to the chat when a suit can't swap to non-functional mode.
/:cl:


modular_zubbers/code/modules/clothing/neck_wearing.dm
```diff
@@ -31,6 +31,7 @@ Use CTRL + SHIFT + LEFT CLICK to turn them on and off.
 	if(!iscarbon(user))
 		return NONE
 	if(only_functional)
+		to_chat(user, span_warning("You can't swap [src] to non-functional mode!"))
```
lessthnthree:
```suggestion
		to_chat(user, span_danger("[src] does not have a non-functional mode!"))
```
makes it more clear imo that this is an intentional block, and not failing to swap due to external factors

lessthnthree (approved):

PR #5054, opened 2025-12-23
Blacklists more items from the nonfunctional neck-slot mode for suits; Makes the text not render on items which cannot be toggled

### LINKED ISSUE #5050: Heretic robes can be worn non-functionally, and still work

aKromatopzia:
...these probably shouldn't have made it past the original look-over. Resolves #5050* and hopefully doesnt come with that caching error.

Linter fails have nothing to do with my PR; my version of the codebase just has bad image caching.

This does not - unfortunately - actually fix the implementation of the nonfunctional neck-slot mode. A complete rewrite is... recommended.

## Why It's Good For The Game

Added items which - off the top of my head - are situation-ally advantageous. The wizard robes don't actually function anyways, but better safe than sorry.

## Proof Of Testing

It's basically a config change. it also works btw; ignore the name I bungled the new interface.
Screenshot_9

## Changelog

:cl:
fix: The prompt to toggle a suit into nonfunctional mode no-longer appears on items which it has been disabled on.
balance: Wizard, Cult (& Eldritch), DOTA, "durability"-parented, armoured mining tailcoat & hooded mining gears may no-longer be worn around necks.
/:cl:


aKromatopzia:
> The others I understand but why the explorer suit, goliath cloak and the mining tailcoat? Afaik they don't carry over armor when switched to non-functionality, and don't have any innate traits as far as I can tell?

the hood for the mining ones; tailcoat is the same as melon armour with having a bonus component.

aKromatopzia:
I HATE THIS CACHING ERROR. 
If it doesnt fix the liners idc

PR #5060, opened 2025-12-25
Adds set of tools upgraded with Proto Nitrate Crystals

amsy2:
## About The Pull Request

Adds new crafting recepies to upgrade experimental power tools by using Proto Nitrate Crystals as necessary component. PN tools have faster tool speed compared to power tool versions.

Jaws 0.7 -> 0.25
Drill 0.7 -> 0.25
Welder 0.5 - 0.25

Other then speed and sprites, PN variants of tools are the same.

## Why It's Good For The Game

So, Proto Nitrate Crystals exist and are made with Crystallizer in atmos but they are effectivly useless since only thing they do, is relaease some air, compared to, among other things, Healium Crystal that straight up reset air completely in large radius.

This is pretty damn lackluster for an item requiring Proto Nitrate, which isn't the hardest to make but it's definitely one of annoying gases, especially if you're not using HFR. Giving use to crystals made with it, would widen use of the gas, which curently, is only used to optimise the SM (although it's like the best gas in the game to do it), having it used for better tools means that all of engineering can benefit from production of crystals.

They still require power tools to make so they still need that lil bit of research, the speed is fast but it's still not as fast as alien tools which have tool speed of 0.1 which is why i feel like the speed is justified.

Theme wise it feels fitting. Proto Nitrate makes SM very efficient, so for the crystal i followed the logic of that increased efficiency, in case of tools it's speed.

## Proof Of Testing

image
image

</details>

## Changelog
:cl:
add: Added new Proto Nitrate tools to crafting menu. Faster then power tools but requiring Proto Nitrate Crystals to make.
/:cl:


amsy2:
Im dumb and misclicked on the close PR button

SapphoQueer:
put my merge approval on (but CI and stuff still needs to be fixed obviously)

amsy2:
> Does the tool speed affect the jaws of life's door opening speed too? Or is that tied to another different timer? If it does then .25 is pretty powerful, as the syndicate jaws of death only have a .5 and they're locked behind antag & 4TC. Proto nitrate crystals are also in the atmos mail lootpool as like, 10% chance or smt like that.

I did not take forcing open airlocks into account. From what I understand looking at code now, its same as tool sppod so yeah, it'd be rather fast. I think I can make it change to 0.5 speed or smth if it's in prying mode only. Syndie jaws are at 0.5 yes but they're also small item, and if you're not an engie then theyre gona be far more valuable to you since its AA that you cant easly get with engie jaws.

SapphoQueer:
Keep them at the same speed as syndie jaws for opening and it'd prolly be fine given the investment needed to make them in the first place plus item size.

KateAnbolho:
The syndiejaws do get the upside of being small enough for a rig and other storage things that store the same size stuff

amsy2:
Errors fixed. I made it so jaws have 0.5 toolspeed as crowbar but when you swap to cutters its 0.25 toolspeed like rest of the kit

PR #5063, opened 2025-12-27
Adds Firebot upgrade to heat/cool air to room temperature 

amsy2:
## About The Pull Request

Adds new Firebot variant and new Freon Firefighting Foam

Freon Firefighting Foam works same as Firefighting Foam (like ones in advanced extingishers) with added beenfit of bringing temperature down/up to 20C even if there's no fire, it also leaves no plasma stains on the floor. This foam is not meant to be available to players and used only by the upgraded Firebot.

You can upgrade standard Firebot by giving it a piece of Hot Ice! Doing so will make it use it's foam if air temperature is below 18C or above 22C, effectively helping keep corridors at room temperature.

## Why It's Good For The Game

Firebots are rarely given a chance to shine, giving them option to effectively work as walking space heaters widens their use and practicality by a lot, especiely on maps like Ice Box notorious for having their corridors go cold all the time. The upgrade also require Hot Ice which is an atmos made material made with Freon, that currently doesn't have many uses, giving atmos techs another way to contribute with their projects to well being of the station.

## Proof Of Testing

image
image
image

## Changelog

:cl:
add: You can upgrade Firebots by giving them Hot Ice. Doing so makes their foam fix air to room temperature.
/:cl:


lessthnthree:
The hot ice is goofy and adorable, I love it.

lessthnthree (changes requested):

lessthnthree (approved):

amsy2:
Ok I think that's all, last minute change was changing description of deafult firebot so people actually know that this upgrade exists 
image

PR #5068, opened 2025-12-29
Wabbajack sources of transformation now uses polymorph code because I got transformed into a honkbot once.

BurgerLUA:
## About The Pull Request

Reworks "Wabbajacking" from all sources and makes it uses polymorphing code instead. When transformed, a 5 minute cooldown starts where you can transform back into your original form again. Dying while transformed just makes you transform back instantly. Getting hit with a polymorph again transforms you back as well.

## Why It's Good For The Game

Wabbajack transformation is a good source of ~~calcium~~ round removal if you get something unlucky, like a cockroach, or if you get transformed and there is no way to get yourself recasted. Yes, I know that other skyrat forms of round restoration exists but not many people know about them or even use them. I think it's just better to rework this mechanic.

## Proof Of Testing

Draft because untested and feedback needed.

## TODO

- [ ] Check to see if this is compatible with other sources of transformation.
- [ ] Check how this interacts with cyborgs, AI shells, and AI cores.

## Changelog
:cl: BurgerBB
balance:  Reworks "Wabbajacking" from all sources and makes it uses polymorphing code instead. The jist of it is that getting transformed this way isn't permanent.
/:cl:


BurgerLUA:
Sound seems to break when Polymorphed, which could just be a /tg/ issue. I'll test tomorrow.

PR #5081, opened 2026-01-01
Reworks how Voidwalker kidnapping works because I died to it once.

BurgerLUA:
## About The Pull Request

Voidwalker capture no longer gives you a space-fearing trauma that places you in gay baby jail when captured, or gives you blackface (It's 2026, blackface is bad!!!).

Voidwalker caputer now does the following:
- You die and get burn-husked.
- You get painted black (washable).
- You teleport to a safe station turf near someone (like how wizard dice works).
- The area near where you are teleported turns to fake space, similiar to how a dimensional anomaly works.

## Why It's Good For The Game

Voidwalker is a really really really bad antagonist type for Bubberstation. Not only are they (generally) treated friendly until they murder, but they're another /tg/ ambush antag on a sea of piss of ambush antags.

Instead of dedicating my New Years Eve fundamentally reworking how Void Walker works for a niche furry erp server, I'm just going to rework what happens when you get scrungled by one so losing to a void walker isn't as awful.

Also I'm pretty sure that there are unintentional bugs with voidwalker that existed for a while. You're intended to be alive during the dumb puzzle you're supposed to do, but what usually ends up happening is that you're just stuck in crit which means you have to either succumb or wait until you die, which is just dumb.

## Proof Of Testing

Draft because untested.

## Changelog

:cl: BurgerBB
balance: Reworks the consequences for getting captured by a voidwalker because I died to it once.
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/antagonists/voidwalker/voidwalker_mob.dm
+++ b/modular_zubbers/code/modules/antagonists/voidwalker/voidwalker_mob.dm
@@ -1,0 +1,24 @@
+/mob/living/basic/voidwalker/take_them(mob/living/victim)
+
+	//Make the victim dark and husk them.
+	victim.reagents?.add_reagent(/datum/reagent/colorful_reagent/powder/black, 5)
+	victim.become_husk(BURN)
+
+	//Visual Effects.
+	victim.flash_act(5 SECONDS, override_blindness_check = TRUE, visual = TRUE, type = /atom/movable/screen/fullscreen/flash/black)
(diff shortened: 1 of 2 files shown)


BurgerLUA:
> Why not just do like the heretic sacrifice does and heal the person a marginal amount for the puzzle? Because you’re changing it now from “minor inconvenience” to “murder”.

The current problem with voidwalker right now is that people don't treat it as a threat because technically it doesn't kill you, it just gives you blackface and an annoying trama.

I want to change it by making people take it seriously while at the same time making the consequences for it consistent and not take time to complete. We shouldn't be making players do minigames for losing.

PR #5085, opened 2026-01-02
Removes the cryosleep head/antag check

KazooBard:
## About The Pull Request

The annoying popup doesn't happen. Made at request of staff

## Proof Of Testing

Trust me bro

## Changelog

:cl:
admin: removes AHELP_FIRST_MESSAGE from command/antags getting into cryosleep pods
/:cl:

### DIFF
--- a/modular_skyrat/modules/cryosleep/code/cryopod.dm
+++ b/modular_skyrat/modules/cryosleep/code/cryopod.dm
@@ -497,10 +497,11 @@ MAPPING_DIRECTIONAL_HELPERS(/obj/machinery/computer/cryopod, 32)
 
 	if(target == user && (tgui_alert(target, "Would you like to enter cryosleep?", "Enter Cryopod?", list("Yes", "No")) != "Yes"))
 		return
-	/*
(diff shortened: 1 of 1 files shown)


modular_skyrat/modules/cryosleep/code/cryopod.dm
```diff
@@ -497,13 +497,14 @@ MAPPING_DIRECTIONAL_HELPERS(/obj/machinery/computer/cryopod, 32)
 
 	if(target == user && (tgui_alert(target, "Would you like to enter cryosleep?", "Enter Cryopod?", list("Yes", "No")) != "Yes"))
 		return
-
+	/*
```
lessthnthree: I believe the request was not to remove the message entirely?

Rather remove antagonist roles from req_admin_notify. The popup should remain for heads of staff, but the message should be changed instead of telling them to adminhelp, simply a reminder to return all the job specific gear before leaving.

lessthnthree (changes requested):

PR #5106, opened 2026-01-08
Adds the Clicker

ibexgoetia:
## About The Pull Request

I was surprised to see that no one had ever added a clicker item, so I decided to take the challenge and tried to add one myself. This little item can be found in the lustwish vendors. It makes a 'click' sound when interacted.
The sprite is inspired by the flash, but remade with the lustwish 'theme'.

Dominant characters who use this are also able to 'reward' submissive characters. The submissive characters will do an emote (which shares the cooldown with the *snap interaction, that being 10 seconds). They will also get a tiny mood buff that will last for two minutes, and doesn't stack.

It works on cyborgs with the submissive module (without the mood buff, clearly), and cyborgs with the dominant module have access to the toy. It can be bought from the lustwish vendor.

clickerpreview

## Why It's Good For The Game

We've got an overwhelming amount of untrained bottoms.

## Proof Of Testing

https://github.com/user-attachments/assets/72b15bf2-d468-4e72-ae64-7216a65cc12c

Here's the borg module version.
image

</details>

## Changelog

:cl:
add: Lustwish presents the Clicker! With this you will now finally be able to train and condition your pets. Visit any Lustwish vendor to buy your own.
/:cl:


xPokee (changes requested):
You need to add the clicker to `modular_skyrat/modules/modular_items/lewd_items/code/lewd_items/_erp_disabled_item_enforcement.dm`, while I am very glad you added it to `modular_zubbers/code/_globalvars/lists/erp_items.dm` it is currently only implemented for the Christmas tree blacklist.

PR #4800, opened 2025-10-09
Moves RedSec to our folder

theselfish:
## About The Pull Request

Title. To be honest this started as just fixing up some errors I made with the bomber coats. Woulda fixed them sooner but telling someone on my own PR that 'I don't care' is apperently ragebaiting.

But that's in the past. This is the future. And the future is red.

I moved everything in the bubber redsec folder to the goofsec folder where the other reskins are handled. I ported Nova's red variants where I could. Removed overrides (like the ugly bulletproof helmet and the PDA colours) and the 'durr no skirt as hos/warden' thing.

I also removed species & joblocks from the Akula outfits. Nothing else has a species lock and the sprites work perfectly fine on the default spaceman.
## Why It's Good For The Game

Redsec is better, actually.

## Proof Of Testing

I've tested it but I wasn't sure what exactly to screenshot with the sheer volume of shit I done. But I tested all I done. It all worked.

## Changelog
:cl:
add: Most, if not all sec clothing now properly has a red variant via the alt click reskin system. 
add: Anyone can wear the Akula outfits now.
del: Removed Bluesec PDA override and the ugly bulletproof helmet.
del: HoS and Warden can wear skirts again.
fix: Atmos Bomber no longer hides legs or slows you down.
/:cl:

PR #5113, opened 2026-01-11
Removes armadyne webbing

CatoChristopherMrow:
## About The Pull Request
Removes armadyne webbing, and the belt.
## Why It's Good For The Game
The belt is being removed, becuase it's just a secbelt that you can already reskin into. Why is it needed there at all?

The webbing is an intentional balance change. https://github.com/Bubberstation/Bubberstation/pull/4800 forgot to move the blue version into the vendor, but all things considered something like this shouldn't exist.

In comparison to other sec belts, it gives two more slots, with the tradeoff of...nothing. The tradeoff it used to have isn't around anymore because of tg updates. It's just a straight upgrade of a TG item for half the price.

I say this as a sec main who always takes this item, this item is too good for what it is, and warrants the axe.
## Proof Of Testing
Works on my machine.
<details>

</details>

## Changelog
:cl:
del: Removes the armadyne webbing
/:cl:

### DIFF
--- a/modular_skyrat/modules/sec_haul/code/peacekeeper/armadyne_clothing.dm
+++ b/modular_skyrat/modules/sec_haul/code/peacekeeper/armadyne_clothing.dm
@@ -156,20 +156,6 @@
 	l_pocket = /obj/item/megaphone/command
 	id = /obj/item/card/id/advanced/armadyne/security
 
-/datum/outfit/armadyne_security/commander/post_equip(mob/living/carbon/human/H, visualsOnly = FALSE)
-	if(visualsOnly)
-		return
-
-	var/obj/item/storage/belt/security/webbing/W = H.belt
-	if(istype(W))
-		W.current_skin = "Black Variant"
-		W.icon = 'modular_skyrat/master_files/icons/obj/clothing/belts.dmi'
-		W.icon_state = "armadyne_webbing"
-		W.worn_icon = 'modular_skyrat/master_files/icons/mob/clothing/belt.dmi'
-		W.worn_icon_state = "armadyne_webbing"
-		W.update_appearance()
-	..()
-
 /datum/outfit/armadyne_security/high_alert
 	name = "Armadyne Corporate Security (High Alert)"
 	belt = /obj/item/storage/belt/security/webbing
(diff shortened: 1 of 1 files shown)


xPokee:
Your title and PR body are both very misleading, given it makes it seem like they are being fully removed from the code, only for them to be removed from a vending machine's stock. Aside from that, I don't see why the security webbing should get to stay either, if extra slots are what you are so worried about.

CatoChristopherMrow:
> Your title and PR body are both very misleading, given it makes it seem like they are being fully removed from the code, only for them to be removed from a vending machine's stock. Aside from that, I don't see why the security webbing should get to stay either, if extra slots are what you are so worried about.

Well, for starters, that vending machine is the only actual way to access them in game. I am removing the ability to obtain it, so.

Secondly, not only is security webbing double the price at 300cr, with less storage then the armadyne item, it's also a TG item.

Infact, due to updates over time, *no* sec belt can hold energy weapons anymore, making this just a twice as effective straight upgrade for half the cost of the intended TG item for this use case.

xPokee:
That doesn't really change my original point, that your PR is misleading. It implies that you are removing it entirely, not just from a vending machine. 

I was not aware however that the other webbing was a /tg/ item. 

I'd honestly rather this just get axed from the code entirely if it's going to be removed, as I doubt it will be used anywhere else, but I'll speak with the other maintainers on this.

CatoChristopherMrow:
> That doesn't really change my original point, that your PR is misleading. It implies that you are removing it entirely, not just from a vending machine.
> 
> I was not aware however that the other webbing was a /tg/ item.
> 
> I'd honestly rather this just get axed from the code entirely if it's going to be removed, as I doubt it will be used anywhere else, but I'll speak with the other maintainers on this.

Ok, fully removed. Two armadyne ERT roles used it, so I've just replaced it with the base TG webbing. Since we already have code that lets us reskin into the sprites the webbing used, I just made it auto reskin on spawn for those roles. 

Sprites not deleted, as we have code that lets the base TG webbing reskin into them (cosmetic only) as I said.

projectkepler-ru (commented):
You honestly should make it a reskinnable item

CatoChristopherMrow:
Merge conflict delt with.

CatoChristopherMrow:
Ok for realsies this time :)

PR #5117, opened 2026-01-12
Adds a new (to us) medium-lowpop map: Ouroboros (Ringworm)

xPokee:
## About The Pull Request
**All the credit for this map goes to Zydras of Nova/Skyrat. See [here](https://github.com/NovaSector/NovaSector/pull/773) for the original PR.**

This map is, as the name hints at, circular! It is pretty compact, about the size of MetaStation or maybe even KiloStation, however, like IceBox, it has two Z-Levels.
## Why It's Good For The Game
We are currently lacking map variety for low to medium population, as we have Box/IceBox, Meta, Tram, and Kilo. That's it. Two of those are different flavors of the same map. Despite having two Z-Levels, this map plays really well on deadpop, as I actually used to run it on a server of my own, so it would work quite well for the deadpop EU hours.
## Proof Of Testing
It compiled and ran locally with no issues. See below for some screenshots taken from the original PR, as I forgot to take screenshots when I tested it.
<details>
image

image

image

image

image

image

image

image

image

</details>

## Changelog
:cl:
add: Added a new map: Ouroboros!
/:cl:

### DIFF
--- a/_maps/map_files/Ouroboros/Ouroboros.dmm
+++ b/_maps/map_files/Ouroboros/Ouroboros.dmm
@@ -9218,10 +9218,6 @@
 /obj/effect/mapping_helpers/broken_floor,
 /obj/structure/fluff/paper/corner,
 /obj/structure/table/wood,
-/obj/item/paper/secretrecipe{
-	pixel_y = 6;
-	pixel_x = -2
-	},
 /obj/item/flashlight/lamp/green{
 	pixel_y = 16;
 	pixel_x = -5


SapphoQueer:
S TIER MAP

BurgerLUA:
Assistants have access to the kitchen back room. Only issue I found.

Though I am concerned about when we eventually get blob/maxcap/meteors on this station because I feel that one breach will space the entire map because of how compact everything is.

MrBobCool:
Security Medical doesn't get a locker inside of their brig medical area.
image

The correction officer lockers aren't locked at all, and contain whole ass sec-belts with equipment within them. This is super easily accessible by perma prisoners that are attempting to escape. They break down one door, go up a ladder, poof, free batongs + sec radio headset (non bowman).
image

latin313:
> Security Medical doesn't get a locker inside of their brig medical area. image
> 
> The correction officer lockers aren't locked at all, and contain whole ass sec-belts with equipment within them. This is super easily accessible by perma prisoners that are attempting to escape. They break down one door, go up a ladder, poof, free batongs + sec radio headset (non bowman). image

Also for the second part, CO as a whole for some reason doesnt have a lock on their locker (like across all maps). Theres always just free gear+swat gear in it thats unlocked and has been that way for probably over a year.

latin313:
> 
> Be the change you want to see... Having to round start hide all of the CO equipment in the adjacent riot locker above the two closets because a prisoner can break one door and go up a ladder to get access to free batongs is rather annoying as sec.

See you would think, but I was talking about the red riot locker being the one that doesnt have a lock. You basically have to manually put a card reader on the COs red locker for it to be locked (Otherwise its just a all-access lock)

xPokee:
> For some reason it seems sec just has all access to robotics

I'm not seeing anything that reflects this on the map. Do you know which round this was? Was it all security or just one officer?

xPokee:
> The vault has no power wire going to it and its electrified windows leaving it to die early.

It looks like it has wires coming down from maintenance? Are you sure there wasn't just a snipped wire somewhere?

image

xPokee:
Okay I added layer manifolds to atmospherics' upper level, they're in the windows, but they ARE there

PR #5152, opened 2026-01-20
Jaws of Rescue Rebalance

sippykot:
## About The Pull Request
The "jaws of rescue" are aggressively bad - they almost never leave the case they start in and are immediately ditched when the "normal" jaws are available; they are bulky, slower and restricted for no upside whatsoever

This PR makes them:

- Slightly faster than normal jaws (still slower than syndicate jaws) so they are good at their actual job; prying doors to rescue people.
- Normal size so they actually fit in the medical belt and can be carried in your backpack like every other tool.

The "silent" version of the jaws was made unobtainable - it was confusing and exclusively something theoretically made by a traitor to disguise their stolen jaws - while still leaving them much slower than normal.

Jaws of rescue are now craftable in their normal, unchanged form, the slightly increased speed makes them actually desirable to make. **YOU NEED NORMAL ENGINEERING JAWS TO CRAFT THEM**

## Why It's Good For The Game
tg added a dead on arrival item;
makes the thing actually usable, hopefully paramedics will fight over it instead of always ignoring it, and then go craft it when they can instead of just carrying yellow jaws

## Proof Of Testing
works on my machine; tested it on localhost
<details>

</details>

## Changelog
:cl:
add: jaws of rescue crafting recipe
del: silent jaws of rescue are no longer obtainable
balance: jaws of rescue are now NORMAL sized instead of BULKY
balance: jaws of rescue are now 0.6 tool speed, up (smaller number = faster) from 1
code: jaws of rescue contain less material to reflect they are NORMAL sized now
code: tweaked the description of jaws of rescue
fix: jaws of rescue were explicitly intended to fit in medical belt but tg made them too big to fit inside
/:cl:


lessthnthree:
Regarding your suggested changes, I would support:
Making the regular jaws of recovery WEIGHT_CLASS_NORMAL (This one could probably be pitched upstream at TG.)
Making regular jaws of recovery printable with same research as jaws of life

I would not support renaming, increasing the speed, or making them small. If you want full power jaws you should still need to cooperate with engineering.

You can then take that engineering jaws of life and upgrade it to the top tier by combining it with a bonesetter.

PR #5162, opened 2026-01-22
Horse bridles

ibexgoetia:
## About The Pull Request

This has been more or less in the works for a bit. This PR adds a horse bridle to the game. It's worn on the face/mask slot, and can be found in the loadout menu (under Masks) or in the lustwish vendor (under Restraints).

It includes blinders and a bit, which will blind and gag the wearer, respectively. By default both are inactive.
Interacting with it by alt clicking will cycle through a few modes: Blinders + Bit, Blinders, Bit, None.
When worn you can also set them individually (though the alt click cycling still works).

The blinders and the bit can also be cycled by others with the strip menu; this will take 4 seconds.
To avoid this from happening, you can interact with the bridle in your hand to secure the straps. This will also prevent you from setting them yourself; you will have to unsecure the straps for it.

## Why It's Good For The Game

It's a very neat addition for equine characters and for people into horse stuff. It's not necessarily an ERP item (in my opinion as much of an ERP item as the collars) so I haven't set it as an ERP item specifically, but I can change it if you think that would be better.

## Proof Of Testing
File is too big for github
https://files.catbox.moe/31vy7y.mp4
## Changelog 
:cl:
add: Added horse bridles to the loadout menu and the lustwish vendors. They come equipped with settable blinders and bit.
/:cl:


BingusTheClown (approved):
Thanks for the latest video!

PR #5179, opened 2026-01-26
[Early Pull] Update list of bad BYOND builds

lessthnthree:
## About The Pull Request

Update the list of broken BYOND builds that have been happening recently


xPokee:
We got this with the last upstream sync.

Issue #3252, opened 2025-03-07
NSFW headshots do not render


SapphoQueer:
### Client Version:

516.1657

### Issue Summary:

NSFW headshots do not render

### Round ID:

7357

### Test Merge Information:

- [fix color picker problems](https://github.com/Bubberstation/Bubberstation/pull/3194)
- [Blood now pools on the floor under mobs](https://github.com/Bubberstation/Bubberstation/pull/2884)
- [[SEMI-MODULAR] Persistence Cyborgs](https://github.com/Bubberstation/Bubberstation/pull/3081)
- [(MODULAR) Adds temporary areas for the BYOS gateway](https://github.com/Bubberstation/Bubberstation/pull/3118)
- [[SEMI-MODULAR] Adds the Psionic Holding cosmetic quirk for roleplaying species with light psionic abilities](https://github.com/Bubberstation/Bubberstation/pull/3133)
- [Disease outbreak incident display](https://github.com/Bubberstation/Bubberstation/pull/3148)
- [Wood Wall Resprite, Biodome changes, new Wooden CatWalks](https://github.com/Bubberstation/Bubberstation/pull/3107)
- [Micro Squishing 

### Reproduction Steps:

Have sfw headshot hosted on catbox
works fine
Have nsfw headshot hosted on catbox
no work fine
nobody else's works fine either.

YakumoChen:
It worked fine for me when I stripped down

Rackunz:
Facing this same problem. SFW headshots work fine, NSFW headshots independent of host all appear as this. I've tried downgrading to 515.1647 to see if this was a 516 irk, but still had the same issue.

![Image](https://github.com/user-attachments/assets/6f092888-c795-46da-8c8c-65a31d77a2a7)

JagOfTroy:
Tested this myself, NSFW picture worked for me when stripped down naked.

PR #5191, opened 2026-01-29
Removing if statement check on NSFW examine panel tab

BingusTheClown:
## About The Pull Request

Previously, NSFW Examine Panel text and NSFW headshots were only shown if certain conditions were met:
1. When the examinee's examine NSFW text prefs were set to 'Always'
2. When the examinee's examin NSFW text prefs were set to 'Nude only' and the examinee has nothing worn in the chest slot

This removes that text, as we do not use any similar pref checks for the NSFW ref to be shown. (The one from the pink icon in the bottom right corner of the examine panel)

## Why It's Good For The Game

Fixes issue: https://github.com/Bubberstation/Bubberstation/issues/3252

## Proof Of Testing

<details>

Before the fix:

With clothes on, and prefs set to 'nude only'
image

With clothes off, and prefs set to 'nude only'
image

**AFTER CHANGES:**

image
vs nude (they are the same) 
image

</details>

## Changelog

:cl:
fix: Fixed NSFW examine text and headshots not displaying when the tab is opened
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/examine_tgui/examine_tgui.dm
+++ b/modular_zubbers/code/modules/examine_tgui/examine_tgui.dm
@@ -140,9 +140,8 @@
 			flavor_text = holder_human.dna.features["flavor_text"]
 			art_ref = holder_human.dna.features["art_ref"]
 			name = holder.name
-			if(show_nsfw_flavor_text == "Always On" || (show_nsfw_flavor_text == "Nude Only" && !(holder_human.w_uniform)))
-				flavor_text_nsfw = holder_human.dna.features["flavor_text_nsfw"]
-				headshot_nsfw = holder_human.dna.features["headshot_nsfw"]
+			flavor_text_nsfw = holder_human.dna.features["flavor_text_nsfw"]
+			headshot_nsfw = holder_human.dna.features["headshot_nsfw"]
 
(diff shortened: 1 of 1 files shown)


ReturnToZender:
Feels as if the solution, in that case, would be to have the NSFW reference mirror the check

BingusTheClown:
> Yes. In my view the fact that the NSFW ref is being shown when the NSFW flavor text should be hidden is the core issue.
> 
> If we proceeded this way (which I don't think is the right direction anyways,) you'd have to remove the entire show_nsfw_flavor_text preference, since it no longer be doing anything.

Totally agree and I didn't do a dive into where that pref was being used.

My suggested solution would be: 
1. Change the checkbox for the ref 'is NSFW' to instead be a link to the NSFW ref, and then seperate out the SFW ref and NSFW ref into the respective tabs.
2. Hide the NSFW tab, based on the check for NSFW Flavour Text (and I mean properly hide it, so that when condition are met, it appears)
          a - Potential bug might be introduced with the onChange function, if someone clothes themselves while the examine panel is there, so that would require additional testing and confirmation to confirm nothing new is introduced
3. Update the wording/description of the NSFW Flavour Text to better align with what it does (e.g. CanSeeNSFW Examine Content)

PR #5202, opened 2026-02-01
Bitrunning Rewards vendor as a research Item

ibexgoetia:
## About The Pull Request

Adds the bitrunning rewards vendor behind the bitrunning research, and allows it to be printed from the cargo fabricator, just like the mining rewards console.

## Why It's Good For The Game

The mining rewards console can be researched and printed, so I don't see why the bitrunning one shouldn't be able to, too. This is especially important for ghost roles that have access to bitrunning, but lack the reward console, such as Port Tarkon, for example.
image

## Proof Of Testing

image
image

## Changelog

:cl:
balance: The Bitrunning Rewards Vendor can now be researched and built.
/:cl:


ibexgoetia:
forgot to close this

PR #5217, opened 2026-02-05
Clear Leg Wraps

xPokee:
## About The Pull Request
Adds clear leg wraps as a loadout item and to the code. This acts identically to the gear harness in that it has no on_mob sprite, so you can wear it but still have the benefit of having the clothing.
## Why It's Good For The Game
People won't need to take a negative quirk to get the benefits of shoes while not wearing them as a digitigrade legs user.
## Proof Of Testing
It worked fine when I tested it.
## Changelog
:cl:
add: Added a new item to the game and loadouts: Clear Leg Wraps, these functionally identically to the gear harness, but for shoes.
/:cl:


xPokee:
> Why not take the SPLURT route and simply make the trait neutral?

The quirk is one of the most unbalanced quirks we currently have in the code, in that pretty much everyone who intends to go barefoot takes it due to the fact it lets you avoid glass shards, broken bottles, etc. I considered removing the quirk outright in favor of this altogether since at least with these, they can be stripped off of a person to gain a combat advantage (or burnt off since they are flammable).

Also, SPLURT has both these and a neutral quirk.

PR #5233, opened 2026-02-08
[Nova Port] Updated Interact Panel UI

xPokee:
## About The Pull Request
Original PR: https://github.com/NovaSector/NovaSector/pull/6736

image

## Why It's Good For The Game
The old UI was super outdated.
## Proof Of Testing
See the image above.
## Changelog
:cl: Adeheuer, xPokee
qol: Updated the Interaction Panel UI.
/:cl:

### DIFF
--- a/modular_skyrat/modules/interaction_menu/code/interaction_component.dm
+++ b/modular_skyrat/modules/interaction_menu/code/interaction_component.dm
@@ -130,6 +130,7 @@
 		data["pleasure"] = user_pleasure
 		data["arousal"] = user_arousal
 		data["pain"] = user_pain
+		data["yourName"] = human_user.real_name
 
 
 	// self - the one who the interaction component belongs to, aka who it's opened on (confusing var name yep)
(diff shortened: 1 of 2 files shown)


tgui/packages/tgui/interfaces/InteractionPanel/tabs/InteractionsTab.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+interface InteractionsTabPropsData {
+  searchText;
+  showCategories;
+}
+
+export const InteractionsTab = ({
+  searchText,
+  showCategories,
+}: InteractionsTabPropsData) => {
```
TealSeer: When we pass props to components we use types instead of interfaces, and they get passed as a single props arg that you then pull data out of, for example in this instance it would be
```ts
type InteractionsTabProps = {
  searchText: string;
  showCategories: boolean;
};

export const InteractionsTab = (props: InteractionsTabProps) => {
  ...
  const { searchText, showCategories } = props;
  ...
};
```

PR #5238, opened 2026-02-10
Makes the blob not end the round upon reaching critical mass

NiklasNeighbor:
## About The Pull Request
Instead of ending the round, the blob now instead triggers Delta Alert and has their resource reserve uncapped. Resources are still gained as normal.
The Blob Roundend scene now instead triggers once the shuttle docks with the station.
## Why It's Good For The Game
I heard someone mention this as something they wanted to change at some point. Might as well do it myself.
Instead of ending the round, the blob now instead triggers Delta Alert and has their resource reserve uncapped. Resources are still gained as normal.
The Blob Roundend scene now instead triggers once the shuttle docks with the station.

## Proof Of Testing
<details>
image
image
image

</details>

## Changelog
:cl:
balance: Blob now no longer immediately ends the round upon reaching critical mass, and can still be fought off.
/:cl:


KateAnbolho:
tbh the hot take is that some round ends aren't the worst thing in the world, like if the crew can't kill the blob maybe they should lose

PR #5245, opened 2026-02-12
Dirty no longer smells

nikothedude:
## About The Pull Request

Title.
## Why It's Good For The Game

Lots of negative feedback said this sucks. So, its going
## Proof Of Testing
<details>

</details>

## Changelog
:cl:
del: Dirty no longer makes you smell like ass
/:cl:

### DIFF
--- a/modular_zubbers/code/datums/quirks/negative_quirks/dirty.dm
+++ b/modular_zubbers/code/datums/quirks/negative_quirks/dirty.dm
@@ -66,10 +66,6 @@
 /datum/quirk/dirty/process(seconds_per_tick)
 	if (cleaned)
 		return
-	for (var/mob/living/iter_mob in get_hearers_in_view(4, quirk_holder))
-		if (HAS_TRAIT(iter_mob, TRAIT_DIRTY) || HAS_TRAIT(iter_mob, TRAIT_ANOSMIA))
-			continue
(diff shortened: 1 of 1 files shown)


nevimer:
honestly you could replace this with like, some effect that spawns (like a dirt stain) when the mobs hugged or hit or whatever. if miasma was neutralized by oxygen i'd even suggest that. 

personally I believe the arguments against the smelly message are disingenuous given there is just so much more content in the game that someone could argue is kink content (xenos, heretics, surgery, disgust, cyborgs, boots, uniforms, just about anything really) so I don't find it to be a compelling argument - instead a red herring to omit their actual reason for disliking it. they should just tell you that they don't like it instead of doing that.

NargaInDisguise:
> honestly you could replace this with like, some effect that spawns (like a dirt stain) when the mobs hugged or hit or whatever. if miasma was neutralized by oxygen i'd even suggest that.
> 
> personally I believe the arguments against the smelly message are disingenuous given there is just so much more content in the game that someone could argue is kink content (xenos, heretics, surgery, disgust, cyborgs, boots, uniforms, just about anything really) so I don't find it to be a compelling argument - instead a red herring to omit their actual reason for disliking it. they should just tell you that they don't like it instead of doing that.

Literally only 1 person ever said it was kink content and they're obviously not being taken seriously. Most of the complaints did in fact just stem from "We don't like the mechanical mood debuff"

PR #5248, opened 2026-02-12
Adds a reagent that lets you remove c4 stuck to people because I saw an officer get stuck once.

BurgerLUA:
## About The Pull Request

Adds a reagent that lets you remove c4 stuck to people. It can be made by mixing Space Lube, Space Cleaner, and Drying Reagent together and spraying it on someone. Note that this does not defuse the c4.

You can no longer attach more than one stick of c4 on someone.

## Why It's Good For The Game

There is actually no way to remove c4 from people once it is applied. Once it's on someone, the only way to get it off is for it to be detonated. Now you might be thinking that it might be possible via surgery or other nonsense, but no. Literally what happens in the code is that the c4 gets moved to nullspace (inaccessible) and then the c4 creates the explosion on the thing it's attached to.

With absurdly long c4 timers existing, you can have a c4 attached for very long periods of time, which isn't really fun when you probably want to do other things.

## Proof Of Testing

image

Pictured: Monkey getting stuck with c4 then getting sprayed with the thing.
Not pictured: My dumb ass losing the screenshot showing the reagent can be created. Trust me, it works.

## Changelog

:cl: BurgerBB
add: Adds a reagent that lets you remove c4 stuck to people. It can be made by mixing Space Lube, Space Cleaner, and Drying Reagent together and spraying it on someone.
remove: You can no longer attach more than one stick of c4 on someone.
/:cl:


BurgerLUA:
> I know you can remove them via showering regardless, but does this remove stickers as well?

It does not

modular_zubbers/code/game/objects/items/grenades/plastic.dm
```diff
@@ -0,0 +1,88 @@
+/obj/item/grenade/c4/
+	var/datum/component/c4_body_tracker/body_tracker
+
+/obj/item/grenade/c4/plant_c4(atom/bomb_target, mob/living/user)
+
+	if(active)
+		return FALSE
+
+	if(isliving(bomb_target) && bomb_target.GetComponent(/datum/component/c4_body_tracker))
+		to_chat(user,span_warning("It would be pretty fucked up and evil to plant a second c4 bomb on someone..."))
```
nevimer: this is a bit too out of character for the games message delivery style

PR #5253, opened 2026-02-14
Alt PKA balance pass

MajManatee:
## About The Pull Request
Tweaks the alt-PKA damage values to accomodate for the fact some of them fire in special ways, or fire more projectiles

The shockwave is now no longer able to fit any variant of an AOE mod, to stop people from one-shotting everything,

The shotgun damage has been reduced as well, making up for the fact it fires three shots.

Comments have been added for the existing variants explaining the reasoning for their current damage level.

Adds a new var for restricted MOD types, allowing for other pkas having this added in the future.
## Why It's Good For The Game
It is not fun for other miners when someone three shots a boss they havent even had a chance to think about yet.
## Proof Of Testing
<details>

# change

</details>

## Changelog
:cl:
balance: rebalanced PKA-Alts
fix: the shockwave PKA can no longer attach AOE kits.
/:cl:

### DIFF
--- a/modular_zubbers/code/game/objects/items/more_pkas.dm
+++ b/modular_zubbers/code/game/objects/items/more_pkas.dm
@@ -200,7 +200,7 @@
 /obj/projectile/kinetic/shockwave
 	name = "concussive kinetic force"
 	icon_state = null
-	damage = 20 // Half normal dmg, 8 projectiles in a 360 around you. Good for CC and rock clearing.
+	damage = 40 // 8 projectiles in a 360 around you. Good for CC and rock clearing.
 	damage_type = BRUTE
 	armor_flag = BOMB
 	range = 1


modular_zubbers/code/game/objects/items/more_pkas.dm
```diff
@@ -190,7 +190,7 @@
 /obj/projectile/kinetic/glock
 	name = "light kinetic force"
 	icon_state = null
-	damage = 10
+	damage = 10 // Low dmg, high modularity. Can be made great, especially when dual weilded.
```
SapphoQueer:
```suggestion
	damage = 10 // Low dmg, high modularity. Can be made great, especially when dual wielded.
```

NargaInDisguise:
Please do NOT lower the railguns damage. It is balanced around the fact that it has an INCREDIBLY long recharge time, HAS to be two-handed, is BULKY, and HAS to take up your belt slot if you want to not hold it in your hands at any point.

Dropping from 100 to 80 damage changes the threshold of almost every single relevant lavaland/icemoon mob from a 3 shot to a 4 shot, making vent defenses so so so much worse than before. Judging it based on the crusher really isn't fair because that can also get upgrades while this cannot.

WHAT I DO think you should do, is drop its damage to FIFTY, and DOUBLE it against lavaland mobs! That's the ACTUAL change it needs, to stop it from being CBT in pvp situations.

Other than that, I think this is an unjustifiable nerf that I'm sure more than just I would disagree with.

MajManatee:
I did a bit of testing and where the others can compensate for the damage drop with modkits, the railgun cant.

NargaInDisguise:
> Actually, just realized you're never realistically going to hit the full 640 damage with the shockwave. It's 8 projectiles, 1 around you each, and they all hit twice. It's not 8 projectiles all centralized in one spot.

Unless the corner tech has been truly, genuinely patched, you can sorta jank a corner attack into shooting all 8 shockwaves into the same corner.

PR #5275, opened 2026-02-22
Tone down the recent buffs from biodegrade (melting cuffs)

muchatras:
## About The Pull Request
 Last december, TGstation added new effects to the handcuffs melting ability that changelings have.

Upon using it while cuffed, the changeling will now remove every possible restraints they have instead of a singular one, which means:
- Handcuffs
- Bolas 
- Shoecuffs
- Straightjackets (new) 
- Anything holding them

Then, it also added the ability to hurt, stun, and break free from any grab by throwing acid upon anyone (except other lings) holding the changeling.

This PR doesn't remove the full restrains melting buff, only the new ability to freely stun and acid the grabber.
## Why It's Good For The Game

This new ability buff allows a changeling to break free from all restraints while giving it a free card to obliterate any officer attempting to subdue it. Changelings are already incredibly hard to deal with, and are one of the strongest (not counting heretics) antag to combat with sometimes only a single security officer joining. Getting two to three armblade hits for totally free with a bonus acid hit is understandable for a server in which you can powergame defenses with a full population able to target you, but bubberstation is much less arcadey than upstream.

Hence why I believe only keeping the full restraints melt (which is the goal of the ability to begin with) is already a decent buff.
## Proof Of Testing
<details>
Ran locally, the code simply bypass the proc for everything mentioned so it won't mess up too much if the file is changed by upstream in the future

</details>

## Changelog
:cl:
balance: Toned down biodegrade buff
/:cl:

### DIFF
--- a/code/modules/antagonists/changeling/powers/biodegrade.dm
+++ b/code/modules/antagonists/changeling/powers/biodegrade.dm
@@ -70,9 +70,11 @@
 		playsound(user, 'sound/items/tools/welder.ogg', 50, TRUE)
 		. = TRUE
 
-	if(space_invader)
-		punish_with_acid(user, space_invader)
-		. = TRUE
+//----------------BUBBER CHANGE------------------------
+	//if(space_invader)
+		//punish_with_acid(user, space_invader)
+		//. = TRUE
+//---------------END OF BUBBER CHANGE------------------
 	return .
 
 /// Spawn green acid puddle underneath obj, used for callback


KazooBard:
The ability, which makes up for 20% of a changeling's ability budget by selecting it, and iirc 1/3 chemicals spent isn't by any means a 100% free escape. An officer attempting a solo-capture of a changeling, which made itself weaker in combat to prepare for this possibility, isn't running away with a get-away-free-card, but spending consciously. This is synonymous to the freedom implant, which in my opinion, doesn't make it a buff, but rather a fix to the oversight of using other methods that would have logically also been affected to keep the one antag that will be round removed on sight in the game for a while longer.

While upstream is more "arcadey", the way a ling's life ends is the same on both servers. It gets swarmed, tries to acid out of cuffs past revives, gets killedgets cremated/gibbed while fullstripped. 

Plus, I think you underestimate how easy it is to get 2-3 free armblade hits for an ACTUALLY free cost, such as a shovestun, slip, or god forbid you step on punji.  


In short, why tone down an ability which is expensive twice-over, barely keeps up with the traitor version which is fast, silent AND removes all the same items post-buff, specifically toning down the fact that it can stun you for the same duration as a shovestun, using the argument of low sec as a balancing factor? My brother in christ, every single target stun is going to be OP when you're fighting an antag alone (sleepy pen, basic bitch baton, soap).

PR #5294, opened 2026-02-28
Milsim Fixes

LordHookie:
## About The Pull Request
A downstream, a fork of a downstream, and a fork of a fork of a downstream all walk into Github.
The contributor says; _"I'm gonna have to fix this bug three times, aren't I?"_

Milsim CIN NPCs used an old way of causing mobs to drop corpses by putting them in their loot table. This was eventually deprecated, and their subtype was eventually given a **corpse** variable to keep their bodies in. Milsim NPCs were not updated to account for this, causing them to drop the default corpse and their own corpse. This fixes this by simply replacing the loot definition with a corpse definition.

There was a reference to an item that doesn't exist in Milsim. That's been replaced with what it was almost certainly supposed to have been in the first place.

I assume this doesn't need to be modular, since these are Skyrat files. Correct me if I'm wrong, I do not play or know anything particularly about Bubber.

## Why It's Good For The Game
fix bug

## Proof Of Testing
<details>
one corpes
image
</details>

## Changelog
:cl:
fix: Virtual CIN Soldiers will no longer undergo mitosis upon death.
fix: Fixed a non-existent item in Milsim.
/:cl:

### DIFF
--- a/_maps/RandomRuins/SpaceRuins/skyrat/vaulttango.dmm
+++ b/_maps/RandomRuins/SpaceRuins/skyrat/vaulttango.dmm
@@ -1437,14 +1437,6 @@
 	dir = 1
 	},
 /area/ruin/space/has_grav/vaulttango)
-"WB" = (
-/obj/effect/turf_decal/trimline/neutral/filled/line{
-	dir = 4
-	},
-/obj/structure/cable,
-/obj/machinery/light/small/directional/east,
-/turf/open/floor/iron/dark,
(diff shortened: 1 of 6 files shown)


SapphoQueer:
Nothing regarding the PR itself but I read the title as "muslim fixes" and i was like "what"

KateAnbolho:
> Nothing regarding the PR itself but I read the title as "muslim fixes" and i was like "what"

what the fuck

PR #5309, opened 2026-03-04
Kills airlocks lights

Arturlang:
## About The Pull Request
airlocks update their lights every time it calls update_overlays which seems pretty often, perf-wise this doesn't seem worth it with how common they are, i don't think they contribute much
Keep in mind, this keeps the emissives and overlays and whatever, and only removes the outward lights emitting from the airlock

## Why It's Good For The Game
tracy told us that lightning can use a helluvalot of time (old trace though)

## Proof Of Testing
um

<details>

</details>

## Changelog

:cl:
del: for performance reasons, airlocks no longer emit their own lights
/:cl:


lessthnthree (changes requested):
If you're going to remove this, perform a proper removal. You're leaving dead code above where we needlessly set a now unneeded pre_light_color var and a bunch of color defines.

lessthnthree (approved):

PR #5343, opened 2026-03-14
Adds the Off-Station Deathrattle

ibexgoetia:
## About The Pull Request

_"Going on an adventure far from the comforts of your station? Unsure you'll make it back in one piece? Kahraman Industries and Nanotrasen present the Off-Station Deathrattle implant! This little chip connected to your vitals will send an automatic distress signal in case of your early demise. It will only work outside of Station areas. Warranty is void if death results from falling into pits or diving into lava or plasma lakes."_
---

The Off-Station Deathrattle Implant is, in fact, a Deathrattle. Its only differences are that instead of sending the message to other people with the implant, it instead sends the message to all silicons, as well as everyone wearing a radio with Cargo OR Medical encryptions. It even works when the radios are malfunctioning! It only works outside of the Station (And Centcom/Interlink) Z level. Space and Space ruins also work, so this implant is a viable option for spacetiding, as well.

It can be bought from the Mining Rewards Vendor, or alternatively from the supply console. In both cases it has the same exact price as the Kheiral cuffs.

On lowpop shifts, when the active crew is less than 10, the message will be sent to everyone's radio, regardless of encryption key.

## Why It's Good For The Game

Gigantic miner QOL. Miners are oft left to rot in lavaland, and this is especially the case on lowpop shifts, where people simply forget about the miner until the materials stop rolling in, or extremely late into the round, requiring admin intervention.

Now the moment a miner (or spacetider) perishes, the whole medical and supply crew will be notified.

## Proof Of Testing

image
image

## Changelog

:cl:
add: Added the Off-Station Deathrattle Implant, for use by Miners and Spacetiders. It will notify Cargo, Medical and Silicons of your demise via radio.
/:cl:


modular_zubbers/code/modules/implants/code/implant_deathrattle_offstation.dm
```diff
@@ -0,0 +1,176 @@
...
+		return TRUE
+
+	if(is_mining_level(owner_z) || is_spaceruins_level(owner_z))
+		return TRUE
+
+	for(var/trait in offstation_traits)
+		if(SSmapping.level_trait(owner_z, trait))
+			return TRUE
+
+	return FALSE
```
Majkl-J: Please compress your whitespaces, this is just one example but if the whole block is just checks you really don't need to put spaces inbetween
```suggestion
	if(is_centcom_level(owner_z) || is_away_level(owner_z) || istype(get_area(owner_turf), /area/centcom/interlink))
		return FALSE
	if(isspaceturf(owner_turf))
		return TRUE
	if(is_mining_level(owner_z) || is_spaceruins_level(owner_z))
		return TRUE
	for(var/trait in offstation_traits)
		if(SSmapping.level_trait(owner_z, trait))
			return TRUE
	return FALSE
```

PR #5347, opened 2026-03-17
Removes Hot Potato from maintenance loot because I died to it once.

BurgerLUA:
## About The Pull Request

Removes Hot Potato from maintenance loot because I died to it once.

## Why It's Good For The Game

When I initially added this to the maintenance loot list, I thought it just blew you up. I didn't think it would gib.

## Proof Of Testing

If it compiles, it werks.

## Changelog

:cl: BurgerBB
del: Removes Hot Potato from maintenance loot because I died to it once.
/:cl:


BurgerLUA:
> Commenting - isnt there already a 'fake' one that spawns on some stations? (like a toy version) Can't it just be replaced with that one?

Already in maint loot

PR #5366, opened 2026-03-22
fixes the digitigrade sprites unit test

Phantastic-Swan:
## About The Pull Request

As per the title. The unit test was never skipping any typepaths, regardless of whether or not they were defined in the modular folders. This made it so that even TG clothing was being checked. The regex which was used to find the typepaths was also broken, as it was prematurely cutting off at the letter `s`, and was also matching with typepaths that weren't actual definitions/overwrites. It also stopped searching for typepaths in the file after the first one was found.

## Why It's Good For The Game

Fixes a unit test.

## Proof Of Testing

image

It compiled and run the test fine. Checking with the debugger confirmed that the `typepaths_to_check` list (renamed from `typepath_files`) was being properly filled with clothing from `modular_skyrat` and `modular_zubbers` folders.

This screenshot shows the status of the `typepath_files` list before the changes:
image
As can be seen, it is empty, meaning the `if(subtype_string in typepath_files)` line is never going to fire, because there is no item in the list.

This screenshot shows the status of the `typepaths_to_check` list after my changes:
image
as you can see, it has 599 entries, all of which were defined or overwritten in `modular_zubbers` or `modular_skyrat` files.

### DIFF
--- a/code/modules/unit_tests/~skyrat/digitigrade_sprites.dm
+++ b/code/modules/unit_tests/~skyrat/digitigrade_sprites.dm
@@ -1,3 +1,7 @@
+/// the regex.find() proc returns 0 if it has found nothing.
+/// So we're replacing the magic number with a name.
+#define NO_MATCH 0
+
 /datum/unit_test/modular_digitigrade_sprites
 	var/type_to_test = /obj/item/clothing/under
 	var/list/modular_folders = list(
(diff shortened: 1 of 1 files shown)


Phantastic-Swan:
Yes, however the issue was that TG clothing WAS being checked. As I said, the `typepath_files` list was always empty because the regex was used incorrectly

Also the logic WAS inverted. The code was set to skip anything in the `typepath_files` list, and the rest of the code implies that that was the list which stored skyrat and bubber items.

PR #5384, opened 2026-03-28
brings back changing emitter icons

Phantastic-Swan:
## About The Pull Request

We used to have emitters change their colors depending on if they're on, off or underpowered. This has been removed some time ago (I assume as a result of an upstream merge), and this PR brings that back

## Why It's Good For The Game

Emitters changing colors:
- looks nice
- is helpful

## Proof of testing

No power in the wire
image
Disabled
image
Firing
image

## Changelog

:cl: Swan
fix: fixes emitters not changing their appearance depending on the state they're in
/:cl:


Phantastic-Swan:
I updated it so that it uses overlays instead, thanks TealSeer

TealSeer (approved):

PR #5409, opened 2026-04-06
death degradation quirk fixes

ghost:
## About The Pull Request
Makes it such death degradation stasis and formaldehyde multipliers do anything now because it wasn't working before.

Also removes the information about getting degradation reduction speedup when you have no ways of reducing it.

And also fixes the quirk to correctly load preferences if you have special characters like underscore in your ckey on round start because they don't load if you ready up and game starts. they only load if you late join in that case. (I tested this, this happens because at roundstart, ckey is different than source.ckey for some spaghetti code reasons and it works fine after you update them manually which is pure annoyance to remember this every time)
## Why It's Good For The Game
Fixes good
## Proof Of Testing
<details>
resting multiplier texts doesnt show when you have no way of reducing it in first place
multgone

Stasis and formeldehyde now works, previously it didn't load prefs from quick configuration and were hard coded to 0x multiplier(fully stopped degradation)
multipliers

</details>

## Changelog
:cl:
fix: death degradation stasis and formaldehyde multipliers now actually work
add: death degradation resting multiplier info doesn't show when you have no way of reducing it in first place
fix: death degradation preferences load correctly if you readied up on round start and have special characters in your ckey
/:cl:


explosivekitty:
I also added a fix that makes prefs load automatically when you ready up round start if you have special characters in ckey because that's a thing now apparently, explained in more detail in about pr

PR #5433, opened 2026-04-14
Adds tongues with special speech bubbles to character editor

ghost:
## About The Pull Request
Adds tongues with special speech bubbles to loadout just like the title says, those tongues are (relatively) identical to the regular human tongue so you don't get any special effects like knowing another language that you don't have enabled in language tab or have different taste for food than your species, i only copied basic stuff from their regular speech bubble counterparts so they shouldn't differ that much.

All of them can be found in character editor -> markings/organs -> tounge tab, and screenshots how they look are below
## Why It's Good For The Game
I don't like how our only character editor variety is aquatic tongue speech bubble so i added more. Most of those bubbles are very rarely seen in game so they will see more sunlight
## Proof Of Testing
<details>
list

New tongue list
(Excuse the screenshots with not matching species to the tongue but i'm not making special character for each screenshot)

xenospeechbubble

Xeno speech bubble

roboticspeechbubble

Robotic speech bubble(Borg speech bubble is blue and this one is gray)

holographicspeechbubble

Holographic speech bubble(Only used by guardian from lavaland mysterious shard, holoparasite uses different one)

slimespeechbubble

Slime speech bubble(Blob and blobbernauts use different one, more angry and red)
</details>

## Changelog
:cl:
add: Added new tongues with special speech bubbles to character editor
/:cl:


xPokee (changes requested):
Why not just add these to the default organs modularly instead of making new subtypes?

explosivekitty:
> Why not just add these to the default organs modularly instead of making new subtypes?

maybe some people want to use default speech bubble so they can optionally take it from organs tab if they want and the default organs of those types force learn you new language so i'd have to remove additional stuff they give or else they would be species specific. it's just easier to make a new subtype than having to removing all of the additional stuff from inheritance

PR #5451, opened 2026-04-17
Inari Tail

lukevale:
This adds in two tails, the Inari 10 tailed kitsune. An extremely large set of tails. Probably best to not use these in departments like sec or medical, but if you are so inclined to do so, simply by wagging it'll swap to the single tailed variant. Which means that if you want to be generous, you can run excitable, so if someone wants your tails to get out of the way they can simply headpat you.

The alternative is added in as the oversized fluff tail, a general use canine tail of large size. 
This is simply more options of customization. While it can be admittedly a bit large, it's mostly for flavor and people who it won't really be an issue with. If anything it'll be more problematic to THEM. And the single tail has been something I wanted for a while, something a bit more then the fox tails, but not like the snake tail.
Ran on a local to make sure it compiles and works fine.
<details>
image
image
image
image
image
image

</details>

## Changelog
:cl:
add: Inari Kitsune tail
add: Oversized Fluff tail
/:cl:


projectkepler-ru:
fluffy

xPokee (changes requested):
I said it on the Discord as well, but this sprite just looks like a big blob of color. It needs shading and definition between each tail.

lukevale:
Got sick so it'll take a hot minute, have everything prepped.

lukevale:
As said on the discord, the shading has been completed and committed

xPokee (approved):

lukevale:
No tail? 😔

PR #5459, opened 2026-04-18
Trustworthy heads

ArrisFairburne:
## About The Pull Request

The non-mindshielded heads of staff (HOP, RD, CMO, CE, QM) now roundstart with mindshields.

Moving alert level up to amber now causes all head of staff headset keys to temporarily gain security access. 

## Why It's Good For The Game

On Bubberstation we have "heads of staff = non antagonist" as a design paradigm 99% of the time. This is in opposition to the code we get from tg where heads of staff can potentially totroll (at least, to my knowledge). Thus, mindshields are being used to indicate this.

Heads of staff should generally have station matters in mind and so when antagonists start getting feisty (amber alert is called) heads of staff should be considering how they can contribute to keeping the station safe and secure. Additionally, security should be getting equipment from other depts. when things start getting hairy; this facilitates that instead of security only ever relying on themselves.

## Proof Of Testing
<details>

(all of these are fresh spawns)
image

capture

</details>

## Changelog
:cl: Arrisfairburne
balance: All bridge crew members are mindshielded.
add: Heads of staff encryption keys now get security comms access at amber alert and above. 
/:cl:


Arturlang:
this makes the objective of converting a head of staff for bloodsucker much harder, makes hypnotizing them not possible without more effort, just noting that

ArrisFairburne:
> this makes the objective of converting a head of staff for bloodsucker much harder, makes hypnotizing them not possible without more effort, just noting that

Mmmmm. Maybe if it's a big problem then the bloodsucker ability can be changed to doafter, then remove the mindshield before it hypnotizes?

NanoCats:
I think this is a good change

ibexgoetia:
Also, I've made a very tiny fix to the changelog. Simply turned 
> add: All bridge crew members are mindshielded.

into

> balance: All bridge crew members are mindshielded.

PR #5498, opened 2026-04-27
More bags and belts for loadout belts and bigger fannypacks

ghost:
## About The Pull Request
Adds more belts, bags and whatnot to loadout and increases fannypack slots back to default so they are actually useable
Also i had to unclogs CIN chest rig code so it doesn't appear as error in loadout
## Why It's Good For The Game
You might think that this fannypack buff is actually insane BUT we always had empty frontier medical kit in loadout that goes on your belt slot and has 7 slots for small sized items and noone complained about it so why bother picking other options? now they are actually comparable to everything else we have, chest rig can be bought in cargo for 50 credits round start so that's not game breaking either, you just save yourself 5 minutes by picking it in loadout instead running to cargo every shift and imminently buying it after you spawn and rest of the belts and bags are easily accessible anyway for the jobs that would take them, they either spawn with those or plenty of them are just laying in their department, you can just rename and change description for them in loadout because normally belts can't have their name edited (i also plan making everything recolorable in loadout so you will be able to customize the color without spraycan). and no- security webbing and explorer webbing can hold more than just small sized items so they are not here
## Proof Of Testing
<details>
screenshot
</details>

## Changelog
:cl:
add: Added more options to loadout belts
balance: Increased how much stuff fannypacks can hold
code: Moved empty frontier medical kit to belt tab in loadout
code: unclogged surplus chest rig icon_state to not error in loadout
/:cl:

### DIFF
--- a/code/datums/storage/subtypes/belts.dm
+++ b/code/datums/storage/subtypes/belts.dm
@@ -287,7 +287,7 @@
 
 ///Fanny pack
 /datum/storage/fanny_pack
-	max_slots = 3
+	//max_slots = 3
 	max_specific_storage = WEIGHT_CLASS_SMALL
 	silent = TRUE
 
--- a/modular_skyrat/master_files/code/modules/loadout/categories/pocket.dm
+++ b/modular_skyrat/master_files/code/modules/loadout/categories/pocket.dm
@@ -142,10 +142,6 @@
 	name = "Random Pizza Box"
 	item_path = /obj/item/pizzabox/random
 
-/datum/loadout_item/pocket_items/deforest_frontiermedkit
-	name = "Empty Frontier Medical Kit"
-	item_path = /obj/item/storage/medkit/frontier
-
 /datum/loadout_item/pocket_items/ingredients
 	name = "Wildcard Ingredient Box"
 	item_path = /obj/item/storage/box/ingredients/wildcard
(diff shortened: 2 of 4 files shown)


YakumoChen:
Once again we're proving skyrat-coded loadout options were a mmistake and only leading to a slippery slope of free universal belts in loadout (obstinately a gameplay benefit)

Fannypacks are supposed to not be good because they're purchased from clothes vendors

Also why is it always a bad thing that people have to order from cargo? Isn't there a whole job based around that?

explosivekitty:
> Once again we're proving skyrat-coded loadout options were a mmistake and only leading to a slippery slope of free universal belts in loadout (obstinately a gameplay benefit)
> 
> Fannypacks are supposed to not be good because they're purchased from clothes vendors
> 
> Also why is it always a bad thing that people have to order from cargo? Isn't there a whole job based around that?

Then maybe we should remove all of the skyrat-coded loadout options to not provide any gameplay benefit by mistake in that case... or just remove loadout because we have clothing vendors and cargo on station anyways... people already knew about using frontier kit as a loadout belt storage and yet i didn't see anyone complaining about it or everyone picking it as a top 1 extreme-ultra-whatever-meta pick that you had to take or else, and now atleast you will be able to edit name, description and possibly recolor the belt if you wanted to. we are not on tg to have competitive rules for loadout so if people won't want to use bigger fannypacks they just don't have to

MrGloopy:
permission to reopen this with some modifications?

MrGloopy:
ok nvm im getting shot by maints

PR #5486, opened 2026-04-22
Adds a quirk thats let you not have character mood

ghost:
## About The Pull Request
Adds a free quirk that locks you out of other mood quirks, hides judging mood face from your screen and moves hunger meter to where mood face was. All of the negative and positive bonuses are disabled with this quirk and mood slow down will be replaced with slow down if you are hungry
## Why It's Good For The Game
I'm huge hater of this feature since the day i started playing this game and i absolutely despise seeing the mood face on my screen and with this quirk you will no longer have to worry about some arbitrary mood meter saying how your character feels like, you can roleplay how you are feeling like and do whatever you please. and no this is not turbo broken, the positive mood you can have has way more mechanic value than having mood always set to neutral so you are locking yourself out of the positives and negatives
## Proof Of Testing
<details>
<summary>Screenshots/Videos</summary>
quirk equipped

hungry

quirk selection

</details>

## Changelog
:cl:
add: Added new quirk that lets you disable and hide your mood from your hud
/:cl:

PR #5501, opened 2026-04-28
Revert "Adds a quirk thats let you not have character mood"

SapphoQueer:
Reverts Bubberstation/Bubberstation#5486

As it turns out, there are far more mood interactions than initially thought and it is in fact breaking things. 

As well as tg adding more and more meaningful interactions for mood it's not a good idea to keep this right now.


KateAnbolho:
Amazing

PR #5507, opened 2026-04-30
Makes .310 AP Unprintable

xPokee:
## About The Pull Request
This ammo does 25 damage per shot to a redsuit. The other maintainers agreed that this probably shouldn't be printable from the reloading bench.
## Why It's Good For The Game
This ammo probably shouldn't be able to be easily printed, and saved as an OPFOR only item instead.
## Proof Of Testing
I didn't have time to test it but if CI passes it works. One line change.
## Changelog
:cl:
balance: .310 AP ammo is no longer printable from the ammo bench.
/:cl:


KateAnbolho:
hee ho.

PR #5522, opened 2026-05-03
Makes secmed grant medical and security hours, instead of just medical

KateAnbolho:
## About The Pull Request

As a secmed, you're supposed to do security work. It makes no sense that you don't get hour requirements done by playing it.

Upon further testing it appears it is not possible to provide two sets of hours. It now properly provides sec hours, however.  

## Why It's Good For The Game

secmeds dont get sec credit for playing it

## Proof Of Testing

should work

## Changelog
:cl:
add: Secmed now actually grants sec hours
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/security/secmed/security_medic.dm
+++ b/modular_zubbers/code/modules/security/secmed/security_medic.dm
@@ -10,7 +10,7 @@
 	exp_requirements = 120
 	exp_required_type = EXP_TYPE_CREW
 	exp_required_type_department = EXP_TYPE_MEDICAL
-	exp_granted_type = EXP_TYPE_SECURITY
+	exp_granted_type = list(EXP_TYPE_SECURITY, EXP_TYPE_MEDICAL)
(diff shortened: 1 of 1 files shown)


BingusTheClown:
Question: Based on your `about` this only fixes secmeds getting sec hours? Is that still the case, or does it now also do what the title says?

If its now fixed to secmeds only getting sec hours instead of medical, i'd ask that the title be updated to reflect that.

CatoChristopherMrow:
This change was made intentionally ages ago since you'd have people who would play minimum amount of secmed and then jump to OTHER roles like HoS or bluey, more or less skipping it by just playing a med focused role

KateAnbolho:
> This change was made intentionally ages ago since you'd have people who would play minimum amount of secmed and then jump to OTHER roles like HoS or bluey, more or less skipping it by just playing a med focused role

Ya, but currently the way secmeds are expected to do their duties with security work primarily

PR #5523, opened 2026-05-04
[BOUNTY] Allows family heirloom users to mark & rename items instead of getting a random one

nikothedude:
## About The Pull Request

Title. Bounty link: https://discord.com/channels/1059199070016655462/1398375166244819014

Adds a new pref option that, if disabled, gives you a one-use action that marks a held item as your heirloom and also gives a chance to rename and redescribe it (the original name is retained in parenthesis).
## Why It's Good For The Game

Gives wider variety for family heirloom users. I can finally break out the set of family jordans...
## Proof Of Testing
<details>

image

https://github.com/user-attachments/assets/059c1a97-926e-4922-82b0-4fdd8aade9fd

</details>

## Changelog
:cl:
add: Family heirloom users can now use a quirk preference to, instead of getting a random item, mark an item inhand as an heirloom.
/:cl:

### DIFF
--- a/code/datums/quirks/negative_quirks/family_heirloom.dm
+++ b/code/datums/quirks/negative_quirks/family_heirloom.dm
@@ -13,6 +13,12 @@
 /datum/quirk/item_quirk/family_heirloom/add_unique(client/client_source)
 	var/mob/living/carbon/human/human_holder = quirk_holder
 	var/obj/item/heirloom_type
+	// BUBBER EDIT ADDITION BEGIN - lets users manually designate heirlooms
+	if (!client_source?.prefs?.read_preference(/datum/preference/toggle/random_heirloom))
+		mark_action = new /datum/action/mark_family_heirloom(src)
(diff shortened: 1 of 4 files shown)


SapphoQueer:
I think this is interesting, is there anything in place to stop someone from making say, their PDA/Backpack the heirloom though? I know the ID itself has the check.

xPokee:
I think /tg/ would probably accept this, although they may want it reworked a bit.

nikothedude:
> I think this is interesting, is there anything in place to stop someone from making say, their PDA/Backpack the heirloom though? I know the ID itself has the check.

I could add those to the blacklist if youre worried.

> I think /tg/ would probably accept this, although they may want it reworked a bit.

Id really rather not deal with TG maintainers (I sometimes get in a infinite review cycle) but if youd really rather me take it upstream I coooould.

NargaInDisguise:
This is something I've wanted asked about for so long, holy shit

PR #5531, opened 2026-05-06
Tele-Port: Ports Telepathy changes from NovaSector

Jinshee:
## About The Pull Request

This ports changes from https://github.com/NovaSector/NovaSector/pull/6016 and https://github.com/NovaSector/NovaSector/pull/6398. 

- The quirk no longer automatically applies the Genetics power to the quirk user. Prior to this PR, communication was mostly anonymized and only _not_ so when one possessed the gene. This presented a clear downgrade for individuals who had the genetic variant over the quirk variant. With this PR, communication through the quirk is not anonymous, but adding the gene on top of the quirk will allow a user to upgrade it into being anonymous.
- Every time someone uses the quirk everyone else (in sight) will see where their attention is sent. This does not add a visual cue besides the warning in the chat, so it's still somewhat discrete.
- Added a new quirk, Psionic Dampener. This blocks most forms of telepathic communication.
- Telepathy was made multiline for readability.

## Why It's Good For The Game

Currently, Telepathy as a quirk has a discrepancy where non-synths receive the mutation Telepathy, while synths receive a different version. The mechanical differences of these created a problem where, depending on who was using it, this either did or did not anonymize the identity of the caster. Additionally, these changes allow the quirk holder who seeks the genetic variant in-round to properly receive the benefits of such.

Other changes allow people to opt-out of receiving telepathic messages, in order to avoid negative interactions from those who abuse it.

Multiline telepathy is a desirable feature for the sake of proof reading.

## Proof Of Testing

image

## Changelog

:cl:
add: Added a new quirk Psionic Dampener, affecting telepathy from quirks, genetics, slimes, xenomorphs, revenants and so on.
balance: Telepathy (Quirk) no longer automatically gives the gene.
qol: The telepathic communication power granted by the quirk or the genetic power is only anonymous if the caster has the mutation, not the target.
qol: Telepathy (Quirk) now shows when the user uses the power and on whom, but not the contents of the message.
qol: Telepathy is now multiline.
/:cl:


BingusTheClown:
This PR is now testmerged!

BingusTheClown:
One bit of noted feedback is that your PR says telepathy costs 2 points, but it is listed in the game as 3.

I'd suggest double checking that everything else also aligns!

PR #5561, opened 2026-05-11
Affection sect for chaplain

NiklasNeighbor:
## About The Pull Request

Adds a new sect for the chaplain to use. The Affection Sect.
Should probably be testmerged for balance first.

**Sacrifices:**
- You gain favor by being interacted with positively through hugs, headpats, boops, etc. or through any interaction done to you with the Interaction Panel while adjacent. The amount of favor you gain starts at 1 per interaction and is doubled while in the chapel or quadroupled if it's done by the last person you healed and they haven't given you affection since. This has a per-mob cooldown of 1 Minute to prevent just spam-hugging endlessly and encourage variety in who you get affection from.

**Rites:**
- Upgrade Favor Gain, increases the amount of favor you gain from affection by one per interaction. The price is multiplied by 1.5 with each upgrade.
- Summon call necklace creates a necklace that when worn allows the wearer to destroy the necklace to call the chaplain to their location via teleportation. The chaplain is returned to where they were before if they either use the return ability provided or walk more than 3 tiles away from where they were summoned.

**Book Effect**
- Hitting someone with a bible lets you reliably heal others by consuming 1 favor for each HP healed. This goes up to 10 HP per smack, though without the risk of giving someone brain damage.

## Why It's Good For The Game

More chaplain content good.

## Proof Of Testing
It runs and I don't see any major bugs.
- Tested favor gain through help intent and interaction panel
- Tested chapel favor multiplier
- Tested summon necklace teleportation
- Tested automatic return for necklace
- Tested healing via bible attacks
TODO: balance testing

<details>
image
image
image

</details>

## Changelog

:cl:
add: New Chaplain Sect: Affection God. You get paid in headpats.
/:cl:


modular_zubbers/modules/religion/affection/teleport_necklace.dm
```diff
@@ -0,0 +1,160 @@
...
+	worn_icon = 'icons/mob/clothing/neck.dmi'
+	worn_icon_state = "beads"
+
+	var/datum/weakref/chaplain_ref
+
+/obj/item/clothing/neck/affection_necklace/equipped(mob/user, slot)
+	. = ..()
+
+	if(slot == ITEM_SLOT_NECK)
+		user.AddComponent(/datum/component/necklace_summon_action, src)
```
nikothedude: theres no need to add a component here to add an action. see /datum/action/item_action and actions_types (item var)

or hell just add the action on equipped and delete it on dropped if you need custom behavior

(im not an authority on action code i just know this is definitely not the way)

modular_zubbers/modules/religion/affection/teleport_necklace.dm
```diff
@@ -0,0 +1,160 @@
+/obj/item/clothing/neck/affection_necklace
```
nikothedude, line 1: i would personally add an examine blurb that says, if used, it summons the chaplain, mainly so we dont get scenarios where prisoners insist on keeping their prized necklace and get broken out by the chaplain in perma

nikothedude (changes requested):

modular_zubbers/modules/religion/affection/teleport_necklace.dm
NiklasNeighbor: I can't quite remember why, but I know there was a reason I didn't use item_action... I'd have to look at the code again

nikothedude: id like if you did, because this is not how we use components

NiklasNeighbor: Yup, could remove the component. It now just adds the ability directly.

nikothedude:
yeah thats a lot better and more sensical, i say lets give this a shot, worst case scenario you have to do more changes

MILDLY concerned over the balance of the teleport amulet because im always wary of teleports but having a range of 3 should be okayyyy??? its not like you can engage in much combat with that at any rate

also some concern about the favor gain being kinda easy but like.......... its not too hard to get favor from other paths

NiklasNeighbor:
> ... also some concern about the favor gain being kinda easy but like.......... its not too hard to get favor from other paths

Yeah, other maintainers also said they didn't like the current version because of the potential hug spam, but I fixed that today by making it so that a mob can only grant favor once a minute

PR #3899, opened 2025-05-26
CUM 4 - The Backending

sippykot:
## About The Pull Request

As you know, every ss13 spessman is a system of beakers connected with tubes; usually 2. This PR adds to that;
The genital organs used to operate on snowflake shitcode that was basically trying to replicate the reagent system while not really doing it, and not even being used. This PR aims to make the genitals (which for some reason include breasts, don't ask) actually use normal reagent code, and lays the groundwork for further PRs implementing species-specific and perhaps quirk-selected functionality.

From the player side; you won't notice that much changing other than how much you cum into beakers and some arousal tweaks (mostly removing the really dumb stuff) - the only major feature is that you can cum into someone's mouth.

## Why It's Good For The Game

shitcode = bad, more robust functionality = good
the whole pr mostly exists to enable other stuff down the line

## Proof Of Testing

I did test it on localhost, seems to be working fine

## Changelog

:cl:
add: cumming into someone's mouth actually makes them ingest cum
code: orgasm and arousal tweaks
code: completely changed how organs/genital work under the hood, now actually uses reagent system
/:cl:

PR #5565, opened 2026-05-12
Return fluid levels to pre-mid-2025 and fix weird output logic. Allow more logical downstream modification

Muffindrake:
## About The Pull Request
This fixes the weird fluid output calculation which introduces two multiplicative factors _mod and _mult that did the same thing which makes no sense from a VV perspective and which would be the primary reason to have those variables.

```x * _mod * _mult```
this makes barely any sense and neither of those variables are even used anywhere else in the codebase
It's changed to to
```(x + _mod) * _mult + _flat```
which makes far more sense conceptually. It'll allow future ingame chemical/item modification to output in a more logical fashion.

The resulting volume is also dependent on reagents.maximum_volume, so keep that in mind if anyone wanted to VV-edit them. It's messy. That is probably subject to a future PR which overhauls the system yet again.

It also restores the liquid levels back to what they were in mid-2025, which was apparently an undiscussed change in #3899 when it was introduced and it cut liquids in half (which is lovely if you're taking a break from the game at the time), and not mentioned as a balance change in the changelog, even. If there are people regularly flooding the station with no good faith effort to contain their mess (because they're throwing beakers), that sounds like a moderation issue.

The changes don't touch fluids for breasts, since they already produce quite a bit at large sizes, and they operate somewhat differently. Giving them the testicle treatment (moving the calculation to a different proc) would be a good idea if someone cares enough.

The initialization size caps are also removed - it makes little sense to have yet another check in here, since the character setup already clamps sizes to legal (reasonable-ish) values.

I'm not entirely happy with the code pre- and post-change because the output depends on too many variables, and makes it slightly difficult to modify them in downstream code. Using the reagent system is a good idea, but the existing code still depends on internal_fluid_maximum, which is not part of that system.

## Why It's Good For The Game
The changes allow for easier future modification through items/chemicals and clean up existing code. Don't kneecap the fun of people who enjoy liquid play.

## Proof Of Testing

I don't implement changes without testing them.

## Changelog
:cl:
balance: Return liquid levels to pre-mid-2025
refactor: Pass on testicle liquid handling
code: Remove redundant genital size caps
/:cl:


sippykot:
i think i mostly did it it to be more sane conceptually? i forgot
it was a modification of really old code that gets very insane especially surrounding like breasts lol

Muffindrake:
> i think i mostly did it it to be more sane conceptually? i forgot it was a modification of really old code that gets very insane especially surrounding like breasts lol

Yeah there are a lot of headscratchers in old code.

PR #5579, opened 2026-05-14
Disables lone op from being spawned via the disk being unsecured

nikothedude:
## About The Pull Request

Title.

No, this is not a salt PR, I just realized this might actually get accepted and am willing to try my luck.
## Why It's Good For The Game

I have never really thought lone op via disk encourages anything interesting. For _us_, the only effect it has is to occasionally bite us in the ass in the case of incompetent command, ending the round with no real input from Anyone Else. It just causes bitterness.

Yes, its soulful, yes its "cool", but I just dont think soul and cool factor really justify it in this case.
## Proof Of Testing
<details>
<summary>Screenshots/Videos</summary>

</details>

## Changelog
:cl:
del: The nuke disc can now be unsecured and wont spawn a lone operative.
/:cl:

PR #5583, opened 2026-05-14
Replaces lone operative from disk with a name&shame + paycut + spawns inspector

nikothedude:
## About The Pull Request

Alternative to https://github.com/Bubberstation/Bubberstation/pull/5579 **WHICH I STILL BELIEVE IS THE BETTER PR.**

This PR exists in the case that PR won't be merged.

Putting the disk in a bed prevents the paydock. Sleep tight disky.
## Why It's Good For The Game

Same justification as the other PR. This one however should make more people happy while maintaining a in-world reason to keep that disk secured.
## Proof Of Testing
<details>

image

image

</details>

## Changelog
:cl:
add: The nuke disc no longer spawns a lone operative if unsecured, instead naming and shaming command and docking their pay.
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/antagonists/nukeop/equipment/nuclear_authentication_disk.dm
+++ b/modular_zubbers/code/modules/antagonists/nukeop/equipment/nuclear_authentication_disk.dm
@@ -1,10 +1,59 @@
-/obj/item/disk/nuclear/unsecured_process(last_move)
+#define UNSECURED_DISK_COMMAND_PAY_PENALTY 0.5
+#define GRACE_PERIOD 35 MINUTES
 
+GLOBAL_VAR_INIT(did_sleepy_disk_announcement, FALSE)
+
+/obj/item/disk/nuclear/unsecured_process(last_move)
 	//If there is no assigned captain, then don't run the event.
 	if(!SSjob || !SSjob.assigned_captain)
 		return
 
-	. = ..()
+	/// How comfy is our disk?
+	var/disk_comfort_level = 0
+
+	//Go through and check for items that make disk comfy
+	for(var/obj/comfort_item in loc)
+		if(istype(comfort_item, /obj/item/bedsheet) || istype(comfort_item, /obj/structure/bed))
+			disk_comfort_level++
+	var/comfy = disk_comfort_level >= 2
+
+	if(last_move < world.time - GRACE_PERIOD)
+		if (comfy)
+			if (GLOB.did_sleepy_disk_announcement)
+				return
+			priority_announce(
+				"Attention [GLOB.station_name]. It has come to our attention that your nuclear authentication disk has been stationary for... wait - did you \
+				tuck it into bed and... awww. Awwwww. You know what? We'll let it slide just this time. Sleep tight, disky.",
+				"Central Command Asset Protection",
+				'modular_skyrat/modules/alerts/sound/alerts/alert1.ogg',
+				has_important_message = TRUE
+			)
+			GLOB.did_sleepy_disk_announcement = TRUE
+		else
+			reprimand_command()
+
+/obj/item/disk/nuclear/proc/reprimand_command()
+	priority_announce(
+		"Attention [GLOB.station_name]. It has come to our attention that your nuclear authentication disk has been stationary for an egregious period of time. \
+		There is no excuse. Your command staff will have their pay docked and employment reviewed. Failure to secure the disk in good time will result \
+		in further consequence.",
+		"Central Command Asset Protection",
+		'modular_skyrat/modules/alerts/sound/alerts/alert1.ogg',
+		has_important_message = TRUE
+	)
+
+	for (var/datum/bank_account/iter_account as anything in assoc_to_values(SSeconomy.bank_accounts_by_id))
+		var/datum/job/our_job = iter_account.account_job
+		if (isnull(our_job))
+			continue
+
+		if (!(our_job::title in SSjob.chain_of_command))
+			continue
+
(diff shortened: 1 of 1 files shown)


SapphoQueer:
commoand

anyways this seems preferred amongst maints so approvaling this

modular_zubbers/code/modules/antagonists/nukeop/equipment/nuclear_authentication_disk.dm
```diff
@@ -1,10 +1,59 @@
...
+			)
+			GLOB.did_sleepy_disk_announcement = TRUE
+		else
+			reprimand_command()
+
+/obj/item/disk/nuclear/proc/reprimand_command()
+	priority_announce(
+		"Attention [GLOB.station_name]. It has come to our attention that your nuclear authentication disk has been stationary for an egregious period of time. \
+		There is no excuse. Your command staff will have their pay docked and employment reviewed. Failure to secure the disk in good time will result \
+		in further consequence.",
```
lessthnthree:
```suggestion
		in further consequences.",
```

sippykot:
putting the disc in a secure briefcase should also count
image

lessthnthree (approved):

modular_zubbers/code/modules/antagonists/nukeop/equipment/nuclear_authentication_disk.dm
```diff
@@ -1,10 +1,59 @@
...
+
+	//Go through and check for items that make disk comfy
+	for(var/obj/comfort_item in loc)
+		if(istype(comfort_item, /obj/item/bedsheet) || istype(comfort_item, /obj/structure/bed))
+			disk_comfort_level++
+	var/comfy = disk_comfort_level >= 2
+
+	if(last_move < world.time - GRACE_PERIOD)
+		if (comfy)
+			if (GLOB.did_sleepy_disk_announcement)
```
lessthnthree, line 21: if (GLOB.did_sleepy_disk_announcement) should be at the start of the proc.
No point checking if it's in bed if we've already done the announcement.

lessthnthree (changes requested):

modular_zubbers/code/modules/antagonists/nukeop/equipment/nuclear_authentication_disk.dm
nikothedude: not entirely correct bc if we early return via did_sleepy_disk_announcement at the top we miss the entire lower block handling a non-cozy disk

lessthnthree: if we've already announced and shamed command, why do we care at all? we don't need to handle anything.

should be up there right after checking if there's an AC.

nikothedude: because in the edge case where someone puts a disk to sleep, then takes it out, and leaves it unsecured, they should be reprimanded

lessthnthree: okay

CB35SSC:
If nobody secures the disk they deserve armageddon.

BubbleTrroubles:
Skyratifcation one PR at a time

Zeerum:
I have to wonder why we put so many safeguards in place to prevent loneop spawns on lowpop, only to later remove it anyway. 

Not securing the disk is failrp.  This PR comes after five command members all collectively forgot to secure the disk, btw.

Mewchild:
The pay cut can be reset via the accounts console in the HoP office. There's literally no downside to not doing the bare basics of your job, with this. In addition, very few command actually use their personal budget, so financial punishment is barely a slap on the wrist.

XiskaVZ:
I heavily dislike this PR in all its elements, It does nothing to promote RP, it does nothing to promote command, infact it makes command have literally 0 responsibilities. The entire name of the game is to PROTECT DAT DISK. also financial incentive doesnt even make sense in canon given how little NT employees are paid to begin with.

theselfish:
I was checking here to grab some stuff from my previous PRs and I'm delighted to see that this server will never stop having a ying yang about Lone Ops.

Just grab the disk. Jesus fucking Christ.

nikothedude:
Im not the biggest fan of compromising, but seems people really want more consequence. It spawns a inspector now with an objective to evaluate command and force the disk to be secured, with power to demote command if they deem them unfit. Could make it worse in the future

XiskaVZ:
> Im not the biggest fan of compromising, but seems people really want more consequence. It spawns a inspector now with an objective to evaluate command and force the disk to be secured, with power to demote command if they deem them unfit. Could make it worse in the future

I dont really like the idea of adding an inspector, it really devalues the war that the Syndicate is waging against nanotrasen, because by removing Lone op you've removed all in lore reason to keep the disk secured, there's nothing to inspect beside command incompetency, and sure the first few times it'll be an ok rp prompt, but it'll just boil down to the same 3 scenarios. And I highly doubt people would wanna end up rping a gold class ert just to yell at a captain who won't listen, so it doesn't feel like anything will happen with this idea. I saw someone propose that if the disk isn't secured after the announcement, a nuclear operative should spawn. Which is a much better idea than the inspector.

BubbleTrroubles:
I dont like the inspector either, but I'd rather have incompetent command demoted for not securing the disk because that's like your one and main objective as captain or acting captain. If you can't do a single simple job of keeping the disk in your survival box when you take up the mantle then you don't need to be playing a command position.

nikothedude:
un-tmed as maintainers debate about what we're gonna do about lone op

PR #5596, opened 2026-05-16
Makes Well-Trained Characters Stay Laying Down After *snap3

MintyPhresh:
## About The Pull Request
Switches the Well-Trained response to a dom doing *snap3 from a 0.1 second knockdown to just toggling laying, in addition to its other effects.

## Why It's Good For The Game

One of the minor things that irks me about this effect is that, as it is currently, the sub character will just get back up as soon as the knockdown's finished. I don't want to have to test my reaction speed just to stay laying down when my character is told to.

## Proof Of Testing

<details>
image

</details>


BingusTheClown:
Heyo, i'm gonna tag this one as pending maintainer discussion, because we already have ongoing talks about the quirk! Let me just bring it up to maints real quick.

PR #5624, opened 2026-05-19
digi legs labcoats subtype fix

Danielance:
## About The Pull Request

A [recent PR of mine](https://github.com/Bubberstation/Bubberstation/pull/5603) unintentionally resulted in a lot of labcoat related clothing becoming missing textures that looked fine previously. This is because I assigned the base type of labcoats to use the "CLOTHING_DIGITIGRADE_VARIATION" flag, which looks for digi sprites, otherwise becomes a missing texture. 

I did not catch this in my PR because I was unaware that there were more labcoat subtypes than the ones I edited. This PR fixes those labcoats by moving the flag to just the clothing items I edited, leaving the affected clothing to go back to using "CLOTHING_DIGITIGRADE_NO_NEW_ICON", which as the name implies, uses the base non-digi icon and doesn't result in a missing texture.

Note that this PR means that the base labcoat will not look right on digi, but I will fix this again at a later point.

## Why It's Good For The Game

Fixing the bugs I created!
## Proof Of Testing

Works on my machine.

This is an affected clothing item (researcher's lab coat), reverted to a working state.
image

The other applied fixes from the last PR (excluding the base labcoat) still work
image

## Changelog
:cl:
fix: fixed some labcoat subtypes becoming missing textures on digitigrade legs
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/clothing/suits/labcoat.dm
+++ b/modular_zubbers/code/modules/clothing/suits/labcoat.dm
@@ -1,6 +1,3 @@
-/obj/item/clothing/suit/toggle/labcoat
-	supports_variations_flags = CLOTHING_DIGITIGRADE_VARIATION
-
 /obj/item/clothing/suit/toggle/labcoat/cmo
 	worn_icon_teshari = 'modular_zubbers/icons/mob/clothing/suits/labcoat_teshari.dmi'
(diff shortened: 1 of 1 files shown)


nikothedude:
lgtm

god damn non-abstract basetypes

lessthnthree:
I was wondering what broke.

PR #5634, opened 2026-05-21
Muridae Traits Update: Phobia

Jinshee:
## About The Pull Request

This adds a new phobia that makes the holder afraid of mousetraps, using the built-in phobia system. Mobs with the Muridae Traits quirk are given this phobia on add().

## Why It's Good For The Game

It's really funny

## Proof Of Testing
image
image

## Changelog
:cl:
add: Added phobia: Mousetraps
qol: Mousetraps phobia applied to Muridae Traits
/:cl:

### DIFF
--- a/code/_globalvars/phobias.dm
+++ b/code/_globalvars/phobias.dm
@@ -24,6 +24,7 @@ GLOBAL_LIST_INIT(phobia_types, list(
 	"spiders" = "Arachnophobia",
 	"strangers" = "Xenophobia",
 	"the supernatural" = "Phasmophobia",
+	"mouse traps" = "Cleithrophobia", // BUBBER ADDITION
 ))
 
 GLOBAL_LIST_INIT(phobia_regexes, list(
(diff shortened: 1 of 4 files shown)


nikothedude:
This is definitely turning this quirk into a negative one - I would either give it the same amount as the phobia quirk, put it IN the phobia quirk, or make this a config option on the quirk

Jinshee:
> This is definitely turning this quirk into a negative one - I would either give it the same amount as the phobia quirk, put it IN the phobia quirk, or make this a config option on the quirk

the quirk itself has both upsides and, with this, one downside. to make it a negative quirk wouldnt accurately represent those upsides (faction_rat and the sniff spell).

Jinshee:
As a part of the phobia system, it is already now a part of the phobia quirk itself. While I could remove the forced trauma on the muridae quirk to make it more of a choice, the interaction with muridae traits was the entire intention behind adding it in the first place

nikothedude:
oh i didnt know it gave you stuff. its probably okay then?

Jinshee:
> oh i didnt know it gave you stuff. its probably okay then?

Thank you. I'll take my first maint approval now 😊

MrGloopy:
phobia is a 0 point quirk anyway, isn't it?

Jinshee:
> phobia is a 0 point quirk anyway, isn't it?

yes

PR #5640, opened 2026-05-21
Another MC crash fix.

CatoChristopherMrow:
## About The Pull Request
Mod over slotting was causing a infinite loop leading to a MC crash in some cases. This fixes that.
## Why It's Good For The Game
I don't like crashes.
## Proof Of Testing
yarrrrrrg
<details>

</details>

## Changelog
:cl: UvvU
fix: fixed ANOTHER MC crash
/:cl:

### DIFF
--- a/code/modules/mod/mod_activation.dm
+++ b/code/modules/mod/mod_activation.dm
@@ -79,7 +79,7 @@
 		playsound(src, 'sound/machines/scanner/scanbuzz.ogg', 25, TRUE, SILENCED_SOUND_EXTRARANGE)
 	if(part_datum.can_overslot)
 		var/obj/item/overslot = wearer.get_item_by_slot(part.slot_flags)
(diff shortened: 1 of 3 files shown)


SapphoQueer:
Given this is only editing upstream content, why not pr this there?

lessthnthree:
> Given this is only editing upstream content, why not pr this there?

This

CatoChristopherMrow:
> Given this is only editing upstream content, why not pr this there?

It's not. The bubber edit is causing the crash. 
image

That is to say, that doing this was not done on TG for a reason, but I added the extra stuff to preserve the additional functionality ya'll wanted while still avoiding the crash.

CatoChristopherMrow:
Or...hm. One moment please, I need to >:( at this harder. By all means, you seem to be right, this actually SHOULD be crashing on TG too...

PR #5656, opened 2026-05-23
Makes regen extracts not heal physical damage entirely

nikothedude:
## About The Pull Request

Title.

Burn/brute/tox are now healed by 10 instead.
## Why It's Good For The Game

Im going to be honest. I hate this thing. It trivializes medbay, entirely. If you get a box of regen cerulean extracts, you basically dont have an excuse to do your job anymore, for any kind of injury. Simultaniously, there are scenarios where a regen extract should be used. So instead of gutting it Im doing this.

A more ideal solution would for there to be a massive downside to using it taht would discourage its use outside of pretty bad scenarios, like a bugged revival. But I cant think of it right now, so the most Im doing it making it not esaily usable in combat.
## Proof Of Testing
<details>

image

</details>

## Changelog
:cl:
balance: Regen extracts now only heal 10 of the major damage types.
/:cl:

### DIFF
--- a/code/modules/research/xenobiology/crossbreeding/regenerative.dm
+++ b/code/modules/research/xenobiology/crossbreeding/regenerative.dm
@@ -8,7 +8,7 @@ Regenerative extracts:
 	desc = "It's filled with a milky substance, and pulses like a heartbeat."
 	effect = "regenerative"
 	icon_state = "regenerative"
-	effect_desc = "Completely heals your injuries, with no extra effects."
+	effect_desc = "Completely heals your injuries (excluding physical damage), with no extra effects." // BUBBER EDIT CHANGE - cant be used in combat effectively
 
 /obj/item/slimecross/regenerative/proc/core_effect(mob/living/carbon/human/target, mob/user)
 	return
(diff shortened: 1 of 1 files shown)


MintyPhresh:
Based on the different effects that the colored regenerative extracts have, it does seem like they were designed primarily as combat heals rather than as medbay heals. I agree with reducing how much raw damage they heal, but 10 is too low. I'd suggest 30, personally

KateAnbolho:
I'm not gonna say that it's outright unneeded but it feels like this is crazy nerfed to the point of uselessness

nikothedude:
things the extract still heals:
everything besides damage and replacing your organs

wounds, diseases, mutations, radiation, organ damage, traumas

this is still very usable this s just one step to making it not a cureall

KateAnbolho:
Why wouldn't you just use a good mixed chem instead of like 8 of these cores

nikothedude:
because it instantly heals with no drawbacks and is mass producable? doesnt trigger allergies, doesnt interact with chems? 

edit: also it STILL HEALS MORE THAN *ANY CHEM CAN* like permatraumas and wounds and diseses??? i fail to see how you came to the conclusion this thing is useless

nikothedude:
ok maybe not mass producable per sae but you can get a lot of these

shayoki:
Alternatively, add a long do_after for this to weaken its usability mid-comat.

PR #5679, opened 2026-05-26
Allows Lycans to wear Neck Items

HometownFunky:
## About The Pull Request
It's a simple one line change that allows the Lycan species to wear items on the neck. This also means that neck items won't be dropped upon transformation.

A few things this PR also enables for clarity:
1. You can wear any piece of neck clothing, which means cloaks and collars (even the NSFW ones) are now kept on you if you transform into an Lycan.
2. ~~Allows Acolyte Lycans to be anti-magic collar'd, making it so the collar is still on post-transformation...~~ Cursekin transformation is inhibited by the collar.
3. ...and also allows Acolyte Lycans to wear a neck focus.
## Why It's Good For The Game
Customization, and NSFW shenanigans. 

When it comes to Acolytes it's also a balance question but I figured that the base Lycan isn't really that strong on its own, so them having access to a focus ~~while also patching the issue of using the TF button to just yeet your collar for free as a cursekin~~ isn't that much of a powercreep.
## Proof Of Testing
<details>
stays on

lycan
The item stays on after the transformation.
</details>

## Changelog
:cl:
balance: lycans can wear neck items
/:cl:


HometownFunky:
Been combing neck items for other potential balance concern stuff that might be affected by this change so far:
* I was worried about the memento mori + boosted lycan combo but going into hardcrit (which you can still do with the mori) knocks you out of the WW form, so I believe it's a non issue.
* The wolf TF cloak from icebox portal loot IS technically a neck item but you need the hood up to use it, and Lycans can't wear hoods.
* NRI police cloak HAS, for some reason, the ability to hold cuffs/ammo boxes? Idk why but that is the case.

nikothedude:
I should clarify antimagic collars inhibit lycan form now, but I guess this is a okay change? I dont think lycans should be able to wear clothes, but thats more a side effect of our decision to let you wear suit items on the neck slot.

HometownFunky:
I could try adding a check on the ctrl-shift-clicking to check if the user is a lycan, but that wouldn't stop a non-lycan from giving a lycan a non-functional neck item. I don't think I have enough code knowledge to prevent the latter. 

I do wanna say though, there are some items like the tribal outer clothing that imo look really good on lycans.

PR #5684, opened 2026-05-27
Fixes alt breast onmob sprites being GAGSed wrong

NanoCats:
# About The Pull Request
The greyscale ramp on the secondary (nipple/areola) color channel in modular_zubbers/icons/mob/sprite_accesory/genitals/breasts_onmob.dmi had its shadow floor set too dark. The darkest pixels in the secondary channel sat at greyscale value 103, meaning the darkest visible shading was rendered at only 40% of the player's chosen color. This made the outer edge of the areola nearly unrecognizable as the chosen color, turning it into a muddy greyish color no matter what. 

I personally believe that this went unresolved primarily because we primarily play host to furries with exotic nipple colors compared to humans, It's a lot more noticeable on humans than a dragon with teal nipples. 

The fix remaps the secondary channel's greyscale floor from 103 up to 170 (67% of chosen color), while leaving the highlight at 247 unchanged. The shading shape is preserved; only the floor is raised. No code changes were required. This affects all 480 secondary frames across all breast sizes and directions.

In less scientific terms, "someone screwed up when making these and I fixed it" 

# Why It's Good For The Game

My theory is that the original GAGS ramp was apparently designed and tested primarily against anthropomorphic characters, whose nipple color choices tend toward saturated non-human-skin hues like cyan or orange. For those players, a very dark shadow edge just reads as "deep shading" and goes unnoticed. For humanoid characters picking subtle rosy or brownish tones close to their skin color, the dark floor collapsed the chosen color into a kind of mess. The breast color picker has effectively been broken for human-presenting characters since these sprites were introduced. This fix makes the color selection actually work as intended for all players.

# Proof Of Testing

image

Before:

image

After:

image

# Changelog
:cl:
fix: Fixed breast (alt) secondary color (nipple/areola) appearing as an unrecognizable dark smear instead of the player's chosen color, most noticeable on humanoid characters using near-skin tones.
/:cl:


NanoCats:
I'm like 90% sure these sprites are good to go but if anyone wants to double check that'd also be cool.

PR #5691, opened 2026-05-28
Proteans can temporarily transform their wearers to look like them.

StrangeWeirdKitten:
## About The Pull Request

See title and proof of testing.

## Why It's Good For The Game

Listen, I've been racking my brain on how to do this since I first coded this species. I finally figured out a way. Besides, I think the component itself will prove useful for any future stuff. This adds more protean content that I feel people will enjoy.

## Proof Of Testing

dreamseeker_UezPpr0cxQ

Also I won't lie, hardlight stuff for the suits is kind of a task on itself on needing to piece together. That's one icon event that can't just be solved by sane conventional methods.
## Changelog

:cl:
add: Proteans can temporarily transform their wearers to look more like them.
/:cl:


tgui/packages/tgui/interfaces/ProteanUI.tsx
```diff
@@ -96,6 +108,14 @@ export const Protean = () => {
           <Table.Cell textAlign="center">
             <Button color="default" icon="heart" onClick={() => act('heal')} />
           </Table.Cell>
+          <Table.Cell textAlign="center">
+            <Button
+              color={transformation ? 'good' : 'default'}
+              icon="video-camera"
+              tooltip="Make your current wearer look like you. Ping @dangerkitten for feedback." // TEMPORARY DESCRIPTION ADDED. MARK THIS IN REVIEW BEFORE FULL MERGING
```
StrangeWeirdKitten: this

code/modules/mob/living/carbon/carbon_update_icons.dm
```diff
@@ -481,7 +481,10 @@
 	update_wound_overlays()
 	var/limb_count_update = 0
 	var/list/new_limbs = list()
-	for(var/body_zone, limb_untyped in get_bodyparts_by_zones())
+	var/datum/component/transformation/transform = GetComponent(/datum/component/transformation) // BUBBER EDIT START - Transformation component override.
```
nikothedude, line 484: getcomponent is gross but this probably isnt the worst use, idk if a signal would be appropiate

nikothedude: maybe this would be better as a status effect rather than a component?

StrangeWeirdKitten: For the first comment: The code needs to be able to interpret data from a datum so that I can properly register different checks. A signal is like something screaming from a mountain, and there needs to be an ear for it to listen to it. It would be way more inefficient is I gave everyone a signal to listen to something. While regenerate icon is somewhat expensive code, it's not really that hot for code. It just checks a list to see if the component is inside. (Besides, I need everything contained inside a decent datum)

For the second comment, a component is needed specifically to store, modify, and apply the information that's given to it. (And for initialization parameters) Especially to for the cache data and communication to the dummy mob spawned in null space. While this could be any datum, a component is a very easy way for anything to give this status to everything else and all you need to feed into it is a source.

StrangeWeirdKitten: While I could have the ui_act attach the signal to the mob, I need something actively intercepting every event of regeneration icon and switching it to something else. It's hacky, but it's definitely not an easy thing to do with how rigid rendering for carbons is. The CHEAPEST possible solution would be to cut someone's overlay and replace it with someone else. But there's not going to be any icon updating or them wearing their actual clothes. Think the cardboard prop you can carry. 

Aka, I think no matter what we do, it's going to be a little hacky. But I did this in the most modular way possible with a simple component, as all that chaos is just a component that can be applied.

StrangeWeirdKitten:
image

I fixed removing limbs not showing. This PR is pretty much ready for review.

LordSaladin:
@StrangeWeirdKitten I’ve noticed a bug in which that if you enable it when the examine panel is open, their character sprite does not refresh along with everything else.

StrangeWeirdKitten:
> @StrangeWeirdKitten I’ve noticed a bug in which that if you enable it when the examine panel is open, their character sprite does not refresh along with everything else.

There's really not that good of a fix for this because the whole transformation is illusionary. There's few actual changes happening to the player. Nor can I use the mob that's in nullspace because their clothes are only just enough to properly reflect a few icon events (like hair hiding)

My only thought would for it to become a static image based off the original.

Maia-J:
We are pushing this hacked together frankensteined mess that is our pref code to the absolute limits and it both makes me love it and awakes the generational trauma I built up when trying to fix anything about it. Hats off, shark.

PR #5692, opened 2026-05-28
Fix compile changelogs running once a week

TealSeer:
TG changed this because they set up TGS to auto trigger the compile changelogs workflow on build, we don't have that


StrangeWeirdKitten:
why the fuck is this PR failing screenshot checks, and I re-ran the checks.

PR #5718, opened 2026-05-30
Regal Condor nerf

SomeCabbage:
## About The Pull Request

Changes the Regal Condor to be less absolutely broken. Changes are as follows:
- Damage multiplier on the gun: 1.25 -> 1.00 (effective damage from 62.5 -> 50)
- Armour Penetration: 40 -> 30
- Projectile Speed: Hitscan -> 1.75 (a bit faster than average)
- Crafting recipe for the gun has been altered: Requires a parts kit purchasable from an uplink instead of 4tc directly where said kit costs 4tc and cannot be discounted, along with being restricted to higher player numbers.
- Crafting recipe for the magazines has been tweaked, now requiring a telecrystal per each additional magazine you want to make.
- To compensate for the exclusivity of the parts kit, a credit-purchasable one has been made available to the persistence.

## Why It's Good For The Game

The regal condor is a ridiculously broken gun and was fortunately not used to it's full potential up until recently, where it became very obvious that it's current state was unacceptable.

## Proof Of Testing

Recipe works as intended, and the condor's projectile and damage has successfully changed. The required item is available in the uplink and the percy express console. Actual validation for the uplink restrictions will require live testing, unfortunately.

## Changelog
:cl:
balance: Nerfed the Condor's damage performance, and tweaked it's recipe.
/:cl:


nikothedude:
a good first step, we can nerf it further if its still a problem

are you sure its not hitscan, though? i dont see a varchange that would indicate that its changed: hitscan = FALSE

SomeCabbage:
I've made and referenced a completely new projectile type that based off the base 10mm rather than the reaper

PR #5720, opened 2026-05-31
Removes the Regal Condor because I got my literal ass organ blown off my body once.

StrangeWeirdKitten:
## About The Pull Request

Removes the regal condor from gameplay. Alternative to #5718 

## Why It's Good For The Game

Alright, so I'll explain why questing items are incompatible and should just be removed. These aren't normal weapons. These are items that are designed to be purposely overpowered to the aforementioned requirements. It's a gun that's either overpowered, or it's not. (because it was nerfed) and becomes a sink for players to waste their time with. Yes, this gun was legitimately designed to mop the floor with everyone. It was given hitscan just for you to be scared shitless the second it comes on your screen. Now I can understand the appeal to bring the gun to a more "sane" level. But in this case, that implies it's a normally obtainable gun, which it's not. I feel like any nerfing done to it will effectively neuter the thing anyway. It becomes "less overpowered", and it becomes a trap for less aware traitors looking through the emagged cargo console or uplink when it remains in the game. The margins for a nerf would be really tight, and there's not going to be a lot of usage of the weapon for any reasonable oversight on how effective the nerf actually is or if it's over-nerfed. It did exist in the codebase for over two years until people started to learn about it.

I don't think #5718 is a bad PR, it's well thought out and seems fairly reasonable. It does add a pop limit to the gun, which is something I mentioned as a possibility. But I still think questing items that reward you with overpowered shit is some really dogshit gameplay design and incompatible with this server. (aka, they should all just be removed)

I said what I had to say regarding the condor, this PR exists as an option if the maintainers want to consider it. Otherwise, feel free to close at discretion. 

## Proof Of Testing

I hit spawn and it didn't appear

## Changelog

:cl:
del: removes the regal condor from being available in cargo.
/:cl:

### DIFF
--- a/_maps/bubber/automapper/templates/IceBoxStation/iceboxstation_persistence.dmm
+++ b/_maps/bubber/automapper/templates/IceBoxStation/iceboxstation_persistence.dmm
@@ -13806,10 +13806,6 @@
 	dir = 1
 	},
 /obj/structure/safe,
-/obj/item/book/granter/crafting_recipe/regal_condor{
-	pixel_x = -6;
-	pixel_y = 11
-	},
 /obj/item/bombcore/syndicate/large{
 	pixel_x = -2;
 	pixel_y = -1
--- a/_maps/bubber/automapper/templates/lavaland/lavaland_persistence.dmm
+++ b/_maps/bubber/automapper/templates/lavaland/lavaland_persistence.dmm
@@ -19332,10 +19332,6 @@
 	dir = 1
 	},
 /obj/structure/safe,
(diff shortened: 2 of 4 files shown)


CB35SSC:
Nerf, don't remove content

YakumoChen:
Almost like everything TG adds behind a time gate (questing items, the economy, etc) is incompatible with a server that has a typical round time 4x as long

nikothedude:
Not currently interested. Nerf is in. If it continues to be ass we'll consider this.

PR #5723, opened 2026-06-01
Adds blob_act() logic to protean modsuits.

StrangeWeirdKitten:
## About The Pull Request

If you're inside your suit, alive, and a blob spreads over you. You will die.

There's a /tg/ bug that will need to be fixed with the blob destroying indestructible items. I already opened the PR. It doesn't totally stop it yet, but that'll come in eventually.

## Why It's Good For The Game

RRing players is bad.

## Proof Of Testing

image

## Changelog
:cl:
fix: Protean suits are considered indestructible items.
fix: Blobs will properly murder proteans if it spreads over them while inside their suit.
/:cl:


KateAnbolho:
oh god did a protean get turned into debris?

PR #5735, opened 2026-06-03
Gives boosted lycans the ability to use their claws on objects + ups to 15 TC

nikothedude:
## About The Pull Request

Title.

Lycans gain an ability that gives them a claw item that deals slightly increased damage in combat, is slower, and deals extra damage to structures.

Note while this does partially counter chairs, its only partial, because this mode is less effective in overall combat and is mostly for breaking shit.

Pacifism suppresses this mode, to give some counterplay once you catch the lycan. (Pacificiation makes them containable)

## Why It's Good For The Game

Further leans into the garou powerfantasy. Every time I played lycan I realized it had little utility outside combat (and even then you have NOTHING for silicons). This gives you a massively damaging anti-object ability and should give ways to function and DO things while in lycan form that youd usually need tools for.

## Proof Of Testing

<details>

https://github.com/user-attachments/assets/11cf6f5a-1ec3-46d5-8616-46174f93d59a

Ignore the doorprying, that was removed

</details>

## Changelog

:cl:
balance: Boosted lycans can now use their claws on objects as long as theyre not pacified
balance: Lycanthropy booster is now 15 TC
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/customization/species/lycans/_lycan_species.dm
+++ b/modular_zubbers/code/modules/customization/species/lycans/_lycan_species.dm
@@ -1,3 +1,5 @@
+#define DOAFTER_SOURCE_LYCAN_DOOR_PRY "lycan door pry"
+
 /datum/species/lycan
 	id = SPECIES_LYCAN
 	examine_limb_id = SPECIES_LYCAN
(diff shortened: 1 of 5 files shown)


nikothedude:
Razor claws sprite doesnt work due to the way its formatted in the dmi

Rykka-Stormheart:
Why was door prying removed?

And will regular, non-boosted lycans still get claws, just weaker?

nikothedude:
> Why was door prying removed?
> 
> And will regular, non-boosted lycans still get claws, just weaker?

Realized I couldnt make it cleanly interact with the pacifism thing. And no, crew lycans are unlikely to eb given this, because an unrevocable ability to smash all windows would render all lycans uncontainable (unless pacified which we DO NOT WANT)

PR #5745, opened 2026-06-06
Victoria's Seikret

lukevale:
Adds in a new muzzle, and a tail. As well as a new hair gradient.
More customization BABEEEEE
## Proof Of Testing
<details>
image
image
image
image
image
image
image

</details>

## Changelog
:cl:
add: Seikret Snout
add: Seikret Tail
add Seikret tail (alt)
add: Cutout hair gradient
add: Harpy large (alt)
add Harpy large (dual color. Alt and non alt)
/:cl:


nikothedude:
honestly im not really a fan of the tail but i cant really articulate why... i guess it just... looks like... really flat? no structure? the words fail me. im not a spriter

lukevale:
Perfectly fine and understandable. But the people I've shown it to are happy with how it came out.
If I ever get the itch to really go in and do MORE I might.

AstrumLernaean:
The side view of 'snout' looks rather odd: it doesn't quite match the look of what a Seikret has, their fascial structure appears a bit more boney and very much scaled compared to the rest of their body; here it appears way too... smooth? There could be a bit more shading in the greyscale/coloration segments to indicate how the colors shift without it basically melding with the rest of the body in a way, the beak doesn't look much like a beak for example on the side profile (the front profile is pretty on point though but that's a given, still could use a _tiny_ bit of shading though with how many pixels it occupies in the front perhaps? Uncertain on this point.). On that note they also have 3 pronged "spikes"/horns on the sides of their heads, the largest being at the top poking slightly above their head, the smallest one in the middle, and the second largest at the bottom near their "chin" or closer to their neck rather.

The tail does also feel a little flat and doesn't quite reflect that a majority of it is mainly occupied by feathers: the thickest portion of their tail is near the base of the tail, where it quickly shrinks down but has an almost consistent width and flow of feathers. We'd suggest potentially shrinking how much of the scaled "underbelly" is present beneath the tail around or just before the halfway point of it, and bringing it closer to the tail-base, while thinning out the side-profile of the last half of the tail with primarily the feather colors with a more sort of "spikey" pattern to them along the edges, and have it not appear as "thick" as it moves towards the end of the tail?

Just some suggestions in hopes it helps improve them a bit more.
Example images with outlines for reference:
image
image

lukevale:
TBF I didn't go directly based off the design as it is exactly in MH:W
Since these are for anthros I did mostly look at how art of them is betrayed and what looked good there. Since the thick base of the tail meets with the underbelly it makes sense why it's so much thicker in the ACTUAL raptor body. But for being used in an anthro with it being yes, BASED on a seikret, but not a direct 1 to 1 (because god damn I do not have the spriting skill to capture that reliably)

The beak was one of those absolutly annoying things to handle. SS13 is. Well. SS13. And I used other muzzles as bases, and in the end I think I got good enough. I can't replicate the style of head easily with the space I'm given on the sprite, and I tried alot of diffrent variations to eventually end up where I'm at now. 

For me it comes down to just the lack of resolution with sprites and ending with what felt comfy. ((It's why I posted these multiple times in the spriting channel for feedback and no one really said anything. LOL))

lukevale:
Oh on the note of me basing the tail off of the art I was able to find? There's like. 5 different way people portray their tails.
Personally just picked what I like the most, though I can always trim and slim it down and see what works. I'll give it a try tomorrow.

gonna give the beak a second (9th)  pass as well.

AstrumLernaean:
All good! And yeah lol we rarely look in the spriting channel, there's a lot we tend to keep up with and that isn't one of them, github is though, but yeah!
It's pretty good to try getting the general vibe of the original but we totally get it, just wanting to provide some feedback since it seemed no one else had at the time. Looking forward to seeing your skills improve, kudos to taking on the task!

lukevale:
On second look I actually 100% get what Nico means LOL yeah I'm gonna give it another pass. Raise the end up more. And gonna make a alt that's the same overall design but thins out much closer to how the Serikret is.

lukevale:
Did the beak
image
@AstrumLernaean

AstrumLernaean:
Oh nice! That beak looks a lot better already! How's the tail looking?

lukevale:
did some minor adjustments, to it. The alt one will be made soonish
image

lukevale:
adding new stuff to this as well because of course I am
😔

lukevale:
new tail in
image
image

lukevale:
And alt and duel color wings because they go with the set. :D 
image
image
image

AstrumLernaean:
> And alt and duel color wings because they go with the set. :D 

Holy damn that new tail is absolutely _**perfect**_, well done!!

Those wingarms DO feel quite a bit too big though for a seikret, and they do have some claws at the forearms too, do keep that in mind in case you feel like doing another pass on it, no biggy if not and that's how you'd like the wings to be though! You're definitely going above and beyond as is, great work!

lukevale:
@AstrumLernaean The harpy wings are wings that already exist. I simply implemented the dual color and removed the patch that bridges them across the back. (that causes it to clip atop of hair. Which can be good or bad but that's why it's an ALT)

I like both tails, and each will have their uses! :D

AstrumLernaean:
> @AstrumLernaean The harpy wings are wings that already exist. I simply implemented the dual color and removed the patch that bridges them across the back. (that causes it to clip atop of hair. Which can be good or bad but that's why it's an ALT)
> 
> I like both tails, and each will have their uses! :D

OH we see now, did not catch that at first!
Well either way it's probably fine, well done 👍

Arturlang (approved):
code good, slightly worried about big sprite being in the way

Arturlang:
we have far bigger tail sprites

PR #5769, opened 2026-06-11
Allows CSL to work with every species

xPokee:
## About The Pull Request
This PR (inspired by https://github.com/NovaSector/NovaSector/pull/6278, but reworked) allows every species to be able to select the Common Second Language quirk.

This also allows you to use any language you select as your second language (I'm not sure if that was already a thing, but if not, you can do it now).

When using the CSL quirk you MUST have at least two languages selected, or you will be UNABLE TO SPEAK.
## Why It's Good For The Game
More people can take a quirk.
## Proof Of Testing
This was on a ghoul. Ghouls used to be unable to take the quirk at all.
image
## Changelog
:cl:
add: Humans, and everyone that uses human_basic for a language will now speak uncommon by default. This can still be removed in the language menu.
fix: Common Second Language (CSL) now works with every species. Make sure you have two languages selected for the quirk to work!
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/languages/_language_holder.dm
+++ b/modular_zubbers/code/modules/languages/_language_holder.dm
@@ -28,3 +28,15 @@
 		/datum/language/spinwarder = list(LANGUAGE_ATOM),
 	)
 	selected_language = /datum/language/spinwarder
+
+// Allows all species(?) to work with Common Second Language.
+/datum/language_holder/human_basic/New()
+	. = ..()
+	understood_languages |= list(
+		/datum/language/common = list(LANGUAGE_ATOM),
+		/datum/language/uncommon = list(LANGUAGE_ATOM),
+	)
+	spoken_languages |= list(
+		/datum/language/common = list(LANGUAGE_ATOM),
+		/datum/language/uncommon = list(LANGUAGE_ATOM),
+	)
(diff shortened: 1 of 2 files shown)


KateAnbolho:
pogging my mind out

xPokee:
Apparently neither side of the posted bounty is actually able to pay, so I am closing this to avoid further frustration.

EDIT: I've been paid for this now.

modular_zubbers/code/modules/languages/_language_holder.dm
```diff
@@ -28,3 +28,15 @@
...
+// Allows all species(?) to work with Common Second Language.
+/datum/language_holder/human_basic/New()
+	. = ..()
+	understood_languages |= list(
+		/datum/language/common = list(LANGUAGE_ATOM),
+		/datum/language/uncommon = list(LANGUAGE_ATOM),
+	)
+	spoken_languages |= list(
+		/datum/language/common = list(LANGUAGE_ATOM),
+		/datum/language/uncommon = list(LANGUAGE_ATOM),
```
nikothedude, line 41: im assuming this isnt giving all humans the uncommon language?

xPokee: It would give uncommon to humans by default due to the nature of the changes. If players don't want it they can still just remove the language though, so it's no big deal in my opinion.

If you have a better solution I'd happily try it.

nikothedude: A better alternative would be to override is_species_appropriate to be simply true, wouldnt it? That way all species are considered to be... appropiate, and we dont have a side effect.

xPokee:
Closing this to open another PR (same fix) instead of reverting all my changes across multiple commits.

PR #5780, opened 2026-06-12
[TM ONLY] Revert "Upstream 6 2 2026 [IDB IGNORE] [MDB IGNORE]"

nikothedude:
Reverts Bubberstation/Bubberstation#5733


LordSaladin:
unscalp the synths

PR #5791, opened 2026-06-14
New offices for the NTC and the Blueshield

ibexgoetia:
## About The Pull Request

I've remade (or added) new offices for the Nanotrasen Consultant and the Blueshield in most maps.
Here are the offices I've remade:
- Icebox
- Tramstation
- Delta Station
- Void Raptor
- Metastation
- Boxstation
- Kilostation
- Catwalk
- Nebula
- Biodome

I've ignored the more recent maps/maps that are being actively worked on.
For the sake of compatibility, I have also removed the blueshield's locker from the pool of items that automatically spawn in every map. WHAT THIS MEANS is that all maps need to have a blushield's locker mapped in, or it will not spawn.

## Why It's Good For The Game

Most maps are missing offices for the blueshields, and the ntc offices could use a wee update

## Proof Of Testing
Some examples of the reworked offices; They have most likely been modified and changed slightly. I suggest using mapdiffbot to see the new offices.

Tramstation's Blueshield's office
image

Icebox
image

## Changelog

:cl:
map: The offices of the Blueshield and the NTC have been reworked on most maps.
/:cl:


shayoki (approved):
The map changes are good for the most part. The automapper replacing the aux tool storage in ice box is rough, but we'll see what the players think.

The BlueShield armory being a staple now is fine as well, after some access were revoked from them, this is fair. I should warn you about the second mapped in blueshield combat kit for balance reasons. If there will be one in the locker, as well as a shelf, then there will be two of these. I'd personally like for a single one on the station (BS locker) at all times due to its extreme size capacity (It is a reskinned nukie combat kit, aka 28 slots of medical equipment).

And in an ideal world, the praetorian MOD would be in said suit storage in the BS armory, but this should be expected if every single map ever has an armory.

Overall, I was impressed by good use of space to bring in a BlueShield office to upstream maps.

ibexgoetia:
> The map changes are good for the most part. The automapper replacing the aux tool storage in ice box is rough, but we'll see what the players think.

I was trying to figure out where else to place it, with most other places being either too small or too much in the way. I could look into it again, however.

As for the kits, I didn't think too much about giving them two kits since there already are a couple maps with spare combat kits. TO MY KNOWLEDGE the surgical kit that all doctors get should also have the same space as the blueshield kit.

I suppose I could look into replacing the kits with a regular first aid kit, as opposed to the bluey one.

I could also code in a blueshield suit locker with a praetorian in it, yeah, and just slap the modsuit in the locker if for any reason there isn't enough space for it.

Many of my concerns were in regards to how other maps would have to handle these changes, given that now, for example, blueshields NEED to have a locker and a spawn mapped in (as opposed to it being automapped). But given that I don't believe we will be getting any more upstream maps any time soon, I suppose this could be doable.

ibexgoetia:
Oh and, another todo I have is to make a randomised loot pool of fluff to give to the safes, instead of being always the same. I'd need to look into how maint loot spawners work, and add a bunch of innocent but cool things to give them, like cigars, plushies, documents, alcohol, et cetera.

ibexgoetia:
image

This was my first idea, as a replacement for the AUX tool storage area. The little cranny between the library and the barber; but it'd make a 2x4 space (or 3x4 if I make it eat a little of the space to the entrance of the library, which I am not a fan of)

lessthnthree (approved):

PR #5799, opened 2026-06-15
Allows Lycans to wear underwear

xPokee:
## About The Pull Request
Does as the title says.
## Why It's Good For The Game
I don't want to flash bang people with my genitals every time I transform, and I'd reckon some people don't want to be flash banged by lycan genitals either. Also, @nikothedude told me that them not having underwear causes a genital visibility issue of some sort.
## Proof Of Testing
See below.
<details>
image

</details>

## Changelog
:cl:
add: Lycans can now wear underwear.
/:cl:


nikothedude:
bug being that lycan slots can sttill select underwear if not originally a lyacn slot and this bugs out and hides genitals

PR #5821, opened 2026-06-21
Fixes power cells being unable to receive plasma injections

NanoCats:
## About The Pull Request

Fixes plasma injection into charged power cells. TG introduced a bug via refractor where you could only inject plasma into a dead cell, which made the sabotage mechanic (inject plasma, put cell in thing, cell detonates on next use) basically impossible.

The root cause is a regression from TG upstream PR #94861, which rewrote the plasma rigging system from a `rigged` boolean with signal hooks into the current `spark_act` approach. Somewhere in that transition the `INJECTABLE` flag gets stripped from the reagents holder on charged cells specifically, leaving uncharged cells unaffected. The explosion mechanic itself is intact and a rigged cell still detonates correctly when its charge is altered but the injection step was broken.

The fix overrides `is_injectable` on `power_store` to bypass the flag check entirely, gating only on whether the reagents holder exists at all, which it does.

## Why It's Good For The Game

The plasma cell sabotage is funny. The mechanic only working on a dead cell was an unintentional restriction that made the mechanic worthless. 

## Proof Of Testing

image

## Changelog

:cl:
fix: Plasma can once again be injected into charged power cells.
/:cl:

### DIFF
--- a/code/modules/power/power_store.dm
+++ b/code/modules/power/power_store.dm
@@ -46,7 +46,8 @@
 
 /obj/item/stock_parts/power_store/is_injectable()
 	return !!reagents
-
+/obj/item/stock_parts/power_store/is_injectable()
+	return reagents
 /obj/item/stock_parts/power_store/Initialize(mapload, override_maxcharge)
 	. = ..()
 	create_reagents(5, INJECTABLE | DRAINABLE)
--- a/code/modules/power/power_store.dm
+++ b/code/modules/power/power_store.dm
@@ -45,9 +45,8 @@
 	return .
 
 /obj/item/stock_parts/power_store/is_injectable()
-	return !!reagents
-/obj/item/stock_parts/power_store/is_injectable()
 	return reagents
+
 /obj/item/stock_parts/power_store/Initialize(mapload, override_maxcharge)
 	. = ..()
 	create_reagents(5, INJECTABLE | DRAINABLE)
--- a/code/modules/power/power_store.dm
+++ b/code/modules/power/power_store.dm
@@ -44,6 +44,9 @@
 	. += NAMEOF(src, corrupted)
 	return .
 
+/obj/item/stock_parts/power_store/is_injectable()
+	return !!reagents
+
 /obj/item/stock_parts/power_store/Initialize(mapload, override_maxcharge)
 	. = ..()
 	create_reagents(5, INJECTABLE | DRAINABLE)


code/modules/power/power_store.dm
```diff
@@ -44,6 +44,9 @@
 	. += NAMEOF(src, corrupted)
 	return .
 
+/obj/item/stock_parts/power_store/is_injectable()
+	return !!reagents
+
```
Maia-J:
```suggestion
/obj/item/stock_parts/power_store/is_injectable()
	return reagents
```
Any nonzero value is truthy

Maia-J (commented):
> Somewhere in that transition the INJECTABLE flag gets stripped from the reagents holder on charged cells specifically, leaving uncharged cells unaffected. The explosion mechanic itself is intact and a rigged cell still detonates correctly when its charge is altered but the injection step was broken.

That is just hiding the issue instead of fixing the INJECTABLE/REFILLABLE flag being lost

code/modules/power/power_store.dm
Maia-J: I think github is having a stroke with this suggested change, what
image

NanoCats:
> > Somewhere in that transition the INJECTABLE flag gets stripped from the reagents holder on charged cells specifically, leaving uncharged cells unaffected. The explosion mechanic itself is intact and a rigged cell still detonates correctly when its charge is altered but the injection step was broken.
> 
> That is just hiding the issue instead of fixing the INJECTABLE/REFILLABLE flag being lost

so do you just want that thing applied or what

Maia-J:
> > That is just hiding the issue instead of fixing the INJECTABLE/REFILLABLE flag being lost
> 
> so do you just want that thing applied or what

Ideally I want the actual issue to be traced and fixed but I just dropped a little bit of extra wisdom in

code/modules/power/power_store.dm
```diff
@@ -44,6 +44,10 @@
 	. += NAMEOF(src, corrupted)
 	return .
 
+/obj/item/stock_parts/power_store/is_injectable()
+	return !!reagents
+/obj/item/stock_parts/power_store/is_injectable()
+	return reagents
```
Arturlang, line 48: you're overriding the same proc twice, seems like a mistake

Arturlang (changes requested):

code/modules/power/power_store.dm
NanoCats: I think that was Majik's suggested change

NanoCats:
> > > That is just hiding the issue instead of fixing the INJECTABLE/REFILLABLE flag being lost
> > 
> > 
> > so do you just want that thing applied or what
> 
> Ideally I want the actual issue to be traced and fixed but I just dropped a little bit of extra wisdom in

While I understand your broader point about wanting the root cause traced here; I've already done more investigation than most would bother with on a wee bug inherited from upstream. I believe that this override is a clean, defensible lil' solution. If you want to dig further yourself you are welcome to.

I just want cells to be able to go spoomy boomy.

code/modules/power/power_store.dm
Maia-J: I do not know why but requesting changes is just borked for me and uses the wrong lines. Probably the Github sloppification

StrangeWeirdKitten:
why can't this be fixed upstream?

NanoCats:
> why can't this be fixed upstream?

For reasons which are my own I will not be engaging Tgstation directly! If you wish to submit a PR to correct this there, you are welcome to!

StrangeWeirdKitten:
Ok, /tg/ bug fixes don't get merged.

PR #5826, opened 2026-06-22
Maybe bioscramblers SHOULDN'T give you genitals

xPokee:
## About The Pull Request
When looking at the causes of flaky errors I saw https://github.com/Bubberstation/Bubberstation/issues/5754 which upon investigation seems to be caused by genitals not being a blacklisted organ. Even if it doesn't fix the flaky test failure, bioscramblers definitely shouldn't give you genitalia.
## Why It's Good For The Game
Bugs that prefbreak me are bad.
## Proof Of Testing
Untested.
## Changelog
Not player facing.


xPokee:
I'm going to kill the flaky screenshot tests failures with a hammer.

PR #5858, opened 2026-06-28
Body Morpher quirk + Protean Shapeshift

ibexgoetia:
## About The Pull Request
I have partly ported the Body Morpher quirk from SPLURT, with some fixes. It's a neutral quirk, which costs 0 points. It's an alternative to the Polymorph program for the NIF, meant to be used by slimey/shapeshifter characters. The quirk is blacklisted for proteans, because proteans now innately have a shapeshift ability, since they're nanite slimes, as well as slime people obviously.

## Why It's Good For The Game
More flavour for shapeshifting characters.

## Proof Of Testing
image
image

</details>

## Changelog
:cl:
add: New Body Morpher neutral quirk.
add: Proteans are now able to shapeshift at will.
/:cl:


ibexgoetia:
forgor to blacklist slimepeople from having the quirk as well

BingusTheClown:
Will this have the same rules and restrictions as the NIF Polymorph?

ibexgoetia:
> Will this have the same rules and restrictions as the NIF Polymorph?

yes, it's exactly the same as the NIF polymorph (which is, in turn, exactly the same as the slime polymorph)

BingusTheClown:
Is it possible to add logging for when this is used? Would be helpful for the admin side of things.

ibexgoetia:
> Is it possible to add logging for when this is used? Would be helpful for the admin side of things.

It already does! It inherits the same exact functionalities, including the logging. :D

image

PR #5861, opened 2026-06-28
Adds an item to match Proteans and wearers

MintyPhresh:
**In an attempt to make the most out of their Protean staff's unique abilities, Nanotrasen is launching Protean Match, an opt-in program that pairs Proteans with willing wearers. Enjoy the perks of a battery-unlimited MOD without having to know one personally-- or be the living assistant to an experienced and knowledgeable crewmember. Join today!**

## About The Pull Request

This PR adds Protean Match tags to the Loadout and RoboDrobe, which allow crew (both Protean and otherwise) to opt-in to a system that will automatically match them with an available Wearer or Protean.

When used, the tag adds you to a pool of candidates waiting to match with another. If there's already someone in the pool you can match with, it instead picks randomly from all other tags in the pool.

You can unclaim the tag at any time by using them in hand. This will both remove the tag from the pool, and unpair you from your match, if you're matched with anyone. 

These tags can't be used if you have the Entombed quirk, are a snailperson, have no ID, have no sec record, or are a prisoner.

_(To add this to the premium tab of the RoboDrobe, I had to change the way Protean organs' pricing was set, otherwise these tags would cost 220 credits. This should hopefully be saner.)_

_(I also went and fixed some spacing and capitalization in code that's adjacent to what I changed.)_

## Why It's Good For The Game

Finding a wearer as a Protean can be tough-- advertising over radio often has a bystander effect that makes people less likely to respond, and approaching someone directly can feel risky. These tags circumvent that, and act as a way to signal openness to being worn/wearing without saying a word.

Getting Proteans on peoples' back is good because it's fun! Nothing encourages RP like hanging off someone else's back (or having them hang off you), and it's generally more attractive than other forms of team-up RP (like pAI) since you can get up and walk away whenever you want, and can use existing, independent characters for it.

## Proof Of Testing

Suffer my evil image editing.
<details>
image

</details>

## Changelog

:cl: Mintyphresh
add: Added Protean Match tags to the loadout and RoboDrobe. Use them in hand to search for Proteans to wear, or crew to be worn by. Or, just wear them on their own to signal you're open to the idea.
/:cl:


MintyPhresh:
I promise these failed CIs are not my fault I'm just gonna keep refreshing

code/modules/loadout/loadout_helpers.dm
```diff
@@ -44,13 +44,13 @@
 		// BUBBER EDIT ADDITION BEGIN
 		if(item.restricted_roles && equipping && !(equipping.title in item.restricted_roles))
 			if(preference_source.parent)
-				to_chat(preference_source.parent, span_warning("You were unable to get a loadout item([initial(item.item_path.name)]) due to job restrictions!"))
+				to_chat(preference_source.parent, span_warning("You were unable to get a loadout item ([initial(item.item_path.name)]) due to job restrictions!"))
```
StrangeWeirdKitten, line 47:
```suggestion
				to_chat(preference_source.parent, span_warning("You were unable to get a loadout item([initial(item.item_path.name)]) due to job restrictions!"))
```

StrangeWeirdKitten: I know it's a single space but it'll still cause github to conflict if our upstreams touch this file.

code/modules/loadout/loadout_helpers.dm
```diff
@@ -44,13 +44,13 @@
...
 			loadout_datums -= item
 			continue
 
 		if(item.blacklisted_roles && equipping && (equipping.title in item.blacklisted_roles))
 			if(preference_source.parent)
-				to_chat(preference_source.parent, span_warning("You were unable to get a loadout item([initial(item.item_path.name)]) due to job blacklists!"))
+				to_chat(preference_source.parent, span_warning("You were unable to get a loadout item ([initial(item.item_path.name)]) due to job blacklists!"))
```
StrangeWeirdKitten, line 53:
```suggestion
				to_chat(preference_source.parent, span_warning("You were unable to get a loadout item([initial(item.item_path.name)]) due to job blacklists!"))
```
Dito

code/modules/loadout/loadout_helpers.dm
```diff
@@ -60,6 +60,12 @@
 			loadout_datums -= item
 			continue
 
+		if(item.blacklisted_species && (dna.species.id in item.blacklisted_species))
+			if(preference_source.parent)
+				to_chat(preference_source.parent, span_warning("You were unable to get a loadout item ([initial(item.item_path.name)]) due to species blacklists!"))
+			loadout_datums -= item
+			continue
+
```
StrangeWeirdKitten, line 68:
```suggestion
                // BUBBER EDIT START
		if(item.blacklisted_species && (dna.species.id in item.blacklisted_species))
			if(preference_source.parent)
				to_chat(preference_source.parent, span_warning("You were unable to get a loadout item ([initial(item.item_path.name)]) due to species blacklists!"))
			loadout_datums -= item
			continue
		// BUBBER EDIT END
```
This is a /tg/ file, thus we have to indicate it is modified for future maintainability and that this code belongs to us.

code/modules/loadout/loadout_items.dm
```diff
@@ -70,6 +70,8 @@ GLOBAL_LIST_INIT(all_loadout_categories, init_loadout_categories())
 	var/list/blacklisted_roles
 	/// If set, is a list of species which can get the loadout item
 	var/list/restricted_species
+	/// If set, is a list of species which can't get the loadout item
+	var/list/blacklisted_species
```
StrangeWeirdKitten, line 74:
```suggestion
	/// BUBBER EDIT - If set, is a list of species which can't get the loadout item
	var/list/blacklisted_species
```
Same as above.

StrangeWeirdKitten (changes requested):
part 1 until i can sit down and look more in depth at the code.

modular_zubbers/code/modules/clothing/accessory/accessory.dm
```diff
@@ -172,6 +172,148 @@
...
+// This list contains either unmatched Wearers or unmatched Proteans, waiting for the opposite type to join and match them.
+// It should never ever ever contain both, since if the second type tries to join the list it'll just get paired.
+// So if we need to differentiate between if there's Proteans or Wearers in the pool, we just check the first entry with the pool_polarity() proc.
+#define POOL GLOB.protean_match_tag_pool
+#define UNASSIGNED "0"
+#define IN_POOL "1"
+#define MATCHED "2"
+
+/obj/item/clothing/accessory/dogtags/protean_match
+	name = "\improper Protean Match tag"
```
StrangeWeirdKitten, line 190:
```suggestion
	name = protean match tag"
```
Why?

modular_zubbers/code/modules/clothing/accessory/accessory.dm
```diff
@@ -172,6 +172,148 @@
...
+/obj/item/clothing/accessory/dogtags/protean_match/proc/unassign(unmatch_reason)
+	match_progress = UNASSIGNED
+	POOL -= src
+	assignee_name = initial(assignee_name)
+	assignee_protean = initial(assignee_protean)
+	name = initial(name)
+	paired_tag = null
+	update_static_data_for_all_viewers()
+	if(unmatch_reason)
+		say("You have been unmatched. Reason: [unmatch_reason].")
```
StrangeWeirdKitten, line 332: I'm assuming both tags hold a ref for each other? The classic pitfall is circular refs causing some hard deletes. Create and Destroy wouldn't catch that in this case. So you need to either use weakrefs or ensure the other tag ``paired_tag`` is also set to null.

This explains weakrefs, that's the safer option. But you can also do paired_tag.paired_tag = null
https://github.com/Bubberstation/Bubberstation/blob/master/code/datums/weakrefs.dm#L16

StrangeWeirdKitten (commented):
some extra stuff, will have to look at more later.

code/modules/loadout/loadout_helpers.dm
MintyPhresh: There's already a bubber edit start marker on line 44, and a bubber edit end marker on line 86! That also puts the spacing changes within the bubber edit, too

code/modules/loadout/loadout_items.dm
MintyPhresh: Existing bubber edit start/end markers at lines 64 and 81

modular_zubbers/code/modules/clothing/accessory/accessory.dm
MintyPhresh: I had this as a proper name because "Protean Match" is the name of the NT initiative that's distributing these tags

StrangeWeirdKitten: sounds good

code/modules/loadout/loadout_items.dm
StrangeWeirdKitten: Put the bubber edit on the actual var you've made. It's way clearer on what var we're adding.

code/modules/loadout/loadout_helpers.dm
StrangeWeirdKitten: Move it down to purely around the edited code block. it serves as a legit marker for what lines we've exactly touched.

MintyPhresh: What I'm saying is that this is already in the edited code block. the block is 44-87 and these lines are 63-68

code/modules/loadout/loadout_items.dm
MintyPhresh: Should I retroactively add bubber edit to the other vars within this edit block, too?

code/modules/loadout/loadout_helpers.dm
StrangeWeirdKitten: ok, nvm about this then.

code/modules/loadout/loadout_items.dm
StrangeWeirdKitten: ugg, I see. It's older code. Sorry, was confused. nvm then.

modular_zubbers/code/modules/clothing/accessory/accessory.dm
MintyPhresh: Code's been edited to use weakrefs and it doesn't seem to be affecting functionality. How's the implementation look?

modular_zubbers/code/modules/clothing/accessory/accessory.dm
```diff
@@ -172,6 +172,151 @@
...
+	if(match_progress == MATCHED)
+		var/obj/item/clothing/accessory/dogtags/protean_match/paired_tag = paired_tag_weakref?.resolve()
+		paired_tag.unassign("paired tag destroyed")
+	return ..()
+
+///Filters out crew that can't wear Proteans or shouldn't have them.
+/obj/item/clothing/accessory/dogtags/protean_match/proc/check_eligibility(mob/source)
+	// ID card is how we check that the user is NT crew and not a prisoner.
+	. = TRUE
+	var/ineligibility_reason = null
```
StrangeWeirdKitten, line 256:
```suggestion
	var/ineligibility_reason
```
 Redundant

modular_zubbers/code/modules/vending/wardrobe.dm
```diff
@@ -137,8 +137,8 @@
 	premium = list(
 		/obj/item/organ/stomach/protean = 1,
 		/obj/item/organ/heart/protean = 1,
+		/obj/item/clothing/accessory/dogtags/protean_match = 6,
```
StrangeWeirdKitten, line 140:
```suggestion
		/obj/item/clothing/accessory/dogtags/protean_match = 4,
```
Are we reasonably going to expect this many pairs?

modular_zubbers/code/modules/clothing/accessory/accessory.dm
```diff
@@ -172,6 +172,151 @@
...
+	playsound(src, "modular_skyrat/modules/emotes/sound/emotes/synth_yes.ogg", 20, FALSE)
+	say("Matched with [(other_tag.assignee_protean == TRUE) ? "Protean" : "Wearer"]: [other_tag.assignee_name]!")
+	paired_tag_weakref = other_tag.create_weakref()
+
+/// Used both when unmatching with a person and when leaving the match pool.
+/obj/item/clothing/accessory/dogtags/protean_match/proc/unassign(unmatch_reason)
+	match_progress = UNASSIGNED
+	POOL -= src
+	assignee_name = initial(assignee_name)
+	assignee_protean = initial(assignee_protean)
```
StrangeWeirdKitten:
```suggestion
	assignee_name = null
	assignee_protean = null
```
Just null the variables

modular_zubbers/code/modules/clothing/accessory/accessory.dm
```diff
@@ -172,6 +172,151 @@
...
+	name = initial(name)
+	paired_tag_weakref = null
+	update_static_data_for_all_viewers()
+	if(unmatch_reason)
+		say("You have been unmatched. Reason: [unmatch_reason].")
+
+/obj/item/clothing/accessory/dogtags/protean_match/examine(mob/user)
```
StrangeWeirdKitten, line 334: Make the examines text some sort of span. Span_notice() would be the best application.

modular_zubbers/code/modules/clothing/accessory/accessory.dm
```diff
@@ -172,6 +172,151 @@
...
+
+/obj/item/clothing/accessory/dogtags/protean_match/Destroy()
+	POOL -= src
+	if(match_progress == MATCHED)
+		var/obj/item/clothing/accessory/dogtags/protean_match/paired_tag = paired_tag_weakref?.resolve()
+		paired_tag.unassign("paired tag destroyed")
+	return ..()
+
+///Filters out crew that can't wear Proteans or shouldn't have them.
+/obj/item/clothing/accessory/dogtags/protean_match/proc/check_eligibility(mob/source)
```
StrangeWeirdKitten: This entire proc bothers me. The item shouldn't just be intrinsically aware that you're non-crew or wanted or an even prisoner. (You restricted the item for them anyway!) The only thing it should do is check eligibility wise is people on the station Z level. 

It's just an item, not corporate security. 

The checks for snail and entombed are fine though. That's genuine mechanical restrictions.

StrangeWeirdKitten (changes requested):

modular_zubbers/code/modules/vending/wardrobe.dm
MintyPhresh: I'd rather not have people worried that supply might run out. Do you foresee any negative consequences to having the number be that high?

modular_zubbers/code/modules/clothing/accessory/accessory.dm
MintyPhresh: I'll admit it was a mechanics-first decision, but I don't think its unreasonable for a device to do a quick check that you actually exist as an NT employee in their records. When I wrote this I was asking myself how the tags could be abused and figured that someone who sits around with it active but doesn't meet with their match would be something I'd want to design against.

Non-crew and prisoners were the most likely candidates for that-- the former are rarely on station, and the latter probably wouldn't be allowed to meet up with their match if they're outside prison. 

Additionally, from a lore perspective, Protean crew are probably the most advanced and expensive form of synthetic life NT employs-- organs cost hundreds, shells cost thousands, and neither can be produced on station. All their other, mundaner MODs are protected behind ID-locked doors at the very least, why would their tags put the burden of verifying employment on the Protean?

~~Agreed that the check for the prisoner job mind is weird. It came up in testing that roundstart prisoners aren't marked as incarcerated in the code, and I wanted that check to cover both roundstart and non-roundstart prisoners. It does that in a bad way, though, and as you said it's not that important to prevent roundstart prisoners from using it when they can't even get it. I'll remove that part next push.~~

The Z-level check is a very good point, though, that's something I'd add next push.

(Edit: Misremembered how I addressed roundstart prisoners-- the code check for the job assignment written on the ID, not checking the user's mind job, so I think it's reasonable to keep)

modular_zubbers/code/modules/clothing/accessory/accessory.dm
MintyPhresh: Does this have a functional difference, or is it just a code style change? I can't say I was consciously thinking about the difference between declaring it null or not when I was writing this, but I think it makes the code a little easier to read to explicitly call out that it starts as null, then checks if its not null after going through all the if statements that would set the ineligibility_reason to something other than null

StrangeWeirdKitten (approved):

PR #5864, opened 2026-06-29
Returns Space Z Level Count To 7

R0bro:
## About The Pull Request

[The count was reduced all the way back in 2024 as a way to help server performance ](https://github.com/Bubberstation/Bubberstation/pull/2510) however I've been told it might be worth upping again.
## Why It's Good For The Game

People yearn for the space age. Tiding gets boring and there's often not a lot in a given round, especially when you have stuff like Tarkon that take up half a large portion of a given z-level.

## Proof Of Testing

The tests don't fail! Might need a test merge to judge the performance impact at scale if there's still concern.

:cl: Robwo
code: Returned space Z-level count back to 7.
/:cl:


KateAnbolho:
This needs to be a config PR I think? but don't quote me on that, staff would need to say.

lessthnthree (approved):

PR #5875, opened 2026-07-01
The Return of the (printable) defib

ibexgoetia:
image

## About The Pull Request
With this I am reverting the removal of the printable defibrillator, a change made by /tg/ some time ago.
Only the **base** defibrillator is printable, not the compact/belt slot defibrillator. It will need to have a cell inserted to work.

## Why It's Good For The Game

The change was initially made to increase the amount of interactions with cargo, but I believe that this only makes the process of obtaining defibrillators bothersome, as the interdepartmental action solely involves pressing a single button on the order console...
Obtaining defibrillators through cargo is still possible; you'll get three with charged batteries in them, without having to spend resources. I'm simply re-adding the old way to get them. This change was disliked both on /tg/ and here.

## Proof Of Testing

image

## Changelog

:cl:
balance: Defibrillators are now printable once again from the medical techfab.
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/research/designs/medical_designs.dm
+++ b/modular_zubbers/code/modules/research/designs/medical_designs.dm
@@ -48,3 +48,15 @@
 	RND_CATEGORY_TOOLS + RND_SUBCATEGORY_TOOLS_MEDICAL_ALIEN
 	)
 	departmental_flags = DEPARTMENT_BITFLAG_MEDICAL
+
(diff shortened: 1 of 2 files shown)


KateAnbolho:
🫡

nikothedude:
tg has this strange idea that "go to desk and ask for something" is good interaction when its just boring boilerplate

lessthnthree (approved):

PR #5892, opened 2026-07-03
Converts the UwU Quirk to an UwU Mutation

xPokee:
## About The Pull Request
Does as the title would suggest.
Only felinids can have this mutation naturally, however, anyone can obtain it via an injector.
## Why It's Good For The Game
Brings the code in-line with our other speech based things, making it a negative mutation instead of a neutral quirk. See: Swedish, Medieval, etc.
## Proof Of Testing
See below.
<details>
image

</details>

## Changelog
:cl:
refactor: The uwu quirk is now a negative mutation instead. It can be found in felinids, scan one today!
/:cl:


Boopideedoo:
LMAO the one thing I add has people already trying to lock it as a mutation

Boopideedoo:
wouldn't it be better to lock it behind the race of cats since that's what it was intended for if you're going to add it to the DNA pool I wouldn't usually see humans using it nor lizards. just some thought

xPokee:
> wouldn't it be better to lock it behind the race of cats since that's what it was intended for if you're going to add it to the DNA pool I wouldn't usually see humans using it nor lizards. just some thought

I can do that, yeah.

lessthnthree (approved):

PR #5924, opened 2026-07-09
THE WITCH IS DEAD (Mini egun is gone from cargo)

nikothedude:
## About The Pull Request

Title.

## Why It's Good For The Game

Ok so the mini egun is just straight up better than the laser pistol in literally every single way except for charge speed. Does more damage, has more ammo, has a disabler, has a seclite, is SMALL (CAN FIT IN BOXES), and is only somewhat more expensive. You can hoard these in a box and get really silly.

By removing it from cargo altogether we can sidestep all of these issues by making it an exclusive item you cant easily replace. Balance doesnt matter as much when its limited in number.

## Proof Of Testing
<details>

image

</details>

## Changelog
:cl:
balance: Mini egun is gone from cargo
/:cl:

### DIFF
--- a/code/modules/projectiles/guns/energy/energy_gun.dm
+++ b/code/modules/projectiles/guns/energy/energy_gun.dm
@@ -39,7 +39,6 @@
 	ammo_x_offset = 2
 	charge_sections = 3
 	single_shot_type_overlay = FALSE
-	projectile_damage_multiplier = 0.8 // BUBBER EDIT ADDITION
 
 /obj/item/gun/energy/e_gun/mini/add_seclight_point()
 	// The mini energy gun's light comes attached but is unremovable.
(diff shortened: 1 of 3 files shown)


nikothedude:
yes it says 1100 cr thats a old image

CatoChristopherMrow:
This kinda feels like a "choose one" moment.

>Ok so the mini egun is just straight up better than the laser pistol in literally every single way

Less capacity on lethal, I believe.

KateAnbolho:
This seems sorta like a gigantic nerf tbh, 20% reduction is wild. Maybe we should more approach cell size/drain per shot? Also people will just buy the regular single pack which isn't that big of a deal.

StrangeWeirdKitten:
20 percent is a massive nerf and that's kind of the difference between someone using it or leaving it inside their locker or on the floor.

Charge doesn't matter because I can fit 5 of the things on me. Money doesn't matter because money really doesn't stop anything.

My solution, remove it from cargo. There's plenty of places it shows up and it suddenly has scarcity. You're also indirecting nerfing roles like bridge assistant otherwise, and worse, giving them a pea shooter to noob trap them with.

StrangeWeirdKitten (approved):

xPokee (approved):

PR #5969, opened 2026-07-19
Bounty selection no longer runtimes for Blacksmith and Prisoner jobs

Kaltf4:
## About The Pull Request

Edits the new version of tgstation's civilian_bounty_trims.dm to add "if" cases for the Smith and Prisoner job types, as each job type check is hardcoded as "if" statements in the original file.

## Why It's Good For The Game
Once again allows Blacksmiths and Prisoners to get Personal Bounties, giving both roles a potential goal to do as non-antags to get some cash and fund the station.
On the backend, cuts down on runtime errors.

Risk: Blacksmithing bounties feel much easier and more lucrative than many other station jobs. **This does not address that issue,** so it's possible this fix does weird things to the station's economy.

## Proof Of Testing

<details>

Spawned as various jobs and inserted my ID into the Civilian Bounty Control Terminal, then clicked the New Bounty button. Two job-appropriate and one random bounty appeared on the resulting list.

Tested the following jobs to make sure bounties applied and no runtimes appeared:
- Blacksmith
- Prisoner
- Captain
- Nanotrasen Consultant
- Blueshield
- Barber
- Bitrunner
- Persistence Sanitation Technician
- Cardboard ID (doesn't slot into terminal)

<summary>Screenshots/Videos</summary>

Bubber_BountyFix1
Bubber_BountyFix2

</details>

## Changelog
:cl: Kaltren
fix: Blacksmiths and Prisoners can once again properly generate Personal Bounties.
/:cl:

PR #5970, opened 2026-07-19
Blacksmith bounty balance tweak

Kaltf4:
## About The Pull Request

~~Won't do anything player-visible without PR #5969 !!! (because the bounties are just plain broken)~~
The above PR is now merged, so no longer a concern.

In the Blacksmithing bounty list, removes the obsolete Cortical Borer Cage and ~~adds the Charging Holster instead, both as a way to point novice blacksmith players to "wait, I can make that?" and to give a high-material, high-cooperation, and low-time bounty to chase~~ replaces it with 20x Railroad Tracks (requiring two Rail Nail smithing completions).

In addition, rebalances the blacksmithing bounties a little, to make the time, clicks, and materials roughly similar ~~(aside from the Charging Holster, which removes the time requirement in favor of unusual collaboration with security)~~.

**Note: In the below table, Glass Globes are a point of reference for effort-value calculations, and are not used in blacksmithing bounties. All values at Legendary Smithing (or Production for the glass globe).**
```
Item        | OldVal | NewVal | Time | Mats
Glass Globe |   1000 |   1000 |  36s | 1 glass -- (Note: Not bounty, but exportable on cargo shuttle)
3x Cuffs    |   1600 |   3000 | 102s | 3 [any]
Staff       |   1000 |   1050 |  34s | 1 [any], 2 wood
3x Sword    |   4800 |   3400 | 102s | 3 [any], 6 wood
2x Armor    |   4000 |   2600 |  68s | 2 [any], 2 lthr
Katana      |   1200 |   1100 |  34s | 1 [any], 2 wood
Rlrd Trcks  |    N/A |   1950 |  50s | 2 [any], 4 wood
```

## Why It's Good For The Game
This prevents a scenario in which blacksmith players choose a Cortical Borer Cage from the bounty list, then find out they're unable to make them.
~~It also soft-advertises the existence of (and ability to make) Charging Holsters, and tells players that it requires Legendary in the skill to do so.~~ [EDIT: We ended up going with something other than Charging Holsters here.]
Finally, it brings the bounty values more in line with each other, with a mild nerf to the top end.

Per discussions with @ArrisFairburne , in the longer term, we probably want to do more with blacksmithing bounties, since the goal should be to have them interact with the crew and make things the crew wants, rather than just making a bunch of stuff to sell to CentCom. That said, I don't think we should wait for that to swap out the cortical borer cages, and swapping out the cortical borer cages is a good opportunity to make the bounties a _little_ less lucrative.

## Proof Of Testing
<details>

Manually verified the following bounty values by selecting bounties from the bounty terminal, creating the items via blacksmithing, and sending the items, then noting the values on the bounty cubes. PlrVal is the value shown to the player, while CbeVal is the value of the bounty cube as a whole. (This doesn't include the speedy return bonus.)

```
Item      | PlrVal | CbeVal |
3x Cuffs  |    900 |   3000 | Tested
Staff     |    315 |   1050 | Tested
3x Sword  |   1020 |   3400 | Tested
2x Armor  |    780 |   2600 | Tested
Katana    |    330 |   1100 | Tested
Rail Trx  |    585 |   1950 | Tested
```

image
image
image
image

1950 + 390 (20%) bonus = 2340
30% of 2340 = 702
2340 - 702 = 1638

</details>

## Changelog
:cl: Kaltren
fix: Cortical Borer Cage is no longer in the blacksmithing bounty list, as it's no longer smithable.
balance: 20x Railroad Tracks are now in the blacksmithing bounty list, worth 1950 total credits (585 player cut).
balance: 3x Handcuffs blacksmithing bounty is now worth 3000 total credits, up from 1600.
balance: 3x Sword blacksmithing bounty is now worth 3400 total credits, down from 4800.
balance: 2x Armor blacksmithing bounty is now worth 2600 total credits, down from 4000.
balance: Staff blacksmithing bounty is now worth 1050 total credits, up from 1000.
balance: Katana blacksmithing bounty is now worth 1100 total credits, down from 1200.
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/cargo/bounties/blacksmith.dm
+++ b/modular_zubbers/code/modules/cargo/bounties/blacksmith.dm
@@ -31,7 +31,7 @@ Todo: Refactor upstream code so that item deletion is decided at the bounty leve
 
 /datum/bounty/item/blacksmith/charging_holster
 	name = "Charging Holster"
-	description = "Our security team doesn't want to deal with finding rechargers. Send us a recharging holster. Only legendary blacksmiths need apply."
+	description = "Our security team doesn't want to deal with finding rechargers. Send us a charging holster. Only legendary blacksmiths need apply."
 	reward = CARGO_CRATE_VALUE * 10
 	required_count = 1
 	wanted_types = list(/obj/item/storage/belt/hip_holster/charging = TRUE)
--- a/modular_zubbers/code/modules/cargo/bounties/blacksmith.dm
+++ b/modular_zubbers/code/modules/cargo/bounties/blacksmith.dm
@@ -29,12 +29,12 @@ Todo: Refactor upstream code so that item deletion is decided at the bounty leve
 
 */
 
-/datum/bounty/item/blacksmith/charging_holster
-	name = "Charging Holster"
-	description = "Our security team doesn't want to deal with finding rechargers. Send us a charging holster. Only legendary blacksmiths need apply."
-	reward = CARGO_CRATE_VALUE * 10
-	required_count = 1
-	wanted_types = list(/obj/item/storage/belt/hip_holster/charging = TRUE)
+/datum/bounty/item/blacksmith/rail_track
+	name = "Railroad Tracks"
+	description = "The tram broke down again. Make us some railroad tracks to help us get around."
+	reward = CARGO_CRATE_VALUE * 9.75
+	required_count = 20
+	wanted_types = list(/obj/item/stack/rail_track = TRUE)
 
 /datum/bounty/item/blacksmith/cuffs
 	name = "Handcuffs"
--- a/modular_zubbers/code/modules/cargo/bounties/blacksmith.dm
+++ b/modular_zubbers/code/modules/cargo/bounties/blacksmith.dm
@@ -29,38 +29,38 @@ Todo: Refactor upstream code so that item deletion is decided at the bounty leve
 
 */
 
-/datum/bounty/item/blacksmith/cage
-	name = "Cortical Borer Cage"
-	description = "One of Nanotrasen's partner stations is undergoing a borer infestation and would like to capture some life specimen for research. Ship some off to CentCom right away."
-	reward = CARGO_CRATE_VALUE * 14
-	required_count = 2
-	wanted_types = list(/obj/item/cortical_cage = TRUE)
+/datum/bounty/item/blacksmith/charging_holster
+	name = "Charging Holster"
+	description = "Our security team doesn't want to deal with finding rechargers. Send us a recharging holster. Only legendary blacksmiths need apply."
+	reward = CARGO_CRATE_VALUE * 10
+	required_count = 1
+	wanted_types = list(/obj/item/storage/belt/hip_holster/charging = TRUE)
 
 /datum/bounty/item/blacksmith/cuffs
 	name = "Handcuffs"
 	description = "One of our wardens lost all the handcuffs again. Help restore order and ship some off for a reward."
-	reward = CARGO_CRATE_VALUE * 8
+	reward = CARGO_CRATE_VALUE * 15
 	required_count = 3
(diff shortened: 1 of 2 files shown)


modular_zubbers/code/modules/cargo/bounties/blacksmith.dm
```diff
@@ -29,38 +29,38 @@ Todo: Refactor upstream code so that item deletion is decided at the bounty leve
 
 */
 
-/datum/bounty/item/blacksmith/cage
-	name = "Cortical Borer Cage"
-	description = "One of Nanotrasen's partner stations is undergoing a borer infestation and would like to capture some life specimen for research. Ship some off to CentCom right away."
-	reward = CARGO_CRATE_VALUE * 14
-	required_count = 2
-	wanted_types = list(/obj/item/cortical_cage = TRUE)
+/datum/bounty/item/blacksmith/charging_holster
```
nikothedude: can you lock this behind the blacksmith being legendary before receiving the bounty?

nikothedude (changes requested):

nikothedude:
Good god I had no idea blacksmiths cuold make this much money. Also, that glass globe bounty is ridiculous. 1k credits... jesus.

The nerfs definitely look fine to me. Still probably a bit overtuned but more within reason.

modular_zubbers/code/modules/cargo/bounties/blacksmith.dm
nikothedude: also asking if this is feasible at all lol

Kaltf4: I did a bit more due diligence and I think the answer is "yes," just gotta figure out how.

There's a proc can_get() that's used elsewhere in bounties, see:
code\modules\cargo\bounties\science.dm
Block:
```
/datum/bounty/item/science/ref_anomaly/can_get(obj/O)
	var/anomaly_type = wanted_types[1]
	if(SSresearch.created_anomaly_types[anomaly_type] >= SSresearch.anomaly_hard_limit_by_type[anomaly_type])
		return FALSE
	return TRUE
```

So once I poke around a bit on getting the player's skill level, I might be able to use logic like this. (I'll probably find and use Arris' logic for whether the charging holster is craftable.)

nikothedude: see what you can do, would be rly nice

Kaltf4:
Yeah. I _can_ nerf it further if desired (the glassblowing file is in a skyrat modular file so it wouldn't touch tg code), but I figured I'd start with a light touch.

(Note for context: Glass globes are exportable, but aren't actually bounty targets. That means the player can export 'em at any time via the cargo shuttle, and if they wrap and tag the globes / their container with a barcode, they can get a cut from 0-50% of the total value, while the cargo budget gets the rest.)

modular_zubbers/code/modules/cargo/bounties/blacksmith.dm
Kaltf4: Okay, it's a heck of a lot harder than I thought; the bounty system grabs from the ID trim, so the user and their mind aren't actually involved in this context. I don't know how to reasonably bring 'em into the context without causing more problems. (The mind needs to be involved to check both skill level and traits, since smiths without the starting smithing skillchip are missing traits necessary to make the charging holster.)

The options I can see right now are to leave as-is (still with the warning text in the description) or just choose a different low-level item (e.g. railroad tracks).

Kaltf4: Went with the producer classic if "if it sucks to do, cut it." Changed it to railroad tracks instead, which are still an interesting item, take fewer strikes to complete, and are available at minimum level / no skillchip.

The benefit I cooked up of the charging holster in bounties being an interesting way to expose it to players, as well as encourages interaction with security, is still _neat_, but has enough tripping factors without the "can craft" check that its benefits probably don't outweigh the drawbacks.

nikothedude:
lgtm

PR #5971, opened 2026-07-19
Takes a knife to window airbags

lessthnthree:
## About The Pull Request

Removes Skyrat window airbags

## Why It's Good For The Game

It never sat right with me that the reinforced windows get an additional set of magic indestructable bubble wrap to eliminate any negative effects from them being broken.

## Changelog

:cl: LT3
del: Removed airbags from reinforced windows
/:cl:


AstrumLernaean:
This is very likely to cause a fair amount of additional grief for players, especially those not expecting the change; could you elaborate further on why this is good for the game? Also would you consider making a ping-less announcement of this change as it affects all stations and future window-interactions?

lessthnthree:
Repairing windows is part of engineering's job, it's good for the game because it brings back a piece of content to a job that doesn't have enough going for it. I don't think it's 'grief' for windows to be breaking.

Like everything, it will get a changelog if it proceeds to be merged yes.

SapphoQueer:
> This is very likely to cause a fair amount of additional grief for players, especially those not expecting the change; could you elaborate further on why this is good for the game?

For one, the airbags do not stop antagonists from breaching stuff in the slightest. If a traitor wants a room spaced, the only difference in what they have to do is to either alt-click the airbag before it goes off, or to poke it with a pen, or anything else really. This is an insignificant amount of resistance.

What this does impact is random events such as rods/meteors. In practice - depending on where it hits - heavily neutering their destructive potential. If it hits a window, the area it impacts will not be spaced. The airbag instantly deploys to effectively negate half the threat of any event which causes breaches. Something engineering has to fix as part of their job, and with how our engineering already tends to get bored when not doing a major project, I dont think limiting the work they do get to do for no reason outside of "well, the room could get spaced and that might annoy people within them" isnt a good tradeoff.

Players reacting to things that occur to the station is a good thing, be it having to evacuate a room or having to fix it, on the other hand, having it automatically resolve itself removes more than it adds.

Steamlined:
This WOULD incentivize the use of the fire locker inflatable walls that nobody uses, and I AM sick of players, particularly in brig cells, deciding to tear our all the window airbags around the brig cell and then throw them at security. I think this is a good change.

nikothedude:
orig pr was a gandalf pr with no justification, good revert

nikothedude:
but. please dont self tm with no approvals they exist for us as wellas player trust

edit: this was discussed internally - i was not a part of the discussion so i was unaware it happened so i apologize for misleading people into thinking this was a true self speedmerge
however, i still think publicly showing signs of this discussion on this pr via at least a flag is important for transparency

lessthnthree:
> but. please dont self tm with no approvals they exist for us as wellas player trust

why don't you maybe talk to me before making snippy comments, and then I can direct you to where we discussed this among maints in the maints channel where it has two other maintainer approvals

MintyPhresh:
Not gonna comment on that other part, but things that aapear to be self tms are just as eroding to player trust as things that actually are self tms -- even just a single comment saying it was discussed in maintainer chat would help

PR #5978, opened 2026-07-20
Oversized Buff/QOL

ibexgoetia:
## About The Pull Request

I've made it so that Oversized characters get a 0.25 debuff to their movement speed (as opposed to 0.50) as well as having removed their MOB_SIZE_LARGE trait, which made them unable to fit into any machine in the game, unless specifically coded in to ignore the mob size (to my knowledge it's the SAD, cryotube, cryopod and cyborg recharger).

I've also preserved the vampiric coffin functionality, made it so that it checks if the character using the coffin has the oversized trait, instead of checking whether they are too large for it.

## Why It's Good For The Game

Oversized is a very difficult trait. While it makes sense for there to be some drawbacks for being so large, I also believe that the negatives were a little too much, making the quirk very challenging to use and punitive. I've tweaked it a bit while at the same time preserving most of the traits that came with the quirk.

## Proof Of Testing

https://github.com/user-attachments/assets/1bc0fba1-2d34-4dba-834d-1ecf8dfcff50

The movement speed change.
Green is regular speed, orange is 0.25 (the speed they will have) and red is 0.50 (the current oversized speed).

## Changelog

:cl:
balance: Oversized characters are now faster, having their speed penalty lowered from 0.50 to 0.25.
balance: Oversized characters no longer have the MOB_SIZE_LARGE trait, allowing them to fit in machines and closets.
/:cl:


xPokee (commented):
If you're making it so oversized can fit into machinery by default you're probably good to remove `var/ignore_size` from `modular_zubbers/code/game/machinery/_machinery.dm` as I believe that is its only use case.

Rykka-Stormheart:
Hey, I'm not a maintainer, so feel free to ignore this, but!
I've always wanted to use Oversized, except for the max 200% size thing. Can you make a second quirk or smth that keeps JUST the MOB_SIZE_LARGE?

(Specifically bc not being able to fit in machines and stuff is pretty neat)

xPokee (approved):

PR #6001, opened 2026-07-26
Disables The Summon Ghosts Wizard Spell

xPokee:
## About The Pull Request
Does as the title would suggest. This **ONLY** disables it from being selected by wizards, it does not stop admins from running it.
## Why It's Good For The Game
While the idea of ghosts being visible is funny on paper, when ghosts start orbiting ERPers, invisible antagonists, or outright leading security to antagonists via pointing a path to them it becomes really unfun for the targets of the ghosts, and can be outright uncomfortable in the case of ghosts orbiting ERPers.
## Proof Of Testing
See below.
<details>

image

image

</details>

## Changelog
:cl:
del: Removed 'Summon Ghosts' from the wizard spellbook. It is still available via adminbus.
/:cl:


KateAnbolho:
1984, but understandable if people start being weird about it.

Odairu (approved):

Issue #5953, opened 2026-07-16
#5764 Seemingly broke species change from synth


nikothedude:
### Client Version:

516.1666

### Issue Summary:

#5764 

When testing polymorph, it was discovered that when the synth species deletes, it deletes the brain in preparation to transition the mob to having a new brain. This results in the mmi of the brain being deleted which deletes the brainmob which ghosts the mob.

### Round ID:

N/A

### Reproduction Steps:

1. Spawn as synth
2. Change to human
3. Get ghosted

nikothedude:
@Vanilla1040

PR #6003, opened 2026-07-26
Fixes Teshari Sudden Death Syndrome (TSDS)

NanoCats:
## About The Pull Request

Once upon a time I was doing some kink shit and all the sudden what was some innocent debauchery turned sour when I discovered that using the Polymorph NIFsoft on a Teshari would just kill them sometimes(???). No damage, no explanation, just a dead guy and everyone standing around confused. I went digging. It turns out there wasn't one bug, there were three, probably more given the nature of Polymorph being a mess and only one of them had anything to do with Teshari.

**1. Missing organs were coming back broken enough to kill you.**

When you transform, the game snapshots your injuries so you don't get a free Aheal out of it. Sensible. The snapshot works by asking each organ how hurt it is, and if the organ isn't there at all it records maximum damage instead, so you can't use the Self-Actualization Device as a free organ vending machine. Also sensible.

Except it can't tell the difference between "someone cut this out of me" and "my species doesn't come with one of these."

Proteans don't have lungs. So a protean turning into anything with lungs got a brand new set of lungs pre-installed at exactly the damage value that counts as failing. You'd finish the transformation, get told you can't breathe, and suffocate to death while wondering what you did wrong.

Now the snapshot checks species, asks whether it was ever supposed to have that organ. If it wasn't, that's not an injury and we don't carry it over. If it genuinely was and the organ was genuinely gone, you still get a wrecked organ, just one that stops a hair short of outright failing, so you need a doctor instead of a coffin. No worries there.  I almost wanted to make it display a warning when this happens, but this is just a simple bug fix I can't go all out every time. 

**2. Runtime errors everywhere.**

The Unblinking quirk was turning blinking back **on** when it got removed, unconditionally. Trouble is, some eyes never had eyelids to begin with (Teshari, insects, serpentids), and the eyelids only get built when the eyes are created. So flipping the switch on afterwards left the blinking code reaching for eyelids that were never there. Every transformation involving one of those species vomited runtimes into the log.

It now puts the setting back the way it found it.

## Why It's Good For The Game

People were dying. Not "took some damage", not "bumps and bruises", just "NIF that lets you instantly kill Teshari", with no feedback explaining why, from a purely cosmetic feature that requires both parties to consent. That's about the worst possible outcome for a thing whose entire purpose is letting people play a different character for a while.

I want to be upfront that this is a first pass. The Polymorph NIFsoft has more problems than this. There's a synth brain that doesn't get swapped back when you stop being a synth, which leaves an open-skull sprite hovering over your head. There's a quirk system warning that fires every single transformation. There's at least one case where I ended up with no head, which I intend to go back and look at later. 

But none of those kill anybody, and this stuff does, so it goes first.

## Testing

Tested offline with the Body Morpher quirk, which grants the same ability and can target the caster, so I didn't need a testin partner. Hopped between a Teshari, a protean, a synth, and back again, in several orders.

- Protean into Teshari now finishes with working lungs and no suffocation.
- Teshari transformations no longer throw runtimes.
- Deliberately removing an organ before transforming still leaves you with a badly damaged replacement, or it should.

Nothing to take a picture of here, what do you want, a picture of a perfectly alive Teshari?

## Changelog

:cl:
fix: Shapeshifting no longer breaks organs that your old species simply never had, so proteans stop suffocating if Tfing into humans or similar
fix: Cleaned up a pile of runtime errors thrown by shapeshifting.
/:cl:

### DIFF
--- a/code/controllers/subsystem/ambience.dm
+++ b/code/controllers/subsystem/ambience.dm
@@ -122,6 +122,10 @@ SUBSYSTEM_DEF(ambience)
 		return
 
 	var/area/my_area = get_area(src)
+	// BUBBER EDIT ADDITION - a mob sitting in nullspace (a brainmob inside an organ being deleted) has no area to buzz in
+	if(isnull(my_area))
+		return
+	// BUBBER EDIT ADDITION END
 	var/sound_to_use = my_area.ambient_buzz
 	var/volume_modifier = client.prefs.read_preference(/datum/preference/numeric/volume/sound_ship_ambience_volume)
 
--- a/code/controllers/subsystem/ambience.dm
+++ b/code/controllers/subsystem/ambience.dm
@@ -122,10 +122,6 @@ SUBSYSTEM_DEF(ambience)
 		return
 
 	var/area/my_area = get_area(src)
-	// BUBBER EDIT ADDITION - a mob sitting in nullspace (a brainmob inside an organ being deleted) has no area to buzz in
-	if(isnull(my_area))
-		return
-	// BUBBER EDIT ADDITION END
 	var/sound_to_use = my_area.ambient_buzz
 	var/volume_modifier = client.prefs.read_preference(/datum/preference/numeric/volume/sound_ship_ambience_volume)
 
--- a/code/modules/surgery/organs/internal/eyes/_eyes.dm
+++ b/code/modules/surgery/organs/internal/eyes/_eyes.dm
@@ -513,11 +513,6 @@
 	if (!blink_animation || HAS_TRAIT(parent, TRAIT_PREVENT_BLINKING))
 		return
 
-	// BUBBER EDIT ADDITION START - eyes that spawn without eyelids can have blink_animation flipped on later, don't blink with nothing
-	if (isnull(eyelid_left) || isnull(eyelid_right))
-		return
-	// BUBBER EDIT ADDITION END
-
 	eyelid_left.color = eyelid_color
 	eyelid_right.color = eyelid_color
 	eyelid_left.render_target = "*[REF(parent)]_eyelid_left"
--- a/code/modules/surgery/organs/internal/eyes/_eyes.dm
+++ b/code/modules/surgery/organs/internal/eyes/_eyes.dm
@@ -513,6 +513,11 @@
 	if (!blink_animation || HAS_TRAIT(parent, TRAIT_PREVENT_BLINKING))
 		return
 
+	// BUBBER EDIT ADDITION START - eyes that spawn without eyelids can have blink_animation flipped on later, don't blink with nothing
+	if (isnull(eyelid_left) || isnull(eyelid_right))
+		return
+	// BUBBER EDIT ADDITION END
+
 	eyelid_left.color = eyelid_color
 	eyelid_right.color = eyelid_color
 	eyelid_left.render_target = "*[REF(parent)]_eyelid_left"
--- a/code/modules/surgery/organs/organ_movement.dm
+++ b/code/modules/surgery/organs/organ_movement.dm
@@ -68,7 +68,12 @@
 
 	var/obj/item/organ/replaced = receiver.get_organ_slot(slot)
 	if(replaced)
-		replaced.Remove(receiver, special = TRUE)
+		// BUBBER EDIT ADDITION START - an organ we are about to delete must not hand anything off on the way out, there is nothing left to hand it to
+		var/replaced_movement_flags = movement_flags
+		if(movement_flags & DELETE_IF_REPLACED)
+			replaced_movement_flags |= NO_ID_TRANSFER
+		replaced.Remove(receiver, special = TRUE, movement_flags = replaced_movement_flags)
+		// BUBBER EDIT ADDITION END - original: replaced.Remove(receiver, special = TRUE)
 		if(movement_flags & DELETE_IF_REPLACED)
 			qdel(replaced)
 		else
--- a/code/modules/surgery/organs/organ_movement.dm
+++ b/code/modules/surgery/organs/organ_movement.dm
@@ -68,12 +68,7 @@
 
 	var/obj/item/organ/replaced = receiver.get_organ_slot(slot)
 	if(replaced)
-		// BUBBER EDIT ADDITION START - an organ we are about to delete must not hand anything off on the way out, there is nothing left to hand it to
-		var/replaced_movement_flags = movement_flags
-		if(movement_flags & DELETE_IF_REPLACED)
-			replaced_movement_flags |= NO_ID_TRANSFER
-		replaced.Remove(receiver, special = TRUE, movement_flags = replaced_movement_flags)
-		// BUBBER EDIT ADDITION END - original: replaced.Remove(receiver, special = TRUE)
+		replaced.Remove(receiver, special = TRUE)
 		if(movement_flags & DELETE_IF_REPLACED)
 			qdel(replaced)
 		else
--- a/modular_skyrat/master_files/code/datums/components/damage_tracker.dm
+++ b/modular_skyrat/master_files/code/datums/components/damage_tracker.dm
@@ -151,7 +151,7 @@
 
 	return ..()
 
-/// Returns the damage of the `organ_to_check`, or a sentinel describing why the organ isn't there.
+/// Returns the damage of the `organ_to_check`, or a special value describing why the organ isn't there.
 /mob/living/carbon/human/proc/check_organ_damage(obj/item/organ/organ_to_check)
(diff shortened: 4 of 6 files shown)


CB35SSC:
Your contribution to the medical community has been noted.

KateAnbolho:
VeyMed's very concerned

NanoCats:
> VeyMed's very concerned

Why do they have something against Teshari?

KateAnbolho:
> Why do they have something against Teshari?

While that would be really funny, it's more that their systems have killed several people by giving them organs that don't work

NanoCats:
Small update, turns out that my unintended victim died because they were a *skeleton*, I couldn't tell because they were decked out in mining gear. This PR fixes it anyways by virtue of the way it was programmed.

nikothedude:
Closes https://github.com/Bubberstation/Bubberstation/issues/5953?

code/controllers/subsystem/ambience.dm
```diff
@@ -122,6 +122,10 @@ SUBSYSTEM_DEF(ambience)
 		return
 
 	var/area/my_area = get_area(src)
+	// BUBBER EDIT ADDITION - a mob sitting in nullspace (a brainmob inside an organ being deleted) has no area to buzz in
+	if(isnull(my_area))
+		return
+	// BUBBER EDIT ADDITION END
```
Maia-J: This hides the underlying issue, the brainmob should be in the contents for a lot more reasons than ambiance.

code/modules/surgery/organs/internal/eyes/_eyes.dm
```diff
@@ -513,6 +513,11 @@
 	if (!blink_animation || HAS_TRAIT(parent, TRAIT_PREVENT_BLINKING))
 		return
 
+	// BUBBER EDIT ADDITION START - eyes that spawn without eyelids can have blink_animation flipped on later, don't blink with nothing
+	if (isnull(eyelid_left) || isnull(eyelid_right))
+		return
+	// BUBBER EDIT ADDITION END
```
Maia-J: This also feels like hiding an underlying issue with eyelidless eyes having blinking toggled somehow, which should not happen in the first place

modular_skyrat/master_files/code/datums/components/damage_tracker.dm
```diff
@@ -146,10 +151,32 @@
...
+/// Returns the damage of the `organ_to_check`, or a sentinel describing why the organ isn't there.
+/mob/living/carbon/human/proc/check_organ_damage(obj/item/organ/organ_to_check, organ_slot)
 	var/obj/item/organ/organ_to_track = get_organ_by_type(organ_to_check)
-	if(!organ_to_track)
-		return 100 //If the organ is missing, return max damage. we have this here so that if the SAD replaces an organ, it's broken.
+	if(organ_to_track)
+		return organ_to_track.damage
+
+	// a species that never grows this organ isn't hurt, it's just built that way, so there's no injury to carry across
+	if(organ_slot && isnull(dna?.species?.get_mutant_organ_type_for_slot(organ_slot)))
```
Maia-J: you could just use the type::slot instead of passing the organ_slot, makes it easier to maintain into the future

modular_skyrat/master_files/code/modules/client/preferences/brain.dm
```diff
@@ -28,7 +28,12 @@
 	new_brain.modular_persistence?.owner = new_brain
 	old_brain.modular_persistence = null
 
-	new_brain.Insert(target, movement_flags = DELETE_IF_REPLACED)
+	// pull the old brain ourselves with NO_ID_TRANSFER instead of letting DELETE_IF_REPLACED do it. otherwise the brain
+	// hands the occupant off to its brainmob on the way out and we immediately delete the brainmob underneath them
+	old_brain.Remove(target, special = TRUE, movement_flags = NO_ID_TRANSFER)
+	qdel(old_brain)
+
+	new_brain.Insert(target)
```
Maia-J: This also feels like an issue with a certain brain type needing special behaviour on being replaced via Insert(), this just hides the underlying issue until some other brain replacement code triggers the issue (like `/mob/living/carbon/human/proc/psykerize()` for example. 
Also may have been caused by missing the special flag, which is meant to signal that what we're doing is not going to boot us out in the first place.

Maia-J (changes requested):

code/modules/surgery/organs/organ_movement.dm
```diff
@@ -68,7 +68,12 @@
 
 	var/obj/item/organ/replaced = receiver.get_organ_slot(slot)
 	if(replaced)
-		replaced.Remove(receiver, special = TRUE)
+		// BUBBER EDIT ADDITION START - an organ we are about to delete must not hand anything off on the way out, there is nothing left to hand it to
+		var/replaced_movement_flags = movement_flags
+		if(movement_flags & DELETE_IF_REPLACED)
+			replaced_movement_flags |= NO_ID_TRANSFER
+		replaced.Remove(receiver, special = TRUE, movement_flags = replaced_movement_flags)
+		// BUBBER EDIT ADDITION END - original: replaced.Remove(receiver, special = TRUE)
```
Maia-J:
```suggestion
		replaced.Remove(receiver, special = TRUE)
```
This still doesn't fix the issue that was happening, see #6051 (I was pawed at to fix it, it required quite a lot of stack tracing and debugging to figure out)

No need to try and fix this specific issue now, as #6051 got it. Other stuff is still great work, just needs comment cleanup

modular_skyrat/master_files/code/datums/components/damage_tracker.dm
```diff
@@ -1,3 +1,8 @@
+/// Stored instead of a damage value when the organ was absent because the species simply doesn't have one.
+#define ORGAN_ABSENCE_SPECIES_NORMAL -1
+/// Stored instead of a damage value when the organ should have been there and wasn't.
+#define ORGAN_ABSENCE_INJURY -2
```
Maia-J:
```suggestion
/// The organ was absent because the species simply doesn't have one.
#define ORGAN_ABSENCE_SPECIES_NORMAL -1
/// The organ should have been there and wasn't.
#define ORGAN_ABSENCE_INJURY -2
```

comment cleanup

modular_skyrat/master_files/code/datums/components/damage_tracker.dm
```diff
@@ -146,10 +151,32 @@
 
 	return ..()
 
-/// Returns the damage of the `organ_to_check`, if the organ isn't there, the proc returns `100`.
+/// Returns the damage of the `organ_to_check`, or a sentinel describing why the organ isn't there.
```
Maia-J:
```suggestion
/// Returns the damage of the `organ_to_check`, or a special value describing why the organ isn't there.
```
Ditto, easier to grasp wording, plus it technically isn't a sentinel (Nerd emoji)

modular_skyrat/master_files/code/datums/components/damage_tracker.dm
```diff
@@ -146,10 +151,32 @@
...
+	if(organ_to_track)
+		return organ_to_track.damage
+
+	// a species that never grows this organ isn't hurt, it's just built that way, so there's no injury to carry across
+	if(isnull(dna?.species?.get_mutant_organ_type_for_slot(organ_to_check::slot)))
+		return ORGAN_ABSENCE_SPECIES_NORMAL
+
+	return ORGAN_ABSENCE_INJURY
+
+/// Puts a tracked damage value back onto an organ, translating the absence sentinels from [check_organ_damage].
```
Maia-J:
```suggestion
/// Puts a tracked damage value back onto an organ
```
Ditto

modular_skyrat/master_files/code/datums/components/damage_tracker.dm
```diff
@@ -146,10 +151,32 @@
...
+/mob/living/carbon/human/proc/restore_tracked_organ_damage(organ_slot, stored_damage)
+	if(stored_damage == ORGAN_ABSENCE_SPECIES_NORMAL)
+		return
+
+	var/obj/item/organ/organ_to_damage = get_organ_slot(organ_slot)
+	if(!organ_to_damage)
+		return
+
+	// the organ was genuinely gone, so the replacement arrives wrecked rather than free, but stops short of outright failing
+	if(stored_damage == ORGAN_ABSENCE_INJURY)
```
Maia-J, line 175:
```suggestion
		return

	if(stored_damage == ORGAN_ABSENCE_INJURY)
```
This code is self documenting

Maia-J (changes requested):
Last bit of cleanup

Maia-J (approved):

PR #6028, opened 2026-07-29
Dusting Sickness fixes

ibexgoetia:
## About The Pull Request

I have fixed a couple issues with the quirk as is; firstly, plasmamen and abductors can no longer pick this quirk, to not instantly kill them upon spawning (as they technically don't have blood). Secondly, being dusted now makes you drop your things the moment before, instead of dusting them with you.

## Why It's Good For The Game

The species not being restricted seemed like an oversight.

Making it so that the quirk doesn't delete your items makes the quirk a lot less of a pain for heads of staff specifically, as it could lead to the loss of critical and/or unique items.
## Proof Of Testing

Here's Cpt. Pinkmann (any correlation to the Breaking Bad character is purely coincidental) being syphoned of their precious blood. Don't worry, they're fine.

https://github.com/user-attachments/assets/8538a13c-3853-4956-b6c1-f4b1d843d664

</details>

## Changelog
:cl:
balance: Being dusted via Dusting Sickness now makes you drop your equipment.
fix: Abductors and Plasmamen are now blacklisted from picking the Dusting Sickness quirk.
/:cl:

### DIFF
--- a/modular_zubbers/code/datums/quirks/negative_quirks/bloodloss_dusting.dm
+++ b/modular_zubbers/code/datums/quirks/negative_quirks/bloodloss_dusting.dm
@@ -2,7 +2,7 @@
 	name = "Dusting Sickness"
 	desc = "If you run out of blood to the point where a normal person would die, you turn to dust."
 	value = -8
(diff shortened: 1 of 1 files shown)


xPokee (commented):
Make sure to add that you blacklisted those two species to your changelog, but otherwise this looks good.

PR #6030, opened 2026-07-29
Storyteller config cleanup and finer config control

Phantastic-Swan:
## About The Pull Request

I was very annoyed at there being so many vars and configs on the storyteller that do absolutely fucking nothing so I removed them. While I was at it, I decided to fix some configs having wrong names, and add config control over the storyteller types, so that they can be controlled more finely via configs, rather than through code.

## Why It's Good For The Game

Removes unnecessary configs that literally do not affect anything, so that admins don't get confused by tweaking values and not having anything change. Also makes it easier to control the storytellers via configs

## Proof Of Testing

It compiled. I checked if the configs properly multiply the thresholds and they do.

## Changelog

:cl: Swan
config: Added configs for finer control over the various storytellers, and renamed wrongly named configs. Removed configs which weren't tied to anything.
/:cl:


Maia-J (approved):
Someone needs to update the config, but it looks good codewise

PR #6032, opened 2026-07-29
[BOUNTY] Removes the pairing mechanic from sec dept rolls

Maia-J:
## About The Pull Request

This completely does away with the pairing system for security department rolls. The pairing system always forced you together with someone if they were going to be alone as a department sec.

What this means is sec officers will be distributed among all the departments, with the system trying to put preference first, but if there's two going for the same one, one will inevitably lose out and be given to another department.

- Benefits are that more departments get their secoffs.
- Disadvantage is that, well, you're still not guaranteed to get your preferred dept, but it at least makes sense now and won't just force you together with someone.

This PR has an alternative version, this is Option 1 and my personally preferred and endorsed option.

closes #6033 (alternative PR)

## Why It's Good For The Game

It makes the system less confusing by no longer forcing you in with someone even though your preferred department was empty. Also makes it that more departments get their secoffs.

## Proof Of Testing

This works in theory, and the procs do react as they should in testing, but I can't simply summon 7 secoffs to my test server so I need this TMd to be sure

## Changelog

:cl:
balance: Sec officers no longer get forced into a department with another officer when an empty department is available, making the dept preference system make sense.
/:cl:

### DIFF
--- a/code/modules/jobs/job_types/security_officer.dm
+++ b/code/modules/jobs/job_types/security_officer.dm
@@ -320,7 +320,7 @@ GLOBAL_LIST_EMPTY(security_officer_distribution)
 	 * This is the function that is responsible for taking the list of preferences,
 	 * and spitting out what to put them in.
 	 *
-	 * However, it should, wherever possible, prevent solo departments.
+	 * However, it should, wherever possible, prevent solo departments. // BUBBER EDIT: This was patched out, allowing solo departments
 	 * That means that if there's one medical officer, and one engineering officer,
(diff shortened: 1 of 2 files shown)


MajManatee:
The pairing does feel like a really terrible way to have done it in the first place.

Maia-J:
> The pairing does feel like a really terrible way to have done it in the first place.

It does make sense for TG where you're kinda guaranteed to get 1-2 antags in every department but other than that, yeah

PR #6034, opened 2026-07-30
Adds a new quirk, Holographic Nature

Odairu:
## About The Pull Request
Other servers such as Nova/Oculis have a species that has the hologram effect overlayed on them, which is cool, however as a teshari main myself, I cannot be these things, this makes this affect takeable on everyone

This offers no mechanical changes to a character, its purely visual, the only 'con' to taking this is you will no longer be able to hide in dark areas as you are lit up.

This also makes use of the Color_override variable on makeHologram()

## Why It's Good For The Game
Adds more customization options for a character, without locking it to only specific species so that anyone can enjoy it

## Proof Of Testing
image
image

<details>

</details>

## Changelog

:cl:
add: holographic nature quirk
/:cl:


SomeCabbage:
let's watch as someone gets misidentified as a glitch and gets evaporated

YakumoChen:
At least made sense with a dedicated species / synth but I can't help but feel this ruins my immersion to deal with normal, regular carbons that look like holograms. I find it hard to imagine doing holographic surgery on a holographic engineer, doing a holographic coronary bypass on their holographic heart, questioning if I'm in a simulation while I do a holographic revival. Are these people even real? Are they any more alive than the simplemob russians in void maint bar?

MrGloopy:
...Yeah honestly I have to agree with the above point. Why would they have blood? A beating organic heart? If they're supposed to have some sort of emitter then that needs to be reflected in gameplay, like the doppler species has.

I can't imagine it'd be easy to quirkify the species but that at least makes sense the way it is with an actual emitter and their own custom blood, along with the forcing all internal organs to be synthetic (from what I can see).

PR #6041, opened 2026-07-31
Makes gun cases use their open icon state when appropriate. 

NanoCats:
## About The Pull Request

Gun cases have had an "open" sprite sitting in the icon file for years and they're basically vestigial, they only show when you use the case in hand, which, why the hell would you? Consequently, the sprite has been quietly gathering dust.

Now the lid does what a lid ought to do. Open a case and it opens. Rummage away. Close it and it shuts again. You know, like a real case. Traditionally you need to open a box to look inside it. 

As a proactive step, I've rigged it so that it works out for itself whether a case has an open sprite to show. If it does, the case uses this state. If it does not, nothing changes for it at all, nothing changes for TG cases, but any that operate on the newer skyrat spec should work automatically. Any new cases added in the right way can hook this for free. 

## Why It's Good For The Game

Free polish. The art was already on the cutting room floor, it just needed somebody to plug it in. It also reads nicely at a glance: an open case means the container is being accessed.

Artur asked me to do this also.

## Proof Of Testing

https://github.com/user-attachments/assets/49cfb05c-7bd4-41d4-aaf5-efa4df85be38

## Changelog

:cl: Aeri
fix: Gun cases now utilize their existing "open" sprites where appropriate. 
/:cl:


modular_zubbers/code/game/objects/items/storage/toolboxes/guncase_lid.dm
xPokee, line 1: We don't use master_files and we should probably aim to get rid of it. Please put your file in somewhere like `modular_zubbers/code/game/objects/items/storage/toolboxes/guncase_lid.dm`

xPokee (changes requested):

modular_zubbers/code/game/objects/items/storage/toolboxes/guncase_lid.dm
NanoCats: Good to know, in my defense I didn't know I wasn't allowed to do that... Moved as you wanted.

PR #6042, opened 2026-07-31
The Great dna.features["body_size"] Killing

xPokee:
## About The Pull Request
Kills every instance of the body_size feature and changes everything to use current_size, RESIZE_DEFAULT_SIZE, and update_transform.

Also cleans up oversized code big time. The only player-facing change from this is that the oversized specific organs are now gone.
## Why It's Good For The Game
The oversized code was a mess and having snowflakey features when the proc and such already exists caused me pain.
## Proof Of Testing
See below.
<details>

https://github.com/user-attachments/assets/017fb9e3-aef9-48ce-a64e-b24a99ae49eb

https://github.com/user-attachments/assets/4cdaa48c-0237-4447-a943-cabb222be25b

https://github.com/user-attachments/assets/3bab0b0c-5098-4d51-b77e-994e21bcf87e

</details>

## Changelog
:cl:
del: Removed the oversized specific organs.
/:cl:


Maia-J (approved):
Didn't notice anything off

lessthnthree (approved):

Phantastic-Swan:
This broke the size collar.

First of all, `mob_size` on `/mob/living/carbon/human` doesn't actually refer to the scale of the mob. It seems to be essentially a categorization of mobs into sizes for certain functionalities. That value is set by default to 2 on humans.
image
image

Second of all, `update_transform(resize = RESIZE_DEFAULT_SIZE)` doesn't apply `resize` as the size of the mob, it MULTIPLIES the current size of the mob by `resize`. Ergo, if we have a mob with a scale of 1, their `original_size` is set to 2, and then every time the size collar tries to apply the `original_size` it doubles the scale of the mob.

I will also use this opportunity to salt at the removal of these 3 defines:
```
#define BODY_SIZE_NORMAL 1.00
#define BODY_SIZE_MAX 1.5
#define BODY_SIZE_MIN 0.8
```
was this seriously necessary? Could we really not have kept them as a convenient define of the maximum and minimum sizes we can give our characters in the character creator? Especially since you essentially replaced those with magic values. I am personally of the opinion that
```
minimum = BODY_SIZE_MIN
maximum = BODY_SIZE_MAX
```
more accurately conveys the intent than
```
minimum = RESIZE_DEFAULT_SIZE * 0.8
maximum = RESIZE_DEFAULT_SIZE * 1.5
```
especially when the defines COULD have been defined as `RESIZE_DEFAULT_SIZE * 0.8/1.5`. Which would give us the same result, but it would be easier to read.

Also also, I do think having the original size of the mob stored in their dna is a good idea. This makes it possible for the size collar to always be certain that the size value it is looking at is the original size value of the mob, rather than a value that was possibly modified by other things.

xPokee:
> This broke the size collar.

Then make a bug report.

> First of all, mob_size on /mob/living/carbon/human doesn't actually refer to the scale of the mob. It seems to be essentially a categorization of mobs into sizes for certain functionalities. That value is set by default to 2 on humans.

`mob_size` being used instead of `current_size` was a mistake on my end.

Phantastic-Swan:
I am already making a PR to fix it.

PR #6043, opened 2026-08-01
Adds Piss system

NanoCats:
## About The Pull Request

That's right. The brand new Pre-Initialization Skin System!

image

A lot of items in this codebase have alternate skins. The mech pilot suit is the obvious one and also the most important one for me particularly: red, white, blue, black. Normally, you get at them by alt-clicking the item in game. This is kind of a pain in the ass if you happen to have multiple items which require reskinning in your loadout (Every damn shift!!!), and I saw the potential to make things just a little more frictionless. When I started investigating the issue, I saw it:  the loadout menu already had a skin picker for this. The sprite buttons, the "Reskinnable" badge, the saving, the applying at spawn. 

Everything I needed, all of it, already built, sitting right there next to Rename and the GAGS colors. The problem is that the loadout menu only switched this function on if a loadout entry declared `reskin_datum` by hand, and exactly one item in the entire codebase ever did that: the pride pin.

What this PR does is make the functionality automatic by default. Any item that already has a `reskinable_item` component now offers its skins in the loadout config screen automatically. Items that can be reskinned in this way have a color swatch icon on them. 

### Opting out

If, for some unknowable reason you don't want an item you add to be reskinnable, I really cannot fathom why you would want this. I thought of that too and `LOADOUT_FLAG_BLOCK_RESKIN` can be appended to your item to turn detection off for a single loadout entry. Two things needed it because they're made in a way where the game thinks they have alternate sprites, but don't actually. 

1. NIFSoft datadisks inherit the colored floppy disk skins. Picking one made the disk invisible. This issue is fixed as collateral of this PR. 
2. The black turtleneck donator variant tries to use the "tacticool" skins, but it cannot actually do that. So it fails, I took the easy way out because I didn't want to devote any additional time to solving the problem. 

I checked other items for this problem, but only those two are loadout items, and I already solved the problem, this is just something to bear in mind if you happen to be adding reskinnable loadout items spcifically. 

## Why It's Good For The Game

Getting a character set up at round start is annoying. The game already tends to spawn our colorful blorbos wearing a whole mess of things they immediately trash because they don't actually want to wear them. One small solution is this, reducing the need to alt-click half your loadout to put it back the way you like it (Dtramatization). There is no decision in that, and no gameplay in it either. It is just a small tedium task. 

The other half of it is that this feature was already made. Someone built the whole thing and connected it to one item. This PR just enables vestigial functionality. 

## Proof Of Testing

Compiled and run locally. Loaded myself up with as many reskinnable items as possible before joining, and it worked just fine. 

image
image

## Changelog

:cl: Aeri
qol: You can set an item's alternate skin in the loadout menu now, in the same place you rename it or pick its colors. Anything you could alt-click to reskin in game will offer its skins up front.
/:cl:

### DIFF
--- a/code/__DEFINES/preferences.dm
+++ b/code/__DEFINES/preferences.dm
@@ -166,9 +166,11 @@
 #define LOADOUT_FLAG_GREYSCALING_ALLOWED (1<<2)
 /// Allows the item to be renamed by the player.
 #define LOADOUT_FLAG_ALLOW_NAMING (1<<3)
+// BUBBER EDIT ADDITION START - Automatic reskin detection
 /// Prevents automatic detection of reskin support on this loadout item.
 /// Use when the item inherits a reskin component whose skins don't suit it, or when its skins are deliberately gated in game.
 #define LOADOUT_FLAG_BLOCK_RESKIN (1<<4)
+// BUBBER EDIT ADDITION END
 
 // Loadout item info keys
 // Changing these will break existing loadouts
--- a/code/__DEFINES/preferences.dm
+++ b/code/__DEFINES/preferences.dm
@@ -166,6 +166,9 @@
 #define LOADOUT_FLAG_GREYSCALING_ALLOWED (1<<2)
 /// Allows the item to be renamed by the player.
 #define LOADOUT_FLAG_ALLOW_NAMING (1<<3)
+/// Prevents automatic detection of reskin support on this loadout item.
+/// Use when the item inherits a reskin component whose skins don't suit it, or when its skins are deliberately gated in game.
+#define LOADOUT_FLAG_BLOCK_RESKIN (1<<4)
 
 // Loadout item info keys
 // Changing these will break existing loadouts
--- a/code/__DEFINES/preferences.dm
+++ b/code/__DEFINES/preferences.dm
@@ -166,9 +166,6 @@
 #define LOADOUT_FLAG_GREYSCALING_ALLOWED (1<<2)
 /// Allows the item to be renamed by the player.
 #define LOADOUT_FLAG_ALLOW_NAMING (1<<3)
-/// Prevents automatic detection of reskin support on this loadout item.
-/// Use when the item inherits a reskin component whose skins don't suit it, or when its skins are deliberately gated in game.
-#define LOADOUT_FLAG_BLOCK_RESKIN (1<<4)
 
 // Loadout item info keys
 // Changing these will break existing loadouts
--- a/code/modules/loadout/loadout_items.dm
+++ b/code/modules/loadout/loadout_items.dm
@@ -55,9 +55,6 @@ GLOBAL_LIST_INIT(all_loadout_categories, init_loadout_categories())
 	/// Base typepath to what reskin datum this item can use to reskin into
 	/// Doesn't verify that the item_path actually has these reskins
 	var/datum/atom_skin/reskin_datum
-	// BUBBER EDIT ADDITION - Tracks whether reskin auto detection has already run for this datum.
-	var/reskin_datum_checked = FALSE
-	// BUBBER EDIT ADDITION END
 	/// A list of greyscale colors that are used for items that have greyscale support, but don't allow full customization.
 	/// This is an assoc list of /datum/job_department -> colors, or /datum/job -> colors, allowing for preset colors based on player chosen job.
 	/// Jobs are prioritized over departments.
@@ -336,7 +333,6 @@ GLOBAL_LIST_INIT(all_loadout_categories, init_loadout_categories())
 /datum/loadout_item/proc/on_equip_item(obj/item/equipped_item, list/item_details, mob/living/carbon/human/equipper, datum/outfit/outfit, visuals_only = FALSE)
 	if(isnull(equipped_item))
 		return NONE
-	try_detect_reskin() // BUBBER EDIT ADDITION - Populate reskin_datum before applying a saved skin
 
 	if(!visuals_only)
 		ADD_TRAIT(equipped_item, TRAIT_ITEM_OBJECTIVE_BLOCKED, TRAIT_SOURCE_LOADOUT)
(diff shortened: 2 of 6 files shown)


KateAnbolho:
this PR is a title of all time

muchatras:
I despise the ALT-click reskin so much, this is an awesome addition

NanoCats:
> I despise the ALT-click reskin so much, this is an awesome addition

Yeah I don't like it either, I got annoyed with having to do it every single shift, so there you go.

code/__DEFINES/preferences.dm
```diff
@@ -166,6 +166,9 @@
 #define LOADOUT_FLAG_GREYSCALING_ALLOWED (1<<2)
 /// Allows the item to be renamed by the player.
 #define LOADOUT_FLAG_ALLOW_NAMING (1<<3)
+/// Prevents automatic detection of reskin support on this loadout item.
+/// Use when the item inherits a reskin component whose skins don't suit it, or when its skins are deliberately gated in game.
+#define LOADOUT_FLAG_BLOCK_RESKIN (1<<4)
```
Maia-J, line 174: Missing bubber edit comment

code/modules/loadout/loadout_items.dm
```diff
@@ -380,15 +395,71 @@ GLOBAL_LIST_INIT(all_loadout_categories, init_loadout_categories())
...
 
+// BUBBER EDIT ADDITION START - Automatic reskin detection
+/**
+ * Checks item_path for a reskinable_item component and adopts its skin type, so that any item
+ * which supports alt click reskinning in game also offers skin selection in the loadout.
+ *
+ * This cannot run in New(). Loadout datums are built by GLOBAL_LIST_INIT, which fires before
+ * SSatoms initialises, and at that point atom/New() deliberately skips Initialize(). No
+ * Initialize means no setup_reskins(), so the component would not exist yet to be found.
+ * Running on first use instead means SSatoms has long since finished.
```
Maia-J: Datums aren't atoms which is why they don't run `Initialize()`. They don't even have the proc for it, that one's only for atom and its children.

Thus, this whole code comment is incorrect.

code/modules/loadout/loadout_items.dm
```diff
@@ -380,15 +395,71 @@ GLOBAL_LIST_INIT(all_loadout_categories, init_loadout_categories())
...
+	if(reskin_datum_checked || reskin_datum || (loadout_flags & LOADOUT_FLAG_BLOCK_RESKIN))
+		return
+	reskin_datum_checked = TRUE
+	if(!ispath(item_path, /obj/item))
+		return
+	// Mob holders carry a live creature. Building one spawns the creature inside it, and deleting it
+	// then tries to drop that creature on the floor. There is no floor here, so it throws. They never
+	// have skins anyway, so leave them alone.
+	if(ispath(item_path, /obj/item/mob_holder))
+		return
```
Maia-J: If mob holders do that then this is what's behind the Mr. Fluff runtime that I saw once or twice in the runtime log. I assume they're not getting force deleted correctly by the loadout system?

Either way, comment isn't really needed. An issue report into this would be nice though

```suggestion
	if(ispath(item_path, /obj/item/mob_holder))
		return
```

code/modules/loadout/loadout_items.dm
```diff
@@ -380,15 +395,71 @@ GLOBAL_LIST_INIT(all_loadout_categories, init_loadout_categories())
...
+		reskin_datum = reskin_comp.get_base_reskin_type()
+	qdel(probe)
+// BUBBER EDIT ADDITION END
+
 /**
  * Returns a formatted list of data for this loadout item.
  */
 /datum/loadout_item/proc/to_ui_data() as /list
 	SHOULD_CALL_PARENT(TRUE)
+	try_detect_reskin() // BUBBER EDIT ADDITION - Populate reskin_datum before the UI data is built
```
Maia-J, line 453: Since `to_ui_data()` runs for (seemingly every) loadout datum singleton just once, right upon them actually being initialized for showing via the preferences system, you might be able to run this just here and avoid the whole extra variable and init order issue. `on_equip_item` relies on preferences already running, and by then every item was (probably, I didn't look into this super deep) iterated through via this.

Maia-J (changes requested):

code/modules/loadout/loadout_items.dm
NanoCats: Beg your pardon, Mr fluff was causing a lot of issues with this pr. This was an attempt to behave proactively.

Maia-J: Oh no, keep the code, just the comment isn't really needed, and an issue report into this would be appreciated.

BubbleTrroubles:
> this PR is a title of all time

Had to log in for this one.

Maia-J (approved):

FridayRemada:
These items are currently broken/do not display properly from both the loadout screen and when spawning in game:

midnight gown (/obj/item/clothing/under/dress/bubber/midnight_gown)
 • displays and properly sets those positions/alts, but deletes the color out of it entirely
giant scarf (/obj/item/clothing/under/dress/bubber/giant_scarf)
 • seems to change the actual display sprite when its on hte ground but then makes it invisible when you wear it, could be a simple mis-match
security shorts (/obj/item/clothing/under/rank/security/peacekeeper/trousers/shorts) 
 • these have an alt that seem to be broken/missing a texture entirely.

KateAnbolho:
> These items are currently broken/do not display properly from both the loadout screen and when spawning in game:
> 
> midnight gown (/obj/item/clothing/under/dress/bubber/midnight_gown) • displays and properly sets those positions/alts, but deletes the color out of it entirely giant scarf (/obj/item/clothing/under/dress/bubber/giant_scarf) • seems to change the actual display sprite when its on hte ground but then makes it invisible when you wear it, could be a simple mis-match security shorts (/obj/item/clothing/under/rank/security/peacekeeper/trousers/shorts) • these have an alt that seem to be broken/missing a texture entirely.

There's a few that'll have missing/broken alts because TG moment as well:tm:

NanoCats:
Yeah I'll probably have to make a follow-up. It's not perfect but it's still cool to have.

PR #6050, opened 2026-08-03
Small Dolphin and Carp event adjustments

Maia-J:
### Dolphin migration
Weight 15->5
Track Moderate->Mundane

### Space Carp
Tags += Space, Combat

## Why It's Good For The Game

Dolphins are harmless and low impact so having them as a fairly rare mundane track pick makes way more sense than just stealing from the moderate track.

Space carp, as the name implies, should have the space tag (Ground stations don't even have spawn points for them in the first place). They also bite you and force a bit of combat to deal with them, so.

## Proof Of Testing

It compiles. It's variable and weight adjustments anyhow

## Changelog

:cl:
balance: Space carp event is now actually tagged as SPACE and COMBAT
balance: Space dolphins are now in a different track, at a lower weight, to make them rarer.
/:cl:


xPokee:
Good PR. Space dolphins roll so frequently, especially on lowpop. Almost every round I see with under 35 players rolls dolphins at some point.

lessthnthree (approved):

PR #6060, opened 2026-08-05
Fixes Size Collar + Cleanup

xPokee:
## About The Pull Request
Oops. Turns out I broke the size collar.
## Why It's Good For The Game
Bugs bad
## Proof Of Testing
See below.
<details>
<summary>Screenshots/Videos</summary>

https://github.com/user-attachments/assets/ab8bc0ee-bea1-41a9-ba80-bfecddd205c1

</details>

## Changelog
:cl:
fix: Fixed the size collar.
/:cl:

PR #6061, opened 2026-08-05
Fixes the size collar

Phantastic-Swan:
## About The Pull Request

#6042 broke the size collar. The `original_size` variable was assigned improperly, resulting in the sprite size being doubled each time the size collar was unequipped. The PR also added a `current_body_size` variable to dna, but doesn't assign it to the dna upon character creation so I corrected that too (which I then used to assign the `original_size` variable in the size collar) That variable is used in one more place in code but from my limited testing my change SHOULDN'T break it.

That PR ALSO removed 2 defines that I think were pretty neat, so I brought them back too.

## Why It's Good For The Game

Fixes the size collar, removes some magic numbers.

## Proof Of Testing

<details>
<summary>Videos</summary>

https://github.com/user-attachments/assets/1cf8e05a-ec3f-4898-ace6-99fbb96ec823

</details>

## Changelog

:cl: Swan
fix: Fixes the size collar
/:cl:

### DIFF
--- a/code/__DEFINES/~~bubber_defines/mobs.dm
+++ b/code/__DEFINES/~~bubber_defines/mobs.dm
@@ -56,3 +56,7 @@
 #define PENIS_LAYER_OFFSET 0.03
 #define BELLY_LAYER_OFFSET 0.02
 #define BREASTS_LAYER_OFFSET 0.01
+
+// Maximum and minimum size we can set in the character creator
+#define BODY_SIZE_MIN (RESIZE_DEFAULT_SIZE * 0.8)
+#define BODY_SIZE_MAX (RESIZE_DEFAULT_SIZE * 1.5)
--- a/modular_skyrat/master_files/code/modules/client/preferences/body_size.dm
+++ b/modular_skyrat/master_files/code/modules/client/preferences/body_size.dm
@@ -2,8 +2,8 @@
 	category = PREFERENCE_CATEGORY_SECONDARY_FEATURES
 	savefile_identifier = PREFERENCE_CHARACTER
 	savefile_key = "body_size"
-	minimum = RESIZE_DEFAULT_SIZE * 0.8
-	maximum = RESIZE_DEFAULT_SIZE * 1.5
+	minimum = BODY_SIZE_MIN
+	maximum = BODY_SIZE_MAX
 	step = 0.01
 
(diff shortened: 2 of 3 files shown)


code/__DEFINES/~~bubber_defines/mobs.dm
```diff
@@ -56,3 +56,7 @@
 #define PENIS_LAYER_OFFSET 0.03
 #define BELLY_LAYER_OFFSET 0.02
 #define BREASTS_LAYER_OFFSET 0.01
+
+// Maximum and minimum size we can set in the character creator
+#define BODY_SIZE_MIN (RESIZE_DEFAULT_SIZE * 0.8)
+#define BODY_SIZE_MAX (RESIZE_DEFAULT_SIZE * 1.5)
```
Maia-J, line 62: If these are used in only one place, we really don't need them. It just makes compilation take longer with no real benefits

modular_skyrat/master_files/code/modules/client/preferences/body_size.dm
```diff
@@ -2,8 +2,8 @@
 	category = PREFERENCE_CATEGORY_SECONDARY_FEATURES
 	savefile_identifier = PREFERENCE_CHARACTER
 	savefile_key = "body_size"
-	minimum = RESIZE_DEFAULT_SIZE * 0.8
-	maximum = RESIZE_DEFAULT_SIZE * 1.5
+	minimum = BODY_SIZE_MIN
+	maximum = BODY_SIZE_MAX
```
Maia-J, line 6: ditto

Maia-J (changes requested):
Looks good otherwise

code/__DEFINES/~~bubber_defines/mobs.dm
Phantastic-Swan: I'm personally of the opinion that magic numbers should be avoided, and that naming numbers like this is good. Also it makes a central place where we can set the min/max body size values, so if anyone wants to add more in the future, or change it, they do it here, rather than in many other files.

Alternatively, would it be better if I also brought these values back to `modular_zubbers/code/modules/mob/living/carbon/human/species_types/slimepeople.dm`?
image

xPokee:
Already fixed by https://github.com/Bubberstation/Bubberstation/pull/6060.

PR #6065, opened 2026-08-06
[TM-ONLY][REMOVE ONCE WE GET LAW RACKS FROM UPSTREAM] Fixes unique AI not rolling with laws

Odairu:
## About The Pull Request
So it turns out that MMIs call set_config_laws() in their initialization, which happens before the station subsystem gets initialized. This means that laws are set up and good to go, then unique_ai gets applied, and is unable to run through its proc chain because laws already being set up. 
There's 3 other places that setup laws, and all of them init after the station subsystem, so plenty of sanity checks to make sure laws actually get written.
## Why It's Good For The Game
Bug fix go brrr
## Proof Of Testing
image
<details>

</details>

## Changelog
:cl:
fix: Unique AI now rolls its laws
/:cl:


Odairu:
this is in upstream

PR #6072, opened 2026-08-07
Cleaning up some ticker code

Phantastic-Swan:
## About The Pull Request

I was looking into changing how the round start ping is being generated so that it isn't dependant on the QOTR message (on which I eventually just gave up). While doing so I stumbled upon 3 separate overrides for the same proc of the ticker, in 3 different files. This induced a headache of immense proportions and I had to fix it because jesus christ what the fuck is that?

While I was at it, I grabbed some other ticker procs strewn around `modular_zubbers` and `modular_skyrat` and threw them into a single file responsible for handling the ticker.

## Why It's Good For The Game

Do you understand how cancer inducing it is to try and figure out how code is executed when there are 3 separate overrides for a single proc?

I am of the opinion that any codebase should override any proc AT MOST once. Multiple overrides for a proc make for worse stack traces, less clarity in the order of code execution, and more difficulty in finding original declarations of procs. They also make it harder to keep procs up to date with upstreams if say, the arguments change (I know BYOND doesn't enforce proc overrides to have the same arguments as the original proc; however I consider that a load of bullshit and a major flaw in the language)

## Proof Of Testing

I did compile the code, and the round ends normally, and the debugger tells me that code executes normally. Though, given that this does mess with subsystem code, maybe TM'ing is in order?

## Changelog

No player facing changes


Maia-J:
> Do you understand how cancer inducing it is to try and figure out how code is executed when there are 3 separate overrides for a single proc?

Oh yes we do. There's plans to finally pack the two modular folders together into one big cleaned up folder. Sadly TG just keeps beating us up with big refactors of icon and mob code so we're a bit slow.

That is to say, thank you

Phantastic-Swan:
See the worst part is that most of those overrides aren't even in the `modular_skyrat` folder, there are literally 3 overrides for the same proc in the `modular_zubbers` folder

modular_zubbers/code/controllers/subsystem/ticker.dm
```diff
@@ -39,3 +53,63 @@
 		player_ready_data.Insert(1, "Job Estimation:")
 		player_ready_data.Insert(1, "")
 	return player_ready_data
+
+/datum/controller/subsystem/ticker/declare_completion(was_forced = END_ROUND_AS_NORMAL)
+	handle_antag_tickets()
+	handle_quote_of_the_round()
+	handle_credits()
+
+	. = ..()
```
Maia-J: Can this be pushed up in the file on like line 18 so that it's ordered as proc supercalls/overrides followed by our new procs?

Maia-J (changes requested):
Seems great otherwise

modular_zubbers/code/controllers/subsystem/ticker.dm
Phantastic-Swan: done

Maia-J (approved):

PR #6085, opened 2026-08-10
Fixes marking issues 

### LINKED ISSUE #3864: Right hand markings will not work in setup character
### Client Version:

_No response_

### Issue Summary:

I was on the setup character menu, making a new character. But for some reason, no matter what I do, the markings on the right hand will not appear. The markings on the left had appear just fine, but not the right hand. I have no idea what is going on, but it just will not work.

### Round ID:

_No response_

### Test Merge Information:

_No response_

### Reproduction Steps:

Go to setup character, go to Augments+, try to select  a marking for the right hand (for example, Hyena) the marking on the right hand just wont appear.

### LINKED ISSUE #5879: "sticky" hand markings bug
### Client Version:

_No response_

### Issue Summary:

Sometimes, hand markings on a character will persist between characters. This is irreversible and only resets when the server does. Adding hand markings and removing them does not get rid of it.

Started appearing a few upstreams ago.

Image

(No hand markings, and yet, has hand markings. This persists ingame as well)

### Round ID:

_No response_

### Test Merge Information:

_No response_

### Reproduction Steps:

(Note: This is tricky to replicate as it does not always happen, but here are ways to make it more likely)
1. Make/have a character with markings on the hands
2. Switch to a different character (Perhaps try switching to a character who doesn't have hand markings?)
3. See if the hands are incorrect

### LINKED ISSUE #5880: can only add hand markings if arm markings exist
### Client Version:

_No response_

### Issue Summary:

You can only add hand markings to a character if you add arm markings first.

Image

Pictured: Right hand marking, no right arm marking.

Image

Pictured: Right arm marking added, right hand marking now appears.

### Round ID:

_No response_

### Test Merge Information:

_No response_

### Reproduction Steps:

1. Make a character
2. Try to add a hand marking
3. Try to add an arm marking

### LINKED ISSUE #6047: Unremoveable hand markings spreading into other character slots.
### Client Version:

516.1685

### Issue Summary:

The right hand chitin marking (and maybe more) shows up on my sprites regardless if they have the marking to begin with. This unremovable marking seems to spread to other character slots seemingly at random and on occasion taking the color from another character's hand marking. 

Image

Image

Image

### Round ID:

12601

### Test Merge Information:

_No response_

### Reproduction Steps:

It doesn't seem to work 100% of the time but...

1: Make an anthromorph (humans and some other species don't seem to work)
2: Add the chitin right hand and chitin right arm.
3: Remove the markings. 

You should be left with a unremovable marking.

NanoCats:
## About The Pull Request

Hand markings have been broken in a few different ways for a while now. This fixes them.

I've tracked down three unique symptoms that all have the same cause. 

1. Hand markings would refuse to display unless you also had a marking on the matching arm.
2. Once a hand marking appeared, removing it did nothing. It would remain until server restart.
3. Hand markings would show up on other characters that never had them.

The arm/hand link was the clue. Bubberstation caches limb icons in one shared list, and the cache key is built in `generate_icon_key()`. That key includes the limb's own markings, but it never included the hand markings, which the arm also draws. So the game had no idea anything had changed when you touched a hand dropdown. Changing an arm marking altered the key, forced a real redraw, and the hand marking came along for the ride. That is why there seemed to be a chain between the hand and arm marks.

The cache is shared across every mob, including the preview dummies in character setup, which is how one character's hands ended up on another character.

There was also a smaller problem in `update_limb()`. The hand marking lookup sat inside the arm's `if` block, and `markings_alpha` was only set in that same block. Alpha has no default, and a null alpha draws nothing. Both are now handled on their own.

Closes https://github.com/Bubberstation/Bubberstation/issues/5880
Closes https://github.com/Bubberstation/Bubberstation/issues/3864
Closes https://github.com/Bubberstation/Bubberstation/issues/5879
Closes https://github.com/Bubberstation/Bubberstation/issues/6047

## Proof of Testing
Behold, the amazing Technicolor Dream Man. (I made him specifically to test if this fix works, it does) 
image

## Why It's Good For The Game

Markings you applied to your character should appear on your character. Markings you deleted should go away. Neither of those was true, and the bugs that were caused by these fixes not being in the game caused markings to be smuggled unintentionally between character slots.

## Changelog

:cl: Aeri
fix: Hand markings now work without needing an arm marking, can be removed, and no longer leak onto other characters.
/:cl:

### DIFF
--- a/code/modules/surgery/bodyparts/_bodyparts.dm
+++ b/code/modules/surgery/bodyparts/_bodyparts.dm
@@ -1225,13 +1225,19 @@
 		limb_alpha = human_owner.dna.features[limb_alpha_key]
 	// BUBBER EDIT ADDITION END
 
+	// BUBBER EDIT ADDITION START - Hand marking fixes
+	markings_alpha = owner_species.markings_alpha // Has no default, and a null alpha renders nothing
+
 	if(body_zone in owner_species.body_markings)
 		markings = LAZYCOPY(owner_species.body_markings[body_zone])
-		if(aux_zone && (aux_zone in owner_species.body_markings))
-			aux_zone_markings = LAZYCOPY(owner_species.body_markings[aux_zone])
-		markings_alpha = owner_species.markings_alpha
 	else
 		markings = list()
+
+	if(aux_zone && (aux_zone in owner_species.body_markings))
+		aux_zone_markings = LAZYCOPY(owner_species.body_markings[aux_zone])
+	else
+		aux_zone_markings = list()
+	// BUBBER EDIT ADDITION END
 	// SKYRAT EDIT END
 	return TRUE
 
--- a/modular_skyrat/modules/customization/modules/surgery/bodyparts/_bodyparts.dm
+++ b/modular_skyrat/modules/customization/modules/surgery/bodyparts/_bodyparts.dm
@@ -17,6 +17,11 @@
 		. += limb_id == "digitigrade" ? ("digitigrade_1_" + body_zone) : body_zone
 		. += "-[key]_[markings[key][MARKING_INDEX_COLOR]]_[markings[key][MARKING_INDEX_EMISSIVE]]"
 
+	// Hand markings are drawn by this limb too, so they have to be part of its cache key.
+	for(var/key in aux_zone_markings)
+		. += aux_zone
+		. += "-[key]_[aux_zone_markings[key][MARKING_INDEX_COLOR]]_[aux_zone_markings[key][MARKING_INDEX_EMISSIVE]]"
+
 	return .
 
 /**


Maia-J:
Do mind that TG reworked a shitton of icon and overlay generation that we're yet to pull down so I'm putting this on hold until after the upstream comes in

NanoCats:
"You can only make Fix Prs"
"No not like that"
image
\j

code/modules/surgery/bodyparts/_bodyparts.dm
```diff
@@ -1225,13 +1225,19 @@
 		limb_alpha = human_owner.dna.features[limb_alpha_key]
 	// BUBBER EDIT ADDITION END
 
+	// BUBBER EDIT ADDITION START - Hand marking fixes
+	markings_alpha = owner_species.markings_alpha // Has no default, and a null alpha renders nothing
```
Maia-J, line 1229: This is right afer an addition end. Merge the edit blocks into one

Maia-J (changes requested):

PR #6086, opened 2026-08-10
Fixes GAGS previews

NanoCats:
## What this does

Fixes the broken image that shows up in the color menu instead of your item preview in like, I don't know, 50 percent of cases.

The menu makes a preview image, hands the integrated browser a link to it, and sends the image file separately. Those two things are not in step. If the link lands first, the browser asks for a file that doesn't exist, gives up, shits pants, and never tries again. On top of that the server has already written the file down as delivered, so it will not send it a second time. The preview stays broken until you change a color and get a different link.

Now the menu retries a failed image on its own, and there is a refresh button in the corner of the preview if you really want to give it a push (this shouldn't be necessary it SHOULD recover on its own, but I've been wrong before). The refresh also clears the server's note saying you already have the file, so it actually gets sent again.

Oh and nobody ever actually reported this as a bug, but it's been bugging me for like six months or so, it was on the Aeri issue tracker that lives exclusively in my brain. 

## Why this is a Bubber fix

In addition to the fact I refuse to touch TG directly if at all possible; the preview code for our dynamic uniforms lives in modular_zubbers and has no matching proc upstream, so there is nothing over there to fix. Bubbercode also makes vastly greater use of this functionality than upstream does, from the loadout and other places, and our previews rebuild a new image more often. Upstream makes one image per color string and opens the menu from VV or a spraycan, so it breaks less. 

## Testing

Recolored items from the loadout and in round. Caught the preview breaking and watched it come back on its own. Confirmed the refresh button works and throws no runtimes. There's nothing interesting to take a picture of here! OK fine, here, here's a GAGS menu doing nothing out of the ordinary. You can even see the little refresh doodad. 

image

:cl: Aeri
fix: GAGS menus no longer show broken images (and will recover on their own if they somehow do)
/:cl:

### DIFF
--- a/code/modules/admin/greyscale_modify_menu.dm
+++ b/code/modules/admin/greyscale_modify_menu.dm
@@ -33,6 +33,10 @@
 	var/refreshing = TRUE
 	///BUBBER VAR: Optional display labels for color groups, keyed by color index.
 	var/list/color_labels
+	///BUBBER VAR: The asset url of the last preview image we sent. Used to make the client ask for it again.
+	var/last_preview_url
+	///BUBBER VAR: Counts up on every new preview. The menu uses it to reload an image that failed.
+	var/preview_generation = 0
 
 	/**
 	 * Whether the menu is currently locked down to prevent abuse from players.
@@ -122,6 +126,7 @@
 	data["sprites_dir"] = dir2text(sprite_dir)
 	data["icon_state"] = icon_state
 	data["sprites"] = sprite_data
+	data["preview_generation"] = preview_generation //BUBBER ADDITION
 	return data
 
 /datum/greyscale_modify_menu/ui_act(action, list/params, datum/tgui/ui, datum/ui_state/state)
@@ -129,6 +134,13 @@
 	if(.)
 		return
 	switch(action)
+		//BUBBER ADDITION START: Lets the menu ask for the preview image again if it did not load.
+		if("reload_preview")
+			forget_preview_asset()
+			queue_refresh()
+			return TRUE
+		//BUBBER ADDITION END
+
 		if("select_config")
 			var/datum/greyscale_config/new_config = input(
 				usr,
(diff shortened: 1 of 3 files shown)


MrBobCool:
asian-senior-man-old-giving-260nw-1893507727

code/modules/admin/greyscale_modify_menu.dm
```diff
@@ -330,9 +342,49 @@ This is highly likely to cause massive amounts of lag as every object in the gam
...
-	sprite_data["finished"] = icon2html(finished, user, dir=sprite_dir, sourceonly=TRUE)
+	set_finished_sprite(finished) //BUBBER EDIT: was an inline icon2html. See set_finished_sprite.
 	refreshing = FALSE
 
+//BUBBER ADDITION START
+/**
+ * Stores the finished preview image and remembers its url.
+ * The url is what the menu re-requests if the image does not load.
+ */
+/datum/greyscale_modify_menu/proc/set_finished_sprite(image/finished)
```
Maia-J:
```suggestion
/// Stores the finished preview image and remembers its url.
/// The url is what the menu re-requests if the image does not load.
/datum/greyscale_modify_menu/proc/set_finished_sprite(image/finished)
```

Autodocs

code/modules/admin/greyscale_modify_menu.dm
```diff
@@ -330,9 +342,49 @@ This is highly likely to cause massive amounts of lag as every object in the gam
...
+	last_preview_url = icon2html(finished, user, dir = sprite_dir, sourceonly = TRUE)
+	sprite_data["finished"] = last_preview_url
+	preview_generation++
+
+/**
+ * Drops our record of the client already holding the last preview image.
+ * The server marks an asset as delivered the moment it is queued to send, so a send that fails is never retried.
+ * Clearing the record lets the next refresh send the file again.
+ */
+/datum/greyscale_modify_menu/proc/forget_preview_asset()
```
Maia-J:
```suggestion
/// Drops our record of the client already holding the last preview image.
/// The server marks an asset as delivered the moment it is queued to send, so a send that fails is never retried.
/// Clearing the record lets the next refresh send the file again.
/datum/greyscale_modify_menu/proc/forget_preview_asset()
```

Ditto

code/modules/admin/greyscale_modify_menu.dm
```diff
@@ -33,6 +33,10 @@
 	var/refreshing = TRUE
 	///BUBBER VAR: Optional display labels for color groups, keyed by color index.
 	var/list/color_labels
+	///BUBBER VAR: The asset url of the last preview image we sent. Used to make the client ask for it again.
+	var/last_preview_url
+	///BUBBER VAR: Counts up on every new preview. The menu uses it to reload an image that failed.
+	var/preview_generation = 0
```
Maia-J, line 39: Care to explain what the thought behind this is? This is never read as part of any reload check.

Maia-J (changes requested):

code/modules/admin/greyscale_modify_menu.dm
NanoCats: The intent here was basically a reliability/cache-busting measure for the preview.

I can remove it, looking back, you're right that preview_generation isn't actually being used as part of the reload path here, so it isn't accomplishing that as written. 

It already seems to work without it so, great success?

Maia-J: Please remove it then

PR #6089, opened 2026-08-10
Unifies a shit ton of /datum/job overrides

Phantastic-Swan:
## About The Pull Request

I was trying to fix the job display order CI for the upcoming upstream sync. I searched for the `/datum/job/prisoner` and found this

image

> Holy jesus. What is that? WHAT. THE FUCK. IS THAT?! **WHAT IS THAT?!**

As such, I have decided to go and find every override for `/datum/job` in `modular_zubbers` and `modular_skyrat` folders and organised them so that we don't have YES overrides for the same classes.

## Why It's Good For The Game

I shall quote myself from my last PR

> Do you understand how cancer inducing it is to try and figure out how code is executed when there are 3 separate overrides for a single proc?

> I am of the opinion that any codebase should override any proc AT MOST once. Multiple overrides for a proc make for worse stack traces, less clarity in the order of code execution, and more difficulty in finding original declarations of procs. They also make it harder to keep procs up to date with upstreams if say, the arguments change (I know BYOND doesn't enforce proc overrides to have the same arguments as the original proc; however I consider that a load of bullshit and a major flaw in the language)

## Proof Of Testing

This mainly moves code around rather than changes any of it. Regardless, it compiled and it seems that all procs that have been running previously are being hit now, so I assume it works.

## Changelog

No player facing changes


Phantastic-Swan:
I will be real with you chief: I don't think it's my fault rust_g fails to install

Maia-J:
This is currently in conflict, once you resolve that I'll give it a look

icons/map_icons/items/pda.dmi
Maia-J, line 1: Unsure why this has an edit on it

Maia-J (changes requested):
Couldn't find anything else when reviewing this. Minus the fact that the akula outfit system is hell but that's for another PR to fix

icons/map_icons/items/pda.dmi
Phantastic-Swan: For some reason iconforge decided that it doesn't like the position of the NT consultant and Tcomms technician PDA icon states

image

Phantastic-Swan: it's probably because I moved the definition of the NT consultants PDA to `modular_zubbers/code/modules/jobs/job_types/nanotrasen_consultant.dm:122`

Maia-J: Probably? Tbh I got no clue how this happens but since no icons changed it's okay.

Maia-J (approved):
This is just code cleanup so there's no reason to wait for another review, nothing changed for the players.

PR #6110, opened 2026-08-15
Replaced our NPC AI with Behavior Trees. [UPSTREAM PULL]

StrangeWeirdKitten:
This bumps the upstream merge up by one commit. Yeah, all our AI is being reworked. Yipee

:cl:
balance: Houndeye melee damage 15 -> 1. The mob never used it unless player controlled, and its blast is now an ability that player controlled houndeyes can also use.
fix: headcrabs no longer damage eachother when leaping
fix: player controlled headcrabs now transfer control into the zombie upon zombification
fix: Player controlled houndeyes and headcrabs now have full control over the blast and jump abilities
/:cl:


KateAnbolho:
It approaches...

Maia-J:
> It approaches...

I am in pain, you'd think A CHINCHILLA would be easy to refactor but it isn't

PR #6122, opened 2026-08-20
fixes runtime related to barber and lizard gas access

Phantastic-Swan:
## About The Pull Request

The define for barber, lizard gas and a few others was typed as a number, rather than a string. This was causing list index out of bound runtimes in certain places, like interacting with the ore silo (for example, printing circuit boards). This PR fixes it by making the defines strings, rather than numbers

## Why It's Good For The Game

Fixes runtimes. Also moves stuff from skyrat to bubbers files.

## Proof Of Testing

I used the circuit imprinter on runtime station with the jannie debug ID and it stopped runtiming. The accesses I've changed are listed properly now.

<details>
<summary>Screenshots</summary>

image

Before, the lizard gas manager ID had its access set as NONE because they were numbers.

image

The gas lizard manager ID has access listed properly now.

</details>

## Changelog

:cl: Swan
fix: fixes runtimes related to lizard gas station and barber access.
/:cl:


Maia-J (approved):
Oh god it was THAT old and nobody updated it?

PR #6131, opened 2026-08-22
Fix TEG circulator flipping by making it right clickable

Phantastic-Swan:
## About The Pull Request

So it turns out that game verbs are broken and they won't be unbroken until we grab a fix from TG. I noticed it yesterday when I was playing with the TEG and I noticed that the "flip" verb on the circulators doesn't work. Now, I cannot grab the fix from TG because that would require rewriting like 70 files on the codebase, so I will leave that for the next upstream merge. As an interim solution, I simply made the circulators flip behavior happen on right click with an empty hand. And because the controls of the TEG are complicated, I also added the context menu to it and the circulators.

## Why It's Good For The Game

Bandaids a TG bug until it can be merged proper, adds context menu to a complicated machine to make it easier to operate.

## Proof Of Testing

<details>
<summary>Video</summary>

https://github.com/user-attachments/assets/c3702766-d310-429e-9bfd-86e04f52d612

</details>

## Changelog

:cl: Swan
qol: Makes the TEGs circulator flip behavior happen on right click. Also adds context info to it.
/:cl:

### DIFF
--- a/modular_skyrat/master_files/code/modules/power/circulator.dm
+++ b/modular_skyrat/master_files/code/modules/power/circulator.dm
@@ -37,7 +37,22 @@
 	.=..()
 	component_parts = list(new /obj/item/circuitboard/machine/circulator)
 	update_appearance()
+	register_context()
 
+/obj/machinery/atmospherics/components/binary/circulator/add_context(atom/source, list/context, obj/item/held_item, mob/user)
+	. = ..()
+	if(!held_item)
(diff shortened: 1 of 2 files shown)


Phantastic-Swan:
oh god it gave me the feature label it's so over

Maia-J:
> oh god it gave me the feature label it's so over

I gotchu

Odairu (approved):

PR #6155, opened 2026-08-28
Bugfix for Fleshmind points being incorrect when examining

rovertdude:
## About The Pull Request
At this time, there is a bug with Fleshmind that causes examining its cores to give an incorrect number of points needed to level up. This is because the examine code uses the wrong math, doing level * 300 instead of last level up points + 300 which is the actual level up code. This pull request fixes the examine codes math and allows it to correctly calculate the points needed.
## Why It's Good For The Game

This provides a fix to the bug mentioned above, letting spectators and admins grab a correct amount of points needed for it to level up.

## Proof Of Testing

[https://youtu.be/W_2aXQ5Zyq8](https://youtu.be/W_2aXQ5Zyq8)

## Changelog

:cl:Lucky
fix: Fixed a bug where examining the fleshmind core gave an incorrect points needed to level up
/:cl:

### DIFF
--- a/modular_zubbers/code/game/gamemodes/fleshmind/fleshmind_structures.dm
+++ b/modular_zubbers/code/game/gamemodes/fleshmind/fleshmind_structures.dm
@@ -355,7 +355,12 @@
 
 /obj/structure/fleshmind/structure/core/examine(mob/user)
 	. = ..()
+<<<<<<< Updated upstream
 	/// Level * Progress Required - Current Points
+=======
+	/// Last Level Up Points + the points needed to next level, default 300
+	/// typically this means per level it takes around 300 points, may be a bit more if it levelled up fast
+>>>>>>> Stashed changes
 	var/level_calculation = (our_controller.last_level_up_points + our_controller.level_up_progress_required) - our_controller.current_points
 	/// round Cooldown Time / 10
 	var/time_calculation = round(COOLDOWN_TIMELEFT(our_controller, level_up_cooldown) / 10)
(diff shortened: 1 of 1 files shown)


Odairu (changes requested):
Based on your testing evidence, its still largely unclear, this should have some sort of 
points to next level x/y where y is the points at which it will level up.

rovertdude:
> Based on your testing evidence, its still largely unclear, this should have some sort of points to next level x/y where y is the points at which it will level up.

added a proper video and a comment!

modular_zubbers/code/game/gamemodes/fleshmind/fleshmind_structures.dm
Odairu: I appreciate the video, it makes it a lot clearer what its doing, do we think we can change this line to say "Points to next level" rather than progress.
Progress to next level implies it should be counting up, but it counts down

Odairu (changes requested):

modular_zubbers/code/game/gamemodes/fleshmind/fleshmind_structures.dm
rovertdude: i think it would be fine to, yeah! added that change!

rovertdude: the idea is also there to maybe add a "current points total" or whatnot if you're admin'd and examine it, but that info can also be seen via view variables

Odairu (approved):

PR #6084, opened 2026-08-09
FEATURE FREEZE STARTING ABOVE THIS PR - NON FIX PRS WILL BE CLOSED

Odairu:
Between this upstream taking quite a while, and the next two containing large features we will have to parse through, as well as the ever growing issue's page and backlog of PRs, we will be implementing a feature freeze for the time being while we (maintainers) sort through what we need to.
Opening a PR that isn't related to bug fixing during the feature freeze will lead to it being closed!

PR #6161, opened 2026-08-29
Reverts Hold up mechanic to TG base + adds the SR hold up reticle to TG base

SomeCabbage:
## About The Pull Request

Removes Skyrat's holdup system where you shift+middle-mouse to point your gun at someone. Consequently allows the 'Disk Please!!' achievement to be gotten.

using Hold up will now give a reticle similar to how it functioned before, and the damage bonus from holding someone up scales down with the distance it was initiated at. (max damage under TG was 2.5x, now it scales from 2.5x at 1 tile away to 1x at 6 tiles away, losing 0.3x per tile of distance.) - Odairu

## Why It's Good For The Game

I wanted to get the Disk Please!! achievement and, decided to make a fix for the skyrat holdup system, but I was told to just remove it and bring back the /tg/ one instead.

Mechanics are supplemental to roleplay, and the amount of immersion they offer is tied to what they do mechanically. While the SR hold up does its job, the main issue with it (citing the SR PR that introduced it) is that it clashed with their CI rules. We have recently gotten rid of CI, as such, that clashing no longer exists. With the revert back to TG, having someone be held up at gun point gives that little extra bit of immersion by having there be actual consequences to simply ignoring whoever has a gun pointed at you (It makes no sense that someone can point a gun at you, and you can just go "nope" and walk away).
The reason the damage scales down is due to how its going to be used and by who (usually). Antags are more likely to hold someone up from closer range, these are your 'hostage' situations, in which case, they should get to keep the damage bonus associated. Whereas sec are more likely to hold someone up from range, these are your 'freeze scumbag' situations where they draw their weapons as soon as they see you, as such, the antag shouldn't be mechanically punished for engaging in roleplay with security (too much).
Another sort of interaction I can see happening, with the advent of detectives being able to flip their IDs to look like assistants, they can use this for further undercover work, getting in close with the antag before holding them up from close range, and getting the full damage bonus.
There's many other such scenarios that this can be used to further the immersion in a scene, and the ways to just 'game' it are pretty non-existent. - Odairu

## Proof Of Testing
Screenshot 2026-08-29 070852
Screenshot 2026-08-29 070858
Screenshot 2026-08-29 070950
Screenshot 2026-08-29 070957

WITH RETICLE

image

## Changelog
:cl:
add: Adds SR hold up overlay effect to the TG aim in feature
balance: Hold up bonus damage now scales down with distance initiated
balance: Hold up can be activated from 9 tiles away, up from 2
del: Removed Skyrat's holdup system of shift+middle-mouse clicking someone. Reverts to /tg/'s holdup system where you right-click a person while adjacent or one tile away.
/:cl:


YakumoChen:
Feature freeze but also the holdup is a good tool for roleplay, makes it very clear when you're trying to threaten someone, and I'd argue the tg holdup is not conducive in its own right because you can walk up to someone and essentially force mechanics under the guise of roleplay.

thegrb93:
https://github.com/Bubberstation/Bubberstation/pull/6084

thegrb93:
Are we sure there aren't weapons we have that depend on right click? TG's control scheme sounds like it might easily conflict with something.

SomeCabbage:
only a few things I can think of use right click, such as guns with the scope component but I'm unsure how those interact with it.

YakumoChen:
I don't think that the reversion of a genuinely useful feature is worth a single achievement (that has no game value or persistence across servers).

Can you explain how this is good for the game under those terms? How the tg system is better for the game than the one in use?

SapphoQueer:
> Feature freeze

A maint gave them permission to do this despite the freeze to my knowledge

However, I do think I'd rather have the achievement tied into the skyrat holdup mechanic, and _MAYBE_ a lesser version of the damage multiplier added to it.

YakumoChen:
> > Feature freeze
> 
> A maint gave them permission to do this despite the freeze to my knowledge
> 
> However, I do think I'd rather have the achievement tied into the skyrat holdup mechanic, and _MAYBE_ a lesser version of the damage multiplier added to it.

https://github.com/Skyrat-SS13/Skyrat-tg/pull/12624

I went down and hunted the actual PR that changed it all the way back on skyrat.

Removing all the mechanics from gunpoint turns it into purely a roleplaying tool, signaling intent in roleplay rather than forcing mechanics. Adding mechanics to skyrat gunpoint ruins the use as a roleplaying tool

Tg-holdup doesn't need this system removed to work, skyrat has managed it before, I don't see why we cannot restore and preserve the line between roleplay and mechanics.

SomeCabbage:
I dunno, the moment has passed and I don't really care what happens to this PR.

AstrumLernaean:
Oh that's what the TG holdup is. Yeah we'd rather keep the skyrat version of this and simply add the mechanics to it if we're being honest, this seems a lot worse.

SapphoQueer:
afbeelding

K3 has plans and is going to implement it like this

YakumoChen:
cam we keep the purely roleplay holdup mechanic though?

I like to have a visible way of threatening people with guns without forcing combat.

I don't really see this being used in a way that will benefit the roleplay/mechanics divide for the same reason it was removed. If I point a gun at someone from two tiles away just to be threatening it's now treated like a mechanical action because if they move at all they take big damage.

Odairu:
> cam we keep the purely roleplay holdup mechanic though?
> 
> I like to have a visible way of threatening people with guns without forcing combat.
> 
> I don't really see this being used in a way that will benefit the roleplay/mechanics divide for the same reason it was removed. If I point a gun at someone from two tiles away just to be threatening it's now treated like a mechanical action because if they move at all they take big damage.

It was removed because it went against SR's CI rules, we do not have CI anymore and follow standard escalation. Using the Aim in feature from skyrat *is* a mechanical escalation, even if you're not actually shooting.

Mechanics are supplementary to roleplay, and the immersion they create lends to that, TGs hold up is more immersive than skyrat's version. Plus this gives incentive for the person to actually stand there and engage with you, which is good for both antags and sec who want to RP (there is a mechanical punishment to *not roleplaying*).

Fundamentally I think this feature has a lot of cool applications when used in prime conditions. Is it always going to be used in prime conditions? No, but its also a feature that can't really just be gamed so to speak, and the way I will change it then reimplement it will reflect that.

SapphoQueer:
Make sure to edit the PR body/changelog/title appropriately

SomeCabbage:
just rember that at base, the tg holdup only works from 0-1 tiles away unless you like, changed it

SapphoQueer:
> just rember that at base, the tg holdup only works from 0-1 tiles away unless you like, changed it

This has been changed by odairu to be 9, with the damage bonus scaling based on how close you are to the target.

YakumoChen:
> It was removed because it went against SR's CI rules, we do not have CI anymore and follow standard escalation. Using the Aim in feature from skyrat is a mechanical escalation, even if you're not actually shooting.

I don't understand this at all, since skyrat holdup was changed to do nothing mechanical beyond put a reticle on the target and give the person aiming a slowdown (arguably making it mechanically disadvantageous for the aimer). It doesn't offer reflexive shooting.
But it doesn't seem like this will matter if your mind is set

Range shouldn't be longer than 7, 9 goes beyond the non-wide viewscreen (it can be janky to interact on the wide-screen strips)

I'm also not sure how this is going to interact well with zoom functions if aiming is so far, I can see it leading to a lot of accidental mechanics.

AstrumLernaean:
Oh right, this should also probably go back to being Shift+Middle click so there's no unintentional interactions with certain weapons yeah?

Odairu:
> Oh right, this should also probably go back to being Shift+Middle click so there's no unintentional interactions with certain weapons yeah?

I don't believe there'd be anything funky, as their rightclick would override this and they'd just do w/e they do. Tg has weapons that do stuff on right click.
If it turns out to be an issue I'll move it over.

SapphoQueer:
> If it turns out to be an issue I'll move it over.

Between player muscle memory and, "it's the point button with a gun", I think shift + mmb makes more sense, personally.

Maia-J:
> Oh right, this should also probably go back to being Shift+Middle click so there's no unintentional interactions with certain weapons yeah?

I checked this, there's no weapons binding rightclick signals, and the only one that's not TG that uses the rightclick procs differently overrides them completely.

code/__DEFINES/projectiles.dm
```diff
@@ -68,7 +68,7 @@
 #define CALIBER_40MM_SHELL "40MM_shell"
 
 /// For gunpoints, how many tiles around the target the shooter can roam without losing their shot
-#define GUNPOINT_SHOOTER_STRAY_RANGE 2
+#define GUNPOINT_SHOOTER_STRAY_RANGE 9 //BUBBER EDIT - Hold up range is screenwide
```
xPokee, line 71: Make sure to include the original values when changing stuff.
```suggestion
#define GUNPOINT_SHOOTER_STRAY_RANGE 9 //BUBBER EDIT - ORIGINAL: 2 - Hold up range is screenwide
```

code/modules/projectiles/gun.dm
```diff
@@ -356,7 +356,7 @@
 	return fired
 
 /obj/item/gun/interact_with_atom_secondary(atom/interacting_with, mob/living/user, list/modifiers)
-	if(user.combat_mode && isliving(interacting_with))
+	if(user.combat_mode && isliving(interacting_with) && get_dist(user, interacting_with) < 2) //Bubber edit - distance check so you can hold up in combat mode
```
xPokee, line 359: Ditto
```suggestion
	if(user.combat_mode && isliving(interacting_with) && get_dist(user, interacting_with) < 2) //Bubber edit - ORIGINAL: if(user.combat_mode && isliving(interacting_with)) - distance check so you can hold up in combat mode
```

xPokee (changes requested):

PR #6039, opened 2026-07-31
Tweaks the VOKSHOD's armor values, removes the wound 30 on the unpowered HEV suit, and makes them all not marked contra. Also adds a seperate order for the MOD as a premium.

KateAnbolho:
## About The Pull Request

This PR sorta futzes with everything VOKSHOD suit related. 

Tweaking the armor to make it a little more viable against ranged damage sources, nerfing and tweaking the hardsuit's armor (skyrat gave it wound 30, and the unobtainable powered suit has 50 which is kinda insane), adds a specific order for the MOD for those who can afford to pay a premium and don't wanna quest, and makes the suit exist publicly without a console MOD in order to bring hardsuits back :) 

## Why It's Good For The Game

more things to spend your money is good when you grind bounties all day. 

## Proof Of Testing

image

## Changelog
:cl:
add: independent voskhod suit order
balance: rebalanced voskhod suit armor values
/:cl:

PR #6175, opened 2026-09-02
Revert "Tweaks the VOKSHOD's armor values, removes the wound 30 on the unpowered HEV suit, and makes them all not marked contra. Also adds a seperate order for the MOD as a premium."

BingusTheClown:
Reverts Bubberstation/Bubberstation#6039


BingusTheClown:
Good news, the cargo problems are not caused by this PR!

Issue #6194, opened 2026-09-05
Regular Emergency Shuttles will land wherever they damn please if the VARS-7 is loaded in any capacity


SomeCabbage:
### Client Version:

516.1673

### Issue Summary:

When previewing the VARS-7 emergency shuttle, it's not actually selected or loaded anywhere where it can been seen or done anything with, however, when the normal emergency shuttle arrives with the station after previewing it using the shuttle manipulator, the normal shuttle docks wherever it wants akin to how the VARS-7 does so. Additionally can happen if the VARS-7 is bought, then is replaced via a shuttle catastrophe.
<img width="950" height="1120" alt="Image" src="https://github.com/user-attachments/assets/ee5bb8c3-f8b5-44c2-866c-8efb6a1efef2" />

### Round ID:

12888

### Test Merge Information:

- [Adds new malf AI ability to hack IPCs/Synthetics Version 2.0](https://github.com/Bubberstation/Bubberstation/pull/5750)
- [[TM-ONLY EARLY PULL] Gives AI the option to change the color of their hologram when setting it up.](https://github.com/Bubberstation/Bubberstation/pull/6066)
- [[TM ONLY] Fixes RCDs not working on space tiles](https://github.com/Bubberstation/Bubberstation/pull/6112)
- [[TM-ONLY][REMOVE ONCE WE GET LAW RACKS FROM UPSTREAM] Fixes unique AI not rolling with laws](https://github.com/Bubberstation/Bubberstation/pull/6065)
- [Reverts Hold up mechanic to TG base + adds the SR hold up reticle to TG base](https://github.com/Bubberstation/Bubberstation/pull/6161)
- [Blacksmithing Changes Pt. 2: When Pt.1 Was Fullmerged I Gained Immeasurable Power](https://github.com/Bubberstation/Bubberstation/pull/5902)

### Reproduction Steps:

1. Be in-game.
2. Preview the VARS-7 through the shuttle manipulator or purchase legitimately.
2b. If purchased legitimately, have shuttle be replaced by a shuttle catastrophe.
3. Have the non-VARS shuttle dock with the station.
4. Crashes into a random unannounced location.

KateAnbolho:
Oh no

NanoCats:
Your honor, I plead "funny"

PR #6195, opened 2026-09-05
Fixes the VARS-7 not working as intended

NanoCats:
## About The Pull Request

The VARS-7's crash behaviour was leaking onto the regular emergency shuttle(s). Two causes:

**The crash effect happened if it loaded ever.** The backbone of the VARS was a custom `/obj/effect/station_crash` subtype called "ohno", which runs `shuttle_crash()` in `Initialize()`. That's normal for a "Crash" style shuttle (Lance, Meteor, Monastery, Millionaire and Cube all do the same thing), but it means the roll happens whenever the template's contents load, including a preview in the shuttle manipulator that's never actually selected. Ours is... More severe than that, so it tends to get noticed. I swear to god this was possible in a few semi rare cases on TG. 

**It moved the shared dock somewhere funny and never put it back.** It moves `emergency_home`, which is used by all emergency shuttles. The TG shuttle crasher effect only nudges that dock a few tiles from where it already is, so no other shuttle has ever needed to restore it; this is technically a TG bug if you ask me, but the VARS throws it to a random room anywhere on the station, which turned a mild fender bender into "Aeri your shuttle fucking obliterated the entire map fix it you bitch" 

Fixes:
1. Shuttle no longer fucks up unrelated shuttles for no good reason
2. Fixed some minor shit on the shuttle (the cryo tube setup was nonfunctional)
3. The shuttle now leaves behind damage on the station instead of turning it into a mysteriously immaculate backrooms type arrangement. Most shuttles aren't supposed to land on the station so they deliberately snapshot how it was before they got here, and leave it as they found it, ours is SUPPOSED to fuck it up, now it does. 

## Proof Of Testing

image

image

## Changelog

:cl: Aeri
fix: Previewing the VARS-7 Provodnik in the shuttle manipulator no longer causes the regular emergency shuttle to dock in random places.
add: The VARS-7's crash site now stays wrecked after the shuttle departs, instead of reverting to an untouched room.
/:cl:


BingusTheClown (approved):

BingusTheClown:
Should fix: https://github.com/Bubberstation/Bubberstation/issues/6194

PR #6199, opened 2026-09-07
Various fleshmind fixes and changes

Maia-J:
## About The Pull Request

This makes fleshmind a tad more threatening for everyone involved:

- Spawning now checks a much wider area for players, and searches actual places first before fallbacking to blob spawns, which should give it more of a fighting chance
- Wireweed structures and fleshmind mobs (Including the priest) now use the PASSBLOB flag, meaning they don't get stuck on the fleshmind, and have a much easier time moving and fighting.
- Mechiver spawning is less common (4->3)
- Spawners now spawn new mobs automatically and not just when attacked or commanded by the main controller
- A wireweed wall no longer blocks spread, previously this led to situations where the fleshmind would seal itself in small rooms and not grow out of them

### Fixes

The spawn locations were genuine ass, we had several wireweeds recently just plop down and die immediately due to being spotted. Plus they'd surround their spawn with walls and block spreading that way too.

Since most spawning locations were tight spaces (And still are), the mobs (which are kinda the main thing) getting stuck is just bad, feels stupid, and sometimes their ai would get forever stuck attempting to reinforce but being unable to due to structures in the way.

### Balance

Mechivers in their current state aren't fighters, they're sneaky bastards who stay away from you until it's time to knock you into hard crit and corrupt you. Having them spawn super often doesn't really make sense, especially with their massive health pool. So they were lowered a little bit.

Spawners are kinda an experimental thing, which might be nerfed/buffed according to testing

## Proof Of Testing

It works, I forgot le screenshot. Balance changes to be testmerged

## Changelog

:cl:
balance: Mechiver spawn weight lowered 4 -> 3
balance: Fleshmind spawners now spawn new mobs on a 20-60 second timer
fix: Fleshmind mobs and the priest now pass through the fleshmind structures and walls
fix: Fleshmind spawns no longer use a tiny check for players in the area, and try to pick more interesting locations
fix: Fleshmind wirewalls no longer block the fleshmind itself from spreading
/:cl:


SomeCabbage:
being able to pass through each other is huge

Maia-J:
> being able to pass through each other is huge

I am convinced this was a bug with how many structures are placed down by the wireweed and with the mobs spawning atop the spawners.

BingusTheClown (approved):

PR #6201, opened 2026-09-07
Reskin bugs surfaced by the loadout skin system roundup

NanoCats:
## About The Pull Request

I made it so you can alt click items in the loadout so that you can reskin them in your loadout menu, awsome right? Well yes, but that turned out to be a decent bug detector and I'm now sorta honorbound to fix at least some of those. Most of what follows was _already broken in game_ and simply hidden under a rug.

All of these fix the in-game alt-click path as well as the loadout.

1.  GAGS worn icons would in semi rare cases be thrown away on reskin

We had an override that in some cases, would cause certain items to become invisible because of an inconsistency in guards against the scenario. 

I discovered the solution by manually calling `update_greyscale()` with VV to fix the item in game.

The same guard is added to `clear_skin()`, which had the same omission. **13 items** including but not limited to the midnight gown, giant scarf, pride scarf, and more, could potentially be affected by this. 

2. Mask reskins were reverted by the next icon update

A couple masks didn't set change_base_icon_state and would lose their skin the moment they were updated. Sorted. 
Fixed on the paper mask and the coloured clown mask.

3. Security shorts alt had a transposed sprite state name

The object sheet has `workshorts_red_short`, the worn sheet had `workshorts_short_red`. Blue and white use `workshorts_X_short` in both sheets, so red was the odd one out. The alt rendered on the ground and vanished when worn, because the worn state didn't exist properly. Renamed the state in the worn sheet to match its neighbors.

4. Paper mask reskins runtimed when applied without a user

`/datum/atom_skin/paper_mask/apply()` checks whether the user is holding a pen without first checking there is a user. The loadout applies skins with nobody holding the item, so this threw every time, including on the character preview dummy. Anyone with a paper mask in their loadout was generating these just by opening the menu. Now returns early when there is no user, since there is no pen to check for. I'll also note that the check for a pen never worked and I'm not fixing it right now, if anything someone might want to make a PR that removes the broken pen requirement, this is straying further and further from loadout stuff so I don't want to push it any further. 

5. Mob holders runtimed when deleted somewhere they can't drop their occupant

`release()` called `forceMove(drop_location())` with no null check. A holder sitting in nullspace has nowhere to put the mob inside it, and the character preview builds and discards loadout items in exactly that state. The occupant is now deleted along with the holder, since nothing can reach it there, rather than runtiming or being yote into nullspace. 

I will note that this was an issue discovered with the "Mr Fluff" mothroach, I can no longer replicate the issue or test it because only CentralSmith has access to Mr Fluff, I think somebody changed this at some point.  I personally strongly recommend we purge the rest of the ckey locks but that's a conversation for another day. 

## Why It's Good For The Game

Alternate skins, especially those afforded in the loadout are one of the cheapest forms of self expression we offer, and a decent number of them have had small jagged edges for a while. Several items had significant bugs that needed fixing. The PR I made to enable some extra customization forced a few of these to the surface in a convenient way. 

The two runtime fixes are small but they fire on the character preview, which spoiler alert, happens a lot. 

## Proof Of Testing

Compiled and run locally.

image

## Changelog

:cl: Aeri
fix: Several reskinnable items are now repaired so they aren't completely invisible if you try to reskin them. 
fix: Reskinned masks no longer revert to their default face when they shouldn't. 
fix: The red security shorts alt is no longer invisible when worn.
fix: Several runtimes related to the loadout no longer occur 
/:cl:


Maia-J (approved):

Maia-J:
Fix the conflicts we so graciously introduced and you shall be given a swift marging

code/modules/mob/living/inhand_holder.dm
```diff
@@ -86,7 +86,19 @@
...
+		// BUBBER EDIT ADDITION START - A holder sitting in nullspace has nowhere to put its occupant.
+		// The character preview builds and discards loadout items there, and forceMove(null) CRASHes.
+		// Nothing can reach the mob in that case, so take it with us rather than leaking it.
+		var/atom/drop_target = drop_location()
+		if(isnull(drop_target))
+			UnregisterSignal(released_mob, COMSIG_QDELETING)
+			held_mob = null
+			qdel(released_mob)
+			if(!QDELETED(src))
+				qdel(src)
```
Maia-J: Come to think of that, this might be risky if the holder is held by something else that's using it as part of a unit test, and it suddenly forces itself to delete. Plus this might recursively call Destroy() if it was called from there already.
```suggestion
```

Maia-J (changes requested):
My apolocheese but I did notice something

PR #6210, opened 2026-09-08
Add Quirks - "Criminal Underworlder" and "Underworld Connections"

BingusTheClown:
## About The Pull Request

Adds two new quirks (and some backend code to support them and potentially future quirks) that give more roundstart flavour as well as some fun RP centered bits. 

## Why It's Good For The Game

1. Allows people immediate access to the black market if they want to take a negative quirk to balance it out
2. Allows an RP focus on 'light' criminals and criminal activity

## Proof Of Testing

<details>

image
image
</details>
image

## Changelog

:cl:
add: A new negative quirk called 'Criminal Underworlder' - Player gains access to player exploitables and has note in their sec records denoting they should not be given weapon permits
add: A new positive quirk called 'Underworld Connections' - Players spawn in with a Black Market Uplink
/:cl:


AstrumLernaean:
Interesting.

Could you perhaps elaborate a bit more on where the idea for these quirks may have come from? As well as the reasoning for its addition/some more details on why this may be 'good for the game'?

We're curious about how this might also interact with our self-antagonism rules.

code/datums/quirks/_quirk.dm
```diff
@@ -20,6 +20,10 @@
 	var/medical_record_text
 	///Appears in medical guides for this quirk, but only if the quirk has QUIRK_TRAUMALIKE flag.
 	var/medical_symptom_text
+	//BUBBER ADDITION START
+	///This text will appear in security records for the trait
+	var/security_record_text
```
Odairu: We can put this in a bubber file to modularize it

code/datums/records/record.dm
```diff
@@ -113,6 +113,7 @@
 	physical_status = PHYSICAL_ACTIVE,
 	mental_status = MENTAL_STABLE,
 	quirk_notes,
+	security_note, // BUBBER EDIT
```
Odairu, line 116: This can be moved under the 'skyrat edit' bracket and bubber edit be removed (not necessary but)

code/datums/records/record.dm
```diff
@@ -128,6 +129,7 @@
 	src.physical_status = physical_status
 	src.mental_status = mental_status
 	src.quirk_notes = quirk_notes
+	src.security_note = security_note //Bubber EDIT
```
Odairu: ditto

modular_zubbers/code/datums/quirks/negative_quirks/criminal_underworlder.dm
```diff
@@ -0,0 +1,12 @@
...
+	name = "Criminal Underworlder"
+	desc = "You have criminal connections and experience 'encountering' the law which makes you able to exploit others. You are suspicious to law enforcement and have existing security records. Expect random searches and questioning."
+	value = -2
+	security_record_text = "Person is known to have previous connections with criminals. DO NOT provide them with firearms or weapons!"
+	mob_trait = TRAIT_ALWAYS_SUSPICIOUS
+	icon = FA_ICON_PEOPLE_ROBBERY
+
+/datum/quirk/criminal_underworlder/add(client/client_source)
+	var/mob/living/carbon/human/human_holder = quirk_holder
+	human_holder.mind.has_exploitables_override = TRUE
```
Odairu: human_holder.mind?.has_exploitables_override I believe. ? operator sorta acts like if(human_holder.mind)

Odairu (changes requested):

modular_zubbers/code/datums/quirks/negative_quirks/criminal_underworlder.dm
```diff
@@ -0,0 +1,12 @@
+/datum/quirk/criminal_underworlder
```
Odairu, line 1: All quirks need medical record text, thanks TG

BingusTheClown:
> Interesting.
> 
> Could you perhaps elaborate a bit more on where the idea for these quirks may have come from? As well as the reasoning for its addition/some more details on why this may be 'good for the game'?
> 
> We're curious about how this might also interact with our self-antagonism rules.

The idea for these quirks came from player feedback. That primarily, there should be more avenues to roleplay particular backgrounds, or favour towards particular interactions.

My own personal take is that: Quirks already exist in such a manner that they can alter medical text, but for some reason not security notes, which in turn feels like a huge gap to those who take the time to take a quirk and want some pre-filled information. 

Most notably, unless people PUT IN actual sec records (not everyone does), there's zero reason to ever deny a weapon permit. This quirk immediately shakes that up by having a security note placed on the character's records on spawn. 

I think there needs to be self-antag policy changes FIRST before this could get merged. (Just due to the access to exploitable on the quirk) But I think that then paves the way for 'non' antags, to see such things, and perhaps use them against other players in roleplay scenarios. 

Some examples of where I think this quirk might be encouraged to be used are:

An assistant wants to be a small time drug dealer, and use exploitable as a way to find 'customers' for their product. (Or even uses the second quirk as a balance to fence the drugs on the black market)

A bouncer has a bit of 'vigilante' streak, and wants to pressure people into giving them information on what they've seen and heard which might in turn help them do their job. (But is seen as suspicious for their past behavior going out of line)

YakumoChen:
> We're curious about how this might also interact with our self-antagonism rules.

I've had experience with this quirk while playing on Nova, from personal experience it's mainly used by players as an excuse to self-antagonise security whenever sec players interact with their quirk. It's mainly used by problem players looking for an excuse to cause excessive problems.

NanoCats:
> > We're curious about how this might also interact with our self-antagonism rules.
> 
> I've had experience with this quirk while playing on Nova, from personal experience it's mainly used by players as an excuse to self-antagonise security whenever sec players interact with their quirk. It's mainly used by problem players looking for an excuse to cause excessive problems.

Ah, Just like Monkestation's jailbird quirk :)

SapphoQueer:
> I've had experience with this quirk while playing on Nova, from personal experience it's mainly used by players as an excuse to self-antagonise security whenever sec players interact with their quirk. It's mainly used by problem players looking for an excuse to cause excessive problems.

This is my main concern as well. I would rather not have people antagonise sec and vise versa for having a quirk.

BingusTheClown:
I've gone and removed the Suspicious Status as I also agree, doesn't make sense to 'force' people to interact, and would be equally confusing for newer sec folks

code/modules/mob/living/status_procs.dm
```diff
@@ -518,6 +518,23 @@
...
 	return medical ?  dat.Join("<br>") : dat.Join(", ")
 
+//Bubber Addition START
+/**
+* get_quirk_security_string() is used to add all of the quirks that have security notes into a singular note for security officers
+*
+* Arguments:
+* * Category- Which types of quirks we want to print out. Defaults to everything
+*/
+/mob/living/proc/get_quirk_security_string(category = CAT_QUIRK_ALL)
```
Odairu: Modularize this

Odairu (changes requested):

code/modules/mob/living/status_procs.dm
BingusTheClown: Done

YakumoChen:
> This is my main concern as well. I would rather not have people antagonise sec and vise versa for having a quirk.

That's the problem, the negative quirk is basically "security antagonises you: the quirk", because security has extra reason to check to make sure you're not running a gun without a permit.

Despite being a 'negative' quirk it also provides players with exploitables which lets them self-antag even more.

Maia-J (approved):

PR #6213, opened 2026-09-08
Antag prompting

Arturlang:
## About The Pull Request
makes midround antags be prompted via the same UI ghostset antags are, also adds config options to make it prompt everyone, ignoring antag prefs (disableable on a per-round-basis)
also this prompts EVERYONE who is valid job-wise.
opting out 

## Why It's Good For The Game
The intent here is to encourage more people to antag, and to avoid the stale gameplay of having the same antag each round

## Proof Of Testing
dreamseeker_SioRzKg3ve

## Changelog

:cl:
add: midround antag rolls will now prompt everyone valid on the server
add: a new verb in OOC tab for clearing all of your never for this round preferences
add: midround antag rolls, crewset ones will now use the ghost prompt UI, so you can opt-in to antag, the sound/prefs for it is disableable in prefs, but ignores antag prefs, (on config), right clicking it or clicking never disables it for the round still
/:cl:


modular_zubbers/code/modules/antag_prompt/_defines.dm
Maia-J, line 1: This is technically just part of storyteller, move it there so that we don't bloat and split up everything into 6 different modules like skyrat did

Maia-J (changes requested):

NanoCats:
There's one small problem with this, and you might be OK with this but I just want to highlight it just in case: 

Won't this kind of tip everyone off? Like "HEY GANG, THERE'S ABOUT TO BE +1 LING ON THE STATION!"

SapphoQueer:
> There's one small problem with this, and you might be OK with this but I just want to highlight it just in case:
> 
> Won't this kind of tip everyone off? Like "HEY GANG, THERE'S ABOUT TO BE +1 LING ON THE STATION!"

Kinda but not really? Some people get fake rolls, which do prompt a different antagonist than is actually going to be there.

This is on top of command and sec and other non-valid roles not getting the prompt to begin with.

Danielance:
people can already observe, see everything that's happening in the round, and respawn under the faith that they won't use that knowledge to metagame. i think it's fine tbh

Maia-J:
> Kinda but not really? Some people get fake rolls, which do prompt a different antagonist than is actually going to be there.
> 
> This is on top of command and sec and other non-valid roles not getting the prompt to begin with.

Worst case we just add fake rolls too, but we trust our playerbase not to metagame.

Arturlang:
pls don't TM this without having me confirm

Arturlang:
> > Kinda but not really? Some people get fake rolls, which do prompt a different antagonist than is actually going to be there.
> > This is on top of command and sec and other non-valid roles not getting the prompt to begin with.
> 
> Worst case we just add fake rolls too, but we trust our playerbase not to metagame.

main reason not to do this is making it feel pointless to opt in and too annoying, imo

Arturlang:
testing so far has shown no issues, was testmerged for a round

PR #6220, opened 2026-09-10
Repath obj/item/device holdovers to obj/item

NanoCats:
## About The Pull Request

Robwo pointed out in Discord that `/obj/item/device` is a dead path. TG abandoned it a long time ago, so the only things still hanging off it are leftover Skyrat objects. 

There is no such thing as an  `/obj/item/device` definition anywhere in the codebase.

Two type families moved:

- `/obj/item/device/custom_kit` is now `/obj/item/custom_kit`. This takes its four subtypes with it: `medigun_fastcharge`, `deluxe_hypo2`, `empty_cell`, and the `empty_cell` children `body_teleporter` and `relocator`.
- `/obj/item/device/traitor_announcer` is now `/obj/item/traitor_announcer`, along with its `infinite` adminbus subtype.

I also updated everything that referenced those paths, because you know, duh. 

## Why It's Good For The Game

No player-facing change at all; this is tree cleanup. I barely understand what that means, but somebody asked me to do this so here we are. 

image
Pay up

## Proof Of Testing

image

Compiles clean. Verified in game that the affected items still spawn and work. This is just about the most boring PR possible. 

## Changelog

:cl:
refactor: Migrated the leftover `/obj/item/device` objects to their proper paths.
/:cl:


Maia-J (changes requested):
Please add an UpdatePaths and you shall receive the marging

PR #6229, opened 2026-09-12
Fix the narrate verb not working

Phantastic-Swan:
## About The Pull Request

The narrate verb was broken because an if statement checking if the character was unconcious/in crit was inverted. This made it so the player is unable to narrate while being concious and in full health. This PR fixes it.

Also, moves the narrate verb code to `modular_zubbers`

## Why It's Good For The Game

Fixes a bug.

## Proof Of Testing

I ran the changes locally and they work. I could narrate when concious and I couldn't narrate when unconcious. Though, the warning message about it doesn't go through because the emote code actually terminates early when you're unconcious, but I still left the check in in case something triggers it.

<details>
<summary>Screenshot</summary>

image

</details>

## Changelog

:cl: Swan
fix: Fixes being unable to narrate while concious
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/emotes/narrate.dm
+++ b/modular_zubbers/code/modules/emotes/narrate.dm
@@ -1,0 +1,82 @@
+#define NARRATE_RANGE_SAME_TILE "Same Tile"
+#define NARRATE_RANGE_ONE_TILE "1-Tile Range"
+#define NARRATE_RANGE_MAX "Max Range"
+#define NARRATE_RANGE_CUSTOM "Custom Range"
+
+GAME_VERB_DESC(/mob/living, narrate, "Narrate", "Allows you to send a narration message to a target or area", "IC")
+	usr.emote("narrate")
+
+/datum/emote/narrate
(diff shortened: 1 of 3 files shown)


modular_zubbers/code/modules/emotes/narrate.dm
```diff
@@ -0,0 +1,82 @@
...
+	name = "IC Narrate"
+	key = "narrate"
+	message = null
+
+/datum/emote/narrate/run_emote(mob/living/user, params, type_override = null, intentional = TRUE)
+	if(GLOB.say_disabled)	// This is here to try to identify lag problems
+		to_chat(user, span_danger("Speech is currently admin-disabled."))
+		return
+
+	if(IS_UNCONSCIOUS_OR_CRIT(user))
```
Phantastic-Swan, line 19: this is the line that had the bug. It used to be `if(!IS_UNCONSCIOUS_OR_CRIT(user))`. Notice the exclamation mark which inverted it.

BingusTheClown (approved):

PR #6230, opened 2026-09-12
Re-Enable heretic/acolyte roll for Chaplains

DragonLee66:
## About The Pull Request
Comments out the part of the code that restricts chaplains from rolling acolyte.

## Why It's Good For The Game
It's arguably a subjective change and a biased one on my end, but it's strange to single out the chaplain for not becoming an Acolyte. My first argument is that if you can be a vampire chaplain (who can literally burn in their own church), you should be able to become an acolyte chaplain. I don't think chaplains (at least at the moment) have a whole lot of tools that counter acolyte anyway that you can't get elsewhere. (I'm pretty sure you can get magic immunity by drinking silver or holding holymelons, and nullrods don't do extra damage to heretics)

I also believe a heretical/acolyte chaplain is quite thematic, and would make for interesting roleplay. (Imagine them turning their chapel into horrible ritual grounds, being tested on their beliefs, etc.)

## Proof Of Testing
<details>

</details>

I can't easily prove that it works since I'd need to roll the heretic naturally on a chaplain with enough players in the round, but I have compiled the code many times and gotten no errors.

## Changelog
:cl: Dragon Lee
add: The chaplain can now roll the acolyte/heretic antagonist again.
/:cl:

### DIFF
--- a/modular_zubbers/code/modules/storyteller/event_defines/crewset/heretic.dm
+++ b/modular_zubbers/code/modules/storyteller/event_defines/crewset/heretic.dm
@@ -13,7 +13,7 @@
 	tags = list(TAG_COMBAT, TAG_SPOOKY, TAG_CREW_ANTAG)
 
 /datum/round_event_control/antagonist/solo/heretic/New()
-	protected_roles |= JOB_CHAPLAIN // Would be silly to get chaplain heretics
+//	protected_roles |= JOB_CHAPLAIN // Would be silly to get chaplain heretics
 	. = ..()
 
 #define TIME_CUTOFF 1.2 HOURS


Maia-J:
>  vampire chaplain (who can literally burn in their own church)

That sounds like a bug to me

KateAnbolho:
> > vampire chaplain (who can literally burn in their own church)
> 
> That sounds like a bug to me

that's really funny but is there a way to check for it? You choose your vampire "faction" after the antag

DragonLee66:
>that's really funny but is there a way to check for it? You choose your vampire "faction" after the antag

Yus here you go! As far as I know there's nothing stopping chaplains from choosing this clan.
<details>
<summary>Screenshot1</summary>
image

</details>

<details>
<summary>Screenshot2</summary>
image

</details>

I think it's less of a bug and more uh, an oversight.

Either way I'd rather have it left as is then not. I'd just rather have things that enable fun roleplay then just shutting it down. If the chaplain was DETRIMENTAL to killing magical creature then maybe- but as far as I know that's not the case.

Offwrldr:
I think we should make Chaplains anti-vamp and anti-magic powerhouses personally.

thegrb93:
Evil chaplain is good RP though.

DragonLee66:
>I think we should make Chaplains anti-vamp and anti-magic powerhouses personally.

But that's what ERT Chaplains are no? Souped-up chaplains meant to deal with magical threats that go out of hand. 

Generally though, I think antag banning roles is bad practice since it might just create a rabbit hole where we can start to argue that engineers, roboticists and botanists etc. etc. shouldn't roll antags because they're too strong/should fight against antags.

SapphoQueer:
> But that's what ERT Chaplains are no? Souped-up chaplains meant to deal with magical threats that go out of hand.

Yes, but unless the situation is close to being unsalvagable staff isn't really taught to send ERTs. Players have shared their grievances more than a few times over an ERT being called over minor things. I do not believe this to be relevant much to the decision to allow/restrict them, though, and as such I do think this is a good PR.

Chaplains can be strong against heretics/acolytes, whilst also being allowed to roll them, I believe. Security gets enough tools to deal with them either way, and there's other ways of countering the magic. (Holymelons. Holymelons. Holymelons)

Maia-J:
I'm not against this either, the idea of an evil occult ritual chaplain sounds too cool to have it blocked

However bloodsuckers might see it gone, since there the counter is more direct (Separately from this)

Maia-J (approved):
Gonna wait for one more maintainer input

Maia-J:
image

Waiting for the 24 hours, then merging

PR #6236, opened 2026-09-13
Resolves cargo problems introduced by upstream

NanoCats:
## About The Pull Request

The upstream sync(?) cargo bugs. I ran the big one by Drac who confirmed it should be fixed. 
image

Firstly, sometimes you can buy a crate and not be allowed to open it.

Before the last upstream sync, an order paid for by a dept budget arrived in a dedicated crate that opened for the purchaser's department. TG #97348 removed that crate type and replaced it with `/datum/component/locked_to_account`.

The component itself is fine, but the crate it goes on is now the supply pack's own crate, and the pack's `access` is still stamped onto it on top of the account lock.  TL;DR you can buy something you're allowed to buy as a head of staff for your department and then not actually be able to open it. 

I fixed this by repairing the crate's programming. An order paid for by an account now:

* arrives in a crate themed around the department instead of its original crate or the 'commercial' crate, (This is a minor polish/QOL!) 
* is gated by the that department

`can_unlock()` on `/obj/structure/closet/crate/secure` now defers to the account lock when one is present. Crates without one are untouched and behave exactly as they do on master. 

The check the lock performs is unchanged from what the component already did, it has just been pulled out into `account_matches()` so the latch and the open hook share one implementation instead of two.

Each budget gets its own crate, so you can tell at a glance who paid for what. I think this looks a lot better than how it is now, or even was before, we had a lot of crate types you basically never see because of the fact that nothing uses them hardly.:

| Budget | Crate |
| --- | --- |
| Cargo |  cargo crate |
| Engineering | engineering crate |
| Science |  science crate |
| Defense |  weapons crate |
| Medical |  medical crate|
| Service | hydroponics crate (Closest to 'service crate' I could come up with tbh) |
| Command/Civil | secure command crate |
| Interdyne | interdyne crate (Merely theoretical, I don't think this will ever actually come up) |
| Anything else, and all private purchases | commercial crate (That's what I call the original gold and brown crate) |

The commercial crate reclaims `privatecrate`, which #97348 orphaned when it deleted the type that used it. No new sprites, the lock light is drawn generically by `closet_update_overlays()`.

Goody orders bundled by the supply shuttle now follow the same rule, so a departmental goody order lands in that department's crate rather than a blank secure one. Under the free shipping limit they still go into `/obj/item/storage/lockbox/order`, which already handled departments correctly.

Second issue:  Company imports could not be ordered from the order app

`allow_non_private_purchase` already exists and company imports already set it. It was wired into `orderconsole.dm` and `expressconsole.dm` but missed in `budgetordering.dm`, so the same pack that a cargo console would happily sell you was rejected by the order app with "Small crates may only be purchased by private accounts."

This is technically how it probably should work and I suspect it's an oversight, company imports ≠ goodies, there are perfectly good reasons to put say, an RCF on the engineering budget. 

## Why It's Good For The Game

Heads of staff can requisition on their budget again and not be subject to mysterious unopenable crates, which is how it worked for a long time before the sync and how everyone expects it to work. Being charged for a crate you then cannot open is a miserable way to find out something broke.

The per department crates are a small readability win on the shuttle. A pile of identical secure crates becomes a pile you can sort by eye.

## Testing

yeah I tried it and it works 

image

## Changelog

:cl:
fix: Crates bought with a department budget can be opened by that department again, instead of arriving locked to nobody.
fix: Company imports can be ordered from the NT order app again.
/:cl:


code/datums/components/locked_to_account.dm
```diff
@@ -34,38 +34,60 @@
...
+			return TRUE
+	else if(id_card.registered_account == buyer_account)
+		return TRUE
+
+	if(!silent)
+		deny(source, user, "Bank account does not match with buyer!")
+	return FALSE
+// BUBBER EDIT ADDITION END
 
-	var/is_department = istype(buyer_account, /datum/bank_account/department) // department orders check the user's paycheck department instead of a personal account match.  TRUE if this user is allowed to open the crate
```
BingusTheClown, line 60: Should probably bubber edit out TG's code instead of deleting, so we know it belonged to them.

code/datums/components/locked_to_account.dm
```diff
@@ -34,38 +34,60 @@
...
- * runs when someone tries to open the crate, checks their ID against buyer_account
- * * force: if TRUE, the open bypasses this lock entirely (e.g. admin or code-forced opens)
+ * Returns TRUE if [user] is holding an ID that satisfies this lock.
+ * A departmental lock accepts anyone paid by that department, a personal one only the buyer.
+ * * source: the crate, only used to play the denial sound. Optional.
+ * * silent: if FALSE, tells the user why they were turned away.
  */
-/datum/component/locked_to_account/proc/on_pre_open(obj/structure/closet/source, mob/living/user, force)
-	SIGNAL_HANDLER
-	if(force)
```
BingusTheClown, line 45: This is also probably good to bubber edit out as opposed to delete

BingusTheClown (approved):

PR #6240, opened 2026-09-14
Stunning new advances in the field of brooming

NanoCats:
## About this PR

Adds the advanced push broom, the missing twin to the Advanced mop. 

The broom has powered bristles. Charging them sweeps three tiles abreast instead of one, as if three janitors were walking shoulder to shoulder. The charge sticks you to the floor a little or whatever, so it costs a small amount of movement speed while it is running. It's bluespace, I don't gotta explain shit. 

Comes in stanadlone AND implant version (in the advanced Jannie toolset implant)
1. The standalone broom does nothing until you brace it with two hands, the same as a normal push broom. Bracing it charges the bristles and does the cool wide sweepy thing
2. The implanted broom needs no bracing, since it is bolted to your arm. It sweeps as if wielded while it's out, and using it in hand toggles the extra wide sweep. We only got so many fuckin' buttons on the keyboard. 

Also it sweeps less individual items as a tradeoff for the badass 3-abreast thing. 

Fun fact I had to re-make this PR and branch because github makes me want to pull my hair out. 

## Why It's Good For The Game

The janitorial implant is the cool and good version of the janitor's kit, I always felt like it was missing something, so I made that something real. 

The wide sweep was my best idea for a cool extra feature, I was like "ughhhhh we have the advanced mop.... What aaaam I going to do to make it special...".  

## Testing

Tested in game, it works. Here are the sprites: 
image 

...It's just a slightly edited version of the regular broom. 

## Changelog

:cl:
add: Added the advanced push broom
/:cl:


CB35SSC:
JaniTech finally reaches its deserved conclusion.

JackSage:
Borg upgrade version...? 👉👈

modular_zubbers/code/game/objects/items/advanced_broom.dm
```diff
@@ -0,0 +1,107 @@
...
+
+/obj/item/pushbroom/advanced/on_unwield(obj/item/source, mob/user)
+	. = ..()
+	set_wide_sweep(user, FALSE)
+
+/obj/item/pushbroom/advanced/proc/set_wide_sweep(mob/user, charged)
+	if(wide_sweep == charged)
+		return
+	wide_sweep = charged
+	slowdown = wide_sweep ? ADVANCED_BROOM_SLOWDOWN : initial(slowdown)
```
Maia-J, line 51: Does this actually update just from this?

Boopideedoo:
So from what Im seeing this is a good add to janitors they definently need more and this makes sure thats happening but can janiborgs get this too?

modular_zubbers/code/game/objects/items/advanced_broom.dm
NanoCats: Nope, it's actually broken because i'm stupid, Fixed.

PR #6270, opened 2026-09-19
Resprite departmental guard batons

NanoCats:
## About The Pull Request

Resprites the departmental guard batons with a new appearance.

The new sprites are in my opinion a little cleaner and more visually striking, particularly while the baton is active. Each departmental variant keeps its existing color identity.

I also went back and touched up/restored the original baton sprites rather than removing them entirely. They're still available as an alternate skin for anyone who prefers the old look. And watch this space, because I'm whipping up MORE changes to guards. 

The available styles are:

- **Modern/Nouveau** - New default sprites.
- **Coder Art** - The original departmental baton design, cleaned up and retained as an alternate.

## Why It's Good For The Game

The existing guard batons are pretty old and, frankly, kind of ugly according to a focus group that is real which I didn't make up to support my argument.

The new sprites bring them closer in line with newer equipment.

At the same time, there's no real reason to throw the originals away. Keeping them as an alternate skin preserves the old aesthetic for people who prefer it and gives departmental guards another small customization option. I cleaned them up and made them obey the new rules just to be safe. Both versions are GAGS compatible too! Make the pimp baton, go for it, sure, why not. 

## Proof Of Testing

New:
image

Old:
image

## Changelog

:cl:
image: Resprited departmental guard batons and added their original appearance as an alternate skin.
/:cl:


YakumoChen:
Are the in-hands different? Concerned that blue glow is going to be less striking than the traditional yellow baton in hands that everyone knows.

NanoCats:
> Are the in-hands different? Concerned that blue glow is going to be less striking than the traditional yellow baton in hands that everyone knows.

The inhands will be blue when I finish, you'll just have to learn that dept guard batons have a blue glow instead of orange /shrug

